PR #18, opened 2019-04-24
First pass at HTTP/2 support

tomchristie:


requirements.txt
```diff
@@ -1,5 +1,6 @@
 certifi
 h11
+h2
```
yeraydiazdiaz: We also need to add it to `install_requires` in the setup.py

tomchristie: Good catch - thanks!

PR #19, opened 2019-04-27
Add __str__ method to URL to convert url components to url string

chdsbd:
I think it's slightly nicer to be able to convert a url to a string via `str()`.


tomchristie:
Thanks, yup!
Ended up getting pulled in already elsewhere, but appreciate this all the same!

Issue #10, opened 2019-04-18
Exceptions for decoding errors


tomchristie:
If Gzip, Brotili or Deflate decoding fails it should raise a specific exception type.

PR #20, opened 2019-04-27
Wrap decoding errors with custom exceptions

yeraydiazdiaz:
Addressing #10


httpcore/__init__.py
```diff
@@ -9,6 +9,9 @@
     ResponseClosed,
     StreamConsumed,
     Timeout,
+    DeflateDecodingError,
+    GzipDecodingError,
+    BrotliDecodingError,
```
tomchristie: Let's just roll with a single `DecodingError` exception class.

httpcore/decoders.py
```diff
@@ -77,16 +92,22 @@ def __init__(self) -> None:
...
+            self.decompressor.finish()
+            return b""
+        except brotli.Error as exc:
+            raise httpcore.exceptions.BrotliDecodingError from exc
 
 
 class MultiDecoder(Decoder):
     """
-    Handle the case where mutliple encodings have been applied.
+    Handle the case where mutiple encodings have been applied.
```
tomchristie: We've both typo'ed here. "m-u-l-t-i-p-l-e"  🤣

tomchristie:
Ace, thanks!

Issue #14, opened 2019-04-23
Wait for connection close


tomchristie:
Sync support for `.raw()` doesn't work just yet because the transport close is implemented as as a "call_soon", and we're currently not using `transport.wait_closed`, meaning that the asyncio loop ends up being closed before the pending task completes.

PR #21, opened 2019-04-27
Add `raw` method to SyncResponse

yeraydiazdiaz:
Towards fixing #14, though I haven't been able to reproduce the error.

From this branch, which does not introduce `wait_closed`, and using httpbin.org as a test:

```
import httpcore
http = httpcore.SyncConnectionPool()
response = http.request('GET', 'http://httpbin.org/stream-bytes/1024?chunk_size=8', stream=True)

buffer = b''
for data in response.raw():
    print(len(data), end=" ")
    buffer += data

print(len(buffer))
```

Succeeds with an output similar to : `176 368 208 104 112 56 1024`, indicating the iterator is working as expected.

I've tried this on Mac and Python 3.6 and 3.7 docker containers in case it only happened on Linux but the output is similar. Can you share how you got this error?

It's possible, of course, it only happens under specific circumstances, so I could add `await self._writer.wait_closed()` anyway to make sure things are properly closed.


tomchristie:
Appears to work fine for me too - perhaps I was missing something at the time.
Great work, thanks!

Issue #38, opened 2019-05-03
Add base_url to Client


tomchristie:
Client should support an optional `base_url`. If provided then .get() etc should accept relative URLs such as “/homepage”, which should automatically be joined against the base url.

tomchristie:
Let's punt on this for the moment. There'd probably be some other things to think about here - eg. expectation for redirect-off-domain behavior.

PR #46, opened 2019-05-08
Client base_url

taoufik07:
ref #38

### DIFF
--- a/httpcore/client.py
+++ b/httpcore/client.py
@@ -31,6 +31,7 @@ from .status_codes import codes
 class AsyncClient:
     def __init__(
         self,
+        base_url: URLTypes = None,
         ssl: SSLConfig = DEFAULT_SSL_CONFIG,
         timeout: TimeoutConfig = DEFAULT_TIMEOUT_CONFIG,
         pool_limits: PoolLimits = DEFAULT_POOL_LIMITS,
@@ -45,6 +46,8 @@ class AsyncClient:
 
         self.max_redirects = max_redirects
         self.dispatch = dispatch
+
+        self.base_url = None if base_url is None else URL(base_url)
 
     async def get(
         self,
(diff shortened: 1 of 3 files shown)


httpcore/client.py
```diff
@@ -29,6 +30,7 @@
 class Client:
     def __init__(
         self,
+        base_url: str = None,
```
tomchristie: We'd want to use `URLTypes` here.

httpcore/client.py
```diff
@@ -44,6 +46,8 @@ def __init__(
         )
         self.adapter = EnvironmentAdapter(dispatch=redirect_adapter)
 
+        self.base_url = base_url
```
tomchristie: Let's use the `URL` type. Eg. `self.base_url = None if url is None else URL(base_url)`

httpcore/client.py
```diff
@@ -57,6 +61,10 @@ def __init__(
         ssl: SSLConfig = None,
         timeout: TimeoutConfig = None,
     ) -> Response:
+
+        if self.base_url is not None:
+            url = urljoin(self.base_url, url)
```
tomchristie: I think it'd be something like...

```python
if self.base_url is not None:
    url = URL(url, allow_relative=True).resolve_with(self.base_url)
```

tests/test_client.py
```diff
@@ -12,6 +12,25 @@
...
 
+@pytest.mark.asyncio
+async def test_get_base_url(server):
+    base_url = "http://127.0.0.1:8000/hello/"
+    path = "hummans/"
+    async with httpcore.Client(base_url=base_url) as client:
+        response = await client.get(path)
+    assert response.status_code == 200
+    assert response.text == "Hello, world!"
+    assert str(response.request.url) == "http://127.0.0.1:8000/hello/hummans/"
```
tomchristie: Let's just have a single example in the test case, and use `http://127.0.0.1:8000/` as the base URL, and just do `.get('/hello')` or something like that.

(Also just use `str(response.url)` instead of `str(response.request.url)`)

tomchristie:
On reflection I think we should punt on this for now.
We *might* want to think about it in the future, but there may be some interplay between "base_url" and expected behaviors when redirecting to alternate hosts etc.

Issue #26, opened 2019-05-02
100% Test coverage


tomchristie:
Let's get the test coverage up to 100%, and then force-pin it.

Any contributions towards this are welcome.

yeraydiazdiaz:
The merged PRs and #64 is as much as I've been able to increase the coverage, the report on that branch is as follows:

```
+ venv/bin/coverage report -m
Name                                      Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------
httpcore/__init__.py                         10      0   100%
httpcore/client.py                          132      0   100%
httpcore/concurrency.py                      90      3    97%   79, 93-94
httpcore/config.py                           88      0   100%
httpcore/decoders.py                         63      0   100%
httpcore/dispatch/__init__.py                 0      0   100%
httpcore/dispatch/connection.py              58      0   100%
httpcore/dispatch/connection_pool.py         96      3    97%   78-80
httpcore/dispatch/http11.py                  73      3    96%   69, 100-103
httpcore/dispatch/http2.py                   91      0   100%
httpcore/exceptions.py                       17      0   100%
httpcore/interfaces.py                       36      0   100%
httpcore/models.py                          482      0   100%
httpcore/status_codes.py                      2      0   100%
httpcore/utils.py                            22      0   100%
tests/client/test_async_client.py            72      0   100%
tests/client/test_client.py                  86      0   100%
tests/client/test_redirects.py              158      0   100%
tests/conftest.py                            49      0   100%
tests/dispatch/__init__.py                    0      0   100%
tests/dispatch/test_connection_pools.py      68      0   100%
tests/dispatch/test_connections.py           20      0   100%
tests/dispatch/test_http2.py                 32      0   100%
tests/dispatch/utils.py                      58      0   100%
tests/models/test_headers.py                 94      0   100%
tests/models/test_queryparams.py             26      0   100%
tests/models/test_requests.py                61      0   100%
tests/models/test_responses.py              113      0   100%
tests/models/test_url.py                     23      0   100%
tests/test_config.py                         56      0   100%
tests/test_decoders.py                       59      0   100%
tests/test_timeouts.py                       19      0   100%
-----------------------------------------------------------------------
TOTAL                                      2254      9    99%
```

- `httpcore/dispatch/http11.py                  73      3    96%   69, 100-103`
    + Line 69 covers receiving an informational response which I haven't been able to reproduce from our server fixture.
[...]

PR #64, opened 2019-05-14
Include URL.__hash__ in test

yeraydiazdiaz:
Increasing the test coverage #26 ever so slightly.


tomchristie:
I’d probably:

1. Add a test to models/test_url instead of test_request.
2. Test that URL instances can used in a set, rather than testing hash directly. (Since that’s the actual behaviour we care about here)

PR #68, opened 2019-05-16
Add 100-continue test

tomchristie:


jamshid:
Sorry I can't tell, does this provide full support for `100 Continue`, where it does not send the body until the 100 response? 
I just asked in https://github.com/encode/httpx/discussions/1713. This would be a nice advantage of `httpx` over `requests`.

PR #94, opened 2019-06-18
Support WSGI and ASGI apps.

tomchristie:
Add support for using the client directly against WSGI apps:

```python
def hello_world(environ, start_response):
    ...

client = http3.Client(app=hello_world)
client.get('https://www.example.com')
```

And against ASGI apps:

```python
async def hello_world(scope, receive, send):
    ...

client = http3.Client(app=hello_world)
client.get('https://www.example.com')
```

Also note that we've got the option of either testing with `Client` or `AsyncClient` in either case.


jodal:
Are there any plans to extend this to also work with `AsyncClient(app=app)`?

jodal:
For most async test cases, I guess it's totally OK to use `http3.Client(app=app)` and do blocking HTTP calls from the test. However, a use case I have is to:

1. Create an async HTTP client wrapping a web app mock, e.g. made with Starlette in a few lines inside the test.
2. Pass that HTTP client to the code under test.
3. Exercise the code under test, which will then use the HTTP client I provided.

Since the code under test expects an async HTTP client, I must be able to create an async HTTP client in my test setup, wrapping an ASGI web app mock.

tomchristie:
@jodal - Yup, it'll already work just fine with AsyncClient too. 👍

jodal:
Sorry, it just noticed. I might have run into an issue with it though. Will try to make a minimal test case, and if I can reproduce it there, I'll open a proper issue. Thanks for your quick response!

tomchristie:
Gotcha, please do yup.

PR #112, opened 2019-07-08
Minor tweaks for connection keep-alive.

cansarigol:
I was looking at your PR, wanted to add a little decorator to narrow the code down. If it's useful.

### DIFF
--- a/http3/decorators.py
+++ b/http3/decorators.py
@@ -1,0 +1,21 @@
+import functools
+import typing
+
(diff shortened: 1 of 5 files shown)


tomchristie:
Thanks. Actually this would change the behaviour slightly as currently implemented. We’re careful to only wrap the single first “write” - we only want to retry if the connection was dropped without us being able to start the request.

Issue #24, opened 2019-05-02
Text streaming


tomchristie:
Currently all of the streaming APIs we're providing are byte-wise. We ought to also support text-decoding interfaces.

This would likely mean that we'd want `stream(text=True)`. Difficulties here are that we don't have `.apparent_encoding` available to us prior to the content being available. We'd need to use [chardet's incremental decoding](https://chardet.readthedocs.io/en/latest/usage.html#example-detecting-encoding-incrementally) as well as an [incremental decoder](https://docs.python.org/3/library/codecs.html#codecs.getincrementaldecoder)

I'd suggest we'd want to implement this by adding a `TextDecoder(encoding=None)` class to `decoders.py` that takes the `charset_encoding` as it's initial argument. It would implement the same interface as the existing decoders, except returning `str`.

Aside: It's possible that in the future we might start to move more into line with requests' `iter_content()` and `iter_lines()`.

graingert:
having an incrementally decoding File and StreamReader like objects would be amazing contributions to the chardet project https://github.com/chardet/chardet/issues/154

PR #115, opened 2019-07-11
Support for streaming Multipart and files

### LINKED ISSUE #114: Design Multipart helper to stream by default
This is a frequent problem with urllib3 in that the default behavior loads all of the file's data into memory at once before sending resulting in a huge files getting loaded into memory.

It's a hard problem to fix in urllib3 because of how the API was designed, but we can still fix our implementation not to do this.  I've got part of a patch for this currently.

sethmlarson:
Here's my first extremely rough stab at allowing us to support passing a file-like object to `data=` parameter in addition to not loading multipart `files` parameters completely into memory before sending.

You're going to have to be rough on me for the first few reviews because I've only looked at the codebase for ~3 hours. Lots to learn! :sweat_smile:

Fixes #114

### DIFF
--- a/http3/dispatch/http2.py
+++ b/http3/dispatch/http2.py
@@ -103,6 +103,8 @@ class HTTP2Connection:
     async def send_data(
         self, stream_id: int, data: bytes, timeout: TimeoutConfig = None
     ) -> None:
+        if data == b"":
+            return
         flow_control = self.h2_state.local_flow_control_window(stream_id)
         chunk_size = min(len(data), flow_control)
         for idx in range(0, len(data), chunk_size):
--- a/http3/interfaces.py
+++ b/http3/interfaces.py
@@ -17,6 +17,7 @@ from .models import (
     Response,
     URLTypes,
 )
+from .utils import async_streamify
 
 
 class Protocol(str, enum.Enum):
@@ -204,6 +205,7 @@ class ConcurrencyBackend:
         raise NotImplementedError()  # pragma: no cover
 
     def iterate(self, async_iterator):  # type: ignore
+        async_iterator = async_streamify(async_iterator)
         while True:
             try:
                 yield self.run(async_iterator.__anext__)
--- a/http3/models.py
+++ b/http3/models.py
@@ -28,7 +28,14 @@ from .exceptions import (
 )
 from .multipart import multipart_encode
 from .status_codes import StatusCode
-from .utils import is_known_encoding, normalize_header_key, normalize_header_value
+from .utils import (
+    async_streamify,
+    get_content_length,
+    is_known_encoding,
+    normalize_header_key,
+    normalize_header_value,
+    streamify,
+)
 
 URLTypes = typing.Union["URL", str]
 
@@ -545,7 +552,7 @@ class BaseRequest:
             if is_streaming:
                 auto_headers.append((b"transfer-encoding", b"chunked"))
             elif content:
-                content_length = str(len(content)).encode()
+                content_length = str(get_content_length(content)).encode()
                 auto_headers.append((b"content-length", content_length))
         if not has_accept_encoding:
             auto_headers.append((b"accept-encoding", ACCEPT_ENCODING.encode()))
@@ -595,9 +602,9 @@ class AsyncRequest(BaseRequest):
             self.is_streaming = False
             self.content = data
         else:
-            assert hasattr(data, "__aiter__")
+            assert hasattr(data, "__aiter__") or hasattr(data, "__iter__")
             self.is_streaming = True
-            self.content_aiter = data
+            self.content_aiter = async_streamify(data)
 
         self.prepare()
 
@@ -614,7 +621,8 @@ class AsyncRequest(BaseRequest):
             async for part in self.content_aiter:
                 yield part
         elif self.content:
-            yield self.content
+            async for part in async_streamify(self.content):
+                yield part
 
 
 class Request(BaseRequest):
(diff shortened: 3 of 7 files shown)


tomchristie:
Awesome - thanks so much for diving in on this!

> You're going to have to be rough on me for the first few reviews because I've only looked at the codebase for ~3 hours. Lots to learn!

Ha, the complementary part to that is go easy on me if you're running into any rougher edges in the codebase. Plenty of polishing and maturing still to be done.

So I'd suggest we tackle this (and other) pull requests incrementally, and try to keep the change footprint low for each as we go. That'll give us a nice feedback loop, and better reviews.

I'd suggest we start with a pull request that *just* adds support for BytesIO/TextIO request streams. (As opposed to just the existing iterating request streams.)

There's actually a bunch of fiddly stuff for us to get right there, esp with the sync/async bridging.

sethmlarson:
Yeah I realized to get the multipart done I'd have to also support file streaming, I'll split this into two.

tomchristie:
I'm going to just sling out thought process on this.

So, where are we at the moment?...

If we perform a `http3.post(url, data=BytesIO(b'abc'))` then I think that'll succeed just fine, since we support iterators, and BytesIO provides an `__iter__` interface. However...

* We ought to have a test case to establish that `BytesIO` interfaces work okay.
* Our buffering when passing a `BytesIO` interface isn't really ideal. Eg. if we do `http3.post(url, data=open('some-large-file', 'rb'))` then we'll iterate in whatever readline-sized chunks. Really we'd like to just iterate uploads on eg. 4096 byte sized chunks. We should probably always check if an iterator has a `read()` method, and use that if possible instead of iterating directly.
* We'll currently fail with `RequestBodyUnavailble` if we get a redirect. We should instead `seek(0)` if possible, and only error if that's not possible. We should have a test case to confirm that aspect of behavior.
* We should support *either* binary iterators *or* text iterators. At least as a first pass we should just encode to `utf-8` by default (use binary if you need more control). I think we should be checking the type of the data we iterate over for `str` or `bytes`, and error with a `TypeError` on anything else.
* A couple of the bullet points here are fiddly because we've actually bridged to `AsyncIterator` by the time we're sending the request *but* there's no stdlib async equivalent of `BytesIO/TextIO`. I guess we want to start out by providing one ourselves.

sethmlarson:
Thanks for the comments, I'll incorporate those into my PR for just file streaming. :)

sethmlarson:
I'm unsure about the last bullet point though, could you expand on that? You're saying we should provide an interface for an async file-like object with `async read()`?

tomchristie:
You're actually hitting on a really awkward area here wrt. us having to really tie down the text/byte stream/iterator sync/async interfaces.

Also related to stuff in #24. Plus eg. *exactly* what interface we should support for streaming responses. (Right now it's a byte iterator, but we'd probably be better off exposing byte stream and text stream interfaces)

sethmlarson:
I actually think we should support `bytes`, `bytes stream`, and `bytes` iterator in that order of preference. You're suggesting we should collapse all of those into a stream interface before handing off to whoever would previously consume the bytes iterator?

tomchristie:
> I'm unsure about the last bullet point though

Yeah, well it's fiddly.

> You're saying we should provide an interface for an async file-like object with async read()?

I think we'll need to, yes. At least both `read()` and `seek()`. We might forget about `read()` and nice buffering on a first pass here, and just aim to support `seek`.

If we pass a bytes iterator to a client, then we've [converted it into an async bytes iterator](https://github.com/encode/http3/blob/0.6.7/http3/client.py#L568-L581) by the time it's being sent.

That's not quite sufficient for our purposes here, because rather than just bridge from iterator -> async iterator, we actually need to bridge for stream -> async stream, so that we have an `async seek()` method available to us, to [call when redirects occur](https://github.com/encode/http3/blob/0.6.7/http3/client.py#L285).

sethmlarson:
You know you and I are really syncing up on these comments? :)

tomchristie:
> > I actually think we should support bytes, bytes stream, and bytes iterator in that order of preference

Yup, that sounds great to me too.

tomchristie:
So, fiddly question here is what's the AsyncClient interface?

tomchristie:
I'd assumed bytes, async byte streams, and async byte iterators but I'm not *actually* sure that's right.

sethmlarson:
`bytes`, `async bytes` stream (whatever this is, tbd), `bytes` streams, `async bytes` iterator and `bytes` iterator?

tomchristie:
> I think we still want AsyncClient to support synchronous operations, cuz why not? We can just coerce them over.

Indeed. I was thinking the same when walking through this.

tomchristie:
Which actually *kinda* changes the internals / approach for this.

tomchristie:
Because we don't want to perform any bridging, here: https://github.com/encode/http3/blob/0.6.7/http3/client.py#L568

But we *do* want to support coercing those types within `AsyncClient` (or `AsyncRequest`).

tomchristie:
At which point my reassessment of the very first PR towards any of this would be:

1. Add byte iterator support to AsyncClient.

And then, once that's complete...

2. Add byte stream support to Client and AsyncClient.

tomchristie:
Does that sound like a reasonable assessment / first steps?

tomchristie:
Also please don't be put off by this, you've just managed to hit one of the most awkward possible spots! 😅

sethmlarson:
That seems reasonable to me.  And don't worry, this isn't my first rodeo. :) Glad to be a part of the swamp management team. ;)

PR #116, opened 2019-07-11
Encoding detection in Response.json

### LINKED ISSUE #74: Finesse json
* Detect encoding from initial bytes, rather than just using "utf-8".
* Allow additional arguments to `response.json()` and review any subtle API diffs to requests.

yeraydiazdiaz:
Fixes #74

This PR replicates requests's logic where we typically use `response.text` and it's encoding detection via headers. In cases where it's not available [a helper function](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L851-L886) is used to detect possible UTF encodings which I ~shamelessly~ humbly copied over along with its tests (many thanks to the requests team for the complete coverage).

### DIFF
--- a/http3/models.py
+++ b/http3/models.py
@@ -29,10 +29,10 @@ from .exceptions import (
 from .multipart import multipart_encode
 from .status_codes import StatusCode
 from .utils import (
+    guess_json_utf,
     is_known_encoding,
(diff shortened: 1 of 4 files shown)


tomchristie:
Amazing! Yes please. 🙏

PR #118, opened 2019-07-12
Support client cert passwords, new TLS options

sethmlarson:
Regarding disabling commonName: commonName support has been deprecated in urllib3 for over 5 years now and Chrome and Firefox don't support commonName for new certificates I think it's time we turn it off where we can (Python 3.7+ and OpenSSL 1.1.0+).

### DIFF
--- a/http3/config.py
+++ b/http3/config.py
@@ -123,8 +123,6 @@ class SSLConfig:
             context.load_verify_locations(cafile=ca_bundle_path)
         elif os.path.isdir(ca_bundle_path):
             context.load_verify_locations(capath=ca_bundle_path)
-        else:
-            context.load_default_certs(purpose=ssl.Purpose.SERVER_AUTH)
 
         if self.cert is not None:
             if isinstance(self.cert, str):
--- a/http3/config.py
+++ b/http3/config.py
@@ -117,8 +117,10 @@ class SSLConfig:
 
         # Disable using 'commonName' for SSLContext.check_hostname
         # when the 'subjectAltName' extension isn't available.
-        if hasattr(context, "hostname_checks_common_name"):
+        try:
             context.hostname_checks_common_name = False
+        except AttributeError:
+            pass
 
         if os.path.isfile(ca_bundle_path):
             context.load_verify_locations(cafile=ca_bundle_path)
--- a/http3/config.py
+++ b/http3/config.py
@@ -39,8 +39,6 @@ class SSLConfig:
     SSL Configuration.
     """
 
-    __slots__ = ("cert", "verify", "ssl_context")
-
     def __init__(self, *, cert: CertTypes = None, verify: VerifyTypes = True):
         self.cert = cert
         self.verify = verify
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -80,6 +85,30 @@ def cert_and_key_paths():
...
+            )
+        )
+
+
+@pytest.fixture
+def cert_and_encrypted_key_paths():
+    ca = CAWithPKEncryption()
+    ca.issue_cert("example.org")
+    with ca.cert_pem.tempfile() as cert_temp_path, ca.private_key_pem.tempfile() as key_temp_path:
+        yield cert_temp_path, key_temp_path
```
tomchristie: Could we use something like this?...

```python
@pytest.fixture
def example_cert():
    ca = CAWithPKEncryption()
    ca.issue_cert("example.org")
    return ca

@pytest.fixture 
def cert_pem_file(example_cert):
with example_cert.cert_pem.tempfile() as cert_temp_path:
    yield cert_temp_path

@pytest.fixture 
def cert_private_key_file(example_cert):
with example_cert.private_key_pem.tempfile() as key_temp_path:
    yield key_temp_path
```

In order to allow us to pass fixtures as `cert_pem_file, cert_private_key_file` rather than unpacking a two-tuple in the test body?

http3/config.py
```diff
@@ -185,6 +215,8 @@ class PoolLimits:
     Limits on the number of connections in a connection pool.
     """
 
+    __slots__ = ("soft_limit", "hard_limit", "pool_timeout")
```
tomchristie: I think let's pass on making `__slots__` changes in PRs.

If we do want to add them in then let's treat em as an independent PR. (Tho I'd be tempted to say we shoud simply ignore them, at least until we're closing in on a 1.0??)

tomchristie:
Fab! Two minor bits to talk through.

http3/config.py
sethmlarson: Sure! I'll revert these.

http3/config.py
```diff
@@ -101,20 +103,24 @@ def load_ssl_context_verify(self) -> ssl.SSLContext:
...
-        context.options |= ssl.OP_NO_COMPRESSION
-
-        context.set_ciphers(DEFAULT_CIPHERS)
-
-        if ssl.HAS_ALPN:
-            context.set_alpn_protocols(["h2", "http/1.1"])
-        if ssl.HAS_NPN:
-            context.set_npn_protocols(["h2", "http/1.1"])
+        context.check_hostname = True
+        context.load_default_certs(purpose=ssl.Purpose.SERVER_AUTH)
```
sethmlarson: We shouldn't be loading default certs unless we don't specify a verify path.

sethmlarson:
I'm deferring the implementation of loading system CA certs because doing what urllib3 does still isn't the right solution. I need to rope Christian in and see what we can even do right now, would like to get this right before we start shipping.

tomchristie:
Loverly stuff!

Issue #119, opened 2019-07-13
DataField.render_data fails when value is not string


ambrozic:
DataField only likes `str` objects as values. DataField.render_data will produce an error when ie. `int` or `bytes` objects are part of posted data or form as well as `files`.

```
  File ".../http3/multipart.py", line 29, in render_data
    return self.value.encode("utf-8")
AttributeError: 'int' object has no attribute 'encode'
```

Test to reproduce
```
@pytest.mark.parametrize(("value"), (123, b"abc"))
def test_multipart_non_str(value):
    client = Client(dispatch=MockDispatch())
    data = {"text": value}
    files = {"file": io.BytesIO(b"<file content>")}
    response = client.post("http://127.0.0.1:8000/", data=data, files=files)
    assert response.status_code == 200
```

or code

```
import http3, io

http3.post(
    url="https://httpbin.org/post",
    data={"a": b"1"},
    files={"file": io.BytesIO(b"<file content>")},
)
```

tomchristie:
Okay. Multipart encoded data can *only* handle string or File data. We could perfectly well either coerce a limited set of scalar types, or just coerce any inputs with str()

tomchristie:
Would probably be a useful point of reference to stay in line with whatever urllib3 currently does.

ambrozic:
https://github.com/urllib3/urllib3/blob/master/src/urllib3/filepost.py#L84-L90

urllib3 isn't doing anything particularly advanced either. i can make a PR with a similar solution if you are fine with it.

tomchristie:
So looks like urllib3 only accepts str/byte values. What does requests do for that case? If it coerces to str, then any idea where it does that?

ambrozic:
Most of the fiddly magic in requests happens in [requests.models:_encode_files](https://github.com/kennethreitz/requests/blob/bedd9284c9646e50c10b3defdf519d4ba479e2c7/requests/models.py#L124-L139). Quite a bit of knowledge implemented which can help us. Prepared data is then sent to [urllib.filepost:encode_multipart_formdata](https://github.com/urllib3/urllib3/blob/master/src/urllib3/filepost.py#L63-L98) we've already checked above. 

Another edge case i came across reading this code are `None` values in `data` which would have to be handled in [http3.multipart:multipart_encode](https://github.com/encode/http3/blob/master/http3/multipart.py#L93)

tomchristie:
Okay, so to keep in the with the behaviour I can see there, we should:

[...]

PR #121, opened 2019-07-14
Multipart data values encoding

ambrozic:
- `None` values and lists of args are handled in `http3/multipart:iter_fields`
- Everything else that comes through being non `bytes` is cast to string
- Added tests for `multipart_encode`. Complex objects parsing into `multipart/form-data` body is same as in requests and urllib3


http3/multipart.py
```diff
@@ -26,6 +26,10 @@ def render_headers(self) -> bytes:
         )
 
     def render_data(self) -> bytes:
+        if not isinstance(self.value, str):
+            if isinstance(self.value, bytes):
+                return self.value
+            self.value = str(self.value)
```
sethmlarson: I'm not sure I'm a big fan of this behavior? Why not raise a TypeError saying what we accept and forcing users to pass values that make sense to this interface?

http3/multipart.py
```diff
@@ -83,9 +89,11 @@ def iter_fields(data: dict, files: dict) -> typing.Iterator[Field]:
         yield FileField(name=name, value=value)
 
 
-def multipart_encode(data: dict, files: dict) -> typing.Tuple[bytes, str]:
+def multipart_encode(
+    data: dict, files: dict, boundary: bytes = None
```
sethmlarson: imo we shouldn't expose the boundary interface publicly. What use-case is there to allowing users to control it?

http3/multipart.py
ambrozic: replied in issue post #119

http3/multipart.py
ambrozic: only reason was to make it easier to test. similar to urlib3. if it is unacceptable it can be removed.

sethmlarson: Can be tested by mocking out `os.urandom()`. I think it should be removed as it shouldn't be a necessary interface.

http3/multipart.py
```diff
@@ -15,9 +15,13 @@ def render_data(self) -> bytes:
...
 
 class DataField(Field):
-    def __init__(self, name: str, value: str) -> None:
+    def __init__(self, name: str, value: typing.Union[str, bytes]) -> None:
         self.name = name
         self.value = value
+        if not isinstance(self.name, str):
+            raise TypeError("Invalid type for name. Expected str.")
+        if not isinstance(self.value, (str, bytes)):
+            raise TypeError("Invalid type for value. Expected str or bytes.")
```
ambrozic: Is there a more elegant way to write this 4 lines above?

sethmlarson: Seems fine to me, maybe if we end up doing this a lot we'd want a function that checks the type of a value and raises a TypeError with the given message if it's not valid?

ambrozic: Such function would be eventually useful. Type hinting could be optionally stricter in sense that could raise runtime errors.

http3/multipart.py
```diff
@@ -73,7 +79,9 @@ def render_data(self) -> bytes:
 
 def iter_fields(data: dict, files: dict) -> typing.Iterator[Field]:
     for name, value in data.items():
-        if isinstance(value, list):
+        if value is None:
```
tomchristie: Let's not do the "omit None values" thing - I think we should just let `None` raise an error, same as any other non-str/bytes type.

tomchristie:
Great stuff - thank you!

PR #123, opened 2019-07-15
Add RequestDataIter for normalizing streams, iterators

sethmlarson:
Makes it so you can pass iterators, async iterators, open files and file-like objects, and raw strings as `data`. Do we need `Request.is_streaming` anymore?

### DIFF
--- a/http3/client.py
+++ b/http3/client.py
@@ -284,7 +284,9 @@ class BaseClient:
         """
         if method != request.method and method == "GET":
             return b""
-        return request.content.rewind()
+        if request.content.seekable:
+            request.content.seek(0, 0)
+        return request.content.data()
 
 
 class AsyncClient(BaseClient):
--- a/http3/client.py
+++ b/http3/client.py
@@ -276,15 +276,15 @@ class BaseClient:
             del headers["Authorization"]
         return headers
 
-    def redirect_content(self, request: AsyncRequest, method: str) -> bytes:
+    def redirect_content(
+        self, request: AsyncRequest, method: str
+    ) -> typing.Union[bytes, typing.IO[typing.AnyStr]]:
         """
         Return the body that should be used for the redirect request.
         """
         if method != request.method and method == "GET":
             return b""
-        if request.is_streaming:
-            raise RedirectBodyUnavailable()
-        return request.content
+        return request.content.rewind()
 
 
 class AsyncClient(BaseClient):
--- a/http3/models.py
+++ b/http3/models.py
@@ -22,6 +22,7 @@ from .exceptions import (
     CookieConflict,
     HttpError,
     InvalidURL,
+    RedirectBodyUnavailable,
     ResponseClosed,
     ResponseNotRead,
     StreamConsumed,
@@ -57,9 +58,13 @@ AuthTypes = typing.Union[
     typing.Callable[["AsyncRequest"], "AsyncRequest"],
 ]
 
-AsyncRequestData = typing.Union[dict, str, bytes, typing.AsyncIterator[bytes]]
+AsyncRequestData = typing.Union[
+    dict, str, bytes, typing.Iterator[typing.AnyStr], typing.AsyncIterator[typing.AnyStr], typing.IO[typing.AnyStr]
+]
 
-RequestData = typing.Union[dict, str, bytes, typing.Iterator[bytes]]
+RequestData = typing.Union[
+    dict, str, bytes, typing.Iterator[typing.AnyStr], typing.IO[typing.AnyStr]
+]
 
 RequestFiles = typing.Dict[
     str,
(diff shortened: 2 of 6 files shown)


http3/models.py
```diff
@@ -1166,3 +1167,106 @@ def info(self) -> email.message.Message:
...
+        if isinstance(wrapped, str):
+            wrapped = wrapped.encode("utf-8")
+        elif hasattr(wrapped, "read"):
+            self._is_stream = True
+        elif hasattr(wrapped, "__next__"):
+            self._is_sync_iterator = True
+        elif hasattr(wrapped, "__anext__"):
+            self._is_async_iterator = True
+        elif not isinstance(wrapped, bytes):  # pragma: nocover
+            raise ValueError("Unsupported data type")
```
sethmlarson: This should probably be a TypeError?

http3/models.py
```diff
@@ -1166,3 +1167,106 @@ def info(self) -> email.message.Message:
             for key, value in self.response.headers.items():
                 info[key] = value
             return info
+
+
+class RequestDataIter:
+    def __init__(
+        self,
+        wrapped: typing.Union[
+            bytes, str, typing.IO[typing.AnyStr], typing.Iterator[typing.AnyStr]
```
sethmlarson: Should just re-use `AsyncRequestData` type hint here.

http3/models.py
```diff
@@ -1166,3 +1167,106 @@ def info(self) -> email.message.Message:
...
+            return self._wrapped
+        raise RedirectBodyUnavailable()
+
+    def __len__(self) -> int:
+        if (
+            self._is_stream
+            and hasattr(self._wrapped, "seek")
+            and hasattr(self._wrapped, "tell")
+        ):
+            start = self._wrapped.tell()
```
sethmlarson: Might be a more efficient way of doing this?

http3/models.py
```diff
@@ -1166,3 +1167,106 @@ def info(self) -> email.message.Message:
...
+            self._wrapped.seek(0, 2)
+            end = self._wrapped.tell()
+            self._wrapped.seek(0, start)
+            return end - start
+        return len(self._wrapped)
+
+    async def __aiter__(self):
```
sethmlarson: Should fix the ordering on these functions.

http3/models.py
```diff
@@ -1166,3 +1167,106 @@ def info(self) -> email.message.Message:
...
+                self._done = True
+                return b""
+        else:
+            return self.chunk()
+
+    def rewind(self) -> typing.Union[bytes, typing.IO[typing.AnyStr]]:
+        if self._is_stream:
+            if hasattr(self._wrapped, "seek"):
+                try:
+                    self._wrapped.seek(0, 0)
```
sethmlarson: This is a bug, need to capture the original starting point of the file before rewinding.

sethmlarson:
Another bigger question is if we move towards having `Request.content` be this interface, should we make it private?

tomchristie:
Something along these lines seems like a good tack, yup.

A thought... perhaps rather than a custom interface, our wrapper class should be an implementation of `IOBase`. (With `UnsupportedOperation` for various stuff that's not relevant.)

Ie.

```python
class ByteStream(io.IOBase):
    """
    Wraps various text/bytes data types up into a single standard byte stream interface.
    """
    def __init__(self, content: typing.Union[
        str,
        bytes,
        typing.Iterator[typing.AnyStr],
        typing.IO[typing.AnyStr],
    ]):
        ...
```

(We'd also end up needing an equivelent async alternative.)

sethmlarson:
I started off on the `IOBase` route and it's a lot more involved. :) I'll grit my teeth through it though.

tomchristie:
Ah okay, that's a shame. I'd figured there'd be a fair bit of functionality that's covered by the built-in mixin implementations. (Also, I think we should probably roll a `streams.py` module if we do go down that route, since it's a biggish chunk of stuff and we probably want to keep it isolated.)

sethmlarson:
I implemented via `IOBase` and this is what I've got right now. I'm not happy with the only way to test both `RequestDataStream` and `AsyncRequestDataStream` in the same testcase is via `request.content._wrapped.read()` but I don't see how else aside from a combined interface as before.  What are your thoughts on this patch @tomchristie?

http3/client.py
```diff
@@ -284,7 +284,9 @@ def redirect_content(
         """
         if method != request.method and method == "GET":
             return b""
-        return request.content.rewind()
+        if request.content.seekable:
+            request.content.seek(0, 0)
```
tomchristie: Presumably we need an `else error` case here?

tomchristie:
Okay, so this is quite large ATM.

There's also a few bits that I think we could seperate out more cleanly. Eg. I don't think `RedirectBodyUnavailable` should have any place in the stream class. Plus I think that the byte stream implementations are a distinct enough piece of work that they should live in an entirely seperate `streams.py` module.

So that we can get nice clean PRs onto all this I'd probably suggest:

* Farm out all the typo any cleanup stuff that's here into a seperate PR.
* We *start* with *just* a PR for a nicely seperate `ByteStream` class - without even hooking it in to anything else, or thinking about the async variant. Just a clean isolated interface, that's got test cases that exercise it.

http3/models.py
```diff
@@ -1166,3 +1184,164 @@ def info(self) -> email.message.Message:
...
+                size = len(data)
+            self._buffer = data[size:]
+            return bytes(data[:size])
+
+    def seek(self, offset: int, whence: int = 0) -> int:
+        if offset != 0 or whence not in (0, 2):
+            raise ValueError(
+                "RequestDataStream doesn't support seek() except "
+                "the beginning or end of the stream"
+            )
```
tomchristie: We *could* support this if we wanted, by reading and throwing away the data.

PR #125, opened 2019-07-16
Start using Nox, Flake8

sethmlarson:
- Switched from scripts to nox
- Started using flake8 to detect issues that autoflake doesn't fix
- Revise Travis config to run lint-check, docs, and tests
- Switch in gitignore for Python
- Fixed all lint issues / unused imports in the code


tomchristie:
Yo. Great stuff in here! 😀
Let's have individual PRs for different areas, tho, since there's all sorts of different bits mixed in.

Roughly:

* listing fixes - wonderful, yes. Let's 100% start by getting those in, against the existing testing/travis structure.
* nox - okay, looks like it could be rather nice, but let's have an isolated PR for it, and discuss.
* setup.py version pinning changes - those don't look correct to me. We should be pinned at a known major version number, not "above or equal too".
* gitignore change - Perhaps. Not esp. wild about the bloat.

sethmlarson:
I'm working on the linting fixes and nox. I'll leave gitignore mostly how it is. I didn't mean to let the setup.py changes go into this PR, sorry! :P

PR #126, opened 2019-07-18
Start using flake8, fix lint issues

sethmlarson:
`autoflake`+`black` fixes most issues but we should still use `flake8` to check for things not fixed by those two.


tomchristie:
Ace. Yes please!

PR #128, opened 2019-07-19
Start using nox for automation

sethmlarson:
I wasn't sure what to do with the `scripts/publish` and `scripts/clean`


sethmlarson:
This PR is actually ready for review now. :)

noxfile.py
```diff
@@ -0,0 +1,56 @@
...
+def test(session):
+    session.install("-r", "test-requirements.txt")
+
+    session.run(
+        "coverage",
+        "run",
+        "--omit='*'",
+        "-m",
+        "pytest",
+        "--cov=http3",
```
StephenBrown2: missed an `http3` rebrand?

tests/client/test_client.py
```diff
@@ -1,9 +1,8 @@
 import asyncio
 import functools
 
-import pytest
```
StephenBrown2: These lines look a lot like a failed isort to me. It _should_ know the known_first_party is `httpx`, but it's classifying it as a third party import.

noxfile.py
```diff
@@ -0,0 +1,56 @@
...
+source_files = ("httpx", "tests", "setup.py", "noxfile.py")
+
+
+@nox.session(reuse_venv=True)
+def lint(session):
+    session.install("autoflake", "black", "flake8", "isort")
+
+    session.run("autoflake", "--in-place", "--recursive", *source_files)
+    session.run(
+        "isort",
```
StephenBrown2: Adding [`--project=httpx` here](https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox#diff-37409d3afed5d01396a659e1b20bdf08R14) _should_ fix the first-party-is-third-party import swap, but for some reason it runs one way on my machine and the other way in a Docker container I built to test. I did end up getting it to be consistent by adding [`seed-isort-config`](https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox#diff-37409d3afed5d01396a659e1b20bdf08R11) and specifying the first party `httpx` in an [`.isort.cfg` file](https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox#diff-170b3556059bb1e212db57e7c98d02b6).

The following branch should be the correct sorting including `httpx`: https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox

The [Dockerfile I used](https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox#diff-3254677a7917c6c01f55212f86c57fbf) for testing is based off the one suggested in the nox docs, but only includes python 3.6, 3.7, and 3.8.

sethmlarson: Awesome! I'll use this as I'm sure it'll be a problem in the future if it's happening now.

StephenBrown2: Cool! I should have also noted that I needed to move `httpx` in the generated config to [known_first_party](https://github.com/sethmlarson/http3/compare/nox...StephenBrown2:nox#diff-170b3556059bb1e212db57e7c98d02b6R2), _then_ isort ran correctly.

sethmlarson: It looks like isort ran correctly for me while using ``--project=httpx`? Should I add it to isort.cfg anyways?

StephenBrown2: I would, because then the config won't conflict with the command line arg. It should be overridden anyways, but it wasn't working for me in this state for some reason.

StephenBrown2: LGTM!

I'll let the real reviewers review now. :-p

sethmlarson: Thank you for your review! :heart:

sethmlarson:
@StephenBrown2 The way I solve that issue is via installing each version using pyenv. Definitely harder if you don't use that though.

StephenBrown2:
Cool deal. `pyenv` is definitely a legit solution.

tomchristie:
This does look like a really nicely done tool, and I'm liking how the travis matrix looks with it.

I'm off now until next week, but will aim to take a proper look over it then.
Please do ping me again on it if it's dropped off my radar anytime.

sethmlarson:
Enjoy your time off! :)

sethmlarson:
I've merged all the changes upstream into this branch. :)

sethmlarson:
I'm gonna merge this PR pending CI. If we want to do changes to this setup we can handle them in a separate PR.

noxfile.py
```diff
@@ -0,0 +1,60 @@
...
+        *source_files,
+    )
+    session.run("black", "--target-version=py36", *source_files)
+
+    check(session)
+
+
+@nox.session(reuse_venv=True)
+def check(session):
+    session.install("black", "flake8", "flake8-bugbear", "mypy")
```
cansarigol: Shell we also add `flake8-comprehensions` here?

.travis.yml
```diff
@@ -4,16 +4,35 @@ language: python
 
 cache: pip
 
-python:
-    - "3.6"
-    - "3.7"
+branches:
+  only:
+    - master
+
```
cansarigol: Does it prevent the pr check before merging the commit?

sethmlarson: I don't understand your question but:
- We run CI against master commits
- For PRs what the resulting merge commit would be and not what the head commit for a given branch is.

noxfile.py
sethmlarson: Probably, let's do that in a separate PR though, this one's gotten out of hand.

.travis.yml
cansarigol: if I'm in `issue-x` branch and I create a merge request, I won't see the result and fix it before its merged. It would be nice if I could see.

cansarigol: I'm trying with a new pr. I'm not sure if I'm right.

cansarigol: Sorry, I thought that it is about contributor's branch but Travis evaluates upstream's branch.

sethmlarson: You'll be able to see the result in the PR? Travis publishes a commit status for PRs even if that branch is disabled.

cansarigol: I can see the result before merge.
Screen Shot 2019-08-13 at 17 19 48

pquentin: If I understand correctly, the only issue here is that you can no longer make sure CI passes on your branch before submitting a PR. This is useful for urllib3 which tests multiples OSes in CI and supports the macOS SecureTransport TLS library, but that's not what httpx does, so the change makes sense!

StephenBrown2: If you're developing on a separate branch on your own fork, this wouldn't apply anyways. This means that it will only run on submitted PRs, not on branches on the origin `encode/httpx` repo. :+1:

cansarigol: you're right. After sending PR, I was wondering that if I see the test result from the merge request page (not from my origin). Because my branch is not master. Travis uses upstream branch so it's ok for me.

PR #129, opened 2019-07-19
Renaming -> httpx

tomchristie:
Screen Shot 2019-07-19 at 13 32 37

Just rejigged the homepage to start with to see how it fits. We good with this?


tomchristie:
Ah, so glad ya like it.
Illustration by this fella https://en.wikipedia.org/wiki/Alpheus_Spring_Packard

tomchristie:
So, releasing to pypi as `httpx` - I'm assuming it'd make most sense to just stick with the version number we're currently on. (Even tho `httpx` on pypi would then start from 0.6.7 onwards.)

That seems fine to me, right?

sethmlarson:
Yeah no problems there. Release the Kraken!

sethmlarson:
Btw I'm not sure if you've got more to do in this PR or not so feel free to merge your own PR. :)

PR #134, opened 2019-07-21
Implement RetryConfig [WIP]

### LINKED ISSUE #108: Retry requests
urllib3 has a very convenient [Retry](https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/retry.py) utility ([docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry)) that I have found to be quite useful when dealing with flaky APIs. http3's Clients don't support this sort of thing yet, but I would love it if they did!

In the meantime, I can probably work out my own with a while loop checking the response code.

sethmlarson:
Early WIP not ready for review yet.  Closes #108.

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -1,18 +1,23 @@
 import asyncio
+import email.utils
+import enum
 import os
+import random
+import re
 import ssl
+import time
 import typing
 
 import certifi
 
-from .__version__ import __version__
+from .exceptions import HttpError, TooManyRetries
+from .models import AsyncRequest, AsyncResponse
+from .status_codes import StatusCode
 
 CertTypes = typing.Union[str, typing.Tuple[str, str], typing.Tuple[str, str, str]]
 VerifyTypes = typing.Union[str, bool]
 TimeoutTypes = typing.Union[float, typing.Tuple[float, float, float], "TimeoutConfig"]
-
-
-USER_AGENT = f"python-httpx/{__version__}"
+StatusCodeTypes = typing.Union[int, StatusCode]
 
 DEFAULT_CIPHERS = ":".join(
     [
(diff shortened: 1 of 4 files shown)


httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
...
+        if retry_after_header is not None:
+            if re.match(r"^\s*[0-9]+\s*$", retry_after_header):
+                seconds = int(retry_after_header.strip())
+            else:
+                retry_after_date = email.utils.parsedate(retry_after_header)
+                if retry_after_date is None:
+                    raise HttpError(
+                        f"'Retry-After' header is invalid: {retry_after_header}"
+                    )
+                seconds = int(time.time() - time.mktime(retry_after_date))
```
sethmlarson: Should be flipped around

httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
...
+            retryable_status_codes=self.retryable_status_codes,
+            retry_after_max=self.retry_after_max,
+            backoff_factor=self.backoff_factor,
+            backoff_max=self.backoff_max,
+            backoff_jitter=self.backoff_jitter,
+            history=new_history,
+        )
+
+    def sleep_for_retry(self, response: AsyncResponse) -> float:
+        number_of_retries = 0
```
sethmlarson: This should be determined from `RetryConfig.history`.

httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
...
+        self.backoff_max = backoff_max
+        self.backoff_jitter = backoff_jitter
+
+        self.history = history
+
+    def should_retry(
+        self, request: AsyncRequest, response: AsyncResponse
+    ) -> typing.Optional[AsyncRequest]:
+
+        """Method to be re-implemented by users in case they'd like to modify the request
```
sethmlarson: Gotta finish this docstring

httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
...
+    Retry values
+    """
+
+    def __init__(
+        self,
+        total_retries=None,
+        *,
+        read_retries: int = None,
+        connect_retries: int = None,
+        retryable_status_codes: typing.Optional[typing.Iterable[StatusCodeTypes]] = (
```
sethmlarson: The type hints on these aren't handled to their fullest extent (`...=None`)

httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
...
+        backoff_factor: float = 0.1,
+        backoff_max: float = 10.0,
+        backoff_jitter: float = 0.0,
+        history: typing.Tuple[RetryHistory, ...] = (),
+    ):
+        self.total_retries = total_retries
+        self.read_retries = read_retries
+        self.connect_retries = connect_retries
+
+        self.retryable_methods = retryable_methods
```
sethmlarson: Should convert the given values here into frozensets.

httpx/config.py
```diff
@@ -243,8 +248,203 @@ def __repr__(self) -> str:
         )
 
 
+class RetryCause(enum.IntEnum):
```
sethmlarson: Switch to using `enum.auto()` here since the values should be opaque.

StephenBrown2:
What's the status of this? Not impatient, just curious if it'll make it into 0.6.9 :-)

sethmlarson:
I've got some local changes but I've been mostly doing maintenance work for the project the past few days. I think tomorrow I'll have a block of time I can devote to working more on it. I'd also like to see it be released soon so we can get feedback early :)

tomchristie:
My rough sense on this is that it should make it in at some point, but that we should be focusing on a narrowly-as-possible scoped 1.0 beta release, and that this is over-stretch.

I don't want us to give ourselves loads of API surface area too early on.

Plus I think that we might want to take a bit of a look more generally at the dispatcher API / if we need some other middleware or hooks interface / if the redirect logic can be more neatly encapsulated as a middleware-ish layer.

If anything we may want to *start* by ensuring that our dispatcher API or anything else we introduce at that kind of level is sufficiently fit-for-purpose that stuff like retries can by implemented as a third party pacakge if needed.

sethmlarson:
Sure, I'll close this for now. We can revisit later!

PR #135, opened 2019-07-22
Additional HTTP Status Codes 418, 451

StephenBrown2:
The status codes 418 and 451 are defined in RFCs 2324 and 7725, and are both fun and practical. I've used 418 as a testing status during development, and 451 is a reasonable inclusion given GDPR and other privacy and legal-related reasons content may be unavailable.

For reference, [Golang][1] and [Symphony][2] implement both, and other languages and frameworks such as [Ruby][3] and [Starlette][4] implement at least one (451).

Both are also documented in the [MDN developer docs][5], so may I should make a PR to Starlette next to add support for 418. :smiley: 

[1]: https://golang.org/src/net/http/status.go
[2]: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Response.php
[3]: https://ruby-doc.org/stdlib-2.6.3/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-HTTP+Response+Classes
[4]: https://github.com/encode/starlette/blob/master/starlette/status.py#L49
[5]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

### DIFF
--- a/httpx/status_codes.py
+++ b/httpx/status_codes.py
@@ -13,6 +13,8 @@ class StatusCode(IntEnum):
         * RFC 2295: Transparent Content Negotiation in HTTP
         * RFC 2774: An HTTP Extension Framework
         * RFC 7540: Hypertext Transfer Protocol Version 2 (HTTP/2)
+        * RFC 2324: Hyper Text Coffee Pot Control Protocol (HTCPCP/1.0)
+        * RFC 7725: An HTTP Status Code to Report Legal Obstacles
     """
 
     def __new__(cls, value: int, phrase: str = "") -> "StatusCode":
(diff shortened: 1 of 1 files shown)


httpx/status_codes.py
```diff
@@ -101,6 +103,7 @@ def is_server_error(cls, value: int) -> bool:
     UNSUPPORTED_MEDIA_TYPE = 415, "Unsupported Media Type"
     REQUESTED_RANGE_NOT_SATISFIABLE = 416, "Requested Range Not Satisfiable"
     EXPECTATION_FAILED = 417, "Expectation Failed"
+    I_AM_A_TEAPOT = 418, "I'm a teapot"
```
sethmlarson: At the risk of being too picky here, could we change the enum value to be `IM_A_TEAPOT`? :)

StephenBrown2: I'm not opposed, but I did want to note that [requests uses both](https://github.com/kennethreitz/requests/blob/9a2e5df000691ad28613524ca7f80aa28a94a8d5/requests/status_codes.py#L72), and I took the constant name from the Symphony status constant on https://httpstatuses.com/418

PR #136, opened 2019-07-22
[WIP] Digest auth

yeraydiazdiaz:
This PR implements Digest authentication, mentioned it was pending work in the README. Tested against [httpbin.org's digest auth endpoint](https://httpbin.org/#/Auth/get_digest_auth__qop___user___passwd___algorithm_) and mimicking [request's implementation](https://github.com/kennethreitz/requests/blob/master/requests/auth.py#L108).

There's a design issue I'd like some feedback on before continuing.

### DIFF
--- a/httpx/auth.py
+++ b/httpx/auth.py
@@ -13,159 +13,16 @@ class AuthBase:
     Base class that all auth implementations derive from.
     """
 
-    def __call__(self, request: AsyncRequest) -> AsyncRequest:
-        raise NotImplementedError("Auth hooks must be callable.")  # pragma: nocover
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ) -> None:
+        self.username = username
+        self.password = password
 
 
 class HTTPBasicAuth(AuthBase):
-    def __init__(
-        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
-    ) -> None:
-        self.username = username
-        self.password = password
-
-    def __call__(self, request: AsyncRequest) -> AsyncRequest:
-        request.headers["Authorization"] = self.build_auth_header()
-        return request
-
-    def build_auth_header(self) -> str:
-        username, password = self.username, self.password
-
-        if isinstance(username, str):
-            username = username.encode("latin1")
-
-        if isinstance(password, str):
-            password = password.encode("latin1")
-
-        userpass = b":".join((username, password))
-        token = b64encode(userpass).decode().strip()
-        return f"Basic {token}"
+    pass
 
 
-class HTTPDigestAuth:
-
-    ALGORITHM_TO_HASH_FUNCTION = {
-        "MD5": hashlib.md5,
-        "SHA": hashlib.sha1,
-        "SHA-256": hashlib.sha256,
-        "SHA-512": hashlib.sha512,
-    }
-
-    def __init__(
-        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
-    ) -> None:
-        self.username = username
-        self.password = password
-
-    def __call__(self, request: AsyncRequest, response) -> AsyncRequest:
-        auth_header = self.build_auth_header(request, response)
-        if auth_header is not None:
-            # TODO: raise instead of returning None?
-            request.headers["Authorization"] = auth_header
-        return request
-
-    def build_auth_header(
-        self, request: AsyncRequest, response: AsyncResponse
-    ) -> typing.Optional[str]:
-        # Retrieve challenge from response header
-        header = response.headers.get("www-authenticate")
-        assert header.lower().startswith("digest")
-        challenge = self._parse_header(header)
-        algorithm = challenge.get("algorithm", "MD5")
-
-        realm = _safe_encode(challenge["realm"])
-        nonce = _safe_encode(challenge["nonce"])
-        qop = _safe_encode(challenge["qop"]) if "qop" in challenge else None
-        opaque = _safe_encode(challenge["opaque"]) if "opaque" in challenge else None
-        username = _safe_encode(self.username)
-        password = _safe_encode(self.password)
-
-        # Assemble parts depending on hash algorithms
-        hash_func = self.ALGORITHM_TO_HASH_FUNCTION[
-            algorithm.replace("-SESS", "").upper()
-        ]
-
-        def digest(data: bytes) -> bytes:
-            return hash_func(data).hexdigest().encode("utf-8")
-
-        def keyed_digest(secret, data) -> bytes:
-            return digest(b":".join((secret, data)))
-
-        A1 = b":".join((username, realm, password))
-        HA1 = digest(A1)
-
-        path = self._get_parsed_path(request.url)
-        A2 = b":".join((request.method.encode("utf-8"), path))
-        # TODO: implement auth-int
-        HA2 = digest(A2)
-
-        # Construct Authenticate header string
-        nonce_count = 1  # TODO: use contextvars to count properly
-        nc_value = b"%08x" % nonce_count
-
-        s = str(nonce_count).encode("utf-8")
-        s += nonce
-        s += time.ctime().encode("utf-8")
-        s += os.urandom(8)
-
-        cnonce = hashlib.sha1(s).hexdigest()[:16].encode("utf-8")
-        if algorithm.lower().endswith("-sess"):
-            A1 += b":".join((nonce, cnonce))
-
-        if algorithm == "MD5-SESS":
-            HA1 = digest(b":".join((HA1, nonce, cnonce)))
-
-        if not qop:
-            to_key_digest = [nonce, HA2]
-        elif qop == "auth" or "auth" in qop.decode().split(","):
-            to_key_digest = [nonce, nc_value, cnonce, b"auth", HA2]
-        else:
-            return None  # handle auth-int
-
-        format_args = {
-            "username": username,
-            "realm": realm,
-            "nonce": nonce,
-            "uri": path,
-            "response": keyed_digest(HA1, b":".join(to_key_digest)),
-        }
-        if opaque:
-            format_args["opaque"] = opaque
-        if algorithm:
-            format_args["algorithm"] = algorithm
-        if qop:
-            format_args["qop"] = "auth"
-            format_args["nc"] = nc_value
-            format_args["cnonce"] = cnonce
-
-        header_value = ", ".join(
-            [
-                '{}="{}"'.format(
-                    key, value if isinstance(value, str) else value.decode("utf-8")
-                )
-                for key, value in format_args.items()
-            ]
-        )
-        return "Digest " + header_value
-
-    def _parse_header(self, header):
-        result = {}
-        for item in header[7:].split(","):
-            key, value = item.strip().split("=")
-            value = value[1:-1] if value[0] == value[-1] == '"' else value
-            result[key] = value
-
-        return result
-
-    def _get_parsed_path(self, url):
-        path = url.path or "/"
-        if url.query:
-            path += "?" + url.query
-        return path.encode("utf-8")
-
-
-def _safe_encode(str_or_bytes) -> bytes:
-    return (
-        str_or_bytes.encode("utf-8") if isinstance(str_or_bytes, str) else str_or_bytes
-    )
-
+class HTTPDigestAuth(AuthBase):
+    pass
--- a/httpx/auth.py
+++ b/httpx/auth.py
@@ -1,7 +1,11 @@
+import hashlib
 import typing
+import time
+import os
 from base64 import b64encode
+from urllib.parse import urlparse
 
-from .models import AsyncRequest
+from .models import AsyncRequest, AsyncResponse
 
 
 class AuthBase:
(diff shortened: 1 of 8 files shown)


httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+            # TODO: raise instead of returning None?
+            request.headers["Authorization"] = auth_header
+        return request
+
+    def build_auth_header(
+        self, request: AsyncRequest, response: AsyncResponse
+    ) -> typing.Optional[str]:
+        # Retrieve challenge from response header
+        header = response.headers.get("www-authenticate")
+        assert header.lower().startswith("digest")
```
sethmlarson: We can't assume that "digest" is being offered as the first challenge: https://tools.ietf.org/html/rfc7235#section-4.1

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+        return result
+
+    def _get_parsed_path(self, url):
+        path = url.path or "/"
+        if url.query:
+            path += "?" + url.query
+        return path.encode("utf-8")
+
+
+def _safe_encode(str_or_bytes) -> bytes:
```
sethmlarson: We really need to add something like to to utils, we accept `typing.AnyStr` in so many locations.

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+            [
+                '{}="{}"'.format(
+                    key, value if isinstance(value, str) else value.decode("utf-8")
+                )
+                for key, value in format_args.items()
+            ]
+        )
+        return "Digest " + header_value
+
+    def _parse_header(self, header):
```
sethmlarson: There's gotta be some function in the `email` module for this?

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+        def digest(data: bytes) -> bytes:
+            return hash_func(data).hexdigest().encode("utf-8")
+
+        def keyed_digest(secret, data) -> bytes:
+            return digest(b":".join((secret, data)))
+
+        A1 = b":".join((username, realm, password))
+        HA1 = digest(A1)
+
+        path = self._get_parsed_path(request.url)
```
sethmlarson: Does the URL object have a property for this already? .target or something?

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+        realm = _safe_encode(challenge["realm"])
+        nonce = _safe_encode(challenge["nonce"])
+        qop = _safe_encode(challenge["qop"]) if "qop" in challenge else None
+        opaque = _safe_encode(challenge["opaque"]) if "opaque" in challenge else None
+        username = _safe_encode(self.username)
+        password = _safe_encode(self.password)
+
+        # Assemble parts depending on hash algorithms
+        hash_func = self.ALGORITHM_TO_HASH_FUNCTION[
+            algorithm.replace("-SESS", "").upper()
```
sethmlarson: Currently if you give MD5-SESS-SESS you'd actually not error out saying we don't support the digest algorithm. Maybe we should be explicit about which algorithms we support even it requires spelling them out exactly.

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+
+        realm = _safe_encode(challenge["realm"])
+        nonce = _safe_encode(challenge["nonce"])
+        qop = _safe_encode(challenge["qop"]) if "qop" in challenge else None
+        opaque = _safe_encode(challenge["opaque"]) if "opaque" in challenge else None
+        username = _safe_encode(self.username)
+        password = _safe_encode(self.password)
+
+        # Assemble parts depending on hash algorithms
+        hash_func = self.ALGORITHM_TO_HASH_FUNCTION[
```
sethmlarson: Should probably have an error emitted if we don't understand the algorithm rather than `KeyError`.

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+                    key, value if isinstance(value, str) else value.decode("utf-8")
+                )
+                for key, value in format_args.items()
+            ]
+        )
+        return "Digest " + header_value
+
+    def _parse_header(self, header):
+        result = {}
+        for item in header[7:].split(","):
```
sethmlarson: The `header[7:]` I don't understand, why is this being used?

httpx/client.py
```diff
@@ -176,6 +179,13 @@ def merge_cookies(
             response = await self.dispatch.send(
                 request, verify=verify, cert=cert, timeout=timeout
             )
+            # TODO: requests also checks for number of 401 calls < 2
+            if response.is_client_error and response.expects_digest_auth:
+                request = self.auth(request, response)  # TODO: remove dependency on self.auth
```
yeraydiazdiaz: We're currently following request's design of using a `BaseAuth` class and, for now, a single concrete implementation with a `__call__` method that injects the authorization header into the request. However this does not quite fit the Digest authentication flow where the authorization header depends on the `www-authenticate` header of the response, which breaks the abstraction.

To get around it temporarily I'm assuming the user creates an `HTTPDigestAuth` object manually [as described in the requests docs](https://2.python-requests.org//en/master/user/authentication/#digest-authentication) and storing it for later use in `self.auth` but ideally we'd redesign the abstraction so it takes the request and optionally a response.

Requests gets around it by using a [`register_hook` callback](https://github.com/kennethreitz/requests/blob/master/requests/auth.py#L292) in the `HTTPDigestAuth` which takes the response and does the tasks we're doing in the `client.send_handling_redirects`.

Thoughts?

httpx/client.py
```diff
@@ -176,6 +179,13 @@ def merge_cookies(
             response = await self.dispatch.send(
                 request, verify=verify, cert=cert, timeout=timeout
             )
+            # TODO: requests also checks for number of 401 calls < 2
+            if response.is_client_error and response.expects_digest_auth:
```
sethmlarson: We're going to need a better way to allow users to use this behavior branch more generally. Realm-based response-aware authentication class that only triggers when certain status codes / headers are available?

httpx/client.py
```diff
@@ -176,6 +179,13 @@ def merge_cookies(
             response = await self.dispatch.send(
                 request, verify=verify, cert=cert, timeout=timeout
             )
+            # TODO: requests also checks for number of 401 calls < 2
```
sethmlarson: Instead of reimplementing is_client_error on response, why not use response.status_code.is_client_error?

httpx/models.py
```diff
@@ -794,6 +794,15 @@ def decoder(self) -> Decoder:
     def is_redirect(self) -> bool:
         return StatusCode.is_redirect(self.status_code) and "location" in self.headers
 
+    @property
```
sethmlarson: I don't think we should have these properties on the response object.

httpx/client.py
```diff
@@ -132,9 +132,12 @@ def merge_cookies(
...
         if auth is not None:
-            if isinstance(auth, tuple):
-                auth = HTTPBasicAuth(username=auth[0], password=auth[1])
-            request = auth(request)
+            if isinstance(auth, HTTPDigestAuth):
+                self.auth = auth
+            else:
+                if isinstance(auth, tuple):
+                    auth = HTTPBasicAuth(username=auth[0], password=auth[1])
+                request = auth(request)
```
yeraydiazdiaz: Are we ok with allowing the users to pass instances of `HTTPBasicAuth` and `HTTPDigestAuth`?

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+        A1 = b":".join((username, realm, password))
+        HA1 = digest(A1)
+
+        path = self._get_parsed_path(request.url)
+        A2 = b":".join((request.method.encode("utf-8"), path))
+        # TODO: implement auth-int
+        HA2 = digest(A2)
+
+        # Construct Authenticate header string
+        nonce_count = 1  # TODO: use contextvars to count properly
```
yeraydiazdiaz: I'm not familiar at all with `contextvars` but from what I could gather from the docs the current context if copied when creating an asyncio Task which *should* match our requirements for this?

httpx/auth.py
```diff
@@ -36,3 +40,132 @@ def build_auth_header(self) -> str:
...
+
+        def keyed_digest(secret, data) -> bytes:
+            return digest(b":".join((secret, data)))
+
+        A1 = b":".join((username, realm, password))
+        HA1 = digest(A1)
+
+        path = self._get_parsed_path(request.url)
+        A2 = b":".join((request.method.encode("utf-8"), path))
+        # TODO: implement auth-int
```
yeraydiazdiaz: I believe [Requests does not currently implement `auth-int`](https://github.com/kennethreitz/requests/blob/master/requests/auth.py#L210). Do we want to?

sethmlarson:
Also feel free to override any of my comments @tomchristie, you're more familiar with how requests handles this than I am.

httpx/auth.py
yeraydiazdiaz: I see, thanks. 👍

httpx/auth.py
yeraydiazdiaz: `MD5-SESS-SESS` should not be a valid value but yeah, it's best to spell them all out. 👍

httpx/auth.py
yeraydiazdiaz: To remove the `Digest ` part of the header and leave only the parameters, this will change to adapt to having more than one challenge as per your comment above.

httpx/client.py
yeraydiazdiaz: Law of Demeter? Frankly I was simply following the existing `is_redirect`, I'm happy to remove it.

httpx/models.py
yeraydiazdiaz: Would you remove `is_redirect` as well? Maybe `is_client_error` is superfluous but seems like a valid case for redirects.

httpx/client.py
tomchristie: Yes.

httpx/auth.py
tomchristie: Treat it as out-of-scope for now.

httpx/auth.py
tomchristie: No - wouldn't want anything here that's coupled up to concurrency primitives.
Either an incrementing instance attribute, or an incrementing class attribute would be preferable. (I'm not sure which of those two is appropriate without digging into exactly how the nonce count is used)

httpx/client.py
tomchristie: Not sure on first sight, but this is a *really* useful point of reference wrt. what API we need to support. Not comfortable jumping on any particular answer quickly here, and it'd push me a bit towards thinking that we *might* want to rethink the auth API.

tomchristie: Also nudges me a bit more in the direction of thinking that DigestAuth should exist as a third party auth, rather than us building it in. (I don't mind us working through it with a PR, but we may well eventually want to consider pushing it a seperate package.)

httpx/auth.py
yeraydiazdiaz: Right, makes sense. [Requests uses thread locals extensively to hold state](https://github.com/not-kennethreitz/requests/blob/9a2e5df000691ad28613524ca7f80aa28a94a8d5/requests/auth.py#L117-L125) which I naively thought we should carry over.

httpx/models.py
tomchristie: Let's drop `expects_digest_auth` and `is_client_error`. I think they fall over the cutoff line.

httpx/auth.py
tomchristie: May well make sense. Obvs let's treat it as a small independent, nicely scoped PR if we do want to do that. Not clear to me immediately how many places we'd be using it ATM, but certainly could be valid.

httpx/client.py
yeraydiazdiaz: I tend to agree, the Digest flow is complex enough to grant its own package, in a similar vein to [`requests-oauthlib`](https://github.com/requests/requests-oauthlib).

I think I'll start a separate issue to discuss how to best approach it design-wise and close this PR.

tomchristie: Sure thing. Also very welcome to work through it against this pull request (as a useful place for discussion, and to validate test cases against it) and only start pulling it out into a package once you've got something sufficiently mature. I don't mind either ways.

yeraydiazdiaz:
I repurposed this PR to have a stab at refactoring redirects and auth into middlewares. Quick tests using httpbin seems to work for redirects, basic auth and digest auth.

The main idea is:
1. `client.send` sets up the pipeline
2. The request is processed by each middleware in turn
3. The resulting request is dispatched
4. The response is processed by each middleware, which can return a modified response or a new request
5. If a new request is returned go back to 2 otherwise return the response

The biggest problem is that `response.next` was assembled as a callable inside the loop, which by refactoring into the middleware makes it much harder so I temporarily left it broken.

Another problem I see is the manual assembling of the middlewares and how to expose an API for the user assemble new middleware pipelines and how much control do we allow.

This is all very experimental btw I completely fine with going in a different direction, just thought I'd try it out 🙂

sethmlarson:
The "middlewares" idea was something I was also thinking about in terms of extendability of the package. The 3rd party packages can distribute middlewares that can get registered on the client and interact with the requests and responses in the same way that WSGI / ASGI has middlewares to modifying the request and response before delivering to application code / user.

Lots of code to review and maybe this needs to be a separate review before we go ahead and use it to implement DigestAuth, but great work here. :) cc @tomchristie

yeraydiazdiaz:
Superseded by #305 305

PR #138, opened 2019-07-23
Add __truediv__ and __rtruediv__ functions to URL

StephenBrown2:
To allow pathlib-like concatenation: https://github.com/python/cpython/blob/master/Lib/pathlib.py#L909-L913

This will allow me to stop using https://pypi.org/project/urlpath/ in favor of httpx.URL, and thus consolidate my dependencies.

Added using the pre-existing `.join` method, and stole several tests from [urlpath](https://github.com/chrono-meter/urlpath/blob/master/test/test_url.py#L43) to verify functionality.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -214,6 +214,12 @@ class URL:
         url_str = str(self)
         return f"{class_name}({url_str!r})"
 
+    def __truediv__(self, right):
+        return self.join(URL(right, allow_relative=True))
+
+    def __rtruediv__(self, left):
+        return URL(left).join(self)
+
 
 class Origin:
     """
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -25,7 +25,7 @@ def test_url():
     assert new.scheme == "http"
 
 
-def test_url__params():
+def test_url_params():
     url = URL("https://example.org:123/path/to/somewhere", params={"a": "123"})
     assert str(url) == "https://example.org:123/path/to/somewhere?a=123"
 
@@ -42,3 +42,40 @@ def test_url_set():
     url_set = set(urls)
 
     assert all(url in urls for url in url_set)
+
+
+def test_url_div():
+    url = URL("http://www.example.com/path/to/file.ext?query#fragment")
+
+    assert str("http://example.org" / url) == str(url)
+    assert str("http://example.org/over/the/rainbow" / url) == str(url)
+
+    assert (
+        str(url / "https://secure.example.com/path")
+        == "https://secure.example.com/path"
+    )
+    assert str(url / "/changed/path") == "http://www.example.com/changed/path"
+    assert (
+        str(URL("http://example.com/base/") / "path/to/file")
+        == "http://example.com/base/path/to/file"
+    )
+    assert (
+        str(URL("http://example.com/no/trailing/slash/base") / "path/to/file")
+        == "http://example.com/no/trailing/slash/path/to/file"
+    )
+    assert (
+        str(
+            URL("http://example.com/path/?q")
+            / URL("http://localhost/app/?q")
+            / URL("to/content", allow_relative=True)
+        )
+        == "http://localhost/app/to/content"
+    )
+
+    assert (
+        str(URL("http://example.com/name/") / "\u65e5\u672c\u8a9e/\u540d\u524d")
+        == "http://example.com/name/%E6%97%A5%E6%9C%AC%E8%AA%9E/%E5%90%8D%E5%89%8D"
+    )
+
+    url = URL("s3://mybucket") / "some_folder/123_2017-10-30T18:43:11.csv.gz"
+    assert str(url) == "s3://mybucket/some_folder/123_2017-10-30T18:43:11.csv.gz"


sethmlarson:
I think I'm -1 on this, I can't see a use-case where `.join()` isn't easier to read.

tomchristie:
Seems fair, yup. It’s a nice pull request, but the syntax probably isn’t more obvious in my view either.

StephenBrown2:
Sad day. But, I realize now that it was incomplete for my purposes, and `.join()` has the same problem.

Anything I pass to `.join()` replaces the full path, rather than concatenating it as I would expect, unless the left path ends with a slash `/`.

```python
In [1]: from httpx import URL                                                                                                                                                                                                                                                                  

In [2]: class divURL(URL): 
   ...:     def __truediv__(self, right): 
   ...:         return self.join(URL(right, allow_relative=True)) 
   ...:  
   ...:     def __rtruediv__(self, left): 
   ...:         return URL(left).join(self) 
   ...:                                                                                                                                                                                                                                                                                        

In [3]: base = URL("http://example.com")                                                                                                                                                                                                                                                       

In [4]: divbase = divURL("http://example.com")                                                                                                                                                                                                                                                 

In [5]: base.join("root")                                                                                                                                                                                                                                                                      
Out[5]: URL('http://example.com/root')

In [6]: divbase / "root"                                                                                                                                                                                                                                                                       
Out[6]: URL('http://example.com/root')

In [7]: baseroot = base.join("root")                                                                                                                                                                                                                                                           

In [8]: baseroot.join("leaf")                                                                                                                                                                                                                                                                  
Out[8]: URL('http://example.com/leaf')

In [9]: divbase / "leaf"                                                                                                                                                                                                                                                                       
Out[9]: URL('http://example.com/leaf')

In [10]: from httpx import __version__                                                                                                                                                                                                                                                         

In [11]: __version__                                                                                                                                                                                                                                                                           
Out[11]: '0.6.7'
```

The docstring says it will "Return an absolute URL, using given this URL as the base.", but I wonder if that should be updated to say "...using this URL scheme and host as the base."? Or at least fleshed out to describe what `.components.resolve_with` does?

With the following modifications to the `__*div__` functions, it works as I expect:

```
    def __truediv__(self, right): 
        right = URL(right, allow_relative=True)
        if not self.path.endswith("/") and not right.path.startswith("/"):
            return self.copy_with(path=self.path + "/").join(right)
        return self.join(right)

    def __rtruediv__(self, left):
        left = URL(left)
        if not left.path.endswith("/") and not self.path.startswith("/"):
            return left.copy_with(path=self.path + "/").join(self)
        return left.join(self)
```
Usage:
```python
In [1]: from httpx import URL

In [2]: base = URL("http://example.org")

In [3]: baseroot = base / "root"

In [4]: assert baseroot == URL('http://example.org/root')

In [5]: baserootleaf = baseroot / "leaf"

In [6]: assert baserootleaf == URL('http://example.org/root/leaf')

In [7]: baserootnoleaf = baseroot / "/leaf"

In [8]: assert baserootnoleaf == URL('http://example.org/leaf')

In [9]: baserootslash = base / "root/"

In [10]: assert baserootslash == URL('http://example.org/root/')

In [11]: baserootslashleaf = baserootslash / "leaf"

In [12]: assert baserootslashleaf == URL('http://example.org/root/leaf')

In [13]: baserootslashnoleaf = baserootslash / "/leaf"

In [14]: assert baserootslashnoleaf == URL('http://example.org/leaf')

In [15]: basemultipleslashes = base / "root" / "leaf"

In [16]: assert basemultipleslashes == URL('http://example.org/root/leaf')
```

So I suppose I'll simply subclass `httpx.URL` with the div functions, as that works well enough.

sethmlarson:
Basically `.join()` is a "What would the new URL become if I received a redirect response". It's less about concatenation and more about following RFC 3986 Section 5.2.

tomchristie:
Turns out that's not quite what it does at the moment...

```python
>>> httpx.URL('http://www.example.com/some/directory/abc.html').join('../../foo.html')
URL('http://www.example.com/some/directory/foo.html')
```

> The docstring says it will "Return an absolute URL, using given this URL as the base.", but I wonder if that should be updated to say "...using this URL scheme and host as the base."?

Other way around. The docs are correct wrt. intent, the implementation is wrong.

Opened #142 to track this.

StephenBrown2:
Ah, that makes sense then. In that case, `.join()` might be better named `.redirect_to()` or [`.merge()`](https://tools.ietf.org/html/rfc3986#section-5.2.3) (per the spec)?
And the slash operator seems like it would be a good complement, as it is intended for path-joining, not necessarily merging, though the relative example @tomchristie 	gave would also work:

```python
>>> httpx.URL('http://www.example.com/some/directory/abc.html') / '../../foo.html'
URL('http://www.example.com/foo.html')
```

Side note: Reading through the RFC, I saw this: https://tools.ietf.org/html/rfc3986#section-6.2.4 which made sense to me and puzzled me as to why the path _had_ to have a trailing slash to concatenate correctly. Looks like some tests should be crafted from the rfc spec.

sethmlarson:
I was thinking `.redirect_to()` in my head as well, I think it'd be a more semantic choice. :D

tomchristie:
Join makes sense to me here, and is in line with `urljoin`.
Although we happen to only end up using it for redirects, it's a *general purpose* operation, that's needed in plenty of other contexts too.
It's the broken behavior that needs resolving rather than the naming, no?

sethmlarson:
I'm fine with `join()` or `redirect_*()`, just thinking this URL object is being used for HTTP not as a general-purpose URL library. Leaving it as `join()` is no issue.

tomchristie:
@StephenBrown2 #144 gets it how you'd expect...

```python
>>> httpx.URL('http://www.example.com/some/directory/abc.html').join('../../foo.html')
URL('http://www.example.com/foo.html')
```

StephenBrown2:
Yup, that looks like it'll do it. I'm glad to have contributed, even if it was just to inadvertently demonstrate a bug. :-)

PR #139, opened 2019-07-24
Allow string comparison for URL().__eq__

### LINKED ISSUE #113: API design question - `Response.url`
Currently our `Response.url` attribute exposes a `URL` instance.

This is a breaking change from the requests API where it just exposes a plain string.

It's feasible that we should instead only be exposing plain string URLs, in order to aim for drop-in replacement API compatibility w/ requests, *and* in order to keep the API surface area low.

Options here are:

* Expose `request.url` as a URL instance. (Richer information, URL class is also useful in its own right.)
* Expose `request.url` as a str. (Better API compat. Lower API surface area to maintain.)
* Expose `request.url` as a str, and `request.urlinfo` as a URL instance. (Better API compat. High API surface area.)

StephenBrown2:
Resolves #113

This turned out to be super easy, as casting a `str` to a `str` is idempotent, so adding `str` to the isinstance check allows it to pass equality checks with `str`s.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -204,7 +204,7 @@ class URL:
         return hash(str(self))
 
     def __eq__(self, other: typing.Any) -> bool:
-        return isinstance(other, URL) and str(self) == str(other)
+        return isinstance(other, (URL, str)) and str(self) == str(other)
 
     def __str__(self) -> str:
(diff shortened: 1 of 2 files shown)


tomchristie:
Great, yup! 👍

PR #143, opened 2019-07-25
Detect EOF signaling remote server closed connection

### LINKED ISSUE #96: h11._util.RemoteProtocolError: can't handle event type ConnectionClosed when role=SERVER and state=SEND_RESPONSE
I intermittently got this error when load testing uvicorn endpoint.

This error comes from a proxy endpoint where I am also using `encode/http3` to perform HTTP client calls.

```
  File "/project/venv/lib/python3.7/site-packages/http3/client.py", line 365, in post
    timeout=timeout,
  File "/project/venv/lib/python3.7/site-packages/http3/client.py", line 497, in request
    timeout=timeout,
  File "/project/venv/lib/python3.7/site-packages/http3/client.py", line 112, in send
  File "/project/venv/lib/python3.7/site-packages/http3/client.py", line 145, in send_handling_redirects
  File "/project/venv/lib/python3.7/site-packages/http3/dispatch/connection_pool.py", line 121, in send
    raise exc
  File "/project/venv/lib/python3.7/site-packages/http3/dispatch/connection_pool.py", line 116, in send
  File "/project/venv/lib/python3.7/site-packages/http3/dispatch/connection.py", line 59, in send
  File "/project/venv/lib/python3.7/site-packages/http3/dispatch/http11.py", line 65, in send
  File "/project/venv/lib/python3.7/site-packages/http3/dispatch/http11.py", line 109, in _receive_event
    event = self.h11_state.next_event()
  File "/project/venv/lib/python3.7/site-packages/h11/_connection.py", line 439, in next_event
    exc._reraise_as_remote_protocol_error()
  File "/project/venv/lib/python3.7/site-packages/h11/_util.py", line 72, in _reraise_as_remote_protocol_error
    raise self
  File "/project/venv/lib/python3.7/site-packages/h11/_connection.py", line 422, in next_event
    self._process_event(self.their_role, event)
  File "/project/venv/lib/python3.7/site-packages/h11/_connection.py", line 238, in _process_event
    self._cstate.process_event(role, type(event), server_switch_event)
  File "/project/venv/lib/python3.7/site-packages/h11/_state.py", line 238, in process_event
    self._fire_event_triggered_transitions(role, event_type)
  File "/project/venv/lib/python3.7/site-packages/h11/_state.py", line 253, in _fire_event_triggered_transitions
    .format(event_type.__name__, role, self.states[role]))
h11._util.RemoteProtocolError: can't handle event type ConnectionClosed when role=SERVER and state=SEND_RESPONSE
```

yeraydiazdiaz:
Fixes #96 

Raise ConnectionClosedByRemote and handle on `send`

### DIFF
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -108,6 +108,9 @@ class Reader(BaseReader):
 
         return data
 
+    def is_connection_dropped(self) -> bool:
+        return self.stream_reader.at_eof()
+
 
 class Writer(BaseWriter):
     def __init__(self, stream_writer: asyncio.StreamWriter, timeout: TimeoutConfig):
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -4,7 +4,7 @@ import h11
 
 from ..concurrency import TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
-from ..exceptions import NotConnected
+from ..exceptions import NotConnected, ConnectionClosedByRemote
 from ..interfaces import BaseReader, BaseWriter, ConcurrencyBackend
 from ..models import AsyncRequest, AsyncResponse
 
@@ -166,6 +166,9 @@ class HTTP11Connection:
                     )
                 except OSError:  # pragma: nocover
                     data = b""
+                if self.h11_state.their_state == h11.SEND_RESPONSE and data == b"":
+                    # the server closed a keep-alive connection
+                    raise ConnectionClosedByRemote
                 self.h11_state.receive_data(data)
             else:
                 break
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -4,7 +4,7 @@ import h11
 
 from ..concurrency import TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
-from ..exceptions import NotConnected, ConnectionClosedByRemote
+from ..exceptions import ConnectionClosedByRemote, NotConnected
 from ..interfaces import BaseReader, BaseWriter, ConcurrencyBackend
 from ..models import AsyncRequest, AsyncResponse
 
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -4,7 +4,6 @@ import h11
 
 from ..concurrency import TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
-from ..exceptions import NotConnected
 from ..interfaces import BaseReader, BaseWriter, ConcurrencyBackend
 from ..models import AsyncRequest, AsyncResponse
 
@@ -46,12 +45,7 @@ class HTTP11Connection:
     ) -> AsyncResponse:
         timeout = None if timeout is None else TimeoutConfig(timeout)
 
-        try:
-            await self._send_request(request, timeout)
-        except ConnectionResetError:  # pragma: nocover
-            # We're currently testing this case in HTTP/2.
-            # Really we should test it here too, but this'll do in the meantime.
-            raise NotConnected() from None
+        await self._send_request(request, timeout)
 
         task, args = self._send_request_data, [request.stream(), timeout]
         async with self.backend.background_manager(task, args=args):
@@ -117,8 +111,6 @@ class HTTP11Connection:
         Send a single `h11` event to the network, waiting for the data to
         drain before returning.
         """
-        if self.reader.is_connection_dropped():
-            raise NotConnected
         bytes_to_send = self.h11_state.send(event)
         await self.writer.write(bytes_to_send, timeout)
 
@@ -159,8 +151,6 @@ class HTTP11Connection:
         """
         Read a single `h11` event, reading more data from the network if needed.
         """
-        if self.reader.is_connection_dropped():
-            raise NotConnected
         while True:
             event = self.h11_state.next_event()
             if event is h11.NEED_DATA:
@@ -192,3 +182,6 @@ class HTTP11Connection:
     @property
     def is_closed(self) -> bool:
         return self.h11_state.our_state in (h11.CLOSED, h11.ERROR)
+
+    def is_connection_dropped(self) -> bool:
+        return self.reader.is_connection_dropped()
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -157,6 +157,8 @@ class HTTP11Connection:
         """
         Read a single `h11` event, reading more data from the network if needed.
         """
+        if self.reader.is_connection_dropped():
+            raise NotConnected
         while True:
             event = self.h11_state.next_event()
             if event is h11.NEED_DATA:
@@ -166,9 +168,6 @@ class HTTP11Connection:
                     )
                 except OSError:  # pragma: nocover
                     data = b""
-                if self.h11_state.their_state == h11.SEND_RESPONSE and data == b"":
-                    # the server closed a keep-alive connection
-                    raise NotConnected
                 self.h11_state.receive_data(data)
             else:
                 break
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -4,7 +4,7 @@ import h11
 
 from ..concurrency import TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
-from ..exceptions import ConnectionClosedByRemote, NotConnected
+from ..exceptions import NotConnected
 from ..interfaces import BaseReader, BaseWriter, ConcurrencyBackend
 from ..models import AsyncRequest, AsyncResponse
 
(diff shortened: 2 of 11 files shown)


httpx/dispatch/http11.py
```diff
@@ -166,6 +166,9 @@ def __init__(
                     )
                 except OSError:  # pragma: nocover
                     data = b""
+                if self.h11_state.their_state == h11.SEND_RESPONSE and data == b"":
+                    # the server closed a keep-alive connection
+                    raise ConnectionClosedByRemote
```
tomchristie: We can just use `NotConnected` here. It's there to handle the exact same case, it's just that we might *either* see it on sending data, or on receiving data, and we were only dealing with one of those two cases prior to this PR.

tomchristie: We'll need to think about the naming. The exception needs to be the subset of connection closes that occur *only* at the start of reading/writing a response.

There's also something worth digging into here, which isn't clear to me. Why isn't the case being triggered by `_send_request` in the first place before we get here? We're writing to the network, and calling drain, so I'd expect to see the exception already.

httpx/dispatch/http11.py
```diff
@@ -166,6 +166,9 @@ def __init__(
                     )
                 except OSError:  # pragma: nocover
                     data = b""
+                if self.h11_state.their_state == h11.SEND_RESPONSE and data == b"":
```
sethmlarson: Are we sure this is the right state? I thought that after a response is done the server would be in `h11.DONE`. cc: @njsmith @pgjones

tomchristie: Presumably a moot point given https://github.com/encode/httpx/issues/96#issuecomment-515018947

yeraydiazdiaz: FWIW it is in that state on all my local tests, which I believe is correct since we do feed the request event to the state machine according to the [server state machine diagram](https://h11.readthedocs.io/en/latest/_static/SERVER.svg).

sethmlarson: You're right, I was looking at the bottom of the tree for our state without remember that by this point we'd already issued a request. So we should be checking for connection alive-ness before we get to that state then?

yeraydiazdiaz: Yep, stay tuned 😄

httpx/concurrency.py
```diff
@@ -108,6 +108,9 @@ def __init__(
 
         return data
 
+    def is_connection_dropped(self) -> bool:
+        return self.stream_reader.at_eof()
```
yeraydiazdiaz: Probably should clarify this seemingly simplistic implementation. The `StreamReaderProtocol.connection_lost` callback [calls `stream_reader.feed_eof`](https://github.com/python/cpython/blob/master/Lib/asyncio/streams.py#L651) which in turn [sets the EOF flag](https://github.com/python/cpython/blob/master/Lib/asyncio/streams.py#L827) which `at_eof` returns.

httpx/dispatch/http11.py
```diff
@@ -117,6 +117,8 @@ def __init__(
         Send a single `h11` event to the network, waiting for the data to
         drain before returning.
         """
+        if self.reader.is_connection_dropped():
+            raise NotConnected
```
yeraydiazdiaz: Checking the connection status here also prevents us from sending the request event which, [as @tomchristie mentioned](https://github.com/encode/httpx/issues/96#issuecomment-515011313), makes things ambiguous for non-idempotent requests.

tests/dispatch/test_connection_pools.py
```diff
@@ -131,3 +131,35 @@
...
+async def test_connection_closed_by_server_before_receiving_data(server):
+    """
+    If the server closes the connection after sending the response but before reading the data
+    a disconnect error is raised.
+    """
+    async with httpx.ConnectionPool() as http:
+        response = await http.request("GET", "http://127.0.0.1:8000/")
+        await server.shutdown()  # shutdown the server to close the connection before receiving the data
+
+        with pytest.raises(httpx.exceptions.NotConnected):
```
yeraydiazdiaz: Is `NotConnected` the right exception here? Maybe we should wrap it into a more descriptive Exception and error message?

tomchristie:
Okay - this is looking great! Nice work.

Now, rather than using `NotConnected` we should instead expose `.is_connection_dropped()` on the `HTTP11Connection` and `HTTP2Connection` classes, and the general `HTTPConnection`. Then the connection pool should *check* it explicitly on acquiring a keep-alive connection. (Rather than attempting to send and catching `NotConnected` if it occurs.)

yeraydiazdiaz:
That sounds good for the two-request scenario, but wouldn't we still need an exception to cover to the last test case where the connection is lost before the response body is read?

tomchristie:
Not exactly.

We’ll remove `NotConnected` completely, and treat any stream exception that occurs once we start sending as a “disconnected while sending” case.

We *will* want to make sure we’ve got well defined exception cases for those events, but that doesn’t need to be part of this PR since it’s a different issue from the “keep alive connection had expired” case.

njsmith:
Yeah, the right logic is: only put idle connections in the pool immediately after a successful call to `start_next_cycle`. When taking an idle connection from the pool to reuse, first check if the underlying socket is readable (ignoring any TLS layer). Implementing this looks very different on different backends, but they can all do it one way or another. If the socket is readable, close it and discard it and try again; otherwise go ahead and use it.

It seems like y'all are still spending a lot of time reinventing stuff here :-/. Do you have an abstraction over backends yet, or is everything still using asyncio only?

sethmlarson:
Btw @njsmith you meant "is NOT readable" for your close and disregard condition?

njsmith:
No, I mean readable. Remember, we think the connection is supposed to be idle, so there shouldn't be any data to read. If it's marked readable, then that means the server is either sending some data in violation of the protocol, or is sending a 408 Request Timeout, or (most likely) has simply closed the connection. (`select` and friends mark a socket readable when the other side has closed it, which is a little weird, but does make sense – `recv` will return `b""`, but it won't block.)

But, we shouldn't actually try to read from the connection while doing this check, because if it's not readable, then we'll block forever, and you wouldn't expect taking an idle connection out of the pool to be a blocking operation that needs a timeout. So it's a weird special-case concept. Trio sockets actually have a special method just to handle this case: https://github.com/python-trio/trio/issues/760

tomchristie:
Let’s try not to railroad this PR with extraneous stuff.

There’s an isolated issue with the “is this connection still alive” on reaquiry of keep-alive connections, because I’d incorrectly assumed that “try writing to the socket” would be the robust way to check for that case. This PR resolves that, and now just needs a little extra rejigging to stop using an unneeded exception-based flow.

(Nothing to do with h11 state or anything else like that here)

Yup we have a concurrency backend interface, let’s discuss that against a separate issue.

njsmith:
> There’s an isolated issue with the “is this connection still alive” on reaquiry of keep-alive connections, because I’d incorrectly assumed that “try writing to the socket” would be the robust way to check for that case.

Right, the problem with using writing to check for connection liveness is that you can't tell whether the server started processing the request or not. So if a write fails, you have to give up and report that back to the user, and let them decide what they want to do about it. But if you can tell that the server closed the connection *before* you started writing, then you know it's safe to go ahead and open a new connection, so the client library can handle it transparently without getting the user involved.

I believe that the right way to check for readability on different backends is:

- On asyncio, using the asyncio `Stream` API: well, you actually can't do it using public APIs. There's `StreamReader.at_eof`, but that will fail if the HTTP server is polite and sends a 408 Request Timeout before it closes the connection. I think the best you can do is:

    ```python
    def asyncio_stream_is_readable(stream):
        return stream._buffer or stream._eof
    ```

    or else giving up on the `Stream` API and using protocols/transports directly.

- On trio, using the trio `Stream` API: this isn't exposed in Trio's stream abstraction, but we know that we don't have just any stream, but rather a `SocketStream`, possibly with one or more layers of `SSLStream` wrapped around it. So you can do:

    ```python
    def trio_stream_is_readable(stream):
        # Peek through any SSLStream wrappers to get the underlying SocketStream
        while hasattr(stream, "transport_stream"):
            stream = stream.transport_stream
        assert isinstance(stream, trio.SocketStream)
        return stream.socket.is_readable()
    ```

    (This requires the next release of trio, which will be out soon.)

- On raw sockets: If you only care about recent Python 3, then you can do something like: https://github.com/python-trio/trio/blob/a9bb9343487c2c64234267d66e04beeccafe2974/trio/_socket.py#L482

  If you need to be fully portable, then things are more complicated: https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/wait.py

  (But you need most of that anyway to do I/O on raw sockets anyway.)

(Does asyncio provide any way to get at the raw socket underneath a `StreamReader`? If it does then that might let you do something that doesn't use private APIs, though it'll be pretty convoluted.)

tomchristie:
Closing in favor of #145 which adds a last little bit of clean up on top of @yeraydiazdiaz's great work here.

PR #144, opened 2019-07-25
Fix `URL.join(...)`

### LINKED ISSUE #142: URL.join behaviour is incorrect.
Our `URL.join` behavior currently isn't as expected. It should be in line with rfc3986#section-5.2, and Python's existing `urljoin`.

It ain't...

```python
>>> httpx.URL('http://www.example.com/some/directory/abc.html').join('../../foo.html')
URL('http://www.example.com/some/directory/foo.html')
```

tomchristie:
We were calling `.normalize` on relative URLs, which was dropping out relative leading path portions. (So eg. `../foo` was being treated as `/foo`.)

Closes #142
Refs #138

### DIFF
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -31,12 +31,26 @@ def test_url_eq_str():
     assert str(url) == url
 
 
-def test_url__params():
+def test_url_params():
     url = URL("https://example.org:123/path/to/somewhere", params={"a": "123"})
     assert str(url) == "https://example.org:123/path/to/somewhere?a=123"
 
     url = URL("https://example.org:123/path/to/somewhere?b=456", params={"a": "123"})
     assert str(url) == "https://example.org:123/path/to/somewhere?a=123"
+
+
+def test_url_join():
+    url = URL("https://example.org:123/path/to/somewhere")
+    assert url.join('/somewhere-else') == "https://example.org:123/somewhere-else"
+
+    url = URL("https://example.org:123/path/to/somewhere")
+    assert url.join('somewhere-else') == "https://example.org:123/path/to/somewhere-else"
+
+    url = URL("https://example.org:123/path/to/somewhere")
+    assert url.join('../somewhere-else') == "https://example.org:123/path/somewhere-else"
+
+    url = URL("https://example.org:123/path/to/somewhere")
+    assert url.join('../../somewhere-else') == "https://example.org:123/somewhere-else"
 
 
 def test_url_set():
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -1,4 +1,6 @@
 from httpx import URL
+from httpx.exceptions import InvalidURL
+import pytest
 
 
 def test_idna_url():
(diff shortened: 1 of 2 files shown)


tomchristie:
Missing a final commit that fixes `.is_absolute` to `.is_absolute_url`.

I'm assuming GitHub's just lagging here, since it appears pushed to me.

tests/models/test_url.py
```diff
@@ -31,14 +31,28 @@ def test_url_eq_str():
...
-def test_url__params():
+def test_url_params():
     url = URL("https://example.org:123/path/to/somewhere", params={"a": "123"})
     assert str(url) == "https://example.org:123/path/to/somewhere?a=123"
 
     url = URL("https://example.org:123/path/to/somewhere?b=456", params={"a": "123"})
     assert str(url) == "https://example.org:123/path/to/somewhere?a=123"
 
 
+def test_url_join():
```
sethmlarson: Can we also get some tests with other components besides path? (ie `.join("?a=1") -> "https://example.org:123?a=1"` and `.join("#fragment") -> "https://example.org:123#fragment"`)

What about a few `..` and `.` cases as well? (`.join("../../../../../abc") -> "https://example.org/abc")

Also since the base URL is the same for all the test cases would it make sense to only define `url = ...` once?

tomchristie: 👍

StephenBrown2: Taking verbatim the cases from the rfc and replacing `a` with `example.com`, the following tests pass with this PR:

```python

def test_url_join_rfc3986():
    url = URL("http://example.com/b/c/d;p?q")

    with pytest.raises(InvalidURL):
        assert url.join("g:h") == "g:h"

    assert url.join("g") == "http://example.com/b/c/g"
    assert url.join("./g") == "http://example.com/b/c/g"
    assert url.join("g/") == "http://example.com/b/c/g/"
    assert url.join("/g") == "http://example.com/g"
    assert url.join("//g") == "http://g"
    assert url.join("?y") == "http://example.com/b/c/d;p?y"
    assert url.join("g?y") == "http://example.com/b/c/g?y"
    assert url.join("#s") == "http://example.com/b/c/d;p?q#s"
    assert url.join("g#s") == "http://example.com/b/c/g#s"
    assert url.join("g?y#s") == "http://example.com/b/c/g?y#s"
    assert url.join(";x") == "http://example.com/b/c/;x"
    assert url.join("g;x") == "http://example.com/b/c/g;x"
    assert url.join("g;x?y#s") == "http://example.com/b/c/g;x?y#s"
    assert url.join("") == "http://example.com/b/c/d;p?q"
    assert url.join(".") == "http://example.com/b/c/"
    assert url.join("./") == "http://example.com/b/c/"
    assert url.join("..") == "http://example.com/b/"
    assert url.join("../") == "http://example.com/b/"
    assert url.join("../g") == "http://example.com/b/g"
    assert url.join("../..") == "http://example.com/"
    assert url.join("../../") == "http://example.com/"
    assert url.join("../../g") == "http://example.com/g"

    assert url.join("../../../g") == "http://example.com/g"
    assert url.join("../../../../g") == "http://example.com/g"

    assert url.join("/./g") == "http://example.com/g"
    assert url.join("/../g") == "http://example.com/g"
    assert url.join("g.") == "http://example.com/b/c/g."
    assert url.join(".g") == "http://example.com/b/c/.g"
    assert url.join("g..") == "http://example.com/b/c/g.."
    assert url.join("..g") == "http://example.com/b/c/..g"

    assert url.join("./../g") == "http://example.com/b/g"
    assert url.join("./g/.") == "http://example.com/b/c/g/"
    assert url.join("g/./h") == "http://example.com/b/c/g/h"
    assert url.join("g/../h") == "http://example.com/b/c/h"
    assert url.join("g;x=1/./y") == "http://example.com/b/c/g;x=1/y"
    assert url.join("g;x=1/../y") == "http://example.com/b/c/y"

    assert url.join("g?y/./x") == "http://example.com/b/c/g?y/./x"
    assert url.join("g?y/../x") == "http://example.com/b/c/g?y/../x"
    assert url.join("g#s/./x") == "http://example.com/b/c/g#s/./x"
    assert url.join("g#s/../x") == "http://example.com/b/c/g#s/../x"
```

tomchristie: That's amazing - thanks @StephenBrown2!
Pulled that in now.

sethmlarson:
Btw @tomchristie can we add this to our `travis.yml` so we aren't pointlessly building branches on the central repo:

```yaml
branches:
  only:
  - master
```

obv doesn't matter much now since the build is so short but in the future it may. Do we rely on builds occurring on non-master that aren't PR builds?

tomchristie:
> can we add this to our travis.yml so we aren't pointlessly building branches on the central repo

Sorry I'm being dumb - exactly which builds would that exclude?

sethmlarson:
We wouldn't get double builds for branches from the encode/httpx repo that are also a PR. Check the commit statuses on your commits in this PR you'll see .../pr and .../push.

PR #145, opened 2019-07-25
Check disconnections on connection reacquiry

tomchristie:
Check for connection aliveness on re-acquiry from connection pool.

Superseeds #143

Great work on this @yeraydiazdiaz! There was a last little bit of extra cleanup on the connection pool handling that we can do now, which it made sense for me to pick up. I think we're good to go on this one now.

### DIFF
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -108,6 +108,9 @@ class Reader(BaseReader):
 
         return data
 
+    def is_connection_dropped(self) -> bool:
+        return self.stream_reader.at_eof()
+
 
 class Writer(BaseWriter):
     def __init__(self, stream_writer: asyncio.StreamWriter, timeout: TimeoutConfig):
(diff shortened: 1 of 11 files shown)


tomchristie:
I've pulled in a last little bit of extra work on top of this in #145 which I think is now good to go.

One particularly interesting point from @njsmith, here https://github.com/encode/httpx/pull/143#issuecomment-515202982

> There's StreamReader.at_eof, but that will fail if the HTTP server is polite and sends a 408 Request Timeout before it closes the connection.

Okay, that's an *interesting* case. 😑

I'd suggest we start by getting this change in as it stands. We can the confirm if we need to finesse the specific `is_connection_dropped()` implementation details as a separate improvement on top of this.

sethmlarson:
Should we open a separate issue to handle the 408 case?

tomchristie:
Yeah, defo. On it.

tomchristie:
Released as 0.6.8

Issue #124, opened 2019-07-15
Try connecting to HTTPS on schemaless URLs


sethmlarson:
urllib3's behavior with schemaless URLs was to connect via `http`. Maybe we can have some logic that if we have `verify=True` active and we receive a schemaless URL either from a user or a redirect we try `https` first before falling back on `http`? Would have to be implemented within the connection retry logic.

sethmlarson:
Relevant discussion: https://github.com/python-http/discussions/issues/4

sethmlarson:
I'm gonna close this one out now that HSTS preloading is enabled.

PR #151, opened 2019-07-26
Switch hosts in HSTS preload list to HTTPS

sethmlarson:
This does the switch on the URL level as a part of normalization. Adds a new dependency that is under my control currently and potentially will be under the python-http org.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -8,6 +8,7 @@ from http.cookiejar import Cookie, CookieJar
 from urllib.parse import parse_qsl, urlencode
 
 import chardet
+import hstspreload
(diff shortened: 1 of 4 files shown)


sethmlarson:
This is related to #124.

sethmlarson:
I'm gonna merge this, if we decide it's not right we can remove it.

tomchristie:
I'm not completely sure we want to do this at the model level.

It might be that we'd prefer to handle this in the client, rather than at the point of instantiating a `URL` instance itself.

tomchristie:
Nice bit of functionality tho!

PR #153, opened 2019-07-26
Fix bug in calculating maximum frame size

IsaacBenghiat:
When calculating the chunk size for frames, now considers max_outbound_frame_size in addition in order to get an accurate number for the amount of data that can be in a single frame.
Documentation: https://python-hyper.org/projects/h2/en/stable/api.html#h2.connection.H2Connection.local_flow_control_window


sethmlarson:
There's definitely some weirdness happening here. Even with your change I'm still able to get an h2 error to raise when sending >16K bytes.

IsaacBenghiat:
It looks like there are some other issues with sending large posts over http2. For the most part they involve failing to end the stream. I will work on fixing them.

sethmlarson:
Yeah we'd really appreciate you investigating more, thanks! :)

IsaacBenghiat:
I made changes to support and manage the flow control. I also moved the calculation of the chunk size to reflect that the flow control window can change. All of the tests in test_http2.py are currently working on my machine.

blockchainunchained:
Any chance of getting this merged? It fixes an issue I was having but I also need this fix https://github.com/encode/httpx/pull/230 for another issue.

tomchristie:
@blockchainunchained - It looks like there are currently branch conflicts against this PR that aren't trivially obvious to me on first sight. I'm also unclear if the tests pass against this cleanly or not. You're welcome to open a new pull request incorperating this work if you'd like to try to progress the issue, and get it to a mergeable state.

IsaacBenghiat:
I have merged the changes from master into the branch. I don't know what error is occurring during the tests, could someone help clarify what is going wrong?

tomchristie:
The tests are running clean now - it's just a code style failure.
You'll need to run `black` to reformat the code for a consistent style.

Folks, I think this needs to be added to https://www.encode.io/httpx/contributing/ right?

httpx/dispatch/http2.py
```diff
@@ -42,6 +44,7 @@ def __init__(
 
         self.events[stream_id] = []
         self.timeout_flags[stream_id] = TimeoutFlag()
+        self.window_update_received[stream_id] = asyncio.Event()
```
yeraydiazdiaz: I believe this be `self.backend.create_event`, is that right @florimondmanca?

florimondmanca: Correct! We should use methods on the backend instead of calling asyncio operations directly. :)

sethmlarson:
I added a few commits to fix CI @IsaacBenghiat :) This should be ready for review from everyone.

httpx/dispatch/http2.py
```diff
@@ -108,18 +110,19 @@ def initiate_connection(self) -> None:
...
-        chunk_size = min(len(data), flow_control)
-        for idx in range(0, len(data), chunk_size):
-            chunk = data[idx : idx + chunk_size]
-
-            logger.debug(
-                f"send_data stream_id={stream_id} data=Data(<{len(chunk)} bytes>)"
+        while data:
+            flow_control = self.h2_state.local_flow_control_window(stream_id)
+            chunk_size = min(
+                len(data), flow_control, self.h2_state.max_outbound_frame_size - 1
```
nsidnev: I'm not sure that here we need to reduce the value of `max_outbound_frame_size` by one, because according to the `h2` documentation, this is the [maximum value](https://python-hyper.org/projects/h2/en/stable/api.html#h2.connection.H2Connection.max_outbound_frame_size) that can be sent and it [should not](https://python-hyper.org/projects/h2/en/stable/api.html#h2.connection.H2Connection.send_data) cause an error in the form in which it is given to us from `h2` state machine.

httpx/dispatch/http2.py
```diff
@@ -158,6 +159,17 @@ def initiate_connection(self) -> None:
             if isinstance(event, h2.events.DataReceived):
                 self.h2_state.acknowledge_received_data(len(event.data), stream_id)
```
nsidnev: I think that we should to replace `len(event.data)` with `event.flow_controlled_length`, according to [this](https://python-hyper.org/projects/h2/en/stable/advanced-usage.html#working-with-flow-control) part of the h2 documentation, and I also don’t quite understand why the code below with manual control of the window size is required, since the `acknowledge_received_data` should do it by itself as written [here](https://python-hyper.org/projects/h2/en/stable/advanced-usage.html#auto-flow-control).

httpx/dispatch/http2.py
```diff
@@ -173,6 +185,13 @@ def initiate_connection(self) -> None:
                 logger.debug(
                     f"receive_event stream_id={event_stream_id} event={event!r}"
                 )
+                if isinstance(event, h2.events.WindowUpdated):
+                    if event_stream_id == 0:
+                        for window_update_event in self.window_update_received.values():
+                            window_update_event.set()
+                    else:
+                        self.window_update_received[event_stream_id].set()
```
nsidnev: I think that `KeyError` can be raised here, if HTTP/2 on the server will not work correctly. Perhaps we should check that `event_stream_id` value is in the `window_update_received` dictionary.

sethmlarson:
@nsidnev Thanks for making me remember to submit my review and not just make the comments unsubmitted :P Couple of our comments are similar.

httpx/dispatch/http2.py
IsaacBenghiat: In testing an error was thrown when chunk size was `self.h2_state.max_outbound_frame_size` (which appeared to always be 2^14 = 16384) which said that the limit on frame size was the 16383. Adding in the -1 fixed the issue, and I assumed that somewhere the maximum value was being interpreted as strictly greater than as opposed to greater than or equal to.

nsidnev: I think that this error was caused by unhandled `WindowsUpdated` event and error in splitting data into chunks. In `h2` this [checked](https://github.com/python-hyper/hyper-h2/blob/385ced7d30a7ccb9b3f163f1bc4e90cf2f4ee184/h2/connection.py#L850) as larger then, so I don't think that there is a need for reducing value.

httpx/dispatch/http2.py
IsaacBenghiat: Implemented this in commit.

httpx/dispatch/http2.py
IsaacBenghiat: Handled this error in new commit.

httpx/dispatch/http2.py
```diff
@@ -139,16 +142,14 @@ def initiate_connection(self) -> None:
...
-
-        status_code = 200
-        headers = []
-        for k, v in event.headers:
-            if k == b":status":
-                status_code = int(v.decode("ascii", errors="ignore"))
-            elif not k.startswith(b":"):
-                headers.append((k, v))
-        return status_code, headers
+                status_code = 200
```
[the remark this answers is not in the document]
IsaacBenghiat: I made this change because it helped me understand what the code was doing, it is functionally the same as the code before the change. I am fine with whichever version you prefer.

httpx/dispatch/http2.py
```diff
@@ -139,25 +142,27 @@ def initiate_connection(self) -> None:
...
         while True:
             event = await self.receive_event(stream_id, timeout)
             if isinstance(event, h2.events.DataReceived):
-                self.h2_state.acknowledge_received_data(len(event.data), stream_id)
+                self.h2_state.acknowledge_received_data(
+                    event.flow_controlled_length, stream_id
+                )
                 yield event.data
+                if self.h2_state.streams[stream_id].open:
+                    await self.stream.write(self.h2_state.data_to_send(), timeout)
```
nsidnev: Why do we still need these 2 lines? As I see it `.receive_event` will process all write operations inside itself, which means that message about increasing the size of the window will be included in the sending payload. So if the stream is open, then the data will definitely be sent there or even earlier. Am I understanding that properly?

IsaacBenghiat: You are right, these two lines are redundant. I'll remove them.

httpx/dispatch/http2.py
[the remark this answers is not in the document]
IsaacBenghiat: Resolved on other request for changes.

httpx/dispatch/http2.py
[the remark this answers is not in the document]
IsaacBenghiat: Resolved on other request for changes.

IsaacBenghiat: Removed -1 in new commit.

httpx/dispatch/http2.py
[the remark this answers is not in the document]
sethmlarson: I think I prefer the earlier version because the exit point isn't in the `while` loop. I'm going to make the change and then this PR looks good to me. :)

tests/dispatch/test_http2.py
```diff
@@ -65,6 +79,20 @@ def test_http2_post_request():
     }
 
 
+async def test_http2_large_async_post_request(backend):
```
florimondmanca: Can we group these two tests together, like other sync/async pairs? Also the convention in this file seems to be test_async_... for the async test cases. :)

httpx/dispatch/http2.py
```diff
@@ -108,18 +110,19 @@ def initiate_connection(self) -> None:
...
+        while data:
+            flow_control = self.h2_state.local_flow_control_window(stream_id)
+            chunk_size = min(
+                len(data), flow_control, self.h2_state.max_outbound_frame_size
             )
-
-            self.h2_state.send_data(stream_id, chunk)
-            data_to_send = self.h2_state.data_to_send()
-            await self.stream.write(data_to_send, timeout)
+            if chunk_size == 0:
```
florimondmanca: Just a comprehension note — I’m wondering why this condition indicates that we should wait for the window to be updated? Is it happening when the max outbound limit is 0? Or when we receive no data? Maybe a link to an appropriate section of the HTTP/2 standard could be useful here, as well as a general comment under “while data:” to explain what the window is, why and when it can change and why we even need to handle it.

tests/dispatch/test_http2.py
IsaacBenghiat: Changed order and name in new commit.

httpx/dispatch/http2.py
IsaacBenghiat: Added some documentation to the code in new commit.

sethmlarson:
Uvloop is failing to compile for Python 3.8. Gonna merge this anyways! :) Thanks @IsaacBenghiat this is an amazing first contribution!

florimondmanca:
Was going to comment the same @sethmlarson. Definitely related: https://github.com/MagicStack/uvloop/issues/266

Issue #127, opened 2019-07-18
Disable TLSv1 and TLSv1.1 by default


sethmlarson:
TLSv1.2 was released over 10 years ago and is widely deployed.

> The Alexa Top 1 Million Analysis from February 2018 shows that for the sites surveyed, the vast majority support TLSv1.2 (98.9 percent), with a mere 0.8 percent using TLSv1.0 and an even smaller percentage using TLSv1.1.

Quote from [IETF Draft on Deprecating TLSv1 and TLSv1.1 from 2018](https://tools.ietf.org/id/draft-moriarty-tls-oldversions-diediedie-00.html)

We may be able to tighten up the ciphers we're enabling by default as well as a result of this change.

halbow:
I can take a look at this !

sethmlarson:
Great! Reach out if you need any help or have questions. :)

PR #155, opened 2019-07-27
Removed deprecated TLS versions

halbow:
So I just added the two deprecated version to the SSL context, As we support python 3.6 this should be good but this way of building the sslContext seems to be deprecated in 3.7

https://docs.python.org/3/library/ssl.html#ssl.OP_NO_TLSv1
https://docs.python.org/3/library/ssl.html#ssl.SSLContext.minimum_version


sethmlarson:
We'll have to look at using the new `minimum_version` interface. We can accomplish that in a separate PR as it'll take additional checks to make sure we get it right (since it's only available for Python 3.7+ and OpenSSL 1.1.0g+)

sethmlarson:
Closes #127.

PR #156, opened 2019-07-27
Added Support for both brotli libraries

halbow:
I tested separately with each libraries and the unit-tests were passing !

### DIFF
--- a/httpx/decoders.py
+++ b/httpx/decoders.py
@@ -100,7 +100,7 @@ class BrotliDecoder(Decoder):
         try:
             if hasattr(self.decompressor, "decompress"):
                 return self.decompressor.decompress(data)
-            return self.decompressor.process(data)
+            return self.decompressor.process(data)  # pragma: nocover
         except brotli.error as exc:
             raise DecodingError from exc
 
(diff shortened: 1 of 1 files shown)


sethmlarson:
Probably failing due to coverage. This is one of the use-cases for using nox so we can test these alternate combinations easier. :) For now you can mark the lines as `nocover`

httpx/decoders.py
```diff
@@ -85,6 +85,9 @@ class BrotliDecoder(Decoder):
     Handle 'brotli' decoding.
 
     Requires `pip install brotlipy`. See: https://brotlipy.readthedocs.io/
+        or   `pip install brotli`. See https://github.com/google/brotli
+    Supports both 'brotlipy' and 'Brotli' packages since they share an import
+    name. The top branches are for 'brotlipy' and bottom branches for 'Brotli'
```
yeraydiazdiaz: GitHub won't let me add the comment in the right place, but in line 96 below there's an `assert` that you may want to amend.

halbow: Good catch, thanks !

PR #157, opened 2019-07-27
added flake8 plugins.

cansarigol:
I've made a few little changes that I think will be useful.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -434,7 +434,7 @@ class Headers(typing.MutableMapping[str, str]):
         set_value = value.encode(self.encoding)
 
         found_indexes = []
-        for idx, (item_key, item_value) in enumerate(self._list):
+        for idx, (item_key, _) in enumerate(self._list):
             if item_key == set_key:
                 found_indexes.append(idx)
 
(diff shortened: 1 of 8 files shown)


httpx/models.py
```diff
@@ -514,12 +514,12 @@ def encode_data(
         self, data: dict = None, files: RequestFiles = None, json: typing.Any = None
     ) -> typing.Tuple[bytes, str]:
         if json is not None:
-            content = jsonlib.dumps(json).encode("utf-8")
+            content = jsonlib.dumps(json).encode()
```
sethmlarson: Let's revert the all changes related to utf-8. I think what we were doing before is correct.

cansarigol: sure I reverted it, is the reason "explicit is better than implicit"?

sethmlarson: Calling `.encode()` without a parameter isn't the same as with utf-8. We definitely want to be able to encode international characters to utf-8 in these scenarios.

cansarigol: Ah sorry, I didn't know. I changed because of default encode parameter is 'utf-8'. 
https://docs.python.org/3.6/library/stdtypes.html#str.encode

PR #159, opened 2019-07-27
Add headers to BaseClient

### LINKED ISSUE #149: Set Headers like Cookies in Client session for all requests
Assuming one can use `Client`/`AsyncClient` as a drop-in replacement for `requests.Session`, I would like to be able to set headers (and cookies) to be sent for each request, for example for token-based authentication.

With `requests`, I can do:
```python
header_dict = {"X-Auth-Token": "FooBarBazToken"}
session = requests.Session()
session.headers.update(header_dict)
```

`cookies` can already be set on the `httpx.Client`. I would like to be able to also do that with `headers`:
```python
>>> headers = Headers({"X-Auth-Token": "FooBarBazToken", "User-Agent": "python-myclient/0.2.1"})
>>> session = Client(base_url="https://httpbin.org", headers=headers)

>>> get_headers = session.get("/headers")

>>> get_headers.json()
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Cookie': 'MyCookie=MyValue',
  'Host': 'httpbin.org',
  'User-Agent': 'python-myclient/0.2.1',
  'X-Auth-Token': 'FooBarBazToken'}}
```

StephenBrown2:
This PR resolves #149, by adding `headers` as an arg to the `BaseClient` class, as with `cookies`, and also adds a `merge_headers` function to... merge the headers. I started by copying `merge_cookies`, but got a `NoneType is not iterable` error when trying to use it, so I changed the conditional to ensure that both the passed `headers` and `self.headers` were set, and returning one or the other in the case that they are not. This will still return `None` if neither are set.

I haven't added test cases, yet, but I did test against httpbin.org. I could check out adding something based on the existing [cookie tests](https://github.com/encode/httpx/blob/master/tests/client/test_cookies.py), but wanted to get some feedback first.

```
Python 3.7.3 (default, Jun 24 2019, 04:54:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from httpx import Headers, Client, AsyncClient                                                                                                                                                               

In [2]: headers = Headers({"X-Auth-Token": "FooBarBazToken", "Fake-Agent": "python-myclient/0.2.1"}) 
   ...: session = Client(base_url="https://httpbin.org", headers=headers)                                                                                                                                        

In [3]: async_session = AsyncClient(base_url="https://httpbin.org", headers=headers)                                                                                                                                 

In [4]: sync_headers = session.get("/headers")                                                                                                                                                                       

In [5]: async_headers = await async_session.get("/headers")                                                                                                                                                          

In [6]: sync_headers.json()                                                                                                                                                                                          
Out[6]: 
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Fake-Agent': 'python-myclient/0.2.1',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.6.8',
  'X-Auth-Token': 'FooBarBazToken'}}

In [7]: async_headers.json()                                                                                                                                                                                         
Out[7]: 
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Fake-Agent': 'python-myclient/0.2.1',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.6.8',
  'X-Auth-Token': 'FooBarBazToken'}}

In [8]: headers = Headers({"X-Auth-Token": "FooBarBazToken", "User-Agent": "python-myclient/0.2.1"}) 
   ...: session_two = Client(base_url="https://httpbin.org", headers=headers)                                                                                                                                    

In [9]: async_session_two = AsyncClient(base_url="https://httpbin.org", headers=headers)                                                                                                                             

In [10]: sync_headers_two = session.get("/headers")                                                                                                                                                                  

In [11]: async_headers_two = await async_session.get("/headers")                                                                                                                                                     

In [12]: sync_headers_two.json()                                                                                                                                                                                     
Out[12]: 
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Fake-Agent': 'python-myclient/0.2.1',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.6.8',
  'X-Auth-Token': 'FooBarBazToken'}}

In [13]: async_headers_two.json()                                                                                                                                                                                    
Out[13]: 
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Fake-Agent': 'python-myclient/0.2.1',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.6.8',
  'X-Auth-Token': 'FooBarBazToken'}}

In [14]: async_session.headers.update({"Another-Header": "IsAThing"})                                                                                                                                                

In [14]: async_headers_three = await async_session.get("/headers")                                                                                                                                                   

In [15]: async_headers_three.json()                                                                                                                                                                                  
Out[15]: 
{'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Another-Header': 'IsAThing',
  'Fake-Agent': 'python-myclient/0.2.1',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.6.8',
  'X-Auth-Token': 'FooBarBazToken'}}

```

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -49,6 +49,7 @@ class BaseClient:
     def __init__(
         self,
         auth: AuthTypes = None,
+        headers: HeaderTypes = None,
         cookies: CookieTypes = None,
         verify: VerifyTypes = True,
         cert: CertTypes = None,
@@ -95,6 +96,7 @@ class BaseClient:
             self.base_url = URL(base_url)
 
         self.auth = auth
+        self.headers = Headers(headers)
         self.cookies = Cookies(cookies)
         self.max_redirects = max_redirects
         self.dispatch = async_dispatch
@@ -108,6 +110,15 @@ class BaseClient:
             merged_cookies.update(cookies)
             return merged_cookies
         return cookies
+
+    def merge_headers(
+        self, headers: HeaderTypes = None
+    ) -> typing.Optional[HeaderTypes]:
+        if headers and self.headers:
+            merged_headers = Headers(self.headers)
+            merged_headers.update(headers)
+            return merged_headers
+        return headers or self.headers
 
     async def send(
         self,
@@ -527,6 +538,7 @@ class AsyncClient(BaseClient):
         timeout: TimeoutTypes = None,
     ) -> AsyncResponse:
         url = self.base_url.join(url)
+        headers = self.merge_headers(headers)
         cookies = self.merge_cookies(cookies)
         request = AsyncRequest(
             method,
@@ -608,6 +620,7 @@ class Client(BaseClient):
         timeout: TimeoutTypes = None,
     ) -> Response:
         url = self.base_url.join(url)
+        headers = self.merge_headers(headers)
         cookies = self.merge_cookies(cookies)
         request = AsyncRequest(
             method,
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -114,11 +114,11 @@ class BaseClient:
     def merge_headers(
         self, headers: HeaderTypes = None
     ) -> typing.Optional[HeaderTypes]:
-        if headers and self.headers:
+        if headers or self.headers:
             merged_headers = Headers(self.headers)
             merged_headers.update(headers)
             return merged_headers
-        return headers or self.headers
+        return headers
 
     async def send(
         self,
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -318,7 +318,7 @@ class QueryParams(typing.Mapping[str, str]):
         return f"{class_name}({query_string!r})"
 
 
-class Headers(typing.MutableMapping[str, str]):
+class Headers(typing.MutableMapping[str, str], MutableMapping):
     """
     HTTP headers, as a case-insensitive multi-dict.
     """
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -417,8 +417,8 @@ class Headers(typing.MutableMapping[str, str]):
 
     def update(self, headers: HeaderTypes = None) -> None:  # type: ignore
         headers = Headers(headers)
-        for header, value in headers.items():
-            self.headers[header] = value
+        for header in headers:
+            self[header] = headers[header]
 
     def __getitem__(self, key: str) -> str:
         """
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -318,7 +318,7 @@ class QueryParams(typing.Mapping[str, str]):
         return f"{class_name}({query_string!r})"
 
 
-class Headers(typing.MutableMapping[str, str], MutableMapping):
+class Headers(typing.MutableMapping[str, str]):
     """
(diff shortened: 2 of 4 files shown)


httpx/client.py
```diff
@@ -109,6 +111,15 @@ def merge_cookies(
...
         return cookies
 
+    def merge_headers(
+        self, headers: HeaderTypes = None
+    ) -> typing.Optional[HeaderTypes]:
+        if headers and self.headers:
+            merged_headers = Headers(self.headers)
+            merged_headers.update(headers)
+            return merged_headers
+        return headers or self.headers
```
yeraydiazdiaz: The reason you're getting the `NoneType is not iterable error` is because `Headers` does not implement the `update` method, [unlike `Cookies`](https://github.com/encode/httpx/blob/master/httpx/models.py#L1110-L1113), and falls back to the `MutableDict` implementation which does not handle the `None` case. Maybe it's worth implement it if only for consistency?

httpx/models.py
```diff
@@ -309,7 +309,7 @@ def __repr__(self) -> str:
         return f"{class_name}({query_string!r})"
 
 
-class Headers(typing.MutableMapping[str, str]):
+class Headers(typing.MutableMapping[str, str], MutableMapping):
```
yeraydiazdiaz: Why is this needed?

httpx/client.py
StephenBrown2: Sounds good, I've implemented `update` to mimic Cookies now, and it appears to work just the same. :+1:

httpx/models.py
StephenBrown2: I had added it thinking it would inherit the update function, but it turns out that's not how it worked. Removed it now.

StephenBrown2:
Definitely, I've added the param to the api spec, but the docs could use a larger fleshing out/update most likely.

sethmlarson:
Yeah that'll be a pre-1.0 release probably.

sethmlarson:
Looks like CI is failing a few tests, take a look at those.

StephenBrown2:
Good catch, forgot what object I was in, fixed now.

StephenBrown2:
Sure thing. Do we also want tests for cookies sent from the client since it's a similar use case?

sethmlarson:
I added a single test-case that verifies the way that conflicting headers are resolved so we maintain that property of how they've resolved. :)

sethmlarson:
Thank you @StephenBrown2 for this great change! :)

tomchristie:
Great stuff! Related to this - I reckon we'll probably want a `BaseClient.build_request()` method as a public interface.

PR #161, opened 2019-07-27
Update IDNA encoding to 2008 spec

### LINKED ISSUE #150: Use IDNA 2008 instead of IDNA 2003
Using `str.encode("idna")` uses IDNA 2003 which isn't recommended for modern use. We should be using IDNA 2008 provided by the `idna` module (which is a dependency of `httpx` but I don't think we're using it anywhere?)

StephenBrown2:
Resolves #150 and adds tests based on deviations from 2003 spec: https://unicode.org/reports/tr46/#Deviations

This also enables Unicode® Technical Standard #46 to normalize capital letters and such, and a test based on the example given in the [module documentation](https://github.com/kjd/idna#compatibility-mapping-uts-46).

Since the idna encode() function chokes on colons, I've split and rejoined the authority string and added to the tests to make sure the port stays as expected.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -8,6 +8,7 @@ from http.cookiejar import Cookie, CookieJar
 from urllib.parse import parse_qsl, urlencode
 
 import chardet
+import idna
 import hstspreload
 import rfc3986
 
@@ -94,7 +95,11 @@ class URL:
 
         # Handle IDNA domain names.
         if self.components.authority:
-            idna_authority = self.components.authority.encode("idna").decode("ascii")
+            # idna.encode raises InvalidCodepoint when encountering a colon, so split
+            # host and port in case the latter is specified and rejoin after encoding
+            host_port = self.components.authority.split(":")
+            host_port[0] = idna.encode(host_port[0]).decode("ascii")
+            idna_authority = ":".join(host_port)
             if idna_authority != self.components.authority:
                 self.components = self.components.copy_with(authority=idna_authority)
 
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -98,7 +98,7 @@ class URL:
             # idna.encode raises InvalidCodepoint when encountering a colon, so split
             # host and port in case the latter is specified and rejoin after encoding
             host_port = self.components.authority.split(":")
-            host_port[0] = idna.encode(host_port[0]).decode("ascii")
+            host_port[0] = idna.encode(host_port[0], uts46=True).decode("ascii")
             idna_authority = ":".join(host_port)
             if idna_authority != self.components.authority:
                 self.components = self.components.copy_with(authority=idna_authority)
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -8,7 +8,6 @@ from http.cookiejar import Cookie, CookieJar
 from urllib.parse import parse_qsl, urlencode
 
 import chardet
-import idna
 import hstspreload
 import rfc3986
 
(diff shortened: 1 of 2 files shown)


tests/models/test_url.py
```diff
@@ -1,12 +1,41 @@
...
+    assert url.port == 443
+
+    url = URL("http://ශ්‍රී.com:444")
+    assert url == URL("http://xn--10cl1a0b660p.com:444") # IDNA 2008
+    assert url.host == "xn--10cl1a0b660p.com"
+    assert url.scheme == "http"
+    assert url.port == 444
+
+    url = URL("https://نامه‌ای.com:4433")
+    assert url == URL("https://xn--mgba3gch31f060k.com:4433") # IDNA 2008
```
yeraydiazdiaz: Nitpick: I would drop the `IDNA 2008` comments 🙂

httpx/models.py
```diff
@@ -93,7 +94,11 @@ def __init__(
 
         # Handle IDNA domain names.
         if self.components.authority:
-            idna_authority = self.components.authority.encode("idna").decode("ascii")
+            # idna.encode raises InvalidCodepoint when encountering a colon, so split
+            # host and port in case the latter is specified and rejoin after encoding
+            host_port = self.components.authority.split(":")
+            host_port[0] = idna.encode(host_port[0], uts46=True).decode("ascii")
```
yeraydiazdiaz: Reference on the UTS 46 flag https://github.com/kjd/idna#compatibility-mapping-uts-46, maybe we should include it in a comment?

httpx/models.py
```diff
@@ -93,7 +94,11 @@ def __init__(
 
         # Handle IDNA domain names.
         if self.components.authority:
-            idna_authority = self.components.authority.encode("idna").decode("ascii")
+            # idna.encode raises InvalidCodepoint when encountering a colon, so split
+            # host and port in case the latter is specified and rejoin after encoding
+            host_port = self.components.authority.split(":")
```
sethmlarson: I think there's a way we can rely on rfc3986 to hand us the host portion of a URI. The current implementation would choke on IPv6 addresses.

httpx/models.py
sethmlarson: Let's drop the uts46 flag in favor of strict=True and std3_rules=True per RFC 3986

tests/models/test_url.py
StephenBrown2: Certainly. I think I will also parameterize the test to get better ids.

httpx/models.py
StephenBrown2: This is true, I had also assumed that authority would always be a fqdn/hostname. Getting just the host portion to start with would be best.

httpx/models.py
StephenBrown2: Do we want to fail on things like capitalization? I feel we don't necessarily want to be strict, but too much assistance can also be bad.

e.g. This will fail:
```
idna.core.InvalidCodepoint: Codepoint U+004B at position 1 of 'Königsgäßchen' not allowed
```

StephenBrown2:
Looks like rfc3986.IRIReference will in fact do exactly what we want, including handling colons in the authority. I've rebased on master to resolve the conflict with hsts... and switched to using `rfc3986.api.iri_reference(url).encode()` which will handle the IDNA Domain and return a URLReference object as expected.

Ref: convenience function - https://github.com/python-hyper/rfc3986/blob/df21c4b09078580c1cb10c728dc18c3111fcf603/src/rfc3986/api.py#L41
Relevant bit of IRIReference.encode() - https://github.com/python-hyper/rfc3986/blob/df21c4b09078580c1cb10c728dc18c3111fcf603/src/rfc3986/iri.py#L97-L129

httpx/models.py
StephenBrown2: Ended up using rfc3986.IRIReference.encode() instead and bypassing this entirely.

httpx/models.py
```diff
@@ -87,17 +87,14 @@ def __init__(
     ) -> None:
         if isinstance(url, rfc3986.uri.URIReference):
             self.components = url
+        elif isinstance(url, rfc3986.iri.IRIReference):
```
tomchristie: I don't think we should be exposing extra types as part of this pull request.

(In fact I don't think we should be exposing `rfc3986.uri.URIReference` as part of our public interface in the first place - it'd be much better if we could keep that as an internal implementation detail, in order to allow us to completely switch implementations if needed.)

tomchristie:
I guess we *might* also choose to switch our dependency listing to `rfc3986[idna]`, instead of listing the two packages seperately?

That'd stop devs from having to do a "wait where are you actually using the `idna` package?" double take.

httpx/models.py
tomchristie: See #170

tomchristie:
Yas - good stuff!

PR #162, opened 2019-07-28
Added base class HttpxError with request/response attribute

### LINKED ISSUE #141: Add request and response properties to Exceptions
In order to gracefully handle different HttpErrors I would like to know the url and status code of the offending call. Currently I'm wrapping the httpx `raise_for_status` code so this is added. However I think this would be a good addition to the core HttpError exception.

So an HttpError would consist of:
* message
* status code
* url

A drawback is that these additional properties may grow unwieldy as more exception handling cases are discovered. But I think some basic call information is called for.

halbow:
I created a HttpxError as base class for other Httpx exception. And added request for exception happening during the "send" (e.g. ConnectTimeout, ReadTimeout) and response for raise_for_status and errors linked to "too many redirect" (which should be close to the requests API). 

Things that may need attention:
- As the request/response object is separated from the IO, I catch HttpxError during sending in BaseClient and re-raise with the extra request info
- For the tests I added an extra check in methods test_raise_for_status but didn't find tests for redirect or tests mocking network failures, is it by design or still to be added ? 

Closes #141

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -22,6 +22,7 @@ from .exceptions import (
     RedirectBodyUnavailable,
     RedirectLoop,
     TooManyRedirects,
+    HTTPError,
 )
 from .interfaces import AsyncDispatcher, ConcurrencyBackend, Dispatcher
 from .models import (
@@ -78,13 +79,13 @@ class BaseClient:
                 )
 
         if dispatch is None:
-            async_dispatch = ConnectionPool(
+            async_dispatch: AsyncDispatcher = ConnectionPool(
                 verify=verify,
                 cert=cert,
                 timeout=timeout,
                 pool_limits=pool_limits,
                 backend=backend,
-            )  # type: AsyncDispatcher
+            )
         elif isinstance(dispatch, Dispatcher):
             async_dispatch = ThreadedDispatcher(dispatch, backend)
         else:
@@ -147,13 +148,18 @@ class BaseClient:
                 auth = HTTPBasicAuth(username=auth[0], password=auth[1])
             request = auth(request)
 
-        response = await self.send_handling_redirects(
-            request,
-            verify=verify,
-            cert=cert,
-            timeout=timeout,
-            allow_redirects=allow_redirects,
-        )
+        try:
+            response = await self.send_handling_redirects(
+                request,
+                verify=verify,
+                cert=cert,
+                timeout=timeout,
+                allow_redirects=allow_redirects,
+            )
+        except HTTPError as exc:
+            # Add the original request to any HTTPError
+            exc.request = request
+            raise
 
         if not stream:
             try:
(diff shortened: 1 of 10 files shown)


StephenBrown2:
Just as a github note, your referenced "Closes" number is incorrect, it should be #141, not #123

sethmlarson:
I fixed your "Closes X" in your original post @halbow :)

httpx/exceptions.py
```diff
@@ -1,7 +1,20 @@
+class HttpxError(Exception):
```
tomchristie: I'd probably prefer `HTTPError` here.

httpx/exceptions.py
```diff
@@ -1,7 +1,20 @@
...
+    """
+    Base class for Httpx exception
+    """
+
+    def __init__(self, *args, **kwargs):
```
tomchristie: Could we use `request: BaseRequest = None, response: BaseResponse = None` here?
Would be better to have the tighter checking if possible.

httpx/client.py
```diff
@@ -169,19 +170,24 @@ def merge_cookies(
...
+                raise TooManyRedirects(response=history[-1])
             if request.url in [response.url for response in history]:
-                raise RedirectLoop()
+                raise RedirectLoop(response=history[-1])
+
+            try:
+                response = await self.dispatch.send(
+                    request, verify=verify, cert=cert, timeout=timeout
+                )
+            except HttpxError as exc:
```
tomchristie: I think it might be better to include this in the `AsyncClient.send()` method instead, wrapping the `response = await self.send_handling_redirects(...)` call.

httpx/exceptions.py
halbow: I'll probably leave the *args as it allows to pass message to the exception. And I think we have circular import if we try to import BaseRequest and BaseResponse in exceptions, I'll leave them untyped for now if it's OK with you

halbow: Actually there's an entry in the doc for this kind of issue: https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles I'll take a look this evening to fix this.

tomchristie: Ah interesting yeah - that's worth a shot.

httpx/client.py
```diff
@@ -147,13 +148,17 @@ def merge_headers(
...
+            response = await self.send_handling_redirects(
+                request,
+                verify=verify,
+                cert=cert,
+                timeout=timeout,
+                allow_redirects=allow_redirects,
+            )
+        except HTTPError as exc:
+            # Add the original request to any HTTPError
+            raise type(exc)(request=request) from None
```
sethmlarson: Instead of doing this hack why not do:

```python
except HTTPError as exc:
    exc.request = request
    raise
```

halbow: Indeed in this case it seems (much !) simpler that way, I'll update it

PR #163, opened 2019-07-28
changed url validation with rfc3986.validators

cansarigol:


tomchristie:
Thanks for the contribution - certainly a neat idea.
On balance tho I think this is less readable than what we've got at the moment.

PR #167, opened 2019-07-28
Fixed multipart header params encoding

### LINKED ISSUE #164: non-ASCII characters filename in multipart/form-data
I tried to send a file with russian letters in the file name, and `httpx` escaped all non-ASCII characters (as I understood it the way specified in [RFC 7578](https://tools.ietf.org/html/rfc7578)). But this is different from `requests` [behavior](https://github.com/psf/requests/blob/589a82256759018a7e5e289302898dae32544949/requests/models.py#L110), where the file name is simply written as is (using [`RequestField.make_multipart`](https://github.com/urllib3/urllib3/blob/f0d9ebc41e51c4c4c9990b1eed02d297fd1b20d8/src/urllib3/fields.py#L248) from `urllib3`). Maybe `httpx` should just change [`FileField.render_headers`](https://github.com/encode/httpx/blob/3ba2e8c328f05b07ee06ae40f99dce7fe1e8292c/httpx/multipart.py#L57), so as not to escape the file name and instead just put it as bytes?

```diff
def render_headers(self) -> bytes:
-    name = quote(self.name, encoding="utf-8").encode("ascii")
-    filename = quote(self.filename, encoding="utf-8").encode("ascii")
-    content_type = self.content_type.encode("ascii")
     return b"".join(
         [
             b'Content-Disposition: form-data; name="',
-            name,
+            self.name.encode(),
             b'"; filename="',
-            filename,
+            self.filename.encode(),
             b'"\r\n',
             b"Content-Type: ",
-            content_type,
+            self.content_type.encode(),
             b"\r\n",
        ]
    )
```

nsidnev:
This resolves #164. Fixed using non-ASCII letters in `multipart/form-data` request.

### DIFF
--- a/httpx/multipart.py
+++ b/httpx/multipart.py
@@ -1,9 +1,39 @@
 import binascii
 import mimetypes
 import os
+import re
 import typing
 from io import BytesIO
-from urllib.parse import quote
+
+_HTML5_REPLACEMENTS = {'"': "%22", "\\": "\\\\"}
+
+_HTML5_REPLACEMENTS.update(
+    {chr(cc): "%{:02X}".format(cc) for cc in range(0x00, 0x1F + 1) if cc not in (0x1B,)}
+)
+
+
+def _replace_multiple(
+    value: str, needles_and_replacements: typing.Dict[str, str]
+) -> str:
+    def replacer(match: typing.Match[str]) -> str:
+        return needles_and_replacements[match.group(0)]
+
+    pattern = re.compile(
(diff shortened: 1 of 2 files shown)


sethmlarson:
Thanks for creating this PR @nsidnev! I've made some changes just moving code around so there's less layers of indirection. This PR still needs some tests before it's ready to ship. :)

sethmlarson:
Thanks @nsidnev!!!

Issue #152, opened 2019-07-26
Ignore `None` Headers


StephenBrown2:
I wanted to set an optional Header, so sometimes the value is `None`. What I expected was for the header to be omitted from the list when the value is `None`. Instead, I got:
```
AttributeError: 'NoneType' object has no attribute 'encode'
```
`if v is not None` could be added [here](https://github.com/encode/httpx/blob/master/httpx/models.py#L325) and [here](https://github.com/encode/httpx/blob/master/httpx/models.py#L330) to filter the headers on instantiation. [`__setitem__`](https://github.com/encode/httpx/blob/master/httpx/models.py#L428) should probably also be updated to `del` the item if the passed `value` is `None`.

sethmlarson:
What is the use-case for this that can't be accomplished via `headers.pop("name", None)` to remove headers if they exist?

Do any other HTTP libraries have this behavior? urllib3 I know doesn't.

StephenBrown2:
I suppose I could filter the dict before passing it to Headers, but my use was like:
```py
def function(param_one: str, optional_header: Optional[str] = None)
    headers = Headers(
        {
            "X-Auth": param_one,
            "X-Auth-Token": optional_header
        }
    )
    httpx.get(url, headers=headers)
```
Instead I _could_ do:
```py
def function(param_one: str, optional_header: Optional[str] = None)
    header_dict = {
        "X-Auth": param_one,
        "X-Auth-Token": optional_header
    }
    headers = Headers(
        {k: v for k, v in header_dict if v is not None}
    )
    httpx.get(url, headers=headers)
```

but that doesn't seem as clean to me. I hadn't seen it elsewhere, it was just surprising behavior when I tried it.

sethmlarson:
What about?

```python
def function(..., optional_header: Optional[str] = None):
    headers = {
        ...
    }
    if optional_header is not None:
        headers["X-Auth-Token"] = optional_header
    httpx.get(url, headers=headers)
```

?

StephenBrown2:
Yeah, there are several ways to do it. I'd probably make a utils function for myself [like](https://stackoverflow.com/a/54817412):

```py
from collections.abc import Mapping

def filter_none(d):
	""" Generic filter function to recursively remove keys with values of `None`.
	Ref: https://stackoverflow.com/a/54817412
	"""
    if isinstance(d, Mapping):
        return {k: filter_none(v) for k, v in d.items() if v is not None}
    else:
        return d
```

and then:

```py
[...]

Issue #122, opened 2019-07-15
TypeErrors against public interfaces.


tomchristie:
Partly prompted by [this comment](https://github.com/encode/http3/issues/119#issuecomment-511312157).

I think we should *probably* be a bit strict about enforcing types passed to our public interfaces. It'd be nice to raise loud clear errors, rather than failing with some arbitrary AttributeError in the middle of the codebase if a user passes something completely invalid to the interface. This would be stricter than what most Python libraries do, but I think it might be a good trade-off for us, to have a very strict and clearly demarked API.

That'd be easy enough to do since there's not too many points we'd need to enforce.
The big obvious one would be checking everything in `AsyncClient.request()`, which most of the rest of the public API calls into.

The would be some other bits & pieces to check too eg. Enforce that streaming data iterates through bytes/str, and not any other types. Enforce that dispatchers really do return a response type. Type checking on the various models that we expose.

Perhaps I'm being over-zelaous here? Any thoughts?

sethmlarson:
I'd rather be strict about what we accept, then we don't have to handle odd-ball cases down the road as a point of "backwards-compatibility" because it used to work but not on purpose.

tomchristie:
Sounds good to me.

halbow:
I could take a look at this with ! :) I'll probably start with Client.request and AsynClient.request, do you have an idea/thoughts on the implementation ? (I was thinking that a decorator checking func.__annotations__ could maybe do the trick)

sethmlarson:
Yeah start with a small example that Tom and I can take a look at for implementation review before we start applying it to all our interfaces. Are there any examples of this in another library?

tomchristie:
> I was thinking that a decorator checking

Indeed - I reckon a `@typechecked` decoartor approach is worth trying out here.
We can keep it faily limited - it only needs to deal with the cases that we want to use it with.

Wrt. implementation I'd suggest starting out by only handling checking of actual types, and treating any `typing.Union`/`typing.Optional`/`typing.Dict` cases all as `Any`.

Once you've got it working with basic types you could then look at expanding it out to deal with the more complex composite cases.

tomchristie:
We might want to have this at some point, but closing it as out-of-scope for now.

PR #169, opened 2019-07-29
Require values for headers

StephenBrown2:
Not sure if this is too literal an interpretation of "require values within the `headers` parameter" from https://github.com/encode/httpx/issues/152#issuecomment-515816059, but thought it might be useful to enforce.

Error message is just something I came up with.

Feel free to close this as not useful, if we'll simply rely on the `AttributeError: 'NoneType' object has no attribute 'encode'` that's currently raised three lines down.

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -15,6 +15,8 @@ def normalize_header_value(value: typing.AnyStr, encoding: str = None) -> bytes:
     """
     Coerce str/bytes into a strictly byte-wise HTTP header value.
     """
+    if value is None:
+        raise ValueError("NoneType header values are not allowed")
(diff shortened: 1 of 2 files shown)


cansarigol:
Hi, would it be helpful to add it to [this part](https://github.com/encode/httpx/blob/fb17459335b7e703f1895c4186a4fe421a54d815/httpx/multipart.py#L114)?

tomchristie:
Yes to the intent of this but I think we should close this in favor of a more general approach to #122 

I'd like to see us end up with a `@typechecked` decorator that we can simply apply to our public interfaces, and have that enforce strict type checking based on our annotations.

PR #170, opened 2019-07-29
Remove 'URIReference' from public interfaces

tomchristie:
Drops `rfc3986.uri.URIReference` from being referenced in any public interface.
In fact it *wasn't* actually included in `URLTypes` in the first place.

Our `copy_with` and `join` will be less efficient - but if we want to resolve that we should do it in a different way, rather than exposing implementation-level types in our public interface.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -86,24 +86,24 @@ class URL:
         params: QueryParamTypes = None,
     ) -> None:
         if isinstance(url, str):
-            self.components = rfc3986.api.uri_reference(url)
+            self._uri_reference = rfc3986.api.uri_reference(url)
         else:
-            self.components = url.components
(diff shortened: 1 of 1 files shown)


tomchristie:
We might also want to rethink the `components` attribute here - that's leaking implementation details at the moment.

A tighter API here would be for `components` to instead be `_uri_reference`.

We *could* also still expose a `components` interface, but as a plain named-tuple.

tomchristie:
Right, I've also dropped the `url.components` interface here. We don't need it as a public interface, since we've got by-name access anyways. If we *did* want to expose a tuple-like interface too, than let's do that as a plain tuple, rather than exposing `rfc3986` as part of our interface.

Let's keep a really tight API surface area wherever possible.

PR #171, opened 2019-07-29
added jwt token for basic authentication

cansarigol:
I usually use jwt token in my projects. I thought it would be good to be as built-in. I hope :)


tomchristie:
I think we'd *probably?* want to push for alternate auth implementations as third party packages (tho we could certainly link to them prominently from the docs.)

tomchristie:
Feasible that it could make it in at some point in the future, but let's try to keep things tightly scoped at this point in time.

PR #172, opened 2019-07-29
added control that auth instance has to be BasicAuthBase

cansarigol:


cansarigol:
Hi @tomchristie , maybe with `BasicAuthBase`, third party auth implementation would be easier.

httpx/client.py
```diff
@@ -145,6 +145,8 @@ def merge_headers(
         if auth is not None:
             if isinstance(auth, tuple):
                 auth = HTTPBasicAuth(username=auth[0], password=auth[1])
+            elif not isinstance(auth, BasicAuthBase):
```
tomchristie: This check isn't correct here - any `AuthBase` subclass is valid, not just `BasicAuthBase`

tomchristie:
Heya!

I think we'll pass on this.

It's better not to introduce additional differing subclasses here since it just complicates things, as we then have to document two alternate ways of writing an auth implementation.

PR #175, opened 2019-07-30
Tweak query param types

### LINKED ISSUE #154: Change QueryParams(key=None) to not emit a parameter
This one I know is [`requests` behaviour](https://github.com/psf/requests/blob/9a2e5df000691ad28613524ca7f80aa28a94a8d5/requests/models.py#L101):

Take a dict:
```py
>>> from httpx import QueryParams
>>> params = {"thing_one": 1, "thing_two": [1,2,3], "thing_three": None}
>>> QueryParams(params)
QueryParams('thing_one=1&thing_two=%5B1%2C+2%2C+3%5D&thing_three=None')
```
Expected:
```py
QueryParams('thing_one=1&thing_two=%5B1%2C+2%2C+3%5D')
```
Adding `if v is not None` [here](https://github.com/encode/httpx/blob/master/httpx/models.py#L263-L264) will do it.

tomchristie:
Allow primitive data types for query params.

Closes #154

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -20,6 +20,21 @@ def normalize_header_value(value: typing.AnyStr, encoding: str = None) -> bytes:
     return value.encode(encoding or "ascii")
 
 
+def str_query_param(value: typing.Union[str, int, float, bool, type(None)]) -> str:
+    """
+    Coerce a primitive data type into a string value for query params.
+
+    Note that we prefer JSON-style 'true'/'false' for boolean values here.
+    """
+    if value is True:
+        return "true"
+    elif value is False:
+        return "false"
+    elif value is None:
+        return ""
+    return str(value)
+
+
 def is_known_encoding(encoding: str) -> bool:
     """
     Return `True` if `encoding` is a known codec.
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -42,8 +42,8 @@ URLTypes = typing.Union["URL", str]
 
 QueryParamTypes = typing.Union[
     "QueryParams",
-    typing.Mapping[PrimitiveData, PrimitiveData],
-    typing.List[typing.Tuple[PrimitiveData, PrimitiveData]],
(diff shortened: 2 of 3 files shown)


tomchristie:
Later follow ups:

* Allow `typing.Union[PrimitiveData, typing.List[PrimitiveData]]` for eg. `{a: [1, 2, 3]}` => `a=1&a=2&a=3`
* Use same style for urlencoded and multipart data.

tomchristie:
Thanks for the look-over, @pquentin!

sethmlarson:
Sorry for not getting my review in on this! This implementation can't emit a query key without a value (ie: `?q`, different than an empty value `?q=`)

tomchristie:
> This implementation can't emit a query key without a value (ie: ?q, different than an empty value ?q=

Yup. I’m very okay with that - I don’t think we should distinguish between the two cases, although I’d be happy with confirming which of the two styles browser form inputs end up with. If users need that use case for something esoteric, then they can use explicit string query strings.

httpx/utils.py
```diff
@@ -20,6 +20,21 @@ def normalize_header_value(value: typing.AnyStr, encoding: str = None) -> bytes:
     return value.encode(encoding or "ascii")
 
 
+def str_query_param(value: typing.Union[str, int, float, bool, type(None)]) -> str:
```
cansarigol: are there any obstacles to using PrimitiveData? for consistency

tomchristie: Yes and no. We'd need some tweaking to avoid a cyclical dependency.

StephenBrown2:
Following up on [my comment last week](https://github.com/encode/httpx/issues/154#issuecomment-515709242), turns out the API I'm implementing a client for dies if the optional param is present whether or not the key has an `=` after it, so roundtripping is the only sane way for me. :point_up: looks good to me though!

PR #176, opened 2019-07-30
[WIP] Middlewares

yeraydiazdiaz:
Split from #136 [as requested](https://github.com/encode/httpx/pull/136#issuecomment-516207173). Also from that PR with some changes:

A stab at refactoring redirects and auth into middlewares. Quick tests using httpbin seems to work for redirects and basic auth, a digest auth is also working but not included in this PR.

The main idea is:
1. `client.send` sets up the pipeline
2. The request is processed by each middleware in turn
3. The resulting request is dispatched
4. The response is processed by each middleware, which can return a modified response or a new request
5. If a new request is returned go back to 2 otherwise return the response

I've inlined some comments in the problematic areas but main ones are:
- `response.next` was assembled as a callable inside the loop, which by refactoring into the middleware makes it much harder so I temporarily left it broken.
- The manual assembling of the middlewares and how to expose an API for the user assemble new middleware pipelines and how much control do we allow.

Hopefully with this PR is a bit clearer and we can discuss design options 🙂

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -107,6 +107,7 @@ class BaseClient:
         self.max_redirects = max_redirects
         self.dispatch = async_dispatch
         self.concurrency_backend = backend
+        self.additional_middlewares = additional_middlewares or []
 
     def merge_cookies(
         self, cookies: CookieTypes = None
@@ -157,6 +158,8 @@ class BaseClient:
             middlewares.append(HTTPBasicAuthMiddleware(username=auth[0], password=auth[1]))
         elif isinstance(auth, HTTPBasicAuth):
             middlewares.append(HTTPBasicAuthMiddleware(username=auth.username, password=auth.password))
+
+        middlewares.extend(self.additional_middlewares)
 
         while True:
             for middleware in middlewares:
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -24,6 +24,11 @@ from .exceptions import (
     TooManyRedirects,
 )
 from .interfaces import AsyncDispatcher, ConcurrencyBackend, Dispatcher
+from .middlewares import (
+    BaseMiddleware,
+    RedirectMiddleware,
+    HTTPBasicAuthMiddleware,
+)
 from .models import (
     URL,
     AsyncRequest,
@@ -61,6 +66,7 @@ class BaseClient:
         app: typing.Callable = None,
         raise_app_exceptions: bool = True,
         backend: ConcurrencyBackend = None,
+        additional_middlewares: list = None,  # TODO: is this the best way to inject mws?
     ):
         if backend is None:
             backend = AsyncioBackend()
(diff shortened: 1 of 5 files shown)


httpx/client.py
```diff
@@ -61,6 +66,7 @@ def __init__(
         app: typing.Callable = None,
         raise_app_exceptions: bool = True,
         backend: ConcurrencyBackend = None,
+        additional_middlewares: list = None,  # TODO: is this the best way to inject mws?
```
sethmlarson: Can we rename this param to just middlewares?

httpx/client.py
```diff
@@ -131,171 +138,53 @@ def merge_headers(
...
+        middlewares.extend(self.additional_middlewares)
 
         while True:
-            # We perform these checks here, so that calls to `response.next()`
-            # will raise redirect errors if appropriate.
-            if len(history) > self.max_redirects:
-                raise TooManyRedirects()
-            if request.url in [response.url for response in history]:
-                raise RedirectLoop()
+            for middleware in middlewares:
```
sethmlarson: In order to implement caching we'd need a middleware to be able to return an HTTP response on the `process_request()` stage.

sethmlarson:
Definitely want your thoughts on this @tomchristie

httpx/client.py
yeraydiazdiaz: I considered it but thought it might convey that you can override the built-in middlewares so I added the "additional". Happy to change it if we don't think that's the case.

httpx/client.py
yeraydiazdiaz: Good point, is it worth adding it now?

sethmlarson:
Ping @tomchristie I think this interface is really promising.

httpx/client.py
sethmlarson: I think so, might as well get it working with this revision.

tomchristie:
I'm still not really sure at the moment. There's a couple different styles we could use, one is a separate `process_request -> request` and `process_response -> [request|response]` like this. 

Another option would be...

```python
middleware (request, call_next):
    # Do some request processing
    response = call_next(request)
    # Do some response processing
```

I think it's a bit awkward that we're having to set up a middleware chain per-request.

The other thing that ties in here is our existing dispatch interface. The `auth` and `redirect` implementations here *could* alternately be treated as dispatch subclasses, that are provided with the next dispatcher in the chain (and are each setup once per client rather than on every request.)

I'd probably need to make a separate PR to show how that'd look, and why it ties in to me now being sure exactly what we should (or shouldn't) do about a middleware interface.

tomchristie:
Apologies for not being able to be more definitive yet.
It's a really nice piece of work, but I'm having a hard time figuring if it's the right direction or not.

yeraydiazdiaz:
No worries, that's why I put together this PR, so we can think about it and discuss it. 🙂 

> I think it's a bit awkward that we're having to set up a middleware chain per-request.

I agree. The reason it's that way is because we allow `auth` to be passed as arguments to the initializer of `Client` and also to `Client.send`. We also have a `max_redirects` argument to the `Client` initializer but a `allow_redirects` flag in `Client.send`.

Ideally we'd build the necessary middlewares once on the initializer, but we'd have to break the `send` API to do it, which I personally don't think it's worth it but maybe it is?

> The other thing that ties in here is our existing dispatch interface. The auth and redirect implementations here could alternately be treated as dispatch subclasses, that are provided with the next dispatcher in the chain (and are each setup once per client rather than on every request.)

That's interesting, we would avoid introducing a new abstraction at the cost of building the chain from the set of arguments.

I think I'll have a go at this and open up a different PR from what's on this one.

florimondmanca:
Closed in favor of #268

PR #177, opened 2019-07-30
added netrc support

cansarigol:


httpx/config.py
```diff
@@ -39,9 +39,16 @@ class SSLConfig:
     SSL Configuration.
     """
 
-    def __init__(self, *, cert: CertTypes = None, verify: VerifyTypes = True):
+    def __init__(
```
sethmlarson: Let's revert usages of trust_env for SSLConfig? The default behavior in the future should be use system trust stores unless we're specifically told which cert bundle to use.

cansarigol: You're right new approach for environment variables would be nice.
I reverted the last commit.

httpx/utils.py
```diff
@@ -64,3 +66,12 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
             return "utf-32-le"
         # Did not detect a valid UTF-32 ascii-range character
     return None
+
+
+def get_netrc_login(url: str) -> typing.Union[typing.Tuple[str, str, str], None]:
```
tomchristie: Signature should either be `host` or `authority` here, right?
It's not clear to me from a glance at the netrc docs, which of those two is actually the correct thing to use?
Also we should prefer `typing.Optional[typing.Tuple[str, str, str]]` as the return annotation.

httpx/utils.py
```diff
@@ -64,3 +66,12 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
+    try:
+        _netrc = netrc.netrc(os.environ.get("DEFAULT_NETRC_PATH"))  # type: ignore
```
tomchristie: I'd probably prefer a variable name that's not underscore prefixed here. `netrc_info` maybe?

tests/.netrc
```diff
@@ -0,0 +1,3 @@
+machine test_entrc.httpx.com
```
tomchristie: Let's use the `example.org` domain, since it's there for this sort of thing.

tests/.netrc
```diff
@@ -0,0 +1,3 @@
...
+login tomchristie
+password password123
```
tomchristie: I'd probably marginally prefer that we use a fake style username here `example-username` and `example-password` or something. Just feels marginally cleaner.

httpx/utils.py
tomchristie: Is `DEFAULT_NETRC_PATH` a standardized thing? (Or used by requests or something?)

tests/.netrc
cansarigol: If appropriate, I used netrcexample.org. Because some tests are affected by example.org.

httpx/utils.py
cansarigol: I made it up to explain the initial path.

httpx/utils.py
cansarigol: `machine` in the config. I changed it as host I can change again if necessary

cansarigol:
Thanks, I applied your comments.

httpx/utils.py
sethmlarson: I'm pretty sure `NETRC` is the env variable to specify the path

tests/client/test_auth.py
```diff
@@ -67,3 +68,27 @@ def auth(request):
 
     assert response.status_code == 200
     assert response.json() == {"auth": "Token 123"}
+
+
+def test_entrc_auth():
```
StephenBrown2: Typo: `entrc` -> `netrc`

cansarigol: thanks, fixed

tomchristie:
Great stuff, thanks!
Opening up two other issues that would act as further enhancements to this.

PR #184, opened 2019-08-01
Move HSTS preload checking to client

### LINKED ISSUE #180: Perform HSTS in the Client, not the URL.
I think it'd probably be a good idea for HSTS enforcement to be at the `Client` layer, rather than the `URL` model layer.

Not super keen on instantiating a `URL`, and having it end up with a different scheme, but I *am* okay with `client.request` dealing with it.

sethmlarson:
Closes #180

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -2,6 +2,8 @@ import inspect
 import typing
 from types import TracebackType
 
+import hstspreload
+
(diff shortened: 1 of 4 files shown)


tomchristie:
Lovely stuff, yup - I like how you've done this!

Issue #186, opened 2019-08-01
Try both `~/.netrc` and `~/_netrc` files


tomchristie:
Windows uses "_netrc" rather than ".netrc".
Our implementation should simply try both.

We might want something like this...

```python
if 'NETRC' in os.environ:
    NETRC_FILES = (os.environ['NETRC'],)
else:
    try:
        NETRC_FILES = (os.path.expanduser('~/.netrc'), os.path.expanduser('~/_netrc'))
    except KeyError:
        # May be raised if no user home directory is set.
        NETRC_FILES = ()
```

As part of this change we probably also want to make sure that any netrc file permission error or parse error simply results in netrc permissions not being used, rather than hard erroring. 

Related to #177

sethmlarson:
This issue has been fixed

PR #187, opened 2019-08-01
added trust_env attribute on the client instance

### LINKED ISSUE #185: Support `.trust_env` attribute on the client instance.
We should add a `trust_env` attribute and `__init__` keyword argument to `BaseClient`.
This should be used as the default if `trust_env` is not explicitly specified on the request.

The default value for `trust_env` on the per-method calls should then switch to `None`, so that we can distinguish it from "explicitly True" and "explicitly False".

This'd be a nice enhancement on top of #177.

cansarigol:
Closes #185

### DIFF
--- a/httpx/api.py
+++ b/httpx/api.py
@@ -32,7 +32,7 @@ def request(
     cert: CertTypes = None,
     verify: VerifyTypes = True,
     stream: bool = False,
-    trust_env: bool = True,
+    trust_env: bool = None,
 ) -> Response:
     with Client() as client:
         return client.request(
@@ -66,7 +66,7 @@ def get(
     cert: CertTypes = None,
     verify: VerifyTypes = True,
     timeout: TimeoutTypes = None,
-    trust_env: bool = True,
+    trust_env: bool = None,
 ) -> Response:
     return request(
         "GET",
@@ -96,7 +96,7 @@ def options(
     cert: CertTypes = None,
     verify: VerifyTypes = True,
     timeout: TimeoutTypes = None,
-    trust_env: bool = True,
+    trust_env: bool = None,
 ) -> Response:
     return request(
         "OPTIONS",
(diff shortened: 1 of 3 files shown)


httpx/api.py
```diff
@@ -32,9 +32,9 @@ def request(
     cert: CertTypes = None,
     verify: VerifyTypes = True,
     stream: bool = False,
-    trust_env: bool = True,
+    trust_env: bool = None,
 ) -> Response:
-    with Client() as client:
+    with Client(trust_env=trust_env) as client:
```
tomchristie: I'd suggest passing `trust_env` in the `request` call here, instead. That'd feel more obvious to me.

httpx/client.py
```diff
@@ -332,7 +331,6 @@ class AsyncClient(BaseClient):
         cert: CertTypes = None,
         verify: VerifyTypes = None,
         timeout: TimeoutTypes = None,
-        trust_env: bool = True,
```
tomchristie: We don't want to get rid of `trust_env` here - we just want to change the behavior so that if `trust_env == None` then the client default is used, but if `trust_env in (True, False)` then the explicitly requested behaviour is used.

tomchristie:
Typically I'd suggest keeping pull requests for #185 and #186 seperate - that makes for more easily reviewable chunks of work, and means we can progress one, even if the other is blocked.

cansarigol:
Sorry, I thought you want to move `trust_env` into `BaseClient.__init__`.
I separated PRs.

cansarigol:
> Code looks good, can we get one test case that verifies that setting `client.request(..., trust_env=X)` overrides `client.trust_env`?

Could you review again? I added a test that contains `trust_env=True`. False case already existed.

sethmlarson:
Thanks @cansarigol! :D

cansarigol:
You're welcome

PR #188, opened 2019-08-01
Use pathlib.Path instead of os.path functions

StephenBrown2:
Since httpx targets Python 3.6+, I figured it would be good to use the `pathlib.Path` object functions rather than `os.path.*`, for a slightly better interface.

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -1,7 +1,7 @@
 import asyncio
-import os
 import ssl
 import typing
+from pathlib import Path
 
 import certifi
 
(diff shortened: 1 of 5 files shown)


StephenBrown2:
I don't believe so, as the way I changed it it will still expect a `str` and initialize a `Path` object from it. Though I did see the `netrc` PRs come through that use `os.path`, so I'll rebase and update those as well, and make sure to check the types for that also.

StephenBrown2:
Rebased as promised and looked over the definitions. I don't know if `context.load_verify_locations` will accept a `Path` object, so I've left that as being cast back to a string.

sethmlarson:
Thanks @StephenBrown2! Will take a look later :)

PR #189, opened 2019-08-01
added `~/.netrc` and `~/_netrc` files for netrc path

cansarigol:
Issue: #186

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -83,10 +83,24 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
     return None
 
 
+NETRC_STATIC_FILES = tuple(
+    os.path.expanduser(path) for path in ("~/.netrc", "~/_netrc")
+)
+
+
 def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
-    try:
-        netrc_info = netrc.netrc(os.environ.get("NETRC"))  # type: ignore
-    except FileNotFoundError:
+    NETRC_FILES = (
+        (os.environ["NETRC"],) if "NETRC" in os.environ else NETRC_STATIC_FILES
+    )
+    netrc_path = None
+
+    for file_path in NETRC_FILES:
+        if os.path.isfile(file_path):
+            netrc_path = file_path
+            break
+
+    if netrc_path is None:
         return None
 
+    netrc_info = netrc.netrc(netrc_path)
     return netrc_info.authenticators(host)  # type: ignore


httpx/utils.py
```diff
@@ -84,9 +84,25 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
-    except FileNotFoundError:
+    if "NETRC" in os.environ:
+        NETRC_FILES = (os.environ["NETRC"],)
+    else:
+        try:
+            NETRC_FILES = (
+                os.path.expanduser("~/.netrc"),
+                os.path.expanduser("~/_netrc"),
+            )
+        except KeyError:  # pragma: nocover
```
sethmlarson: Why would this return a `KeyError`?

httpx/utils.py
```diff
@@ -84,9 +84,25 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
+            return None
+
+    netrc_path = None
+    for file_path in NETRC_FILES:
+        if os.path.exists(file_path):
```
sethmlarson: These need to be files so let's do `isfile()`

httpx/utils.py
tomchristie: The `requests` codebase has the same case, claiming "os.path.expanduser" can raise a KeyError if no user directory is set. However thinking about it if that ever *was* true then it's likely it would have been treated as a valid bug in Python, so it's unlikely to be true now. The docs state "If the expansion fails or if the path does not begin with a tilde, the path is returned unchanged." so we probably don't want this catch-case.

httpx/utils.py
```diff
@@ -84,9 +84,25 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
-        netrc_info = netrc.netrc(os.environ.get("NETRC"))  # type: ignore
-    except FileNotFoundError:
+    if "NETRC" in os.environ:
+        NETRC_FILES = (os.environ["NETRC"],)
+    else:
+        try:
+            NETRC_FILES = (
+                os.path.expanduser("~/.netrc"),
+                os.path.expanduser("~/_netrc"),
+            )
```
tomchristie: Rather than doing this every time, lets just populate `NETRC_FILES` once, at the module level.

```python
def get_netrc_files():
    if "NETRC" in os.environ:
         return (os.environ["NETRC"],)
    return (
         os.path.expanduser("~/.netrc"),
         os.path.expanduser("~/_netrc"),
    )

NETRC_FILES = get_netrc_files()
```

cansarigol: module level doesn't solve NETRC env changing. change this part could you review again?

Issue #195, opened 2019-08-07
Fallback when faulty Content-Encoding header is set?


rinti:
Hi!

I am trying to use httpx to talk with an API that sends this header:
`Content-Encoding: UTF-8`
which obviously is an invalid header, this however prevents me from getting a response as I get this error:
```
  File "/Users/andreas/repositories/bussen/venv/lib/python3.7/site-packages/httpx/models.py", line 781, in decoder
    decoder_cls = SUPPORTED_DECODERS[value]
KeyError: 'utf-8'
```

If i try using requests instead it works. I'd be happy to try to help if I get a little guidance on what would be a sensible fallback here.

sethmlarson:
Closed via #196

PR #196, opened 2019-08-07
Fail silently if invalid Content-Encoding is sent

rinti:
This is my proposed fix for #195 
It just discards invalid header values for Content-Encoding


tomchristie:
Seems reasonable enough to me.

sethmlarson:
Thanks for fixing this @rinti! :) Hope you'll stick around and help fix a few other issues.

PR #204, opened 2019-08-13
Centralize tooling configuration

humrochagf:
This centralizes all the tooling configurations that will be used by
flake8, mypy, isort, pytest, and coverage tools.

It prevents the config file bloating at the project and also makes the
editor linters to automatic detects the project coding standards.

It also removes the .isort.cfg file and the configurations hardcoded at
the noxfile to prevent duplication.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -9,18 +9,7 @@ def lint(session):
 
     session.run("autoflake", "--in-place", "--recursive", *source_files)
     session.run("seed-isort-config", "--application-directories=httpx")
-    session.run(
-        "isort",
-        "--project=httpx",
-        "--multi-line=3",
-        "--trailing-comma",
-        "--force-grid-wrap=0",
-        "--combine-as",
-        "--line-width=88",
-        "--recursive",
-        "--apply",
-        *source_files,
-    )
+    session.run("isort", "--project=httpx", "--recursive", "--apply", *source_files)
     session.run("black", "--target-version=py36", *source_files)
 
     check(session)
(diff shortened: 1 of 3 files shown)


humrochagf:
@sethmlarson that prevents the configuration duplication that I mentioned at the PR #128 and makes the IDE's and editors to track the project coding patterns

noxfile.py
```diff
@@ -50,13 +37,4 @@ def docs(session):
...
-        "coverage",
-        "run",
-        "--omit='*'",
-        "-m",
-        "pytest",
-        "--cov=httpx",
-        "--cov=tests",
-        "--cov-report=term-missing",
-    )
+    session.run("coverage", "run", "-m", "pytest")
```
StephenBrown2: I've been curious why this is run this way, instead of using [`pytest-cov`](https://github.com/encode/httpx/blob/master/test-requirements.txt#L20) directly, as recommended by coverage docs themselves: https://coverage.readthedocs.io/en/v4.5.x/#using-coverage-py

AFAIK, `pytest --cov=httpx --cov=tests --cov-report=term-missing` will run the same as `coverage run -m pytest...`, and not require `--omit='*'`, which excludes everything from `coverage` and ends up using `pytest-cov`'s run anyways.

sethmlarson: This was something I was doing as it seemed to work to not track coverage relative to the virtualenv. This appears to work though?

humrochagf: I'll change it, the only difference is that `coverage` adds `httpx` to the execution context automatically so you don't need to install it before running the test command.

To run with `pytest` directly it's required to install `httpx` before running like that:

```python
    session.install("-e", ".")

    session.run("pytest")
```
Otherwise, the nox execution will fail when looking for `httpx` at the cases where nox is installed globally

StephenBrown2: In that case, one can also run `("python", "-m", "pytest")`
Src: https://docs.pytest.org/en/latest/usage.html#calling-pytest-through-python-m-pytest

humrochagf: That was clever :smile:

humrochagf: Added these changes at 4ebc0094482c5e760cf79edaf3338924af1d91fe

PR #206, opened 2019-08-13
added cov-fail-under=100 for pytest

cansarigol:


httpx/models.py
```diff
@@ -100,7 +100,7 @@ def __init__(
             if idna_authority != self._uri_reference.authority:
                 self._uri_reference = self._uri_reference.copy_with(
                     authority=idna_authority
-                )
+                )  # pragma: nocover
```
sethmlarson: Instead of disabling coverage here let's create a simple unit test for the branch.

httpx/utils.py
```diff
@@ -98,7 +98,7 @@ def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
             break
 
     if netrc_path is None:
-        return None
+        return None  # pragma: nocover
```
sethmlarson: Can we also create a unit test for here?

httpx/models.py
cansarigol: I couldn't find a test case because if I try to Unicode domain, I get this error: `rfc3986.exceptions.InvalidAuthority: The authority (😁.de) is not valid.` 

Should we change this part if there is no case or could you help me with this?

sethmlarson: What about `ουτοπία.δπθ.gr`? Emojis aren't valid for IDNs for IDNA 2008.

cansarigol: idna encode raises an error if there is any char that greater then 128. (it's not just about emoji [link](https://github.com/python-hyper/rfc3986/blob/2982509402519575282845adf1f39dc0195369e0/src/rfc3986/iri.py#L121)) so this if clause is never used. If you want we can change it with another PR and check `InvalidAuthority` error?

sethmlarson: I think we can literally just get rid of that whole `if` statement branch. `IRIReference.encode()` does everything for us.

cansarigol: added a pr #210

sethmlarson:
Thanks @cansarigol!

PR #208, opened 2019-08-14
Expand disallow_redirects test assert

behrtam:
The `response.is_redirect` is already implicitly tested to be _True_ in the other test case that disallows redirects (https://github.com/encode/httpx/blob/master/tests/client/test_redirects.py#L220) but I thought it would be good to have it explicit in the first simpler test case.


sethmlarson:
Thanks for this! :)

Issue #120, opened 2019-07-14
Add support for a Trio ConcurrencyBackend


sethmlarson:

tomchristie:
Indeed, yes.

I think there’s a conversation to be had about if that ends up in the core package or not.

My strong inclination is that we should support alternate concurrency backends, but that implementations other than asyncio should live in third party packages.

florimondmanca:
Regarding whether alternative loops/IO libraries should be in the core package, I think it makes sense in terms of relying on the community to support their own needs.

One thing I think would be great though, is for httpx to provide a core test suite that alternatives could run their backend against. Don’t know exactly how that would materialize, but it would definitely allow to make sure that backends are compliant with the main API, as well as ease testing on the side of the third party package.

PR #214, opened 2019-08-14
Trio concurency backend PoC

florimondmanca:
Refs #120 

For now I basically copy-pasted the asyncio backend from `concurrency` module and replaced asyncio-specific parts with the trio equivalent. Did the same thing for the async client tests to get a quick idea of how things were going.

Tests are read timing out for now, I'll try and look as to why that is the case tomorrow. :-)

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -79,7 +79,7 @@ class BaseClient:
                 )
             else:
                 dispatch = ASGIDispatch(
-                    app=app, raise_app_exceptions=raise_app_exceptions
+                    app=app, raise_app_exceptions=raise_app_exceptions, backend=backend
                 )
 
         if dispatch is None:
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -14,11 +14,6 @@ import ssl
 import typing
 from types import TracebackType
 
-try:
-    from contextlib import asynccontextmanager
-except ImportError:  # pragma: no cover
-    from async_generator import asynccontextmanager  # type: ignore
-
 from .config import PoolLimits, TimeoutConfig
 from .exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
 from .interfaces import (
@@ -297,15 +292,7 @@ class BackgroundManager(BaseBackgroundManager):
         self.tasks.add(loop.create_task(coroutine(*args)))
 
     def will_wait_for_first_completed(self) -> BaseAsyncContextManager:
-        @asynccontextmanager  # type: ignore
-        async def context() -> None:
-            initial_tasks = self.tasks
-            self.tasks = set()
-            yield
-            await asyncio.wait(self.tasks, return_when=asyncio.FIRST_COMPLETED)
-            self.tasks = initial_tasks.union(self.tasks)
-
-        return context()  # type: ignore
+        return WillWaitForFirstCompleted(self)
 
     async def __aexit__(
         self,
@@ -317,3 +304,17 @@ class BackgroundManager(BaseBackgroundManager):
             await task
             if exc_type is None:
                 task.result()
+
+
+class WillWaitForFirstCompleted(BaseAsyncContextManager):
+    def __init__(self, background: BackgroundManager):
+        self.background = background
+        self.initial_tasks: typing.Set[asyncio.Task] = set()
+
+    async def __aenter__(self) -> None:
+        self.initial_tasks = self.background.tasks
+        self.background.tasks = set()
+
+    async def __aexit__(self, *args: typing.Any) -> None:
+        await asyncio.wait(self.background.tasks, return_when=asyncio.FIRST_COMPLETED)
+        self.background.tasks = self.initial_tasks.union(self.background.tasks)
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -254,7 +254,7 @@ class AsyncioBackend(ConcurrencyBackend):
 
 
 class BackgroundManager(BaseBackgroundManager):
-    def __init__(self, coroutine: typing.Callable, args: typing.Any) -> None:
+    def __init__(self, coroutine: typing.Callable, args: typing.Sequence) -> None:
         self.coroutine = coroutine
         self.args = args
 
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -14,11 +14,15 @@ import ssl
 import typing
 from types import TracebackType
 
-from async_generator import asynccontextmanager
+try:
+    from contextlib import asynccontextmanager
+except ImportError:  # pragma: no cover
+    from async_generator import asynccontextmanager  # type: ignore
 
 from .config import PoolLimits, TimeoutConfig
 from .exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
 from .interfaces import (
+    BaseAsyncContextManager,
     BaseBackgroundManager,
     BaseBodyIterator,
     BaseEvent,
@@ -292,13 +296,16 @@ class BackgroundManager(BaseBackgroundManager):
         loop = asyncio.get_event_loop()
         self.tasks.add(loop.create_task(coroutine(*args)))
 
-    @asynccontextmanager  # type: ignore
-    async def will_wait_for_first_completed(self) -> typing.AsyncContextManager:
-        initial_tasks = self.tasks
-        self.tasks = set()
-        yield
-        await asyncio.wait(self.tasks, return_when=asyncio.FIRST_COMPLETED)
-        self.tasks = initial_tasks.union(self.tasks)
+    def will_wait_for_first_completed(self) -> BaseAsyncContextManager:
+        @asynccontextmanager  # type: ignore
+        async def context() -> None:
+            initial_tasks = self.tasks
+            self.tasks = set()
+            yield
+            await asyncio.wait(self.tasks, return_when=asyncio.FIRST_COMPLETED)
+            self.tasks = initial_tasks.union(self.tasks)
+
+        return context()  # type: ignore
 
     async def __aexit__(
         self,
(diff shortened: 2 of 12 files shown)


sethmlarson:
This is awesome, thanks for being the first to take a stab at this! :)

Something I think we ought to do for a trio backend: have some way we can run all our test cases with the trio backend instead of the asyncio backend so we can double the value of our test suite easily. :)

This'll definitely need @tomchristie's eyes as well.

florimondmanca:
> have some way we can run all our test cases with the trio backend instead of the asyncio backend so we can double the value of our test suite easily. :)

Absolutely. I copy-pasted the async client tests knowing that it was definitely not ideal. I think making use of pytest fixtures and using the backend *inside* fixtures/tests instead of asyncio/trio would allow something like this.

tomchristie:
Ah fabulous, great stuff!

> Something I think we ought to do for a trio backend: have some way we can run all our test cases with the trio backend instead of the asyncio backend so we can double the value of our test suite easily.

Indeed. I'm open to suggestions here around how best to marshal support for alternate backends, but this PR is a *great* start!

florimondmanca:
So, having looked a bit more into this, I think a necessary step before pulling off a trio backend would be to have end-to-end usage of the concurrency backend in the whole code base.

This requires to remove remaining asyncio-specific code, in particular in tests and `ASGIDispatch` — I already have something up and running and will open a separate PR for it, then rebase this PR off it to confirm it would be sufficient to implement a trio backend. :-)

florimondmanca:
This now contains the changes from #217, and there's progress: only tests for the blocking client are now failing! :tada: (Again because of a `ReadTimeout`, indicating that the client can't reach the server running in the other thread.)

florimondmanca:
@sethmlarson Today's update: I realized that the timeouts I was having were due to incorrectly handling the necessary coexistance of the asyncio and trio event loops.

Since the uvicorn server *has* to run on asyncio (because uvicorn is asyncio-only), I figured that:

1. Any test using the parametrized concurrency `backend` should run in its own thread, and that this thread had to be fired via the asyncio threadpool (previously it was fired by the `backend` itself).
1. We can't use `pytest-trio`/`pytest.mark.trio` on tests that use the trio backend, because they *must* run on asyncio anyway.

So what I did is add a pytest hook that automatically defers to the asyncio threadpool tests that use the  `backend` fixture. (It starts to look like a lot of black magic, but once I get my head around all this we can clean things up for maintainability.)

The hook works like a charm on async tests, but I'm having issues with the blocking ones, e.g. `test_client.py`. In that case, `httpx.Client()` ends up calling `concurrency_backend.run()` (keep in mind — this would be running within the thread fired by asyncio in the hook). I'd expect the resulting new trio event loop to do its business as usual (it works on the asyncio backend), but instead I get a mysterious `TrioInternalError`. 🤷‍♂  I tried running the blocking test function in a plain thread as well (with good ol' `threading.Thread()`), but I ended up with the same result.

Any clues on this? Am I missing a crucial piece of info that would allow simplifying things considerably?

sethmlarson:
Your guess is as good as mine. Gonna cc the I/O wizard: @njsmith

njsmith:
I don't really understand what you've been trying – it sounds complicated! – and I can't debug a TrioInternalError without any details :-). So uh, here's some general comments?

- It might be worth considering using hypercorn instead of uvicorn, since hypercorn already supports both asyncio and trio?
- It should be totally fine to run asyncio from a trio thread (`trio.to_thread.run_sync(asyncio.run, ...)`), and to run trio from an asyncio thread (`loop.run_in_executor(..., trio.run, ...)`)
- I'm missing something about why a "blocking test" involves multiple async backends?

florimondmanca:
Thanks for pitching in @njsmith!

Here's a minimal repro example for the `TrioInternalError`:

```python
# app.py
import asyncio

import httpx
from httpx.contrib.trio import TrioBackend


    assert scope["type"] == "http"
    await send(
        {
            "status": 200,
        }
    )


def test_get(backend):
    url = "http://127.0.0.1:8000/"
    with httpx.Client(backend=backend) as http:
        response = http.get(url)


async def main():
    loop = asyncio.get_event_loop()
    await loop.run_in_executor(None, test_get, TrioBackend())


if __name__ == "__main__":
    asyncio.run(main())
```

Run it with:

- `uvicorn app:app`
- `python app.py`

Full traceback below:

```
$ python app.py
  File "app.py", line 51, in <module>
    asyncio.run(main())
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/asyncio/runners.py", line 43, in run
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "app.py", line 47, in main
    await loop.run_in_executor(None, test_get, TrioBackend())
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "app.py", line 40, in test_get
    response = http.get(url)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 750, in get
    trust_env=trust_env,
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 718, in request
    response.read()
  File "/Users/florimond/Developer/python-projects/httpx/httpx/models.py", line 969, in read
    self._content = b"".join([part for part in self.stream()])
  File "/Users/florimond/Developer/python-projects/httpx/httpx/models.py", line 969, in <listcomp>
    self._content = b"".join([part for part in self.stream()])
  File "/Users/florimond/Developer/python-projects/httpx/httpx/models.py", line 980, in stream
    for chunk in self.raw():
  File "/Users/florimond/Developer/python-projects/httpx/httpx/models.py", line 1008, in raw
    for part in self._raw_stream:
  File "/Users/florimond/Developer/python-projects/httpx/httpx/interfaces.py", line 224, in iterate
    yield self.run(async_iterator.__anext__)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/contrib/trio.py", line 175, in run
    return trio.run(coroutine, *args)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1769, in run
    run_impl(runner, async_fn, args)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1918, in run_impl
    runner.task_exited(task, final_outcome)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1404, in task_exited
    raise TrioInternalError
trio.TrioInternalError
```

The issue seems to be with running `async_iterator.__anext__`, where `async_iterator` refers to the result from [H11Connection._receive_response_data](https://github.com/encode/httpx/blob/master/httpx/dispatch/http11.py#L136).

I am seemingly getting the same error if slimming down the repro example to just this:

```python
import asyncio
import trio


def test_get():
    async def gen():
        yield "hi"

    g = gen()
    trio.run(g.__anext__)


async def main():
    loop = asyncio.get_event_loop()
    await loop.run_in_executor(None, test_get)


if __name__ == "__main__":
    asyncio.run(main())
```

Traceback:

```
$ python app.py
  File "app.py", line 54, in <module>
    asyncio.run(main())
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/asyncio/runners.py", line 43, in run
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "app.py", line 50, in main
    await loop.run_in_executor(None, test_get)
  File "/Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "app.py", line 45, in test_get
    trio.run(g.__anext__)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1769, in run
    run_impl(runner, async_fn, args)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1918, in run_impl
    runner.task_exited(task, final_outcome)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1404, in task_exited
    raise TrioInternalError
trio.TrioInternalError
```

Is it an issue to run a step of an async generator with `trio.run()`?

njsmith:
Filed and debugged at: https://github.com/python-trio/trio/issues/1191

(The problem is that Trio is not expecting you to have an async function that's implemented in C, because those are extremely rare, and it's crashing while trying to do some introspection that doesn't work on C code. Then because of some annoying details the error gets lost.)

We can fix that but there's a more fundamental problem: the only way it makes sense to call `trio.run(__anext__)` like this is if you're going to call `trio.run` on *every* iteration. But not only is this super wasteful (trio has to reconstruct the whole async environment on every iteration, including allocating a self-pipe etc.), it's also a violation of trio's semantics – you're not in general allowed to pass trio objects from one `trio.run` call to another.

I assume that this is happening because httpx's sync API is using a strategy where it allocates an asyncio loop and then calls `loop.run_until_complete` inside every sync call. That strategy only works with asyncio. Fortunately, I don't think httpx needs it to work with anything except asyncio? I don't think the sync API should accept a `backend` argument at all; the details of how it does sync I/O should be up to it.

florimondmanca:
Hmm, now that you say it, what would prevent us from always using asyncio for the sync API? I don’t see a use case where we’d use the sync client instead of the async one in a trio (async) context anyway.

I believe the backend argument on Client was added in #60 — any thoughts on this @tomchristie?

sethmlarson:
I think you're totally right on the sync client not accepting a backend @njsmith. Thanks for all the debugging! :)

PR #215, opened 2019-08-15
Accept SSLContext into SSLConfig(verify=...)

### LINKED ISSUE #166: How to configure TLS beyond client certificate and CA root certs?
Typically in requests you can use a __HTTPAdapter__ to hijack and set the SSL context and define what ciphers to use, how does one go about doing the same in __httpx__?

sethmlarson:
Fixes #166. Allows users to configure an `SSLConfig` instance by passing in an `SSLContext` object that they've loaded CA certs, ciphers, etc onto.


tomchristie:
Lovely!

PR #217, opened 2019-08-15
Generalized concurrency backend usage

florimondmanca:
Base for #214

### DIFF
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -14,9 +14,9 @@ import ssl
 import typing
 from types import TracebackType
 
-from .config import PoolLimits, TimeoutConfig
-from .exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
-from .interfaces import (
+from ..config import PoolLimits, TimeoutConfig
+from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
+from ..interfaces import (
     BaseBackgroundManager,
     BasePoolSemaphore,
     BaseReader,
@@ -24,6 +24,7 @@ from .interfaces import (
     ConcurrencyBackend,
     Protocol,
 )
+from .utils import TimeoutFlag
 
 SSL_MONKEY_PATCH_APPLIED = False
 
@@ -47,38 +48,6 @@ def ssl_monkey_patch() -> None:
             _write(self, data)
 
     MonkeyPatch.write = _fixed_write
-
-
-class TimeoutFlag:
-    """
-    A timeout flag holds a state of either read-timeout or write-timeout mode.
-
-    We use this so that we can attempt both reads and writes concurrently, while
-    only enforcing timeouts in one direction.
-
-    During a request/response cycle we start in write-timeout mode.
-
-    Once we've sent a request fully, or once we start seeing a response,
-    then we switch to read-timeout mode instead.
-    """
-
-    def __init__(self) -> None:
-        self.raise_on_read_timeout = False
-        self.raise_on_write_timeout = True
-
-    def set_read_timeouts(self) -> None:
-        """
-        Set the flag to read-timeout mode.
-        """
-        self.raise_on_read_timeout = True
-        self.raise_on_write_timeout = False
-
-    def set_write_timeouts(self) -> None:
-        """
-        Set the flag to write-timeout mode.
-        """
-        self.raise_on_read_timeout = False
-        self.raise_on_write_timeout = True
 
 
 class Reader(BaseReader):
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -17,6 +17,7 @@ from types import TracebackType
 from ..config import PoolLimits, TimeoutConfig
 from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
 from ..interfaces import (
+    BaseAsyncContextManager,
     BaseBackgroundManager,
     BaseBodyIterator,
     BaseEvent,
@@ -236,6 +237,9 @@ class BackgroundManager(BaseBackgroundManager):
         loop = asyncio.get_event_loop()
         self.tasks.add(loop.create_task(coroutine(*args)))
 
+    def wait_first_completed(self) -> "WaitFirstCompleted":
+        return WaitFirstCompleted(self)
+
     async def __aexit__(
         self,
         exc_type: typing.Type[BaseException] = None,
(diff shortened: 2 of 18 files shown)


sethmlarson:
We definitely need to have this before we can start implementing additional backends. Thanks for opening this @florimondmanca!

httpx/concurrency.py
```diff
@@ -244,31 +247,74 @@ def run(
         finally:
             self._loop = loop
 
+    def create_event(self) -> BaseEvent:
+        return asyncio.Event()  # type: ignore
```
sethmlarson: I saw a nice way to do this with `typing.cast(...)` so we don't have to skip the type checking.

tests/conftest.py
```diff
@@ -10,6 +10,14 @@
 from uvicorn.config import Config
 from uvicorn.main import Server
 
+from httpx.concurrency import AsyncioBackend
+
+
+@pytest.fixture(params=[pytest.param(AsyncioBackend, marks=pytest.mark.asyncio)])
```
sethmlarson: This is so smart, nice work here. :)

sethmlarson:
So this is great, I've got a couple thoughts in my head right now:

Can we somehow create a backend that detects the current async framework we're running in and then just use that `AsyncioBackend` or `TrioBackend`? Then there's no need for users to even think about that. I know there's a library under the trio org that can do this. 

Second Im really strongly feeling that we should package trio backend support in a contrib module and then have an extra for it. Easier for us to test and easier for users since they won't have to go track down another package to get what they need to use trio just `python -m pip install httpx[trio]` or if they already have it it'll just work.

httpx/concurrency.py
florimondmanca: Good point! I always forget about `.cast()`.

florimondmanca:
@sethmlarson Agreed with all your points.

Backend auto-detection is a really nice feature to add to the backlog.

One thing I'm thinking about just now — I suppose the `httpx` API would also use the detected backend by default? If so, we'll probably still need a way to manually set the backend used in that case, e.g.

```python
import httpx

httpx.backend = "asyncio"
```

> I know there's a library under the trio org that can do this.

👉 [sniffio](https://github.com/python-trio/sniffio)

> … we should package trio backend support in a contrib module …

I'm also more inclined towards doing this as well, also for increased control and ease of change on the backend interface.

florimondmanca:
@sethmlarson Alright, I completed reworked this PR.

The commit history is clean(er) now, so feel free to go through each commit diff. The main difference with the previous state is `concurrency` turned into a sub-package (ready to welcome `trio.py` & others :tada:).

(This doesn't have backend auto-detection because it seems premature at this stage; we can easily add it once we have `TrioBackend` in.)

httpx/concurrency/asyncio.py
```diff
@@ -247,28 +219,71 @@ def run(
...
+    async def __aexit__(self, *args: typing.Any) -> None:
+        await asyncio.wait(self.background.tasks, return_when=asyncio.FIRST_COMPLETED)
+        self.background.tasks = self.initial_tasks.union(self.background.tasks)
+
+
+class BodyIterator(BaseBodyIterator):
+    def __init__(self) -> None:
+        self._queue = asyncio.Queue(
+            maxsize=1
+        )  # type: asyncio.Queue[typing.Union[bytes, object]]
```
sethmlarson: We prefer inline type hints :)

florimondmanca:
@sethmlarson I fixed the type hint. I also added a test for using the ASGI dispatcher with the `AsyncClient`, which ended up exposing a bug in how pending tasks were cleaned up in the `AsyncioBackend` background manager. :-)

sethmlarson:
Nice! I'm taking a look now. :)

httpx/concurrency/asyncio.py
```diff
@@ -247,28 +219,84 @@ def run(
...
     ) -> None:
-        await self.task
-        if exc_type is None:
-            self.task.result()
+        done, pending = await asyncio.wait(self.tasks, timeout=0)
+
+        for task in pending:
+            try:
+                # Give it a chance to finish soon
+                await asyncio.wait_for(task, timeout=0.01)
```
sethmlarson: If it actually does finish should we move it to `done`?

florimondmanca: Yup. Which made me realize we can simplify this by passing the small `timeout` to the first `.wait()` call instead. :+1:

sethmlarson:
No idea what's going on with our CI, it's building what looks to be the completely wrong commit and I can't get it to do anything else? Weird.

florimondmanca:
Flake8 was reporting a NameError in a file test, and I believe that was the cause of this. (Still strange that Travis is not displaying the branch and latest commit on the build!)

tomchristie:
Great stuff!!!

I've got some thoughts on this:

* I think it might not be a bad idea for us to start with a smaller PR, that doesn't include the test parameterization (which is essentially stubbed out at this point in time, since it's all vs. asnycio in this pull request)
* Rather than `BodyIterator` the concurrency interface should just expose a Queue primitive, with the neccessary `get`/`put` operations.
* When we're using `__enter__` and `__exit__` we should stub out `async start()` and `async close()` methods, and have concrete pre-existing `__enter__` and `__exit__` methods (eg. see the dispatcher interface)

I've got other thoughts on how we might want to finesse, but I think we'd look at those as a follow on:

* Improving the semaphore interface to be more of a standard "give me a semaphore" rather than being coupled into our `PoolLimits` configuration.
* Simplifying the stream interface. (I reckon we probably want to merge the reader/writer stuff into a single `Stream` class.)
* We *might* want to think about pulling the concurrency interfaces into eg. `concurrency/interfaces.py`, given the highly specific technical nature of why we have it, and the fairly large API surface area it exposes. (Ie. we don't *really* want general user implementations here, except for very specific cases eg. a trio backend, a curio backend. Which is pretty different to eg. the dispatcher interface, which is a very public bit of API.)

tomchristie:
Also aware that I'm being a bit rigerous wrt. asking for a smaller footprint PR here first, with the parameterization coming afterwards.

If y'all think I'm being overzealous there, then *potentially* happy to re-consider there. 😃

florimondmanca:
@tomchristie I’m in favor of small PRs as well, and you’re correct saying that test parametrizarion  is very much YAGNI worthy here. We could add it along with the trio backend (although that particular PR will be >400 LOC already).

I’ll look into your other suggestions soon and update this PR. :-)

tomchristie:
> We could add it along with the trio backend

Either that or add it incrementally, so that it's nicely isolated in it's own PR.

I think we're probably still undecided around if trio support should be built-in or third-party right. (Both feasible avenues to my mind) In either case the main thing right now is nailing down a decent concurrency-backend API, and making sure that we're *actually using it everywhere*.

sethmlarson:
Here are my thoughts for this PR, let me know what you think:

- Keep the move from `concurrency.py` -> `concurrency/asyncio.py`
- Split the `interfaces.py` related to concurrency into `concurrency/interfaces.py` as mentioned.
- Keep the changes to the concurrency interface that don't have comments (`create_event()`, etc)

Future PRs (in order of priority?):

- Change `create_semaphore()` to not use `PoolLimits` but instead be a regular semaphore
- Change from a reader/writer to a stream interface (This change would benefit my start_tls() work within proxies!)
- Changes to allow for parametrize testing, add backend to A/WSGIDispatchers
- Add Trio backend implementation, implement parameterized testing.

Thoughts?

florimondmanca:
I’d be okay with reworking the PR towards that, except:

> add backend to A/WSGIDispatchers

This needs to stick in here IMO — assuming the goal of this PR is to make sure finalize the asyncio abstraction later (ie use backends everywhere, except in tests then).

Fun fact about the stream API bit — this is actually how trop does things, ie opening a connection exposes one single stream object (instead of a reader/writer pair) that allows to send and receive data. Overall I think it would be a simpler abstraction to deal with indeed.

tomchristie:
> This needs to stick in here IMO

Indeed, yup

sethmlarson:
Oh yeah leave that in then!

florimondmanca:
I think each item listed by both of you could be its own PR, and that this would be so much easier to review and discuss independantly, so I started rolling out incremental PRs, e.g. with #244.

PR #218, opened 2019-08-15
handle WindowUpdated event to send large requests using HTTP2

nsidnev:
Right now if you send a HTTP2 request with a lot of data via `httpx`, a `FlowControlError` error from `h2` will be thrown. To fix this data should be splited into chunks with right size and also wait for window update if there is too much data to send. I'm not sure how right is the way I implemented this, but it works with a simple HTTP2 ASGI application served by `hypercorn` and in my work case.


sethmlarson:
Thanks for opening this @nsidnev! I'm going to have to look at this one once I'm home from the hospital. I'd like to mess around and see the differences between base and head. :)

httpx/dispatch/http2.py
```diff
@@ -125,6 +133,9 @@ def initiate_connection(self) -> None:
             self.timeout_flags[stream_id].set_read_timeouts()
             if isinstance(event, h2.events.ResponseReceived):
                 break
+            elif isinstance(event, h2.events.WindowUpdated):
+                if event.stream_id and event.stream_id in self.flow_control_events:
+                    self.flow_control_events[event.stream_id].set()
```
sethmlarson: So what if the server gives us a WindowUpdated event but the size updated isn't large enough for us to be able to send? I think we need to track the threshold we're waiting for in addition to the `BaseEvent` object, decrement the threshold we're waiting for each time we receive an event, and then only call `.set()` when we're past the threshold.

Am I understanding that properly?

httpx/dispatch/http2.py
```diff
@@ -141,7 +152,9 @@ def initiate_connection(self) -> None:
         while True:
             event = await self.receive_event(stream_id, timeout)
             if isinstance(event, h2.events.DataReceived):
-                self.h2_state.acknowledge_received_data(len(event.data), stream_id)
+                self.h2_state.acknowledge_received_data(
+                    event.flow_controlled_length, event.stream_id
```
sethmlarson: Nice!

tests/dispatch/utils.py
```diff
@@ -118,5 +128,21 @@ def stream_complete(self, stream_id):
...
+
+    async def send_data(self, data, stream_id):
+        window_size = self.conn.local_flow_control_window(stream_id)
+        max_frame_size = self.conn.max_outbound_frame_size
+        chunk_size = min(len(data), window_size, max_frame_size)
+
+        for idx in range(0, len(data), chunk_size):
+            left_window_size = self.conn.local_flow_control_window(stream_id)
+            if left_window_size < chunk_size:
+                self.wait_events[stream_id] = asyncio.Event()
```
sethmlarson: This may have to be dealt with within #217 once it's merged, just cc @florimondmanca

florimondmanca: @sethmlarson Yup; I haven't looked at genericizing the HTTP/2 mocks in #217 yet, so thanks for the notice.

httpx/dispatch/http2.py
nsidnev: So. Okay. I understand the problem here, but I'm not sure that we should solve it this way. I think it would be better to use new window size that server gave us, and then wait for new updates of the window. So for example done in `hypercorn`

nsidnev: Also about the waiting for the required value. The [RFC 7540](https://tools.ietf.org/html/rfc7540#section-5.2.1) in the flow control section says that the client can request any window size, but it is the receiver who decides what to provide to the client:

> A receiver MAY choose to set any window size that it
       desires for each stream and for the entire connection.  A sender
       MUST respect flow-control limits imposed by a receiver.

Based on this, I think that we just need to use what is given to us, and then again wait for the WindowUpdated event.

sethmlarson: Right, but isn't the current implementation going to wait for any window size update and then send our pre-determined chunk size that we have? If we're going off of what our peer is telling us to use as a window shouldn't we have to recalculate our chunk size whenever we are ready to send the next chunk?

The code as I'm seeing it right now is doing the following:
1. Calculate chunk size once
2. Create an iterator of chunks from the above chunk size. (Not recalculated)
3. If we can't send the current chunk we have then we wait for any WindowUpdated event
4. Even if that WindowUpdated event didn't increase our available window to be large enough for that chunk we send anyways?

nsidnev: Yes, you are right. I changed the implementation of the `.send_data` method, so can you check current version, please.

nsidnev:
@sethmlarson I changed the way data is split into chunks in `.send_data` method. Can you check it please and say if there is anything else that I can fix before merging #217 and changing the tests?

sethmlarson:
@nsidev I can't see anything wrong with the current implementation. :)

nsidnev:
> @nsidnev I can't see anything wrong with the current implementation. :)

That's good, thanks.

I merged changes from the current master using the `BaseEvent` interface that @florimondmanca provided. What's next? This can be merged or wait for the mocks for HTTP/2.

tomchristie:
> These types of issues where our HTTP/2 implementation isn't following standards

Agreed. Can we enumerate those cases?

tomchristie:
> it looks like this PR might need some extra work put into it to get it ready for merge

Could you expand on that?

nsidnev:
@sethmlarson 
Sorry that I did not reply for a while. As I can see, active work has resumed on #153, and now our code is very similar. So maybe this one should be closed so as not to do the same thing twice.

sethmlarson:
@nsidnev No problem! Thanks for checking back in. I'll close this in favor of #153. :)

PR #222, opened 2019-08-17
fixed __repr__ function of Headers and Url classes to not show auth info explicitly.

### LINKED ISSUE #220: Don't render Authorization, Proxy-Authorization headers, password in URLs
We could render them as `[secure]` when calling `__repr__` so that they're not exposed by logging services. Was wondering what others thought of this idea?

cansarigol:
Closes #220

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -210,6 +210,12 @@ class URL:
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
         url_str = str(self)
+        if self._uri_reference.userinfo:
+            url_str = (
+                rfc3986.urlparse(url_str)
+                .copy_with(userinfo=f"{self.username}:[secure]")
+                .unsplit()
+            )
         return f"{class_name}({url_str!r})"
 
 
@@ -489,10 +495,14 @@ class Headers(typing.MutableMapping[str, str]):
         if self.encoding != "ascii":
             encoding_str = f", encoding={self.encoding!r}"
 
-        as_dict = dict(self.items())
-        if len(as_dict) == len(self):
+        sensitive_headers = {"authorization", "proxy-authorization"}
+        as_list = [
+            (k, "[secure]" if k in sensitive_headers else v) for k, v in self.items()
+        ]
+
+        as_dict = dict(as_list)
+        if len(as_dict) == len(as_list):
             return f"{class_name}({as_dict!r}{encoding_str})"
-        as_list = self.items()
         return f"{class_name}({as_list!r}{encoding_str})"
 
 
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -2,6 +2,7 @@ import json
 import os
 
 from httpx import (
+    URL,
     AsyncDispatcher,
     AsyncRequest,
     AsyncResponse,
@@ -100,3 +101,20 @@ def test_trust_env_auth():
     assert response.json() == {
         "auth": "Basic ZXhhbXBsZS11c2VybmFtZTpleGFtcGxlLXBhc3N3b3Jk"
     }
+
+
+def test_auth_hidden_url():
+    url = "http://example-username:example-password@example.org/"
+    expected = "URL('http://example-username:[secure]@example.org/')"
+    assert url == URL(url)
+    assert expected == repr(URL(url))
+
+
+def test_auth_hidden_header():
+    url = "https://example.org/"
+    auth = ("example-username", "example-password")
+
+    with Client(dispatch=MockDispatch()) as client:
+        response = client.get(url, auth=auth)
+
+    assert "'authorization': '[secure]'" in str(response.request.headers)


httpx/models.py
```diff
@@ -482,17 +486,31 @@ def __eq__(self, other: typing.Any) -> bool:
...
         class_name = self.__class__.__name__
 
         encoding_str = ""
         if self.encoding != "ascii":
             encoding_str = f", encoding={self.encoding!r}"
-
-        as_dict = dict(self.items())
-        if len(as_dict) == len(self):
+        clone = self.copy(hide_sensitive_items=True)
+        as_dict = dict(clone.items())
```
sethmlarson: What do you think about grabbing the items and filtering their value based on being included in a list:

ie:
```python
sensitive_headers = {"authorization", "proxy-authorization"}
clone = [(k, "[secure]" if k in sensitive_headers else v for k, v in self.items()]
```

Would this work the same?

httpx/models.py
```diff
@@ -205,7 +206,10 @@ def __eq__(self, other: typing.Any) -> bool:
         return isinstance(other, (URL, str)) and str(self) == str(other)
 
     def __str__(self) -> str:
-        return self._uri_reference.unsplit()
+        url = self._uri_reference.unsplit()
+        if self._uri_reference.userinfo:
+            url = url.replace(self._uri_reference.userinfo, "[secure]")
```
sethmlarson: We might have to reimplement `rfc3986.URIReference.unsplit()` here. Also we shouldn't obscure the username field, just the password field.

cansarigol:
Thanks for help @sethmlarson. I applied your comments

httpx/models.py
```diff
@@ -205,6 +205,13 @@ def __eq__(self, other: typing.Any) -> bool:
...
 
     def __str__(self) -> str:
+        if self._uri_reference.userinfo:
+            userinfo = f"{self.username}:[secure]"
+            return self._uri_reference.copy_with(
+                authority=rfc3986.normalizers.normalize_authority(
+                    (userinfo, self._uri_reference.host, self._uri_reference.port)
+                )
+            ).unsplit()
         return self._uri_reference.unsplit()
```
tomchristie: The `__str__` method shouldn't obscure the URL password. We should only do that for `__repr__`.

(We still want to allow explicit `str()` casting to function normally, we just don't want credentials leaking out of variables representations in tracebacks etc.)

cansarigol: Thanks @tomchristie

httpx/models.py
```diff
@@ -209,7 +209,19 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
-        url_str = str(self)
```
tomchristie: Let's use a proper block `if ... else ...` here. I think it's too opaque otherwise.

httpx/models.py
```diff
@@ -489,11 +501,15 @@ def __repr__(self) -> str:
         if self.encoding != "ascii":
             encoding_str = f", encoding={self.encoding!r}"
 
-        as_dict = dict(self.items())
-        if len(as_dict) == len(self):
+        sensitive_headers = {"authorization", "proxy-authorization"}
+        clone = [
```
tomchristie: Let's call this variable `as_list` so that it mirrors the existing `as_dict`.

tomchristie:
I'm liking this, yup! Coupla minor comments.

cansarigol:
Thanks again @tomchristie, I applied your comments.

httpx/models.py
```diff
@@ -209,7 +209,18 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
-        url_str = str(self)
+        if self._uri_reference.userinfo:
+            url_str = self._uri_reference.copy_with(
```
tomchristie: Can this be simplified to `self._uri_reference.copy_with(userinfo=f"{self.username}:[secure]").unsplit()`?

tomchristie:
General thought: Do we prefer `[secure]` or `********` as the placeholder here?

We use the asterix style in Starlette, and I quite like it as being language independent, and mirroring eg. password input styling. What prior art is there in other places in Python - is there anything similar in eg. werkzeug or urllib or requests, or anything?

httpx/models.py
cansarigol: as far as I understand, we just copy `scheme, authority, path, query, fragment` ([link](https://github.com/python-hyper/rfc3986/blob/2982509402519575282845adf1f39dc0195369e0/src/rfc3986/_mixin.py#L321))

sethmlarson:
Btw I chose `[secure]` just because it's what Travis does to obscure private environment values.

httpx/models.py
sethmlarson: What if `self.username` is `None`?

tomchristie: `copy_with()` supports [a `userinfo=...` argument](https://rfc3986.readthedocs.io/en/latest/user/parsing.html?highlight=userinfo#some-examples).

So, I think this section of code can be...

```python
    if self._uri_reference.userinfo:
        self._uri_reference.copy_with(userinfo=f"{self.username}:[secure]").unsplit()
```

Note that self.username cannot be `None` if some userinfo value is set.

sethmlarson: Nevermind, totally forgot that it's `self.username` not `uri_reference.username`. Carry on! :)

StephenBrown2:
Random comment from the peanut gallery: I'm in favor of `********`, as pydantic uses asterisks for it's Secret types as well: https://pydantic-docs.helpmanual.io/#secret-types

pquentin:
Another data point: [Sentry uses [Filtered]](https://help.sentry.io/hc/en-us/articles/115000154594-Why-am-I-seeing-Filtered-in-my-event-data-). (But `********` is fine.)

cansarigol:
I'm using Elastic APM. it uses `********`

PR #223, opened 2019-08-17
Initial start on HTTP tunnel proxies

### LINKED ISSUE #201: Implement HTTP Tunnel Proxy
Related: #36

sethmlarson:
Closes #201. This is my initial stab at implement HTTP tunnel proxy as a Dispatcher. Will need to implement the `proxies` argument for the `Client` to match Requests.

I'm wondering what the best way to test this implementation would be.


httpx/client.py
```diff
@@ -178,7 +178,8 @@ def merge_headers(
             )
         except HTTPError as exc:
             # Add the original request to any HTTPError
-            exc.request = request
+            if exc.request is None:
```
sethmlarson: This is to not replace the `request` property on a `ProxyError` which is different than the original request.

httpx/concurrency.py
```diff
@@ -225,6 +225,45 @@ def loop(self) -> asyncio.AbstractEventLoop:
 
         return reader, writer, protocol
 
+    async def start_tls(
```
sethmlarson: This is where a lot of the complexity of the functionality comes from. Can we verify that this is possible in a similar way for Trio cc @florimondmanca

florimondmanca: Will need to review this. :) Overall I think trio is very well equipped to deal with I/O situations. So once we get a working asyncio implementation in we’ll mostly have to see how to abstract it away in terms of concurrency backend. What do you think?

florimondmanca: Looking at it a bit more, I’ve already used trio TLS when implementing connect() in #214, but this seems to be something else. Can you confirm what this does is upgrade the existing TCP connection to start using TLS? What do you think of the web proxy example listed here? https://trio.readthedocs.io/en/stable/reference-io.html?highlight=TLS%20upgrade#the-abstract-stream-api

sethmlarson: Yeah this is more than just TLS on a new connection. `.connect()` is for new connections this is for adding TLS to a TCP stream, regardless of whether that TCP stream already has TLS on or not (since this is what occurs when you have an HTTPS proxy and you're connecting to an HTTPS service)

tomchristie:
Crackin' stuff!
Not a detailed review, but some initial thoughts:

* `start_tls` doesn't appear to be getting called into anywhere yet - where will that end up being plugged in?
* It's not clear to me on first pass if it makes more sense to have the make-the-connection implementation take effect on the Dispatcher class, or if it'd actually be cleaner to have the "Make a CONNECT request" be implemented within an alternate `HTTPConnection`.
* If possible we ought to prefer composing implementations rather than inheriting them, since we're being a bit fuzzy about reaching into details of the subclass that we *might* prefer to be able to treat as private API. (Ie. in this cases that'd mean having a dispatcher class which contains an instance of `ConnectionPool` rather than inheriting from it.)

It'll probably be easier to fully review once `start_tls`, and the actually proxy configuration via-the-client stuff is all plumbed in.

httpx/client.py
florimondmanca: I'd be OK with freezing that bit of information as a code comment. :-)

setup.cfg
```diff
@@ -11,7 +11,7 @@ combine_as_imports = True
 force_grid_wrap = 0
 include_trailing_comma = True
 known_first_party = httpx
-known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,nox,pytest,rfc3986,setuptools,trustme,uvicorn
+known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,nox,pytest,rfc3986,setuptools,tests,trustme,uvicorn
```
florimondmanca: I think this is a missing bit on `master` that appeared after #271 — I've seen you fix it in other places and I had to too in #214. Perhaps worth submitting a quick PR as it makes `nox -s lint` crash? I can take it if you'd like. :+1:

florimondmanca: Will do soon. Also mental note that `tests` should actually go in `known_first_party` instead (so that `isort` places it along with other sources instead of topmost like other external packages).

tomchristie:
What's the proposed API for this once it's linked up to the Client?

tomchristie:
> Maybe I can merge the implementation of these two separate proxies into one

From an API point of view I think that having just `HTTPProxy(...)` and `ConnectionPool(...)` as our documented dispatcher classes is nicer than having two different class implementations for proxying yeah. (With a mode control available *if neccessary*?)

tomchristie:
> What are your thoughts on getting the two separate implementations committed and then merging along with wiring up the Client as it's own PR?

Probably would be okay with that since the dispatcher interface is documented already, yeah.

We should probably start exanding out the "Advanced" section of the docs tho.

sethmlarson:
Closed in favor of #259

PR #224, opened 2019-08-17
Fix typos in docstrings

skshetry:
I have fixed some of the typos inside the docstrings, that I found while reading through the source code. 
Also, I changed the comparator inside the `StatusCode` to use chained comparators (suggested by Pycharm), which I can drop it, to only make the PR about typos. :grinning:

### DIFF
--- a/httpx/concurrency.py
+++ b/httpx/concurrency.py
@@ -30,7 +30,7 @@ SSL_MONKEY_PATCH_APPLIED = False
 
 def ssl_monkey_patch() -> None:
     """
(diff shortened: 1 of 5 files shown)


cansarigol:
IMO we should seperate this pr as typo and other fix as you said.

skshetry:
@cansarigol, I have separated those into 2 commits already.  Let's see what others have to say. :slightly_smiling_face:

cansarigol:
I love separated prs by purpose :) you're right

PR #226, opened 2019-08-17
Skip coverage on breaks right before returns for Python 3.8

sethmlarson:
This is a work-around for this bug: nedbat/coveragepy#772 which makes `break` statements not show up in the opcodes if they immediately precede a `return` statement due to peep-hole optimizations. This allows us to say we want 3.8 to always pass on CI. :)

Maybe in the future we should rely on Codecov to aggregate our code coverage so we don't need to do this but that has it's own set of issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,6 @@ matrix:
     - python: 3.8-dev
       env: NOX_SESSION=test-3.8
 
-  allow_failures:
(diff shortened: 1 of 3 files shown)


tomchristie:
Ah, that's a bit of a shame really - bit cronky having to `nocover` those, but good work!

Issue #229, opened 2019-08-17
ConnectionPool not releasing semaphore


oldani:
I'm trying to figure this out and find a fix for this as for right now. To be on context I'm using httpx for microservice communication, some hours ago I started getting `PoolTimeout` error on sentry at first I thought about increasing the `PooLimit` but when I took a look at metrics I was just hitting 0.5 rq/s so I added some logs.

When the error is raised `dispatch.num_connections` is between `4 & 0` while `dispatch.max_connections.semaphore._value` is `0` and the timeout get raised. So somewhere connections are being dropped but the semaphore is not being increased.

sethmlarson:
I'll investigate, is this with 0.7.0?

oldani:
Version is `0.6.8` I'm on it as right now  my guess is around https://github.com/encode/httpx/blob/master/httpx/dispatch/connection_pool.py#L121

oldani:
Yeah, it's actually the `ConnectionPool` I make a path and deployed it's no longer happening I will open a PR.

Basically, when you call `acquire_connection` and a new connection is created you acquire the semaphore but when you pop up that connection and check if it's dropped you need to release the semaphore otherwise it will keep decreasing until it reaches zero and the app starts crashing.

PR #230, opened 2019-08-18
Release semaphore if connection is dropped

oldani:
This is related to #229


oldani:
Will be great to get this deployed asap.

sethmlarson:
Can we get a test case to verify this behavior so we don't regress it later?

sethmlarson:
Thanks for this @oldani! :)

PR #231, opened 2019-08-18
Install package locally in test-requirements.txt

florimondmanca:
`setup.py` already defines the base requirements for `httpx`,  so we can install the package itself for development instead of re-listing those requirements in `test-requirements.txt`. :-)


sethmlarson:
Thanks @florimondmanca!

PR #232, opened 2019-08-18
Enforce that sync client uses asyncio-based backend

florimondmanca:
As suggested in #214 (https://github.com/encode/httpx/pull/214#issuecomment-522241493).

We still need the `backend` argument on `Client` for some of the tests (e.g. the HTTP2 mock backend in `test_http2.py`), so I figured a simple inheritance check would be enough, esp. since the `backend` argument is not documented.

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -70,6 +70,8 @@ class BaseClient:
         if backend is None:
             backend = AsyncioBackend()
 
+        self.check_concurrency_backend(backend)
+
         if app is not None:
             param_count = len(inspect.signature(app).parameters)
             assert param_count in (2, 3)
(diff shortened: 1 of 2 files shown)


sethmlarson:
What do you think about taking away the argument in the constructor but allowing it to be modified?

sethmlarson:
Or I guess there are use-cases for setting it besides what we're doing to test...

sethmlarson:
Thanks @florimondmanca!

florimondmanca:
:+1:

> What do you think about taking away the argument in the constructor but allowing it to be modified?

So what did you decide about that? One extra reason I didn't remove the argument was that we'd have had to redefine the constructor and all of the other parameters in order to keep proper IDE support.

sethmlarson:
I decided against it, it's valid to have other backends they just have to be asyncio. So really we want people going through that interface. :P

PR #237, opened 2019-08-19
added control to calling decode for empty contents.

### LINKED ISSUE #235: Handle HEAD responses with Brotli decoder
Currently if you receive a response with `Content-Encoding: br` set and no body we get an error because Brotli doesn't like being called on an empty stream.

cansarigol:
Closes #235

### DIFF
--- a/httpx/decoders.py
+++ b/httpx/decoders.py
@@ -98,8 +98,12 @@ class BrotliDecoder(Decoder):
             brotli is not None
         ), "The 'brotlipy' or 'brotli' library must be installed to use 'BrotliDecoder'"
         self.decompressor = brotli.Decompressor()
+        self.seen_data = False
 
     def decode(self, data: bytes) -> bytes:
+        if not data:
+            return b""
+        self.seen_data = True
         try:
             if hasattr(self.decompressor, "decompress"):
                 return self.decompressor.decompress(data)
@@ -108,6 +112,8 @@ class BrotliDecoder(Decoder):
             raise DecodingError from exc
 
     def flush(self) -> bytes:
+        if not self.seen_data:
+            return b""
         try:
             if hasattr(self.decompressor, "finish"):
                 self.decompressor.finish()
--- a/tests/test_decoders.py
+++ b/tests/test_decoders.py
@@ -4,7 +4,13 @@ import brotli
 import pytest
 
 import httpx
-from httpx.decoders import TextDecoder
+from httpx.decoders import (
+    BrotliDecoder,
+    DeflateDecoder,
+    GZipDecoder,
+    IdentityDecoder,
+    TextDecoder,
+)
 
 
 def test_deflate():
@@ -77,6 +83,22 @@ def test_streaming():
     response = httpx.Response(200, headers=headers, content=compress(body))
     assert not hasattr(response, "body")
     assert response.read() == body
+
+
+@pytest.mark.parametrize("header_value", (b"deflate", b"gzip", b"br", b"identity"))
+def test_empty_content(header_value):
+    headers = [(b"Content-Encoding", header_value)]
+    response = httpx.Response(200, headers=headers, content=b"")
+    assert response.content == b""
+
+
+@pytest.mark.parametrize(
+    "decoder", (BrotliDecoder, DeflateDecoder, GZipDecoder, IdentityDecoder)
+)
+def test_decoders_empty_cases(decoder):
+    instance = decoder()
+    assert instance.decode(b"") == b""
+    assert instance.flush() == b""
 
 
 @pytest.mark.parametrize("header_value", (b"deflate", b"gzip", b"br"))


tomchristie:
On reviewing this, I'm not sure my "More generally: we should make sure not to call any decoder if the body is empty." comment is the best - approach.

Your PR here fixes one case, but we'd also need to introduce the same for the streaming case.

It's probably more pragmatic for us to make sure that the brotli decoder implementation (and any others) all correctly deal with `""` inputs, and all correctly return `""` for the output in that case.

sethmlarson:
Totally agree with Tom here. Decoders shouldn't explode on empty input.

httpx/models.py
```diff
@@ -710,7 +710,8 @@ def content(self) -> bytes:
             if hasattr(self, "_raw_content"):
                 raw_content = self._raw_content  # type: ignore
                 content = self.decoder.decode(raw_content)
-                content += self.decoder.flush()
```
tomchristie: This line ought to stay the same. If there's an issue with a decoder erroring at the point of `flush()` being called against an empty buffer, then let's fix it on the decoder implementation.

tests/test_decoders.py
```diff
@@ -79,6 +85,18 @@ def compress(body):
...
+@pytest.mark.parametrize("header_value", (b"deflate", b"gzip", b"br", b"identity"))
+def test_empty_content(header_value):
+    headers = [(b"Content-Encoding", header_value)]
+    response = httpx.Response(200, headers=headers, content=b"")
+    assert response.content == b""
+
+
+def test_decoders_empty_cases():
+    for decoder in {BrotliDecoder, DeflateDecoder, GZipDecoder, IdentityDecoder}:
+        assert decoder().decode(b"") == b""
```
tomchristie: Let's change this to cover both `decode()` and `flush()`. So, for example...

```python
instance = decoder()
assert instance.decode(b"") == b""
assert instance.flush() == b""
```

httpx/models.py
cansarigol: I couldn't figure out how to handle it in decoder. Is this make sense?

```
def flush(self) -> bytes:
        try:
            if hasattr(self.decompressor, "finish"):
                self.decompressor.finish()
            return b""
        except brotli.error as exc:  # pragma: nocover
            if str(exc) == "Decompression error: incomplete compressed stream.":
                return b""
            raise DecodingError from exc
```

cansarigol: it comes from [this line](https://github.com/python-hyper/brotlipy/blob/ffddf2ea5adc584c8c353d246bb1077b7e781b63/src/brotli/brotli.py#L463). we can't use `BrotliDecoderIsFinished`. it is not public

tests/test_decoders.py
florimondmanca: My two cents here: consider using parametrization to have one test *per decoder* instead of a single test. Makes it easy to know and debug which decoder is broken when it happens. :-)

```python
@pytest.mark.parametrize("decoder", [BrotliDecoder(), DeflateDecoder(), ...])
def test_decoders_empty_cases(decoder):
    assert decoder().decode(b"") == b""
    assert decoder.flush() == b""
```

cansarigol: super you're right. I forgot. 
Is there any difference between `BrotliDecoder()` and `BrotliDecoder` using?

httpx/models.py
sethmlarson: I think we can implement this by just having a flag that tracks whether the decoder actually received any data and if not just return `b""`.

cansarigol: `def decode(self, data: bytes, with_flush: bool=False) -> bytes`

httpx/decoders.py
```diff
@@ -100,6 +100,8 @@ def __init__(self) -> None:
         self.decompressor = brotli.Decompressor()
```
tomchristie: Let's add `self.seen_data = False` here, then...

httpx/decoders.py
```diff
@@ -100,6 +100,8 @@ def __init__(self) -> None:
...
 
     def decode(self, data: bytes) -> bytes:
+        if not data:
+            return data
```
tomchristie: We can add `self.seen_data = True` here...

httpx/decoders.py
```diff
@@ -113,6 +115,8 @@ def flush(self) -> bytes:
                 self.decompressor.finish()
             return b""
         except brotli.error as exc:  # pragma: nocover
+            if str(exc) == "Decompression error: incomplete compressed stream.":
```
tomchristie: Then, rather than trying to guess if we've seen a null-data error we can do...

```python
def flush(self) -> bytes:
    if not self.seen_data:
        return b""
    ...
```

tests/test_decoders.py
florimondmanca: > Is there any difference between BrotliDecoder() and BrotliDecoder using?

Depends on whether instanciating the decoder performs any kind of stateful stuff? If that is the case you can pass the class and create an `instance` as in Tom's snippet.

cansarigol:
I added a [commit](https://github.com/encode/httpx/pull/237/commits/8f73ce73802e916526e8fc90d213636d17b07f12) Am I continuing with seen_data?

tests/test_decoders.py
cansarigol: After asked, I needed to add `MultiDecoder`. 
`instance = decoder([IdentityDecoder()]) if decoder == MultiDecoder else decoder()`

tomchristie:
> Am I continuing with seen_data?

Please, yup.

The `httpx/models.py` file should not change as part of this PR, and the signature of `def decode(self, data: bytes)` should stay the same too.

httpx/decoders.py
```diff
@@ -99,20 +113,30 @@ def __init__(self) -> None:
...
         self.decompressor = brotli.Decompressor()
 
-    def decode(self, data: bytes) -> bytes:
-        try:
-            if hasattr(self.decompressor, "decompress"):
-                return self.decompressor.decompress(data)
-            return self.decompressor.process(data)  # pragma: nocover
-        except brotli.error as exc:
-            raise DecodingError from exc
+    def decode(self, data: bytes, with_flush: bool = False) -> bytes:
```
tomchristie: I'd strongly prefer not indenting everything in the body here, but instead just adding an early return in the empty case...

```python
def decode(self, data: bytes) -> bytes:
    if not data:
        return b""
... # Rest of the method unchanged.
```

tests/test_decoders.py
```diff
@@ -79,6 +86,23 @@ def compress(body):
...
+@pytest.mark.parametrize("header_value", (b"deflate", b"gzip", b"br", b"identity"))
+def test_empty_content(header_value):
+    headers = [(b"Content-Encoding", header_value)]
+    response = httpx.Response(200, headers=headers, content=b"")
+    assert response.content == b""
+
+
+@pytest.mark.parametrize(
+    "decoder",
+    (BrotliDecoder, DeflateDecoder, GZipDecoder, IdentityDecoder, MultiDecoder),
```
tomchristie: Don't really need to include `MutliDecoder` here - that's a different type of case.

tomchristie:
Updated the issue description to use [GitHub's "closes" keyword](https://help.github.com/en/articles/closing-issues-using-keywords).

tomchristie:
Loverly stuff. Yes please!

PR #240, opened 2019-08-19
Add logging to h2 connection

yeraydiazdiaz:
Also bump h2 version to `>=3.1.1`.

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -1,5 +1,6 @@
 import functools
 import typing
+import logging
 
 import h2.connection
 import h2.events
@@ -8,6 +9,8 @@ from ..concurrency import TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
 from ..interfaces import BaseReader, BaseWriter, ConcurrencyBackend
 from ..models import AsyncRequest, AsyncResponse
+
+logger = logging.getLogger(__name__)
 
 
 class HTTP2Connection:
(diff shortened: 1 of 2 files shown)


sethmlarson:
I was thinking we'd have DEBUG for everything besides data transfer and then set TRACE for readers / writers with raw data going in and out. What do you think about that?

yeraydiazdiaz:
Sorry, I'm not sure what you mean with setting TRACE of readers and writers.

sethmlarson:
Basically the raw data getting sent/received over streams should be `logger.trace()` and everything else should be `logger.debug()` because raw data is pretty verbose.

I had an example logging setup on a different dispatcher that I'd like everyone to look at and see what they think.

yeraydiazdiaz:
That's what I thought you meant but I there's no `trace` method in [logger](https://docs.python.org/3.7/library/logging.html#logging.Logger)? Or am I missing something?

florimondmanca:
@yeraydiazdiaz You're right, although [it has been suggested before](https://bugs.python.org/issue31732).

Logging levels are extensible though — example inspired from [this comment](https://bugs.python.org/issue31732#msg304304):

```python
# logging.py
import logging
TRACE = 5
logging.addLevelName(TRACE, 'TRACE')
```

Then you could write:

```python
from ..logging import TRACE

logger.log(TRACE, "Stream [%d] - Received response events: %r", stream_id, data)
```

Is this the kind of setup you were thinking about @sethmlarson?

PR #241, opened 2019-08-19
Allow configuration of supported HTTP protocol versions.

tomchristie:
Initial pass at adding `protocols=...` support.

The intent here is to be able to support usage such as `Client(http_versions="HTTP/1.1")` or `Client(http_versions=["HTTP/1.1", "HTTP/2"])` (as well as per-request overrides).

- [x] Add coverage
- [x] Support `Client(http_versions=...)`
- [ ] ~Support `.request(http_versions=...)`~ - Actually we may not even want per-request control here. Even if we do then let's treat it as an incremental PR.
- [ ] ~Drop `Protocol` str enum in favor of plain strings.~ - Going to defer this to a seperate pull request.
- [ ] ~Assert valid protocol type returned after connect.~ - Going to defer this to a seperate pull request.


httpx/config.py
```diff
@@ -223,6 +226,40 @@ def __repr__(self) -> str:
...
+            self.protocols = protocols.protocols
+        else:
+            self.protocols = set(sorted(protocols))
+
+        for protocol in self.protocols:
+            if protocol not in ('HTTP/1.1', 'HTTP/2'):
+                raise ValueError(f"Unsupported protocol value {protocol!r}")
+
+    @property
+    def protocol_ident_strings(self) -> typing.List[str]:
```
sethmlarson: Not a big fan of this name. Can we somehow get `alpn` or `identifier` instead of `ident`?

tomchristie: Yup good call

tomchristie:
> Thoughts on being more explicit with HTTPVersion and http_versions=...?

Yup. Updated in line with that! 👍

httpx/config.py
```diff
@@ -236,15 +236,18 @@ def __init__(self, http_versions: HTTPVersionTypes = None):
             http_versions = ['HTTP/1.1', 'HTTP/2']
 
         if isinstance(http_versions, str):
-            self.http_versions = set([http_versions])
+            self.http_versions = set([http_versions.upper()])
         elif isinstance(http_versions, HTTPVersionConfig):
             self.http_versions = http_versions.http_versions
         else:
-            self.http_versions = set(sorted(http_versions))
+            self.http_versions = set(sorted([version.upper() for version in http_versions]))
```
sethmlarson: You can remove the `sorted` call if it's going into a `set` anyways.

tomchristie:
I've narrowed down the scope of this pull requests slightly, so that it *strictly* only involves getting us over the line, with some smaller incremental improvements that can happen seperately.

sethmlarson:
Sounds good, mark as non-draft when you're ready for a final review :)

tomchristie:
I've further de-scoped this so that it is *only* for `Client(http_versions=...)` support, and does not also include per-request configuration. If we decide that we want that too then there's no reason not to have it as an incremental PR, but it's not obvious if we'd *actually* want it available or not.

tomchristie:
Shiny things:

```shell
>>> import httpx
>>> c = httpx.Client(http_versions='HTTP/1.1')
>>> r = c.get('https://www.google.com')
>>> r.protocol
'HTTP/1.1'
>>> c = httpx.Client(http_versions='HTTP/2')
>>> r = c.get('https://www.google.com')
>>> r.protocol
'HTTP/2'
```

We'll want to reconsider the existing `r.protocol` API, presumably rename it to `.http_version`, but I think I'm good with keeping the scope of our pull requests always as narrow as possible, and this work seems like the smallest possible set of coherent valid feature work that we can get in to start with.

tomchristie:
Follow ups would be:

* `r.protocol` - > `r.http_version` for consistency with our new `Client(http_version=...)` argument.
* Drop the existing `Protocol` str enum in favor of plain old strings, for a smaller API surface area.
* Possibly (but not neccessarily) ensure valid protocol types are returned after a connect?
* Perhaps support per-request `http_versions`? (I'm probably leaning against this myself)

sethmlarson:
I'm 100% :+1: to your above comment bullet points.

httpx/config.py
```diff
@@ -72,27 +75,35 @@ def with_overrides(
             return self
         return SSLConfig(cert=cert, verify=verify)
 
-    def load_ssl_context(self) -> ssl.SSLContext:
+    def load_ssl_context(
```
sethmlarson: What are your thoughts on making http_versions a part of the SSLConfig object instead of being piped through?

Would also include updating `__eq__` and hash too.

httpx/client.py
```diff
@@ -57,6 +58,7 @@ def __init__(
         cookies: CookieTypes = None,
         verify: VerifyTypes = True,
         cert: CertTypes = None,
+        http_versions: HTTPVersionTypes = None,
```
sethmlarson: Should we have an explicit DEFAULT_... here instead of None?

httpx/config.py
```diff
@@ -223,6 +236,43 @@ def __repr__(self) -> str:
...
+
+        for version in self.http_versions:
+            if version not in ("HTTP/1.1", "HTTP/2"):
+                raise ValueError(f"Unsupported HTTP version {version!r}.")
+
+        if not self.http_versions:
+            raise ValueError(f"HTTP versions cannot be an empty list.")
+
+    @property
+    def alpn_strings(self) -> typing.List[str]:
```
sethmlarson: Thoughts on naming this `alpn_identifiers`?

httpx/config.py
```diff
@@ -223,6 +236,43 @@ def __repr__(self) -> str:
...
+        """
+        Returns a list of supported ALPN identifiers. (One or more of "http/1.1", "h2").
+        """
+        mapping = {"HTTP/1.1": "http/1.1", "HTTP/2": "h2"}
```
sethmlarson: Should this mapping be in global scope so it can be used to verify whether a http_version is valid as well? Then only one place needs to be updated each time.

httpx/config.py
```diff
@@ -223,6 +236,43 @@ def __repr__(self) -> str:
         )
 
 
+class HTTPVersionConfig:
```
sethmlarson: We should probably implement equality and hash for this object to implement the SSLConfig comment.

httpx/config.py
tomchristie: I don't know. That gets a bit different once we're approaching HTTP/3, or if we ever decided to support HTTP/2 upgrades over http. My feeling is that it's a "possibly", and that we should treat it as a later refactoring if we did decide to do so.

httpx/client.py
tomchristie: For the `None` cases I don't think so, no. We're being consistent with the surrounding context.

httpx/config.py
tomchristie: Yup. I'll sort it. 👍

httpx/config.py
tomchristie: Sure thing, yup.

httpx/config.py
tomchristie: We *might* want contains, iteration, len, equality, and hash on this class. I figure we should probably defer any decisions there vs. this pull request, since it's a bit unclear where we'd want to draw the line.

PR #250, opened 2019-08-20
Response.protocol -> Response.http_version

tomchristie:
Related to #241.


tomchristie:
Indeed - thanks!
I'd checked in the quickstart and hadn't found it there, but it's actually on the homepage, and the API docs.

PR #253, opened 2019-08-20
added test for HTTPVersionConfig type error

cansarigol:
Related #241

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -251,8 +251,16 @@ class HTTPVersionConfig:
             self.http_versions = {http_versions.upper()}
         elif isinstance(http_versions, HTTPVersionConfig):
             self.http_versions = http_versions.http_versions
+        elif isinstance(http_versions, typing.Iterable):
+            self.http_versions = {
+                version.upper() if isinstance(version, str) else version
+                for version in http_versions
+            }
         else:
-            self.http_versions = {version.upper() for version in http_versions}
+            raise TypeError(
+                f"HTTP version should be a string or list of strings, "
+                "but got {type(http_versions)}"
+            )
 
         for version in self.http_versions:
             if version not in ("HTTP/1.1", "HTTP/2"):
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -115,6 +115,16 @@ def test_invalid_http_version():
         httpx.HTTPVersionConfig("HTTP/9")
 
 
+def test_invalid_http_version_type():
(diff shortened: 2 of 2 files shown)


httpx/config.py
```diff
@@ -251,8 +252,13 @@ def __init__(self, http_versions: HTTPVersionTypes = None):
...
         elif isinstance(http_versions, HTTPVersionConfig):
             self.http_versions = http_versions.http_versions
+        elif isinstance(http_versions, Iterable):
+            self.http_versions = {
+                version.upper() if isinstance(version, str) else version
+                for version in http_versions
+            }
         else:
-            self.http_versions = {version.upper() for version in http_versions}
+            raise ValueError(f"Unsupported HTTP version {http_versions!r}.")
```
tomchristie: I guess we should probably raise a type error instead. Something like...

`TypeError("HTTP version should be a string or list of strings, but got {type(http_versions)}")`?

httpx/config.py
```diff
@@ -1,5 +1,6 @@
 import ssl
 import typing
+from collections.abc import Iterable
```
StephenBrown2: Why `collections.abc.Iterable` instead of `typing.Iterable`?
https://docs.python.org/3/library/typing.html#typing.Iterable

httpx/config.py
```diff
@@ -251,8 +252,15 @@ def __init__(self, http_versions: HTTPVersionTypes = None):
...
             self.http_versions = http_versions.http_versions
+        elif isinstance(http_versions, Iterable):
+            self.http_versions = {
+                version.upper() if isinstance(version, str) else version
+                for version in http_versions
+            }
         else:
-            self.http_versions = {version.upper() for version in http_versions}
+            raise TypeError(
+                f"HTTP version should be a string or list of strings, but got {type(http_versions)}"
```
StephenBrown2: You can break this into two strings to appease [flake8](https://travis-ci.org/encode/httpx/jobs/574508042#L214):
```suggestion
                "HTTP version should be a string or list of strings,"
                f" but got {type(http_versions)}"
```

httpx/config.py
cansarigol: I didn't think to use types like this :) thanks

httpx/config.py
cansarigol: thanks, I overlooked flake8 error. if flake8 is not successful, nox can be stopped.

PR #255, opened 2019-08-20
Unify BaseReader and BaseWriter as BaseStream

florimondmanca:
Based on a suggestion from @tomchristie in https://github.com/encode/httpx/pull/217#issuecomment-522571545, simplifies the stream interface by merging `BaseReader` and `BaseWriter` into a unified stream interface.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -1,5 +1,5 @@
 """
-The `Reader` and `Writer` classes here provide a lightweight layer over
+The `Stream` class here provides a lightweight layer over
 `asyncio.StreamReader` and `asyncio.StreamWriter`.
 
 Similarly `PoolSemaphore` is a lightweight layer over `BoundedSemaphore`.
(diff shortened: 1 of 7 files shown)


httpx/concurrency/asyncio.py
```diff
@@ -80,11 +79,15 @@ def set_write_timeouts(self) -> None:
         self.raise_on_write_timeout = True
 
 
-class Reader(BaseReader):
+class Stream(BaseStream):
     def __init__(
-        self, stream_reader: asyncio.StreamReader, timeout: TimeoutConfig
-    ) -> None:
+        self,
+        stream_reader: asyncio.StreamReader,
```
StephenBrown2: A couple `DeprecationWarning`s show up in Travis:
```
DeprecationWarning: StreamReader is deprecated since Python 3.8 in favor of Stream, and scheduled for removal in Python 3.10
DeprecationWarning: StreamWriter is deprecated since Python 3.8 in favor of Stream, and scheduled for removal in Python 3.10
```
Should those be addressed in this PR?

florimondmanca: Good catch, thanks! :+1: I believe this is independant (warnings also show up on the master build), so it can be addressed in a separate PR — I tracked it in #256.

tomchristie:
Great stuff!

I think that our stream interface should also encapsulate the http version info, (eg `get_protocol_info() -> str) so that instead of connect returning a two tuple of (Stream, str), it can just return Stream.

florimondmanca:
Oh sure, good idea — didn't see that the protocol refactor got in. :+1:

PR #263, opened 2019-08-21
Add ConcurrencyBackend.start_tls()

sethmlarson:
This primitive will be used for HTTP CONNECT tunnel proxies to apply TLS onto either a raw socket stream or an existing TLS stream.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -205,7 +205,7 @@ class AsyncioBackend(ConcurrencyBackend):
         loop = self.loop
         if not hasattr(loop, "start_tls"):  # pragma: no cover
             raise NotImplementedError(
-                "AbstractEventLoop.start_tls() available in Python 3.7+"
+                "asyncio.AbstractEventLoop.start_tls() " "available in Python 3.7+"
             )
 
         assert isinstance(stream, Stream)
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -18,8 +18,8 @@ from ..config import PoolLimits, TimeoutConfig
 from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
 from .base import (
     BaseBackgroundManager,
-    BasePoolSemaphore,
     BaseEvent,
+    BasePoolSemaphore,
     BaseQueue,
     BaseStream,
     ConcurrencyBackend,
@@ -193,6 +193,44 @@ class AsyncioBackend(ConcurrencyBackend):
         return Stream(
             stream_reader=stream_reader, stream_writer=stream_writer, timeout=timeout
         )
+
+    async def start_tls(
+        self,
+        stream: BaseStream,
+        hostname: str,
+        ssl_context: ssl.SSLContext,
+        timeout: TimeoutConfig,
+    ) -> BaseStream:
+
+        loop = self.loop
+        if not hasattr(loop, "start_tls"):  # pragma: no cover
+            raise NotImplementedError(
+                "AbstractEventLoop.start_tls() available in Python 3.7+"
+            )
+
+        assert isinstance(stream, Stream)
+
+        stream_reader = asyncio.StreamReader()
+        protocol = asyncio.StreamReaderProtocol(stream_reader)
+        transport = stream.stream_writer.transport
+
+        loop_start_tls = loop.start_tls  # type: ignore
+        transport = await asyncio.wait_for(
+            loop_start_tls(
+                transport=transport,
+                protocol=protocol,
+                sslcontext=ssl_context,
+                server_hostname=hostname,
+            ),
+            timeout=timeout.connect_timeout,
+        )
+
+        stream_reader.set_transport(transport)
+        stream.stream_reader = stream_reader
+        stream.stream_writer = asyncio.StreamWriter(
+            transport=transport, protocol=protocol, reader=stream_reader, loop=loop
+        )
+        return stream
 
     async def run_in_threadpool(
         self, func: typing.Callable, *args: typing.Any, **kwargs: typing.Any
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -205,7 +205,7 @@ class AsyncioBackend(ConcurrencyBackend):
         loop = self.loop
         if not hasattr(loop, "start_tls"):  # pragma: no cover
             raise NotImplementedError(
-                "asyncio.AbstractEventLoop.start_tls() " "available in Python 3.7+"
+                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
             )
 
         assert isinstance(stream, Stream)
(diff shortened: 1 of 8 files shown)


sethmlarson:
It looks like there were a lot of changes related to import order in this PR from me running `nox -s lint`? Maybe they weren't picked up in other PRs.

tests/test_concurrency.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    backend = AsyncioBackend()
+    ctx = SSLConfig().load_ssl_context_no_verify(HTTPVersionConfig())
+    timeout = TimeoutConfig(5)
+
+    stream = await backend.connect("127.0.0.1", 8001, None, timeout)
+    assert stream.is_connection_dropped() is False
+
+    stream = await backend.start_tls(stream, "127.0.0.1", ctx, timeout)
+    assert stream.is_connection_dropped() is False
```
florimondmanca: Is there anything we can do to verify we are indeed using TLS at this stage?

httpx/concurrency/asyncio.py
```diff
@@ -194,6 +194,44 @@ def loop(self) -> asyncio.AbstractEventLoop:
...
+        stream: BaseStream,
+        hostname: str,
+        ssl_context: ssl.SSLContext,
+        timeout: TimeoutConfig,
+    ) -> BaseStream:
+
+        loop = self.loop
+        if not hasattr(loop, "start_tls"):  # pragma: no cover
+            raise NotImplementedError(
+                "asyncio.AbstractEventLoop.start_tls() " "available in Python 3.7+"
```
florimondmanca: Just a bit of phrasing here:

```suggestion
                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
```

tests/test_concurrency.py
```diff
@@ -0,0 +1,29 @@
...
+
+import pytest
+
+from httpx import AsyncioBackend, HTTPVersionConfig, SSLConfig, TimeoutConfig
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 7),
+    reason="Requires Python 3.7+ for AbstractEventLoop.start_tls()",
+)
```
florimondmanca: Perhaps use `pytest.mark.xfail` instead to make sure the test *does* fail on 3.6?

tests/test_concurrency.py
sethmlarson: We could check ALPN probably?

httpx/concurrency/asyncio.py
```diff
@@ -194,6 +194,44 @@ def loop(self) -> asyncio.AbstractEventLoop:
...
+        stream: BaseStream,
+        hostname: str,
+        ssl_context: ssl.SSLContext,
+        timeout: TimeoutConfig,
+    ) -> BaseStream:
+
+        loop = self.loop
+        if not hasattr(loop, "start_tls"):  # pragma: no cover
+            raise NotImplementedError(
+                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
```
yeraydiazdiaz: Are we fine not supporting this on 3.6?

I'm not familiar with the workings of TLS but looking around in aiohttp it seems [they use `loop.create_connection`](https://github.com/aio-libs/aiohttp/blob/8742305de4ca7b159dc452697802e9a05309f298/aiohttp/connector.py#L1061). Since we're creating the stream reader and writer here maybe we can do something similar with [`asyncio.open_connection`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection) and avoid using `start_tls`?

Sorry if I'm missing something obvious, like I said I'm no expert in TLS 🙂

florimondmanca: From what I understand, start_tls allows to « upgrade » an existing connection to be encrypted when it was not previously — notice how we reuse the stream’s transport in the code here.

There may be a workaround, but at first glance I would guess that basing that off open_connection would open another socket, which is not what we want (we want to keep the existing connection and upgrade it over TLS).

From a Python compatibility point of view, the main implication is that HTTP Tunnel Proxy would only be available on 3.7, correct @sethmlarson? (Then I’m thinking maybe it will be worth catching the exception raised here to provide a more user-friendly message.)

sethmlarson: @florimondmanca is correct, this method takes any steam and performs the TLS handshake, even if it already has TLS active!

And you're right that we won't be able to support the HTTP tunnel on Python 3.6 with the asyncio backend. The warning message can recommend either upgrading Python or using the Trio backend perhaps.

sethmlarson:
Codecov is saying 100% for this branch: https://codecov.io/gh/encode/httpx/pull/263
Would appreciate another round of review for merging. :)

tests/test_concurrency.py
```diff
@@ -0,0 +1,31 @@
...
+    ctx = SSLConfig().load_ssl_context_no_verify(HTTPVersionConfig())
+    timeout = TimeoutConfig(5)
+
+    stream = await backend.connect("127.0.0.1", 8001, None, timeout)
+    assert stream.is_connection_dropped() is False
+    assert stream.stream_writer.get_extra_info("cipher", default=None) is None
+
+    stream = await backend.start_tls(stream, "127.0.0.1", ctx, timeout)
+    assert stream.is_connection_dropped() is False
+    assert stream.stream_writer.get_extra_info("cipher", default=None) is not None
```
florimondmanca: Looks like a good way to check that we're using TLS. Could this be turned into a new method on `ConcurrencyBackend`, though, e.g. `.is_encrypted()`, or `.has_tls()`?

httpx/concurrency/asyncio.py
```diff
@@ -194,6 +194,44 @@ def loop(self) -> asyncio.AbstractEventLoop:
...
+
+        assert isinstance(stream, Stream)
+
+        stream_reader = asyncio.StreamReader()
+        protocol = asyncio.StreamReaderProtocol(stream_reader)
+        transport = stream.stream_writer.transport
+
+        loop_start_tls = loop.start_tls  # type: ignore
+        transport = await asyncio.wait_for(
+            loop_start_tls(
```
florimondmanca: Out of curiosity: why is it necessary to store `loop.start_tls` in a separate variable before using it here?

sethmlarson: Mypy complains about AbstractEventLoop not having start_tls() and I want type coverage for the rest of that line.

tests/test_concurrency.py
sethmlarson: I know we've got a lot of effort going on for generalizing test cases but i think we should keep the low level test cases from deciding our ConcurrencyBackend API. What do you think?

florimondmanca: Okay. I suppose we’ll be able to parametrize how to perform this operation within the test file without touching the concurrency backend anyway. :-)

httpx/concurrency/asyncio.py
florimondmanca: Fine! 😄

sethmlarson:
Thanks for the review @florimondmanca!

PR #264, opened 2019-08-22
Rely on codecov for 100% project coverage

sethmlarson:
As can be seen in #263 we can't have all code be covered by Python 3.6 and I don't want to mark that large chunk of code as `no cover`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,5 @@ script:
 after_script:
   - if [ -f .coverage ]; then
       python -m pip install codecov;
-      codecov;
+      codecov --required;
     fi
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+    status:
+        patch:
+            default:
+                target: '100'
+        project:
+            default:
+                target: '100'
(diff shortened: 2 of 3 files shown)


sethmlarson:
cc @tomchristie looks like codecov needs API tokens for this public repo now?? Not sure what that's about.

pquentin:
codecov is currently broken: https://github.com/codecov/codecov-bash/issues/198 (I noticed this on two other projects)

sethmlarson:
Thanks @pquentin!

pquentin:
... and it's back!

sethmlarson:
Not back for us apparently :(

pquentin:
Well, the part that was broken got fixed, see https://codecov.io/gh/encode/httpx/tree/9229e4591f12b13aa1524a525519e291d3574b72 which reports 100%. I don't know why there was no comment here though!

sethmlarson:
I think I may merge this and see if that solves our codecov comment problem? Either way all maintainers (@encode/httpx-maintainers) be aware of coverage check codecov on each PR before merging until this issue is resolved.

PR #265, opened 2019-08-22
executor for io/cpu bound projects

cansarigol:
Hi, I was wondering that how to use `ProcessPoolExecutor, ThreadPoolExecutor` in HTTPX. Does this approach can follow httpx's philosophy, or should users handle it in their codes?

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -12,14 +12,15 @@ import asyncio
 import functools
 import ssl
 import typing
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from types import TracebackType
 
 from ..config import PoolLimits, TimeoutConfig
 from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
 from .base import (
     BaseBackgroundManager,
-    BasePoolSemaphore,
     BaseEvent,
+    BasePoolSemaphore,
     BaseQueue,
     BaseStream,
     ConcurrencyBackend,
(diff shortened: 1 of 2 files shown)


florimondmanca:
Hey,

From what I understand the ability of the concurrency backend to run things in the threadpool mostly exists to support sync dispatchers (and WSGI apps).

So I don't know if there's a pressing need to be able to run things in the process pool, because in most cases web apps *are* I/O-bound applications. And as you noted in the code, this would also mean we'd need to make sure we don't use processes while iterating in the pool, as it would be terribly inefficient.

That said, we *could* consider supporting customizing the threadpool, e.g.:

```python
# User code
backend = AsyncioBackend(executor_factory=lambda: ThreadPoolExecutor(max_workers=4))
client = AsyncClient(backend=backend)

# asyncio.py
class AsyncioBackend:
    def __init__(self, executor_factory=lambda: None):
        self.executor_factory = executor_factory

    async def run_in_threadpool(self, func, *args, **kwargs):
        # ...
        loop.run_in_executor(self.executor_factory(), func, *args)
```

@cansarigol Were you just playing around with idea of customizing executors, or did you have a more specific use case in mind?

cansarigol:
Hi @florimondmanca, I just wanted to add a new feature for CPU bound or max_worker limited projects. I don't have any use case. I agree with you on web apps as you explained. we can just even max_worker. it would be easy.

florimondmanca:
@cansarigol Okay, thanks for your answer. :) I think it’s good that we know users might want to fine tune how sync things are run, but at this early stage in the life of HTTPX it’s probably best to not add functionality that we don’t know will actually be used. I’ll close this, but if you encounter the use case feel free to come back here and we’ll see what we can do. :) Thanks!

PR #267, opened 2019-08-22
Dispatcher middlewares

yeraydiazdiaz:
Alternative to #176 inspired by [a comment from @tomchristie](https://github.com/encode/httpx/pull/176#issuecomment-523509780)

I quite like this approach, it's nice and simple. I only see two caveats:

1. The "middleware" dispatches have different initializers than "pure" dispatchers requiring a `next_dispatcher` as an argument. It's not *really* a problem, just find the fact that we're deriving from `AsyncDispatch` when we're actually using duck/structural typing a bit misleading. Maybe we could formalize it somehow?
2. The logic to "resolve" the dispatcher can get out of hand and is a mixture of client attributes and `send` arguments. That being said the auth logic is probably a lot more complicated than other usages we might have.

I'm keen to hear if this approach would help implementing features like proxy support. I think redirection and auth look a lot simpler this way.


florimondmanca:
I think there might be ways to reduce boilerplate and improve modularity of middleware dispatchers, but overall this is looking very encouraging! I’ll check out the code and play with it. :-)

florimondmanca:
Closed in favor of #295

PR #268, opened 2019-08-22
Dispatcher middlewares (refactored)

### LINKED ISSUE #295: Client middleware API
There's been quite a few PRs trying to take a stab at making the `Client`/`AsyncClient` extensible, mostly with a middleware approach whose idea seems to have originated in https://github.com/encode/httpx/pull/136#issuecomment-516085808 to help solve Digest Auth.

As mentioned in https://github.com/encode/httpx/pull/268#issuecomment-525347835 it might be useful to define and discuss in which direction we'd like to go.

## Problem statement

A few requirements we may want to satisfy, gathered from looking at the various discussions:

1. Support the simplest case: modify the request (before getting a response from the underlying middleware).
1. Support middleware that may return an early response (e.g. a cache middleware), i.e. allow to skip calling the underlying middleware.
1. Support middleware that need to make subsequent requests after getting the initial response (e.g. Redirect, Digest Auth).
1. Allow users to write their own middleware, e.g. for particular authentication schemes.
1. Allow users to assemble a middleware pipeline (with what level of control?).

None of the approaches attempt to solve 4) nor 5) — and that can definitely be dealt with in its own issue once this one is resolved. As a result, the scope of this issue is **designing an internal API to solve 1), 2) and 3)**.

## Existing implementation

A reminder of the features that HTTPX currently supports that may benefit from being refactored as as middleware:

- Basic authentication: add an `Authorization` header based on a `username` and `password`.
- "Custom authentication": process the request via a `req -> req` callable. (I personally think this might be too lose — we could just require the callable to return an `Authorization` header — but let's tighten the scope of this issue and keep things as they are.)
- Redirect: automatically handle redirect responses by making request to the specified location.

These are currently implemented as part of the `BaseClient` itself. Them cluttering the `BaseClient` is, as far as I understand, the main reason for refactoring them into loosely-coupled middleware.

## Proposed solutions

Here's a summary of the various APIs that were proposed up to now and their respective "no-op middleware" for the sake of example:

### #176 : process_request / process_response

```python
class Middleware:
    async def process_request(self, request: AsyncRequest) -> AsyncRequest:
        return request

    async def process_response(
        self, request: AsyncRequest, response: AsyncResponse
    ) -> AsyncResponse:
        return response
```

### #267 : building on top of the dispatcher API

```python
from httpx.base import AsyncDispatcher

class Middleware(AsyncDispatcher):
    def __init__(self, next_dispatcher, **kwargs):
        self.next_dispatcher = next_dispatcher

    async def send(
        self,
        request: AsyncRequest,
        verify: VerifyTypes = None,
        cert: CertTypes = None,
        timeout: TimeoutTypes = None,
    ) -> AsyncResponse:
        return await self.next_dispatcher.send(request, verify=verify, cert=cert, timeout=timeout)
```

### #268: callable middleware

```python
class Middleware:
    async def __call__(self, request: AsyncRequest, get_response: typing.Callable) -> AsyncResponse:
        return await get_response(request)
```

---

I think the various PRs proved that all APIs tick 1), 2), and 3) as defined above, with various degrees of simplicity and boilerplate.

The callable middleware API seems to be the simplest one while offering enough flexibility. Making extra requests based on the initial response can be achieved by recursively calling the middleware, as in the [RedirectMiddleware](https://github.com/encode/httpx/pull/268/files#diff-99d8817d09420081599eef0eaa2cb23aR85). As another PoC, a rough Digest-Auth implementation based on [this snippet](https://github.com/encode/httpx/pull/136/files#diff-a3cb8894acfb8a529021a206ac4ea62aR56-R66):

```python
class DigestAuthMiddleware:
    async def __call__(self, request, get_response):
        response = await get_response(request)

        # Did the server respond with 401 + WWW-Authenticate?
        if self.requires_authentication(response):
            authorization = self.build_authorization_header(request, response)
            request.headers["Authorization"] = authorization
            return await self(request, get_response)

        return response
```

cc @tomchristie @yeraydiazdiaz @sethmlarson

florimondmanca:
Based off #267 
Fixes #295 

This builds on the idea of a callable-based middleware interface:

```python
class TheMiddleware(BaseMiddleware):
    async def __call__(self, request, get_response):
        # Process request (or return an early response, e.g. from a cache)
        response = await get_response(request)
        # Process response
        return response
```

The initial `get_response` callable is built from the client's `dispatch.send`, and then from the inner middleware.

@yeraydiazdiaz Feel free to cherry-pick the commit into #267, or we can take the PR from here (you should have commit rights if you check out the branch from the `httpx` origin). :-)


sethmlarson:
Mental note here to make sure whoever merges this PR use the GitHub co-commiter magic string in the Squash commit. :) Thank you two for tackling this issue!!

httpx/middleware.py
```diff
@@ -1,61 +1,91 @@
...
+) -> Middleware:
+    if isinstance(username, str):
+        username = username.encode("latin1")
+
+    if isinstance(password, str):
+        password = password.encode("latin1")
+
+    userpass = b":".join((username, password))
+    token = b64encode(userpass).decode().strip()
+    print(username, password, userpass, token)
```
yeraydiazdiaz: Leftover print?

httpx/middleware.py
```diff
@@ -1,61 +1,91 @@
...
+    allow_redirects: bool = True,
+    max_redirects: int = DEFAULT_MAX_REDIRECTS,
+    cookies: typing.Optional[Cookies] = None,
+) -> Middleware:
+    return Redirect(
+        allow_redirects=allow_redirects, max_redirects=max_redirects, cookies=cookies
+    ).dispatch
+
+
+class Redirect:
```
yeraydiazdiaz: While I like the idea of middlewares being functional I somehow prefer them to be Dispatcher subclasses. While working on the Digest auth I realized there's quite a bit of state and logic that is fairly intertwined with dispatching.

That being said I do feel there's a bit of awkwardness in the subclassing, somehow I feel we're missing an abstaction but if I had to choose I'd rather have middlewares as dispatchers. 

Again I'd love to hear other people's thoughts on this. 🙂

httpx/client.py
```diff
@@ -172,42 +156,62 @@ def merge_headers(
...
-            elif callable(auth):
-                dispatcher = CustomAuthDispatcher(
-                    next_dispatcher=dispatcher, auth_callable=auth
-                )
-
-        if username is not None and password is not None:
-            dispatcher = BasicAuthDispatcher(
-                next_dispatcher=dispatcher, username=username, password=password
+        get_response = functools.partial(
+            redirect(allow_redirects, cookies=self.cookies), get_response=get_response
```
yeraydiazdiaz: Personally I'm not sure if this is clearer that the class based approach, I go into more detail on the definition for `Redirect` below, but here it seems we're "forcing" redirect to be a callable for no particular gain?

Keen on hearing other peoples thoughts though 🙂

httpx/client.py
```diff
@@ -172,42 +156,62 @@ def merge_headers(
...
             )
 
-        return dispatcher
+        if trust_env:
+            netrc_login = get_netrc_login(request.url.authority)
+            if netrc_login:
+                username, _, password = netrc_login
+                return basic_auth(username=username, password=password)
+
+        return None
```
yeraydiazdiaz: This is method much clearer now, thanks! 👍

httpx/client.py
```diff
@@ -172,42 +156,62 @@ def merge_headers(
...
+    def _get_auth_middleware(
+        self, request: AsyncRequest, trust_env: bool, auth: AuthTypes = None
+    ) -> typing.Optional[Middleware]:
+        if isinstance(auth, tuple):
+            return basic_auth(username=auth[0], password=auth[1])
+
+        if callable(auth):
+            return custom_auth(auth=auth)
+
+        assert auth is None
```
yeraydiazdiaz: I noticed we use assertions around the code, I believe it is discouraged since they can be [skipped when running in optimized mode](https://www.quora.com/How-do-I-use-Python-asserts-even-when-optimization-mode-is-used).

Not particularly important in this case though.

httpx/middleware.py
florimondmanca: Thanks a lot for your review!

In your experience, what could be potential blockers to implementing Digest Auth using this callable interface? I think Redirect shows that we can add arbitrarily complex logic by deferring the implementation to a callable class, provided we only need to interact at the response layer (and not the underlying .send()).

(In retrospect I’d refactor the ad-hoc .dispatch() method on Redirect into a .__call__() method — and perhaps *all* middleware should be written in this style to avoid a mix of functions and callable classes.)

httpx/client.py
florimondmanca: The latest iteration provides a `BaseMiddleware` class, which requires subclasses to explicitly implement `__call__()`. Would you say this make the callable-based implementation clearer?

httpx/client.py
florimondmanca: I guess you're right here — we'd rather want a `TypeError` to be raised or something. I'll update the PR!

httpx/middleware.py
yeraydiazdiaz: I wasn't suggesting it was a blocker, sorry I that's how it came across, I was simply pointing out I found the class-based approach a bit more straightforward to read. I'll take a look your recent commits and re-review. 🙂

florimondmanca:
Thanks @yeraydiazdiaz!

florimondmanca:
About the co-authoring thingy, I found [here](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors) that we'll need to add this line to the squash merge message:

```
Co-authored-by: Yeray Diaz Diaz <yeray's email>
```

@yeraydiazdiaz Is your GH email address the one listed on your website? If it's configured as private, I think you'll need to send us your GitHub `no-reply` address (instructions [here](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors#required-co-author-information)) (but perhaps that will work in all cases?).

yeraydiazdiaz:
Yes, it's the same one. Thanks :)

florimondmanca:
@sethmlarson Could you lend us a quick review to confirm this is on the right track as far as the middleware API is concerned? Having went over it when refactoring, I think the rest of the work originally commited by @yeraydiazdiaz is solid.

tomchristie:
Something that we could perfectly well consider would be adding a middleware interface to the client, but not forcing ourselves to factor redirects etc. out. I think that *might* end up being the best balance of having a comprehensible codebase, plus an extensible API.

Might well be worth drafting up a pull request along those lines, to see how it would look.

I'm also tentatively wondering if taking that approach might lead us to reassessing what the `auth` interface should look like. (Eg. maybe we'd be able to drop the auth classe interface completely, in favor of middleware, while still just keeping redirect behavior as part of the client itself.)

florimondmanca:
I’m starting to think — it would really help to define the scope, goals, API and associated requirements (in a word: the *design*) of the middleware interface, to clear up any misunderstandings and have a shared vision of where this should be going?

I don’t think there was an issue initially but perhaps we could open one and discuss the details there.

httpx/middleware.py
```diff
@@ -0,0 +1,164 @@
...
+import typing
+from base64 import b64encode
+
+from .config import DEFAULT_MAX_REDIRECTS
+from .exceptions import RedirectBodyUnavailable, RedirectLoop, TooManyRedirects
+from .models import URL, AsyncRequest, AsyncResponse, Cookies, Headers
+from .status_codes import codes
+
+Responder = typing.Callable[
+    [AsyncRequest], typing.Coroutine[typing.Any, typing.Any, AsyncResponse]
```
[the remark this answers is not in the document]
florimondmanca: Solved with a simpler `typing.Callable` annotation instead of `Responder`. :-)

httpx/client.py
```diff
@@ -148,188 +147,77 @@ def merge_headers(
...
-        method = self.redirect_method(request, response)
-        url = self.redirect_url(request, response)
-        headers = self.redirect_headers(request, url)
-        content = self.redirect_content(request, method, response)
-        cookies = self.merge_cookies(request.cookies)
-        return AsyncRequest(
-            method=method, url=url, headers=headers, data=content, cookies=cookies
+        get_response = wrap(
+            get_response,
+            RedirectMiddleware(allow_redirects=allow_redirects, cookies=self.cookies),
```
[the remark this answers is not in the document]
florimondmanca: `RedirectMiddleware` also has the responsibility of managing `response.history` and `response.next` — not just automatically making redirect requests — so I don't think it's possible in the current state of things. Perhaps this is worth a comment here?

florimondmanca:
@sethmlarson 

> How do we plan on implementing middlewares that are capable of making more than one request (ie caching, digest auth)? Since the middleware doesn't have access to the calling client it can't make requests on it's behalf.

As I explained in #295, the middleware can make extra requests by merely calling itself recursively, with a different or modified request. I put a Digest Auth PoC at the end of the issue description in #295.

Another PoC off the top of my head: a middleware that would retry sending a failed request within a given maximum of retries:

```python
class RetryMiddleware(BaseMiddleware):
    def __init__(self, max_retries: int):
        self.max_retries = max_retries
        self.retries = 0

    async def __call__(self, request, get_response):
        if self.retries > self.max_retries:
            raise MaxRetriesExceeded

        response = await get_response(request)
        if response.status_code < 500:
            return response

        # Try again later…
        self.retries += 1
        await asyncio.sleep(1)
        return await self(request, get_response)
```

(This example shows that we may want to figure out how to allow middleware to make async I/O operations via the concurrency backend. I think a simple dependency injection mechanism where the client would automatically pass its `backend` to middleware that have it in their constructor would be a good compromise.)

This could be easily packaged in a separate library and mounted onto the HTTPX client, e.g.

```python
client = httpx.Client()
client.add_middleware(RetryMiddleware(max_retries=3))
```

(Then we'd ask: how do we control where exactly the middleware is inserted? Do we even want/need to give that kind of control? I don't think it's in the scope of this PR though — this is very much a refactoring operation for an *internal* middleware API. Once we have this, we can easily design a public-facing middleware API on top.)

florimondmanca:
Thanks all!

PR #270, opened 2019-08-22
Add backend param to ASGIDispatcher

### LINKED ISSUE #247: Make ASGI dispatcher asyncio-agnostic

florimondmanca:
Fixes #247

### DIFF
--- a/httpx/dispatch/asgi.py
+++ b/httpx/dispatch/asgi.py
@@ -48,13 +48,13 @@ class ASGIDispatch(AsyncDispatcher):
         raise_app_exceptions: bool = True,
         root_path: str = "",
         client: typing.Tuple[str, int] = ("127.0.0.1", 123),
+        backend: ConcurrencyBackend = None,
     ) -> None:
         self.app = app
         self.raise_app_exceptions = raise_app_exceptions
         self.root_path = root_path
         self.client = client
-        # This will need to be turned into a parameter on this class at some point.
-        self.backend: ConcurrencyBackend = AsyncioBackend()
+        self.backend = AsyncioBackend() if backend is None else backend
 
     async def send(
         self,


sethmlarson:
Thanks @florimondmanca!

cansarigol:
Hi, it was merged but is there any obstacle for this:`backend: ConcurrencyBackend = AsyncioBackend()` . Just to narrow down #272

sethmlarson:
Yes, you don't want to use mutable objects as defaults for function parameters because instead of getting a new copy on every call it'll use the same instance.

cansarigol:
ah sorry 👍

PR #273, opened 2019-08-23
Parametrize tests with concurrency backend

### LINKED ISSUE #271: Backend test parametrization
Now that almost all of HTTPX internals are relying on the concurrency backend, and before tackling the Trio backend (#120), it would be good to parametrize all tests with a `backend` fixture.

For now, it would only point to `AsyncioBackend`, but we're going to need it for `TrioBackend` anyway — the main goal is to reduce clutter when the trio backend lands a PR.

Once the `backend` fixture is defined, any test that uses `AsyncClient`, `HTTPConnection` or `ConnectionPool` can basically be parametrized with it, e.g.:

```python
# conftest.py
@pytest.fixture(params=[pytest.param(AsyncioBackend, marks=pytest.mark.asyncio)])
def backend(request):
    backend_cls = request.param
    return backend_cls()

# test_async_client.py
async def test_get(backend):
    async with AsyncClient(backend=backend) as client:
```

I'll tackle this look soon, but if anyone has thoughts on this approach let me know. :-)

florimondmanca:
Fixes #271 

Mostly adds a `backend` fixture and use it in the relevant tests, and some refactoring of HTTP/2 and connection pool tests to be more backend-agnostic as well.

I'm thinking that now we don't really test the usual path (e.g. not passing a `backend` to `AsyncClient`). This can be easily fixed by adding `None` as an extra backend to make `AsyncClient` & others use their default backend. This increases the duration of the test suite though (~20%?), so… any thoughts?

### DIFF
--- a/tests/concurrency.py
+++ b/tests/concurrency.py
@@ -1,0 +1,19 @@
+"""
+This module contains concurrency utilities that are only used in tests, thus not
+required as part of the ConcurrencyBackend API.
+"""
+
+import asyncio
+import functools
+
+from httpx import AsyncioBackend
+
+
+@functools.singledispatch
+async def sleep(backend, seconds: int):
+    raise NotImplementedError  # pragma: no cover
+
+
+@sleep.register(AsyncioBackend)
+async def _sleep_asyncio(backend, seconds: int):
+    await asyncio.sleep(seconds)
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -633,12 +633,17 @@ class Client(BaseClient):
         # concurrency backends.
         # The sync client performs I/O on its own, so it doesn't need to support
         # arbitrary concurrency backends.
-        # Therefore, we kept the `backend` parameter (for testing/mocking), but enforce
-        # that the concurrency backend derives from the asyncio one.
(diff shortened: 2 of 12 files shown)


tests/concurrency.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@functools.singledispatch
+async def sleep(backend, seconds: int):
+    raise NotImplementedError  # pragma: no cover
+
+
+@sleep.register(AsyncioBackend)
+async def _sleep_asyncio(backend, seconds: int):
+    await asyncio.sleep(seconds)
```
florimondmanca: @sethmlarson @tomchristie Thoughts on this pattern?

The HTTP/2 mock dispatcher `asyncio.sleep()` but we didn't need it anywhere else — a similar situation checking the usage of TLS in #263.

I figured there should be a way to gather tests-only concurrency utilities without having to reimplement a backend; a `singledispatch` on the backend type seemed like a good fit.

sethmlarson: I haven't seen this pattern before, it looks good though :)

florimondmanca:
@sethmlarson Sounds fair — I added a test for the case when we're not using a backend. :+1:

PR #274, opened 2019-08-24
use properties to allow more easily changing cookies after init

toppk:
coming from requests, I was used to doing session.cookies = LWPCookieJar()
with httpx, I have to do:
 client.cookies = httpx.models.Cookies(LWPCookieJar())
 or I can just do this as part of the Client constructor.

This doesn't solve the second half of the problem, where now I have to do:
  client.cookies.jar.save()
 instead of 
  session.cookies.save()

But just thought I'd through this patch in there.  I suspect there will be others looking to do something similar.  This is easy enough to workaround.


sethmlarson:
Thanks for opening this! This is definitely a question of Requests compatibility which is something we want to maintain. I have no issue with this change but I've never used cookies/cookiejars with Requests so I don't know much in this area. cc @tomchristie

tomchristie:
I'm not sure about this change.

It seems fairly reasonable to me to force the usage of:

```python
client.cookies = httpx.Cookies(LWPCookieJar())
```

Since the type that the property ends up being is an `httpx.Cookies` instance. It might(?) be a bit weird for us to simply coerce property assignments.

```python
client.cookies = LWPCookieJar()
isinstance(client.cookies, LWPCookieJar). # False
```

On the other hand, perhaps it's worth it for simplicity.

I think a good way of reviewing if we want this or not would be to ask the same question of other assignments. Eg. Do we want to support this...

```python
client.headers = {...}
```

I *guess* the answer to that is "yeah, sure", in which case we probably want the same from `.cookies`. Thoughts, anyone?

tomchristie:
Okay, then I'm a "yes" to this pull request, but I'd like it to be updated to handle both `.cookies` and `.headers` in the same way, for consistency.

florimondmanca:
I'd be okay with this (coercing property assignments of `cookies` and `headers`) as well.

To address the "second half of the problem" mentioned by @toppk (which does exist if we go down this road, I think?), we should probably update the [Client API](https://www.encode.io/httpx/api/#client) to document that `client.cookies` and `client.headers` are always returned as a `Cookies`/`Headers` instance.

httpx/client.py
```diff
@@ -121,17 +137,17 @@ def merge_url(self, url: URLTypes) -> URL:
     def merge_cookies(
         self, cookies: CookieTypes = None
     ) -> typing.Optional[CookieTypes]:
-        if cookies or self.cookies:
-            merged_cookies = Cookies(self.cookies)
+        if cookies or self._cookies:
```
tomchristie: On reflection I think we should just use the public `self.cookies`/`self.headers` accessors for these cases.

florimondmanca:
Nice work @toppk and @sethmlarson on this one 👍

Issue #246, opened 2019-08-20
Support multiple background tasks in BackgroundManager


florimondmanca:
The `BackgroundManager` is used in `dispatchers/http11.py` and `dispatchers/http2.py` to concurrently send the request and read the first part of the response. Its usage currently looks like this:

```python
async with backend.background_manager(coro, ["arg1", arg2"]):
    # Do stuff while background task is running
    ...
```

This is limited in that we cannot start more than one task easily. However, we will need this feature to generalize the usage of `ConcurrencyBackend` across the code base.

Here's a more flexible API, inspired by trio's nurseries:

```python
async with backend.background_manager() as background:
    background.start_soon(coro1, "arg1", "arg2")
    # Do stuff while background task is running
```

Usage with multiple tasks:

```python
async with backend.background_manager() as background:
    background.start_soon(coro1, "arg1", "arg2")
    background.start_soon(coro2, "arg3", "arg4")
    # Do stuff while background tasks are running
```

tomchristie:
Seems like a sensible tack yup.

One *alternative* that we could end up exploring at some point would be to use a *different* pattern style here. We're using the background manager in order to allow us to perform simultanous read/writes. The urllib3 async work instead uses a "read and write for a bit" primitive that backends need to implement.

I won't get into digging into that in detail here, but it's feasible that it could be a neater pattern to have our Socket interface expose a `.read_and_write(...)` method that different backends implement, rather than having to go all the way into exposing context blocked background managers.

That's almost certainly more work, since it'd mean rejigging the h2/h11 implementations a bit, but it's also more constrained than exposing full background managers as part of the concurrency interface. (Tho we've also got a related bit of work around the ASGI implementation, which'd need further thinking about.)

florimondmanca:
Actually, we're also going to need the background manager in `ASGIDispatch`, i.e. to start running the app in the background as we're pushing content to the response. Thanks to the refactor getting rid of `FIRST_COMPLETED`, we can get away with `BackgroundManager` only handling a single coroutine, but it looks to me like we're going to need it anyway. So maybe it's best to leave usage of the background manager in h2/h11 as it is now?

florimondmanca:
I'll close this for now since the ability to run multiple background tasks isn't a requirement anymore. :-)

PR #276, opened 2019-08-25
Add trio concurrency backend

### LINKED ISSUE #120: Add support for a Trio ConcurrencyBackend

florimondmanca:
Fixes #120 

**Current status**: we're blocked by the connection pool implicitly relying on asyncio polling sockets in the background to tell if they're still readable. See discussion around https://github.com/encode/httpx/pull/276#discussion_r322002604.

Most of the notes below are outdated.

---

Getting the tests to pass on this one ended up having to tackle more tricky issues than I initially thought. (There's still `start_tls()` to figure out, but that will be for a future PR.)

So, a few notes, from the most straight-forward to the most obscure:

- I took the liberty to refactor `HTTP2Dispatcher.initiate_connection()` into an async function; it allowed to remove the need for `write_no_block()` on the concurrency backend (which trio does not have an obvious equivalent for). Things seem to be okay, but if I missed any reasons why it was not async in the first place, let me know! cc @tomchristie 
- I had to make connections context-managed in HTTPS tests in `test_connections.py` — otherwise they'd stall when using Trio, probably because the underlying TCP stream would be waiting to be closed. I'm not 100% sure I understand the origin of this, though. Anyway, I ended up converting all those tests to use `async with` for consistency. 
- For async tests, I had to figure out a way to allow an alternative I/O library (trio) to coexist with asyncio. The solution I went for is spawning backend.run() in the asyncio threadpool. If the code needs better/more comments there, let me know.
- For the connection pool tests, I had to figure out how to restart the uvicorn server from an async test that could be running on something else than asyncio. I added a detailed comment in the code about the solution I've come up with — let me know if it's clear enough.

If it can help with reviewing, I'd be happy to `pytest.skip()` some tests in this PR and then submit solutions for the tricky parts in separate PRs that resolve those skips. Let me know how you'd like to handle this. :-)

### DIFF
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -78,7 +78,11 @@ class TCPStream(BaseTCPStream):
             # Reset before recursive call, otherwise we'll go through
             # this branch indefinitely.
             self.write_buffer = b""
-            await self.write(previous_data, timeout=timeout, flag=flag)
+            try:
+                await self.write(previous_data, timeout=timeout, flag=flag)
+            except WriteTimeout:
+                self.writer_buffer = previous_data
+                raise
 
         if not data:
             return
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -71,8 +71,12 @@ class TCPStream(BaseTCPStream):
     async def write(
         self, data: bytes, timeout: TimeoutConfig = None, flag: TimeoutFlag = None
     ) -> None:
-        data += self.write_buffer
-        self.write_buffer = b""
+        if self.write_buffer:
+            previous_data = self.write_buffer
+            # Reset before recursive call, otherwise we'll go through
+            # this branch indefinitely.
+            self.write_buffer = b""
+            await self.write(previous_data, timeout=timeout, flag=flag)
 
         if not data:
             return
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -181,7 +181,7 @@ class TrioBackend(ConcurrencyBackend):
         return Queue(max_size=max_size)
 
     def create_event(self) -> BaseEvent:
-        return typing.cast(BaseEvent, trio.Event())
+        return Event()
 
     def background_manager(
         self, coroutine: typing.Callable, *args: typing.Any
@@ -198,6 +198,25 @@ class Queue(BaseQueue):
 
     async def put(self, value: typing.Any) -> None:
         await self.send_channel.send(value)
+
+
+class Event(BaseEvent):
+    def __init__(self) -> None:
+        self._event = trio.Event()
+
+    def set(self) -> None:
+        self._event.set()
+
+    def is_set(self) -> bool:
+        return self._event.is_set()
+
+    async def wait(self) -> None:
+        await self._event.wait()
+
+    def clear(self) -> None:
+        # trio.Event.clear() was deprecated in Trio 0.12.
+        # https://github.com/python-trio/trio/issues/637
+        self._event = trio.Event()
 
 
 class BackgroundManager(BaseBackgroundManager):
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -1,0 +1,214 @@
+import functools
+import math
+import ssl
+import typing
+from types import TracebackType
+
+import trio
+
+from ..config import PoolLimits, TimeoutConfig
+from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
+from .base import (
+    BaseBackgroundManager,
+    BaseEvent,
+    BasePoolSemaphore,
+    BaseQueue,
+    BaseTCPStream,
+    ConcurrencyBackend,
+    TimeoutFlag,
+)
+
+
+def _or_inf(value: typing.Optional[float]) -> float:
+    return value if value is not None else float("inf")
+
+
+class TCPStream(BaseTCPStream):
+    def __init__(self, stream: trio.abc.Stream, timeout: TimeoutConfig) -> None:
+        self.stream = stream
+        self.timeout = timeout
+        self.is_eof = False
+        self.write_buffer = b""
+
+    def get_http_version(self) -> str:
+        if not isinstance(self.stream, trio.SSLStream):
+            return "HTTP/1.1"
+
+        ident = self.stream.selected_alpn_protocol()
+        if ident is None:
+            ident = self.stream.selected_npn_protocol()
+
+        return "HTTP/2" if ident == "h2" else "HTTP/1.1"
+
+    async def read(
+        self, n: int, timeout: TimeoutConfig = None, flag: TimeoutFlag = None
+    ) -> bytes:
+        if timeout is None:
+            timeout = self.timeout
+
+        while True:
+            # Check our flag at the first possible moment, and use a fine
+            # grained retry loop if we're not yet in read-timeout mode.
+            should_raise = flag is None or flag.raise_on_read_timeout
+            read_timeout = _or_inf(timeout.read_timeout if should_raise else 0.01)
+
+            with trio.move_on_after(read_timeout):
+                data = await self.stream.receive_some(max_bytes=n)
+                if data == b"":  # pragma: no cover
+                    self.is_eof = True
+                return data
+
+            if should_raise:
+                raise ReadTimeout() from None
+
+    def is_connection_dropped(self) -> bool:
+        return self.is_eof
+
+    def write_no_block(self, data: bytes) -> None:
+        self.write_buffer += data
+
+    async def write(
+        self, data: bytes, timeout: TimeoutConfig = None, flag: TimeoutFlag = None
+    ) -> None:
+        data += self.write_buffer
+        self.write_buffer = b""
+
+        if not data:
+            return
+
+        if timeout is None:
+            timeout = self.timeout
+
+        write_timeout = _or_inf(timeout.write_timeout)
+
+        while True:
+            with trio.move_on_after(write_timeout):
+                await self.stream.send_all(data)
+                break
+            # We check our flag at the first possible moment, in order to
+            # allow us to suppress write timeouts, if we've since
+            # switched over to read-timeout mode.
+            should_raise = flag is None or flag.raise_on_write_timeout
+            if should_raise:
+                raise WriteTimeout() from None
+
+    async def close(self) -> None:
+        await self.stream.aclose()
+
+
+class PoolSemaphore(BasePoolSemaphore):
+    def __init__(self, pool_limits: PoolLimits):
+        self.pool_limits = pool_limits
+
+    @property
+    def semaphore(self) -> typing.Optional[trio.Semaphore]:
+        if not hasattr(self, "_semaphore"):
+            max_connections = self.pool_limits.hard_limit
+            if max_connections is None:
+                self._semaphore = None
+            else:
+                self._semaphore = trio.Semaphore(
+                    max_connections, max_value=max_connections
+                )
+        return self._semaphore
+
+    async def acquire(self) -> None:
+        if self.semaphore is None:
+            return
+
+        timeout = _or_inf(self.pool_limits.pool_timeout)
+
+        with trio.move_on_after(timeout):
+            await self.semaphore.acquire()
+            return
+
+        raise PoolTimeout()
+
+    def release(self) -> None:
+        if self.semaphore is None:
+            return
+
+        self.semaphore.release()
+
+
+class TrioBackend(ConcurrencyBackend):
+    async def open_tcp_stream(
+        self,
+        hostname: str,
+        port: int,
+        ssl_context: typing.Optional[ssl.SSLContext],
+        timeout: TimeoutConfig,
+    ) -> TCPStream:
+        connect_timeout = _or_inf(timeout.connect_timeout)
+
+        with trio.move_on_after(connect_timeout) as cancel_scope:
+            stream: trio.SocketStream = await trio.open_tcp_stream(hostname, port)
+            if ssl_context is not None:
+                stream = trio.SSLStream(stream, ssl_context, server_hostname=hostname)
+                await stream.do_handshake()
+
+        if cancel_scope.cancelled_caught:
+            raise ConnectTimeout()
+
+        return TCPStream(stream=stream, timeout=timeout)
+
+    async def run_in_threadpool(
+        self, func: typing.Callable, *args: typing.Any, **kwargs: typing.Any
+    ) -> typing.Any:
+        return await trio.to_thread.run_sync(
+            functools.partial(func, **kwargs) if kwargs else func, *args
+        )
+
+    def run(
+        self, coroutine: typing.Callable, *args: typing.Any, **kwargs: typing.Any
+    ) -> typing.Any:
+        return trio.run(
+            functools.partial(coroutine, **kwargs) if kwargs else coroutine, *args
+        )
+
+    def get_semaphore(self, limits: PoolLimits) -> BasePoolSemaphore:
+        return PoolSemaphore(limits)
+
+    def create_queue(self, max_size: int) -> BaseQueue:
+        return Queue(max_size=max_size)
+
+    def create_event(self) -> BaseEvent:
+        return typing.cast(BaseEvent, trio.Event())
+
+    def background_manager(
+        self, coroutine: typing.Callable, *args: typing.Any
+    ) -> "BackgroundManager":
+        return BackgroundManager(coroutine, *args)
+
+
+class Queue(BaseQueue):
+    def __init__(self, max_size: int) -> None:
+        self.send_channel, self.receive_channel = trio.open_memory_channel(math.inf)
+
+    async def get(self) -> typing.Any:
+        return await self.receive_channel.receive()
+
+    async def put(self, value: typing.Any) -> None:
+        await self.send_channel.send(value)
+
+
+class BackgroundManager(BaseBackgroundManager):
+    def __init__(self, coroutine: typing.Callable, *args: typing.Any) -> None:
+        self.coroutine = coroutine
+        self.args = args
+        self.nursery_manager = trio.open_nursery()
+        self.nursery: typing.Optional[trio.Nursery] = None
+
+    async def __aenter__(self) -> "BackgroundManager":
+        self.nursery = await self.nursery_manager.__aenter__()
+        self.nursery.start_soon(self.coroutine, *self.args)
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: typing.Type[BaseException] = None,
+        exc_value: BaseException = None,
+        traceback: TracebackType = None,
+    ) -> None:
+        assert self.nursery is not None
+        await self.nursery_manager.__aexit__(exc_type, exc_value, traceback)
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -55,7 +55,9 @@ class TCPStream(BaseTCPStream):
 
             with trio.move_on_after(read_timeout):
                 data = await self.stream.receive_some(max_bytes=n)
-                if data == b"":  # pragma: no cover
+                # b"" is the expected EOF message for Trio.
+                # The other case is an edge case that occurs with uvicorn+httptools.
+                if data == b"" or data.endswith(b"0\r\n\r\n"):
                     self.is_eof = True
                 return data
 
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -37,7 +37,7 @@ class TCPStream(BaseTCPStream):
 
         ident = self.stream.selected_alpn_protocol()
         if ident is None:
-            ident = self.stream.selected_npn_protocol()
+            return "HTTP/1.1"
 
         return "HTTP/2" if ident == "h2" else "HTTP/1.1"
 
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -24,10 +24,13 @@ def _or_inf(value: typing.Optional[float]) -> float:
 
 
 class TCPStream(BaseTCPStream):
-    def __init__(self, stream: trio.abc.Stream, timeout: TimeoutConfig) -> None:
+    def __init__(
+        self,
+        stream: typing.Union[trio.SocketStream, trio.SSLStream],
+        timeout: TimeoutConfig,
+    ) -> None:
         self.stream = stream
         self.timeout = timeout
-        self.is_eof = False
         self.write_buffer = b""
         self.write_lock = trio.Lock()
 
@@ -54,18 +57,18 @@ class TCPStream(BaseTCPStream):
             read_timeout = _or_inf(timeout.read_timeout if should_raise else 0.01)
 
             with trio.move_on_after(read_timeout):
-                data = await self.stream.receive_some(max_bytes=n)
-                # b"" is the expected EOF message for Trio.
-                # The other case is an edge case that occurs with uvicorn+httptools.
-                if data == b"" or data.endswith(b"0\r\n\r\n"):
-                    self.is_eof = True
-                return data
+                return await self.stream.receive_some(max_bytes=n)
 
             if should_raise:
                 raise ReadTimeout() from None
 
     def is_connection_dropped(self) -> bool:
-        return self.is_eof
+        stream = self.stream
+        # Peek through any SSLStream wrappers to get the underlying SocketStream.
+        while hasattr(stream, "transport_stream"):
+            stream = stream.transport_stream
+        assert isinstance(stream, trio.SocketStream)
+        return not stream.socket.is_readable()
 
     def write_no_block(self, data: bytes) -> None:
         self.write_buffer += data
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -29,6 +29,7 @@ class TCPStream(BaseTCPStream):
         self.timeout = timeout
         self.is_eof = False
         self.write_buffer = b""
+        self.write_lock = trio.Lock()
 
     def get_http_version(self) -> str:
         if not isinstance(self.stream, trio.SSLStream):
@@ -83,7 +84,8 @@ class TCPStream(BaseTCPStream):
 
         while True:
             with trio.move_on_after(write_timeout):
-                await self.stream.send_all(data)
+                async with self.write_lock:
+                    await self.stream.send_all(data)
                 break
             # We check our flag at the first possible moment, in order to
             # allow us to suppress write timeouts, if we've since
--- a/httpx/concurrency/trio.py
+++ b/httpx/concurrency/trio.py
@@ -63,12 +63,19 @@ class TCPStream(BaseTCPStream):
                 raise ReadTimeout() from None
 
     def is_connection_dropped(self) -> bool:
+        # Adapted from: https://github.com/encode/httpx/pull/143#issuecomment-515202982
         stream = self.stream
+
         # Peek through any SSLStream wrappers to get the underlying SocketStream.
         while hasattr(stream, "transport_stream"):
             stream = stream.transport_stream
         assert isinstance(stream, trio.SocketStream)
-        return not stream.socket.is_readable()
+
+        # Counter-intuitively, what we really want to know here is whether the socket is
+        # *readable*, i.e. whether it would return immediately with empty bytes if we
+        # called `.recv()` on it, indicating that the other end has closed the socket.
+        # See: https://github.com/encode/httpx/pull/143#issuecomment-515181778
+        return stream.socket.is_readable()
 
     def write_no_block(self, data: bytes) -> None:
         self.write_buffer += data
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -6,6 +6,7 @@ import h2.events
 
 from ..concurrency.base import BaseEvent, BaseTCPStream, ConcurrencyBackend, TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
+from ..exceptions import ProtocolError
 from ..models import AsyncRequest, AsyncResponse
 from ..utils import get_logger
 
@@ -187,6 +188,10 @@ class HTTP2Connection:
                 logger.debug(
                     f"receive_event stream_id={event_stream_id} event={event!r}"
                 )
+
+                if hasattr(event, "error_code"):
+                    raise ProtocolError(event)
+
                 if isinstance(event, h2.events.WindowUpdated):
                     if event_stream_id == 0:
                         for window_update_event in self.window_update_received.values():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,7 +47,17 @@ def clean_environ() -> typing.Dict[str, typing.Any]:
     os.environ.update(original_environ)
 
 
-@pytest.fixture(params=[pytest.param(AsyncioBackend, marks=pytest.mark.asyncio)])
+backend_params = [pytest.param(AsyncioBackend, marks=pytest.mark.asyncio)]
+
+try:
+    from httpx.concurrency.trio import TrioBackend
+except ImportError:  # pragma: no cover
+    pass
+else:
+    backend_params.append(pytest.param(TrioBackend, marks=pytest.mark.trio))
+
+
+@pytest.fixture(params=backend_params)
 def backend(request):
     backend_cls = request.param
     return backend_cls()
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -81,7 +81,7 @@ class BaseClient:
             if param_count == 2:
                 dispatch = WSGIDispatch(app=app)
             else:
-                dispatch = ASGIDispatch(app=app)
+                dispatch = ASGIDispatch(app=app, backend=backend)
 
         self.trust_env = True if trust_env is None else trust_env
 
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -112,6 +112,20 @@ class TCPStream(BaseTCPStream):
                     raise WriteTimeout() from None
 
     def is_connection_dropped(self) -> bool:
+        # Counter-intuitively, what we really want to know here is whether the socket is
+        # *readable*, i.e. whether it would return immediately with empty bytes if we
+        # called `.recv()` on it, indicating that the other end has closed the socket.
+        # See: https://github.com/encode/httpx/pull/143#issuecomment-515181778
+        #
+        # As it turns out, asyncio checks for readability in the background
+        # (see: https://github.com/encode/httpx/pull/276#discussion_r322000402),
+        # so checking for EOF or readability here would yield the same result.
+        #
+        # At the cost of rigour, we check for EOF instead of readability because asyncio
+        # does not expose any public API to check for readability.
+        # (For a solution that uses private asyncio APIs, see:
+        # https://github.com/encode/httpx/pull/143#issuecomment-515202982)
+
         return self.stream_reader.at_eof()
 
     async def close(self) -> None:
--- a/httpx/concurrency/base.py
+++ b/httpx/concurrency/base.py
@@ -187,3 +187,10 @@ class BaseBackgroundManager:
         traceback: TracebackType = None,
     ) -> None:
         raise NotImplementedError()  # pragma: no cover
+
+    async def close(self, exception: BaseException = None) -> None:
+        if exception is None:
+            await self.__aexit__(None, None, None)
+        else:
+            traceback = exception.__traceback__  # type: ignore
+            await self.__aexit__(type(exception), exception, traceback)
--- a/httpx/dispatch/asgi.py
+++ b/httpx/dispatch/asgi.py
@@ -130,6 +130,7 @@ class ASGIDispatch(AsyncDispatcher):
         await response_started_or_failed.wait()
 
         if app_exc is not None and self.raise_app_exceptions:
+            await background.close(app_exc)
             raise app_exc
 
         assert status_code is not None, "application did not return a response."
@@ -138,7 +139,7 @@ class ASGIDispatch(AsyncDispatcher):
         async def on_close() -> None:
             nonlocal response_body
             await response_body.drain()
-            await background.__aexit__(None, None, None)
+            await background.close(app_exc)
             if app_exc is not None and self.raise_app_exceptions:
                 raise app_exc
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ combine_as_imports = True
 force_grid_wrap = 0
 include_trailing_comma = True
 known_first_party = httpx,tests
-known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,nox,pytest,requests,rfc3986,setuptools,trustme,uvicorn
+known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,nox,pytest,requests,rfc3986,setuptools,trio,trustme,uvicorn
 line_length = 88
 multi_line_output = 3
 
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ setup(
         "idna==2.*",
         "rfc3986==1.*",
     ],
+    extras_require={"trio": ["trio"]},
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Environment :: Web Environment",
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,4 +1,4 @@
--e .
+-e .[trio]
 
 # Optional
 brotlipy==0.7.*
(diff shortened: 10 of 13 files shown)


tests/conftest.py
```diff
@@ -127,17 +168,67 @@ def cert_encrypted_private_key_file(example_cert):
...
+    # * When that is the case, that environment runs in a separate thread.
+    # * But shutdown/startup() needs to run on the loop from which the server was
+    # spawned, i.e. the one we're currently in.
+    # * Even if we launched `AsyncioBackend().run` in the threadpool, that would run
+    # on a *new event loop - in yet another thread!
+    # So we use a queue as a thread-safe communication channel with async tests.
+
+    while True:
+        try:
+            await queue.get()
```
florimondmanca: Note to self: this can’t work, I need to use get_nowait() instead. (But I’m wondering how the call to asyncio.wait() doesn’t block — currently this task should never terminate.)

httpx/dispatch/http2.py
```diff
@@ -58,10 +58,10 @@ def __init__(
     async def close(self) -> None:
         await self.stream.close()
 
-    def initiate_connection(self) -> None:
+    async def initiate_connection(self, timeout: TimeoutConfig = None) -> None:
```
sethmlarson: What do you think of instead of making this function `async` we keep this sync but just don't write anything? We're gonna get a `await stream.write()` within `send_headers()` anyways?

tests/conftest.py
```diff
@@ -13,12 +15,51 @@
...
+try:
+    from httpx.concurrency.trio import TrioBackend
+except ImportError:  # pragma: no cover
+    TrioBackend = None  # type: ignore
+
+
+@pytest.fixture(
+    params=[
+        pytest.param(AsyncioBackend, marks=pytest.mark.asyncio),
+        pytest.param(TrioBackend, marks=pytest.mark.asyncio),
```
sethmlarson: Shouldn't this be `pytest.mark.trio`? And instead of making `TrioBackend` equal to `None` above can we make the parameters determined within the global scope inside that `try-except` import block?

sethmlarson:
:tada: Very exciting!

tests/conftest.py
florimondmanca: > Shouldn't this be pytest.mark.trio?

Ah, I think this is something that needs a comment as well. The asyncio mark ensures that there will be an event loop setup for tests that use the `backend` fixture. Even tests that run on trio need the event loop because they're actually run in the asyncio threadpool. Again, this is all mostly because we *need* asyncio to be the top-level loop, mostly for uvicorn compatibility reasons.

> And instead of making TrioBackend equal to None above can we make the parameters determined within the global scope inside that try-except import block?

As in, building a `backend_params` list and appending `TrioBackend` to it in the import block? That might end up cleaner, I'll try it out. :-)

httpx/dispatch/http2.py
florimondmanca: True that, but wouldn't it become misleading to send data not related to headers in `send_headers()`? (For example, `initiate_connection()` [sends the HTTP/2 settings frame](https://python-hyper.org/projects/h2/en/stable/_modules/h2/connection.html#H2Connection.initiate_connection) in there.)

If we really want to keep `write_no_block()` (which, I start to realize, perhaps initially existed because we'd like to send headers along with the initial frame in one go, instead of accessing the socket twice?), I can look at implementing a simple buffer mechanism in the trio backend.

florimondmanca: > I can look at implementing a simple buffer mechanism in the trio backend.

I did just that — turns out to be very simple, and requiring no changes to the concurrency backend API.

tests/conftest.py
```diff
@@ -127,17 +169,66 @@ def cert_encrypted_private_key_file(example_cert):
...
+
+async def watch_restarts(server, queue):
+    # We can't let async tests run shutdown()/startup() directly because:
+    # * They may be running in a different async environment (e.g. trio).
+    # * When that is the case, that environment runs in a separate thread.
+    # * But shutdown/startup() needs to run on the loop from which the server was
+    # spawned, i.e. the one we're currently in.
+    # * Even if we launched `AsyncioBackend().run` in the threadpool, that would run
+    # on a *new event loop - in yet another thread!
+    # So we use a queue as a thread-safe communication channel with async tests.
```
yeraydiazdiaz: I'm having trouble understanding this flow, let's see if I got it right.

- In master right now we have a single thread that starts the asyncio event loop.
- Each test is an async function that is then scheduled with a server fixture that schedules the server task and yields it to the test.
- Some tests manually call `startup` and `shutdown`.
- On test teardown we signal Uvicorn to exit and await its task.

In this PR we now have multiple concurrency backends, which means:
- If required (i.e. trio) we spawn the environment in a new thread.
- But server fixture cannot be spawned on that separate thread, since it's asyncio only, thus the tests can no longer call `startup` and `shutdown`.
- So to coordinate between test wanting to do this we add the `restart_server` fixture.
- Tests are running now running in the trio thread and when invoking `restart_server` it places `None` in the queue, and blocks until it gets a value.
- This signal is then picked up by `watch_restarts` to now call `server.shutdown` and `server.startup` in the asyncio thread, and finally placing `None` in the queue.
- This is then picked up by `restart_server` and returns.

Is this correct?

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,214 @@
...
+        self.nursery.start_soon(self.coroutine, *self.args)
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: typing.Type[BaseException] = None,
+        exc_value: BaseException = None,
+        traceback: TracebackType = None,
+    ) -> None:
+        assert self.nursery is not None
```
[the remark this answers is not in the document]
florimondmanca: I think this is a case where assert is suitable as a debugging device — once we get the assertion passing in our tests there’s no way users can make it fail again. :-)

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,214 @@
...
+        if timeout is None:
+            timeout = self.timeout
+
+        write_timeout = _or_inf(timeout.write_timeout)
+
+        while True:
+            with trio.move_on_after(write_timeout):
+                await self.stream.send_all(data)
+                break
+            # We check our flag at the possible moment, in order to
```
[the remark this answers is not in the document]
florimondmanca: Copy-pasted from the asyncio backend (which kind of itches me but I think the comment is necessary) — maybe it’s a typo? 🤷‍♂️ Cc @tomchristie?

tests/conftest.py
florimondmanca: You’ve got it absolutely right @yeraydiazdiaz! Just a couple of precisions:

> But server fixture cannot be spawned on that separate thread, since it's asyncio only, thus the tests can no longer call startup and shutdown.

The server fixture is always spawned in the main thread, ie in asyncio — and that cannot change AFAIK because of how pytest fixtures work. Rather, the main reason we can no longer call startup/shutdown in async tests is because they may run inside a different async environment — so we can’t just « await » the server’s methods because it server is running in asyncio.

> So to coordinate between test wanting to do this we add the restart_server fixture.

Yes — although resorting to a fixture here is mainly motivated by abstracting all this mess from the test code.

Everything you said about the handling of restart signals is correct. :-)

Your analysis is super helpful, though — I’m thinking we can refine some of the comments based on my answers here so that it’s crystal clear what is happening. Have you got any suggestions on this regard?

florimondmanca: Also, I have the intuition there may be a better primitively than a queue to implement the « I demand a restart, please notify me when it’s done » flow here.

I think an implementation based on two events (restart_requested and restart_finished) could be more suitable as well as more self-documenting. Basically we would:

- Set restart_requested and wait for restart_finished in restart_server().
- Wait for restart_requested and set restart_finished in watch_restarts().

This implementation should probably be abstracted away from the fixtures as well (eg via a helper class), which would allow us to put all the comments in a single location.

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,214 @@
...
+        self.is_eof = False
+        self.write_buffer = b""
+
+    def get_http_version(self) -> str:
+        if not isinstance(self.stream, trio.SSLStream):
+            return "HTTP/1.1"
+
+        ident = self.stream.selected_alpn_protocol()
+        if ident is None:
+            ident = self.stream.selected_npn_protocol()
```
pgjones: @florimondmanca I'm not sure if NPN is worth considering any more, I've removed the NPN parts from Hypercorn.

florimondmanca: Not worth it as in — NPN isn't in use anymore/is obsolete? I think @tomchristie originally implemented this check in the asyncio backend.

pgjones: Yep, from https://gitlab.com/pgjones/hypercorn/merge_requests/16, which I believe to be correct, 
> It's my understanding that NPN was for SPDY, ALPN is used by non-ancient http/2 clients. Very old http/2 clients might speak http/1 to hypercorn after this change.

florimondmanca: Cool. I think this can be discussed in a separate issue, as it already affects the existing `AsyncioBackend`. :-)

florimondmanca: https://github.com/encode/httpx/issues/280

httpx/concurrency/trio.py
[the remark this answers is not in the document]
tomchristie: "at the _first_ possible moment"? Correct

florimondmanca:
Yup, good idea @sethmlarson. As you mentioned this can (should?) be implemented in a separate PR. :-)

florimondmanca:
I'm encountering a bug when performing a real-life HTTP/2  + TLS request with the `TrioBackend`:

```python
# example.py
import trio
import httpx
from httpx.concurrency.trio import TrioBackend


async def main(url):
    async with httpx.AsyncClient(backend=TrioBackend()) as client:
        await client.get(url)


trio.run(main, "https://example.com")
```

```console
$ python example.py
  File "example.py", line 11, in <module>
    trio.run(main, "https://example.com")
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1783, in run
    raise runner.main_task_outcome.error
  File "example.py", line 8, in main
    await client.get(url)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 363, in get
    trust_env=trust_env,
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 610, in request
    trust_env=trust_env,
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 181, in send
  File "/Users/florimond/Developer/python-projects/httpx/httpx/client.py", line 218, in send_handling_redirects
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/connection_pool.py", line 120, in send
    raise exc
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/connection_pool.py", line 115, in send
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/connection.py", line 57, in send
    response = await self.h2_connection.send(request, timeout=timeout)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/http2.py", line 45, in send
    status_code, headers = await self.receive_response(stream_id, timeout)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/http2.py", line 119, in receive_response
  File "/Users/florimond/Developer/python-projects/httpx/httpx/dispatch/http2.py", line 158, in receive_event
    await self.stream.write(data_to_send, timeout)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/concurrency/trio.py", line 86, in write
    await self.stream.send_all(data)
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_ssl.py", line 684, in send_all
    with self._outer_send_conflict_detector:
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.7/site-packages/trio/_util.py", line 122, in __enter__
    raise trio.BusyResourceError(self._msg)
trio.BusyResourceError: another task is currently sending data on this SSLStream
```

Everything is fine if I force `http_versions=["HTTP/1.1"]` on `AsyncClient`, so HTTP/2 is indeed failing here.

The place where this error is raised alternates between `.receive_event()` (as above) and `.end_stream()`. They run in parallel via the background manager, respectively as part of receiving the response (which also requires sending some data back to the server) and sending request data.

But, quoting the description of [`BusyResourceError`](https://trio.readthedocs.io/en/latest/reference-core.html#trio.BusyResourceError) in the Trio docs:

> If two tasks try to send data through the same socket at the same time, Trio will raise `BusyResourceError` instead of letting the data get scrambled.

And there's no way around it — probably a way for Trio to tell us to do things differently.

…This is where the "read/write for a while" approach that @tomchristie mentioned [here](https://github.com/encode/httpx/issues/246#issuecomment-522930937) as an alternative to `BackgroundMananger` starts to make sense to me. I don't know exactly how it would look if implemented, but hopefully it ensure that we're not writing to the stream at the same time from two different places?

So from my understanding, right now the Trio backend can't be made compatible with HTTP/2, and we need to rework the sending/reception of request/response data to completely unblock this.

Since HTTP/1.1 is working though, what would be everyone's thoughts on disabling HTTP/2 on trio for now temporarily, so we can get Trio + HTTP/1.1 in now, and add HTTP/2 support later?

(I managed to force usage of HTTP/1.1 by executing:

```python
ssl_context.set_alpn_protocols(["h1"])
ssl_context.set_npn_protocols(["h1"])
```

before opening the `SSLStream`.)

httpx/concurrency/trio.py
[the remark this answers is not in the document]
florimondmanca: Fixed :)

tests/conftest.py
```diff
@@ -126,18 +168,70 @@ def cert_encrypted_private_key_file(example_cert):
...
+    # another thread and/or async environment.
+
+    def __init__(self, config):
+        super().__init__(config)
+        self._restart_channel = asyncio.Queue()
+
+    async def restart(self) -> None:
+        # Request a restart, and wait for it to complete.
+        await self._restart_channel.put(None)
+        await self._restart_channel.get()
```
florimondmanca: This is completely broken: *obviously* this returns immediately since we put a `None` in the queue and remove it immediately after. 🤦‍♂ 

I tried an event-based implementation instead, but we're mixing event loops and asyncio doesn't agree.

I'm starting to wonder if we could just turn towards hypercorn for the background server… Would definitely resolve a lot of difficulties here to be able to run everything in the main thread.

florimondmanca: What's *super weird* though is that the tests *passed*, as if restarting the server didn't make any difference. 😬 (And indeed, commenting the `restart_server()` bit in the connection pool tests has no effect: the tests pass.)

florimondmanca:
@sethmlarson Yes, if asyncio does not have guards in place it’s very possible that we’re writing to the stream writer at the same time as well, potentially sending corrupted data (although up to now it doesn’t seem to have surfaced?).

florimondmanca:
I pushed a commit demonstrating reversing the way we do things: serving the server in a separate thread, which allows us to do all async stuff in the main thread and the proper async environment.

(I've made pytest share the server during the whole test session, so as a benefit tests are now super fast since we don't need to set it up/tear it down on every test. Also we don't need to run sync tests in the threadpool anymore.)

The commit also turns the server restart into a no-op to prove it has no effect on the tests.

There's still a failing tests for the case where an exception is raised within the ASGI app. The failure comes from how we're handling that case — mostly not feeding the exception back into the background manager in `response.on_close()`. I tried fiddling around with it but it's tricky — the comment we put there stating that the background manager is a weak design solution is definitely relevant.

tomchristie:
> it allowed to remove the need for write_no_block() on the concurrency backend (which trio does not have an obvious equivalent for). Things seem to be okay, but if I missed any reasons why it was not async in the first place, let me know!

So the reasoning there was because `initiate_connection` needs to be an atomic operation, that's performed exactly once on the first stream, before any other streams start sending messages.

If we can ensure that it's not async then we know that'll be the case since we can't task switch around it (an alternative would be to introduce a lock around it.)

tomchristie:
> I pushed a commit demonstrating reversing the way we do things: serving the server in a separate thread, which allows us to do all async stuff in the main thread and a "native" async environment.

Would there be any sense in creating a pull request that *only* does that, without the rest of this, so that we can get that into the test suite and confirm it, independently of the trio work?

florimondmanca:
Yup @tomchristie, I’ll isolate it into its own PR. :)

sethmlarson:
Just wondering since I just found the project, would `anyio` work for us?

florimondmanca:
@sethmlarson https://github.com/encode/httpx/issues/296 😄

florimondmanca:
I managed to have the failing ASGI test pass — we needed to close the background manager before re-raising the app exception if it occurred.

The remaining blockers are:

- Even with the threaded Uvicorn server, we're getting a `ConnectionResetError` when it restarts in the connection pool tests. Could be unblocked by #290.
- We need to rethink the HTTP/2 read/write handling to comply with Trio's sensible "only one task can write to a stream at a time" policy.

tomchristie:
> We need to rethink the HTTP/2 read/write handling to comply with Trio's sensible "only one task can write to a stream at a time" policy.

Could you expand on that aspect?

florimondmanca:
@tomchristie I tried to explain that in the comment above: https://github.com/encode/httpx/pull/276#issuecomment-525469745

Let me know if there’s anything else to clarify. Didn’t really have the time to think about a solution yet, but it relates to some ideas you’ve been mentioning lately.

tomchristie:
Okay - I guess the answer there is to put locking around the read/writes.

florimondmanca:
So, an update on my investigations regarding the HTTP/2 issue…

> Okay - I guess the answer there is to put locking around the read/writes.

It was the right thing to do — I added a write lock around the failing `.send_all()` call.

After that I was receiving a `PROTOCOL_ERROR` event. (The program was hanging because we didn't handle error events, so I added a fix for that as well — perhaps worth its own PR?). I fixed it by sending buffered data in a separate `.write()` call instead of concatenating it to the data to send.

**tl;dr: HTTP/2 is now operational on trio :tada:**

Updated example I used to compare the flow between asyncio and trio via `HTTPX_DEBUG`:

```python
# example.py
import asyncio
import trio
import httpx
from httpx.concurrency.trio import TrioBackend


async def main(url, backend):
    async with httpx.AsyncClient(backend=backend) as client:
        await client.get(url)


URL = "https://example.com"
asyncio.run(main(URL, httpx.AsyncioBackend()))
input("\nEnter to continue...\n")
trio.run(main, URL, TrioBackend())
```

florimondmanca:
Last item to tackle from my point of view is the failing tests due to restarts of the uvicorn server, which *should* be solvable by using Hypercorn — #290.

Here's the traceback showing the `ConnectionResetError` in one of the tests:

```console
___________________________ test_connection_closed_free_semaphore_on_acquire[TrioBackend] ___________________________

    @contextmanager
    def _translate_socket_errors_to_stream_errors():
        try:
>           yield

venv/lib/python3.7/site-packages/trio/_highlevel_socket.py:30: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <trio.SocketStream object at 0x10de7d2e8>, max_bytes = 4096

    async def receive_some(self, max_bytes=None):
        if max_bytes is None:
            max_bytes = DEFAULT_RECEIVE_SIZE
        if max_bytes < 1:
            raise ValueError("max_bytes must be >= 1")
        with _translate_socket_errors_to_stream_errors():
>           return await self.socket.recv(max_bytes)

venv/lib/python3.7/site-packages/trio/_highlevel_socket.py:144: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <trio.socket.socket fd=54, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 55977)>
args = (4096,), kwargs = {}

    @_wraps(fn, assigned=("__name__",), updated=())
    async def wrapper(self, *args, **kwargs):
>       return await self._nonblocking_helper(fn, args, kwargs, wait_fn)

venv/lib/python3.7/site-packages/trio/_socket.py:348: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <trio.socket.socket fd=54, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 55977)>
fn = <method 'recv' of '_socket.socket' objects>, args = (4096,), kwargs = {}
wait_fn = <function wait_readable at 0x1064b31e0>

    async def _nonblocking_helper(self, fn, args, kwargs, wait_fn):
        # We have to reconcile two conflicting goals:
        # - We want to make it look like we always blocked in doing these
        #   operations. The obvious way is to always do an IO wait before
        #   calling the function.
        # - But, we also want to provide the correct semantics, and part
        #   of that means giving correct errors. So, for example, if you
        #   haven't called .listen(), then .accept() raises an error
        #   immediately. But in this same circumstance, then on macOS, the
        #   socket does not register as readable. So if we block waiting
        #   for read *before* we call accept, then we'll be waiting
        #   forever instead of properly raising an error. (On Linux,
        #   interestingly, AFAICT a socket that can't possible read/write
        #   *does* count as readable/writable for select() purposes. But
        #   not on macOS.)
        #
        # So, we have to call the function once, with the appropriate
        # cancellation/yielding sandwich if it succeeds, and if it gives
        # BlockingIOError *then* we fall back to IO wait.
        #
        # XX think if this can be combined with the similar logic for IOCP
        # submission...
        async with _try_sync():
>           return fn(self._sock, *args, **kwargs)
E           ConnectionResetError: [Errno 54] Connection reset by peer

venv/lib/python3.7/site-packages/trio/_socket.py:588: ConnectionResetError

value = <trio.Nursery object at 0x10df149e8>

    async def yield_(value=None):
>       return await _yield_(value)

venv/lib/python3.7/site-packages/async_generator/_impl.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.7/site-packages/async_generator/_impl.py:99: in _yield_
    return (yield _wrap(value))
tests/dispatch/test_connection_pools.py:180: in test_connection_closed_free_semaphore_on_acquire
    response = await http.request("GET", server.url)
httpx/dispatch/base.py:40: in request
    return await self.send(request, verify=verify, cert=cert, timeout=timeout)
httpx/dispatch/connection_pool.py:126: in send
    raise exc
httpx/dispatch/connection_pool.py:121: in send
httpx/dispatch/connection.py:65: in send
httpx/dispatch/http11.py:53: in send
    http_version, status_code, headers = await self._receive_response(timeout)
httpx/dispatch/http11.py:133: in _receive_response
httpx/dispatch/http11.py:174: in _receive_event
    self.READ_NUM_BYTES, timeout, flag=self.timeout_flag
httpx/concurrency/trio.py:57: in read
    data = await self.stream.receive_some(max_bytes=n)
venv/lib/python3.7/site-packages/trio/_highlevel_socket.py:144: in receive_some
    return await self.socket.recv(max_bytes)
../../../.pyenv/versions/3.7.3/lib/python3.7/contextlib.py:130: in __exit__
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    @contextmanager
    def _translate_socket_errors_to_stream_errors():
        try:
            yield
        except OSError as exc:
            if exc.errno in _closed_stream_errnos:
                raise trio.ClosedResourceError(
                    "this socket was already closed"
                ) from None
            else:
                raise trio.BrokenResourceError(
                    "socket connection broken: {}".format(exc)
>               ) from exc
E               trio.BrokenResourceError: socket connection broken: [Errno 54] Connection reset by peer

venv/lib/python3.7/site-packages/trio/_highlevel_socket.py:39: BrokenResourceError
```

tomchristie:
> failing tests due to restarts of the uvicorn server

Could you link to where that's discussed? I've lost track of it. Ta!

florimondmanca:
@tomchristie I don’t think I discussed it in much more detail, but tl;dr is that the server restart that we’re doing in three of the connection pool tests to close keep alive connections is seemingly causing trio to lose track of the socket.

It makes sense because I *think* shutting down the uvicorn server closes the socket since the underlying asyncio server is shut down entirely.

Is there a way to close the connections in server.shutdown() without shutting down the whole server, and to then leave the sockets untouched? That might do the trick and not require us to switch to hypercorn — which might actually not even solve this problem.

Perhaps this is actually a problem related to uvicorn not closing the connections gracefully enough. The ConnectionResetError seems pretty rough — is uvicorn missing sending some kind of connection termination packet before closing it?

florimondmanca:
Aaaaalright. I pushed a fix for the server restart issue in the latest commit.

I first realized that `.is_connection_dropped()` was returning `False` after restart when it should return `True`, causing the connetion pool to reuse the actually broken connection, leading to the very much relevant `BrokenResourceError` I saw.

Then, after investigating why `.is_eof` was not set properly, I realized that the sentinel value `b""` [that trio expects implementations to return](https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.ReceiveStream.receive_some) wasn't always returned. (Though it's not clear in this context who plays the role of the "implementation" here?)

More precisely:

- uvicorn + httptools has a final line of the body with a `b"0\r\n\r\n"` suffix:

```console
21:58:29.411 - httpx.dispatch.http11 - receive_event event=NEED_DATA
DATA: b'HTTP/1.1 200 OK\r\ndate: Fri, 06 Sep 2019 19:58:29 GMT\r\nserver: uvicorn\r\ncontent-type: text/plain\r\ntransfer-encoding: chunked\r\n\r\nd\r\nHello, world!\r\n0\r\n\r\n'
21:58:29.412 - httpx.dispatch.http11 - receive_event event=Response(status_code=200, headers=[(b'date', b'Fri, 06 Sep 2019 19:58:29 GMT'), (b'server', b'uvicorn'), (b'content-type', b'text/plain'), (b'transfer-encoding', b'chunked')], http_version=b'1.1', reason=b'OK')
21:58:29.413 - httpx.dispatch.http11 - receive_event event=Data(<13 bytes>)
21:58:29.413 - httpx.dispatch.http11 - receive_event event=EndOfMessage(headers=[])
```

- OTOH, uvicorn + h11 includes that final string of bytes in a separate line, which triggers an extra `NEED_DATA` event:

```console
22:01:14.662 - httpx.dispatch.http11 - receive_event event=NEED_DATA
DATA: b'HTTP/1.1 200 OK\r\ndate: Fri, 06 Sep 2019 20:01:14 GMT\r\nserver: uvicorn\r\ncontent-type: text/plain\r\ntransfer-encoding: chunked\r\n\r\nd\r\nHello, world!\r\n'
22:01:14.666 - httpx.dispatch.http11 - receive_event event=Response(status_code=200, headers=[(b'date', b'Fri, 06 Sep 2019 20:01:14 GMT'), (b'server', b'uvicorn'), (b'content-type', b'text/plain'), (b'transfer-encoding', b'chunked')], http_version=b'1.1', reason=b'OK')
22:01:14.667 - httpx.dispatch.http11 - receive_event event=Data(<13 bytes>)
22:01:14.667 - httpx.dispatch.http11 - receive_event event=NEED_DATA
DATA: b'0\r\n\r\n'
22:01:14.668 - httpx.dispatch.http11 - receive_event event=EndOfMessage(headers=[])
```

So I figured, I just need to say "we reached EOF" if we get `b""` OR something that ends with `b"0\r\n\r\n"`.

I'm really not sure what that final string of bytes represents though — is it standard in HTTP? — so not sure whether this is a dirty hack, a proper fix, or an indication that something might be wrong somewhere (us, h11/httptools, uvicorn, etc).

ANYWAY…

The build is still failing because of the uvloop compilation issue on 3.8-dev, BUT all other tests are passing. :tada: :tada: :tada:

florimondmanca:
I think this is finally ready for review from everyone. :)

yeraydiazdiaz:
I gave this branch a go with the top 10 Alexa sites: 9 of them returned a response just fine with most of them using HTTP2. There was a single ConnectionTimeout on qq.com and some sites returned `413 Request Entity Too Large`.

Really cool stuff 🥇

florimondmanca:
Thanks for trying it out! @yeraydiazdiaz! Do you get similar results with the asyncio backend?

yeraydiazdiaz:
Yep, results are the same for the top 10 sites. Here's the quick and dirty script 😅 

```python
import asyncio
import logging
import sys
from pathlib import Path

import trio

import httpx
from httpx.concurrency.trio import TrioBackend

logging.basicConfig(
    format="[%(levelname)s %(asctime)s] (%(name)s) %(message)s", datefmt="%H:%M:%S"
)
log = logging.getLogger("httpx")
log.setLevel(logging.DEBUG)


async def do_request(url: str) -> None:
    cl = httpx.AsyncClient()
    resp = await cl.get(url)
    print("Success:", resp)


async def do_trio_request(url: str) -> None:
    async with httpx.AsyncClient(backend=TrioBackend()) as client:
        resp = await client.get(url)
        print("Success:", resp)


def do_run(backend, url):
    if backend == "asyncio":
        asyncio.run(do_request(url))
    else:
        trio.run(do_trio_request, url)


if __name__ == "__main__":
    limit = 1_000
    try:
        limit = int(sys.argv[1])
    except:
        pass

    backend = "asyncio"
    try:
        backend = sys.argv[2]
    except:
        pass

    print(f"Limiting to {limit} requests...")

    with open(Path(__file__).parent / "alexa-1m.txt") as fd:
        for i, url in enumerate(fd.readlines(), 1):
            url = url.strip()
            if input(f"\n\nAbout to fetch {url}, is that ok? [Y/n]: ").lower() in (
                "n",
                "no",
            ):
                break
            try:
                do_run(backend, url)
            except httpx.exceptions.HTTPError as e:
                log.exception(e)
                print("\n\n\n ERROR \n\n\n")
                pass

            if i == limit:
                break
```

it assumes the [top 1M Alexa sites text file](https://github.com/encode/httpx/files/3521345/alexa-1m.txt.gz) is in the same directory

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,241 @@
...
+        self.is_eof = False
+        self.write_buffer = b""
+        self.write_lock = trio.Lock()
+
+    def get_http_version(self) -> str:
+        if not isinstance(self.stream, trio.SSLStream):
+            return "HTTP/1.1"
+
+        ident = self.stream.selected_alpn_protocol()
+        if ident is None:
```
sethmlarson: NPN is no longer applicable

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,241 @@
...
+            # Check our flag at the first possible moment, and use a fine
+            # grained retry loop if we're not yet in read-timeout mode.
+            should_raise = flag is None or flag.raise_on_read_timeout
+            read_timeout = _or_inf(timeout.read_timeout if should_raise else 0.01)
+
+            with trio.move_on_after(read_timeout):
+                data = await self.stream.receive_some(max_bytes=n)
+                # b"" is the expected EOF message for Trio.
+                # The other case is an edge case that occurs with uvicorn+httptools.
+                if data == b"" or data.endswith(b"0\r\n\r\n"):
```
sethmlarson: The `b"0\r\n\r\n"` is the last chunk of a `Transfer-Encoding: chunked` response. I don't think we should be handling this at this level, this is an HTTP construct.

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,241 @@
...
+    def is_connection_dropped(self) -> bool:
+        return self.is_eof
+
+    def write_no_block(self, data: bytes) -> None:
+        self.write_buffer += data
+
+    async def write(
+        self, data: bytes, timeout: TimeoutConfig = None, flag: TimeoutFlag = None
+    ) -> None:
+        if self.write_buffer:
```
sethmlarson: If this write times out and it's retried don't we lose data?

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,241 @@
...
+        with trio.move_on_after(connect_timeout) as cancel_scope:
+            stream: trio.SocketStream = await trio.open_tcp_stream(hostname, port)
+            if ssl_context is not None:
+                stream = trio.SSLStream(stream, ssl_context, server_hostname=hostname)
+                await stream.do_handshake()
+
+        if cancel_scope.cancelled_caught:
+            raise ConnectTimeout()
+
+        return Stream(stream=stream, timeout=timeout)
```
sethmlarson: Are we implementing `start_tls()` in this PR or waiting until another PR? I'm fine with either.

florimondmanca: I’d rather defer that to another PR. :)

httpx/concurrency/trio.py
florimondmanca: Yes, that’s what I eventually figured that. The difficulty here is that Trio doesn’t expose an equivalent for asyncio’s stream_reader.at_eof(). And since we only call read() when we receive a NEED_DATA event we are never given the chance to receive an empty bytes message... So the question is: how can we otherwise detect that we received the whole response?

httpx/concurrency/trio.py
florimondmanca: Ah true, we should make sure write_buffer isn’t lost in case of a timeout.

httpx/concurrency/trio.py
florimondmanca: (Also the comment here is plain wrong. 😅)

pquentin: cc @njsmith :)

njsmith: It sounds like there's something fundamentally confused here... whenever you get `NEED_DATA` you call `receive_some` and feed what you get to h11. You know you've received the whole response when h11 reports an `EndOfMessage` event.

The one small wrinkle is that when `receive_some` reports EOF, you need to tell h11 that you've gotten an EOF, you need to tell h11 that you've gotten an EOF. Conveniently, the way you do this is by calling `h11_conn.receive_data(b"")`.

florimondmanca: Thanks for chiming in @njsmith.

> when receive_some reports EOF

The thing is — `.receive_some()` never actually reports an EOF, neither on `TrioBackend` nor `AsyncioBackend`. My proof is a modest print statement on the return value of `backend.read()` while running:

```bash
HTTPX_DEBUG=1 pytest -k test_connection_closed_free_semaphore_on_acquire -s -m asyncio
HTTPX_DEBUG=1 pytest -k test_connection_closed_free_semaphore_on_acquire -s -m trio
```

As far as I understand the current `HTTP11Connection` implementation, when we receive an `EndOfMessage` in `._receive_response_data()` we simply `break` out of the response body iterator. As a result, `response.close()` is called, and the next h11 cycle is prepared. Then the connection is put in a keep-alive state, and we're mostly done.

So apparently, never in this teardown process do we call an equivalent of `h11_state.receive_data(b"")`, and I think that makes sense because that would result in a `ConnectionClosed` event, but the connection has *not* been closed.

With this in mind, I truly wonder how come the asyncio `stream_reader` detects the EOF at all. The following raw setup is that it does detect an EOF when receiving empty bytes as well, just like trio:

- Uvicorn server:

```python
    await send(
        {
            "status": 200,
        }
    )
```

- Client script:

```python
import asyncio
import trio

from httpx.concurrency.asyncio import AsyncioBackend
from httpx.concurrency.trio import TrioBackend

from httpx.config import TimeoutConfig, DEFAULT_TIMEOUT_CONFIG


async def main(backend):
    print(f"Going to connect using {backend!r}")
    input("Please start the server, then continue.")
    stream = await backend.connect(
        "localhost",
        8000,
        ssl_context=None,
        timeout=TimeoutConfig(DEFAULT_TIMEOUT_CONFIG),
    )

    assert not stream.is_connection_dropped()
    input("Please close the server, then continue.")

    # This is the interesting part.
    assert not stream.is_connection_dropped()
    assert await stream.read(1) == b""
    assert stream.is_connection_dropped()

    await stream.close()
    print("Done.")


asyncio.run(main(AsyncioBackend()))
trio.run(main, TrioBackend())
```

But like I said, we never seem to `.read()` empty bytes at all in HTTPX, but for some reason `stream_reader.at_eof()` returns `True` when trying to acquire a new connection after the server restarted. Maybe `stream_reader.at_eof()` does extra processing that I don't know about?

Anyway, I'm really confused, indeed. @tomchristie Any clues on a missing part of the read/write flow I'm missing here?

njsmith: asyncio always keeps reading from the socket in the background, whether you call `read` or not, as long as the loop is running. Trio and sync code don't. Unfortunately, httpx currently relies on the asyncio behavior to detect when an idle connection has gone stale. IMO this is a bug in httpx's idle connection handling, but that's the current state – there's some more discussion in this thread, in this comment and the next few below it: https://github.com/encode/httpx/pull/143#issuecomment-515152590

You'll have to ask Tom how he wants to deal with it here...

florimondmanca: Thanks a lot @njsmith. 🥇 From reading the thread you linked to, I learnt about `.is_readable()`. It turns out to exactly what I need for detecting whether the connection has dropped at the time of re-acquiring it.

I opened a separate issue (#327) for correctly handling connection resets in the asyncio case. So, I guess this is fixed?

florimondmanca: Hmm, sorry about that, I was returning `.is_readable()` instead of `not .is_readable()` in `is_connection_dropped()`. 🤦‍♂ 

@njsmith The commit I pushed is here: https://github.com/encode/httpx/pull/276/commits/bea9592217ec98c88769ef715414a94ff56bf55e

So now `is_readable()` returns `True` even after having restarted the server. I'm not sure why that is, but there's probably a missing step because HTTPX does seem to be relying on asyncio's background socket management — at least that's probably the reason why we only check for dropped connections once, i.e. before acquiring a new one.

njsmith: Oh right, I actually misremembered the issue we were discussing in #143 – that was about the distinction between "closed" and "readable", which is much more subtle. So arguably it's a bug that when we want to detect if a connection is reusable, then the asyncio backend checks for EOF, when ideally it should be checking for readability. But the way your patch handles it looks right to me.

httpx/concurrency/trio.py
```diff
@@ -0,0 +1,250 @@
...
+
+        while True:
+            # Check our flag at the first possible moment, and use a fine
+            # grained retry loop if we're not yet in read-timeout mode.
+            should_raise = flag is None or flag.raise_on_read_timeout
+            read_timeout = _or_inf(timeout.read_timeout if should_raise else 0.01)
+
+            with trio.move_on_after(read_timeout):
+                data = await self.stream.receive_some(max_bytes=n)
+                assert data != b"", "Connection was dropped unexpectedly"
```
njsmith: This `assert` is wrong: EOF is a totally valid thing to receive on an HTTP/1 stream (for example, if the server doesn't set `Content-Length` or `Transfer-Encoding`, then EOF is used to signal the end of the response body). You need to pass the EOF through to h11, and it will handle the EOF appropriately.

sethmlarson: We should add a testcase outside this PR for a framing-header-less response to see how HTTP11Connection handles it.

florimondmanca:
So, I think I got confused about what a *readable* socket means — which means the implementation originally pushed last week was correct. I added a bunch of code comments along with the last commit to clarify the asyncio and trio implementations of `.is_connection_dropped()`.

If all looks well, I'd love a new review from @sethmlarson, and @tomchristie's approval for merging this. (I know there were past discussions about whether trio support should land in core, or be released as a third-party package. My position is that we should merge this in core *first*, and then possibly extract it into a separate package once we have enough live experience.)

sethmlarson:
Now that 0.7.3 is out you ready to merge this?

florimondmanca:
Yup, if this sounds good to you then let's merge. I'll add docs in a subsequent PR. :)

sethmlarson:
It's happening!!! :boom:

sethmlarson:
What an amazing mountain of work, thank you @florimondmanca and @tomchristie, @pgjones, @yeraydiazdiaz, @pquentin, and @njsmith for reviewing! :tada:

florimondmanca:
Thanks to everyone who reviewed and commented! 🌟

Issue #33, opened 2019-05-02
Environment support


tomchristie:
Needs to handle anything that requests currently deals with when `trust_env=True` is set.
`.netrc`, `REQUEST_CA_BUNDLE` and anything else relevant.

First step to dealing with this would be to outline *exactly* what set of stuff `requests` includes. If anyone wants to dig into this and comment on this issue comprehensively with what set of behavior we need, then that'd be a great start.

We have `adapters/environment.py` currently stubbed out to deal with adding in this behavior.

jtimmins-post20:
If this is still important, I'd be interested in doing some research here to determine what exactly `requests` includes, and then possibly implementing proper functionality.

jhtimmins:
If this is still important, I'd be interested in doing some research here to determine what exactly requests includes, and then possibly implementing proper functionality.

sethmlarson:
This is definitely important! If you can lend a hand with discovery that'd be great.

cansarigol:
Hi, I tried to include netrc support. #177 what do you think about this approach?

sethmlarson:
We still need a way to grab environment variables and to document those environment variables for this issue to be closed.

florimondmanca:
I did some research in the Requests docs looking for environment variables it supports, and whether we support them already:

- [x] `NETRC`
- [ ] `REQUEST_CA_BUNDLE` (see [here](https://2.python-requests.org/en/master/user/advanced/#ssl-cert-verification))
- [ ] `HTTP_PROXY`, `HTTPS_PROXY` (see [here](https://2.python-requests.org/en/master/user/advanced/#proxies)), [`ALL_PROXY`](https://2.python-requests.org/en/master/community/updates/#id35), [`NO_PROXY`](https://2.python-requests.org/en/master/community/updates/#id81).

(Note that the names of the `*_PROXY` environment variables seems be a [convention](https://superuser.com/questions/944958/are-http-proxy-https-proxy-and-no-proxy-environment-variables-standard).)

To make extra sure we don't miss anything, it would probably be worth cloning the Requests repo and grepping our way using `environment variable`, `os.getenv` and `os.environ.get`.

The scope of this issue is rather broad, so if that's confirmed and as suggested by @sethmlarson in https://github.com/encode/httpx/pull/277#issuecomment-525036637, how about creating issues for each set of environment variables?

florimondmanca:
[...]

PR #277, opened 2019-08-26
Add easier debug logging for users

sethmlarson:
This builds on #240 and adds it to all request-making dispatchers. It can be added to ASGI and WSGI dispatchers as well but I'm less familiar with what information would be useful for debugging those use-cases.

This also adds a new environment variable `HTTPX_DEBUG` which can be turned on to automatically display these debug logs.

A new section `Environment Variables` has been added to the documentation with the intent of future expansion for all environment variables that HTTPX will react to.

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -1,0 +1,47 @@
+Environment Variables
+=====================
+
+The HTTPX library can be configured via environment variables.
+Here are a list of environment variables that HTTPX recognizes
+and what function they serve:
+
+HTTPX_DEBUG
+-----------
+
+Valid values: `1`, `true`
+
+If this environment variable is set to `1` or `true` then
+logging will be turned on by default to stderr about low-level
+details of the HTTP request and responses being sent and received.
+
+This can help you debug issues and see what's exactly being sent
+over the wire and to which location.
+
+Example:
+
+```python
+# test_script.py
+
+import httpx
+client = httpx.Client()
+client.request("GET", "https://google.com")
+```
+
+```console
+user@host:~$ HTTPX_DEBUG=1 python test_script.py
+20:54:17.585 - httpx.dispatch.connection_pool - acquire_connection origin=Origin(scheme='https' host='www.google.com' port=443)
+20:54:17.585 - httpx.dispatch.connection_pool - new_connection connection=HTTPConnection(origin=Origin(scheme='https' host='www.google.com' port=443))
+20:54:17.590 - httpx.dispatch.connection - start_connect host='www.google.com' port=443 timeout=TimeoutConfig(timeout=5.0)
+20:54:17.651 - httpx.dispatch.connection - connected http_version='HTTP/2'
+20:54:17.651 - httpx.dispatch.http2 - send_headers stream_id=1 headers=[(b':method', b'GET'), (b':authority', b'www.google.com'), ...]
+20:54:17.652 - httpx.dispatch.http2 - end_stream stream_id=1
+20:54:17.681 - httpx.dispatch.http2 - receive_event stream_id=0 event=<RemoteSettingsChanged changed_settings:{...}>
+20:54:17.681 - httpx.dispatch.http2 - receive_event stream_id=0 event=<WindowUpdated stream_id:0, delta:983041>
+20:54:17.682 - httpx.dispatch.http2 - receive_event stream_id=0 event=<SettingsAcknowledged changed_settings:{}>
+20:54:17.739 - httpx.dispatch.http2 - receive_event stream_id=1 event=<ResponseReceived stream_id:1, headers:[(b':status', b'200'), ...]>
+20:54:17.741 - httpx.dispatch.http2 - receive_event stream_id=1 event=<DataReceived stream_id:1, flow_controlled_length:5224 data:>
+20:54:17.742 - httpx.dispatch.http2 - receive_event stream_id=1 event=<DataReceived stream_id:1, flow_controlled_length:59, data:>
+20:54:17.742 - httpx.dispatch.http2 - receive_event stream_id=1 event=<StreamEnded stream_id:1>
+20:54:17.742 - httpx.dispatch.http2 - receive_event stream_id=0 event=<PingReceived ping_data:0000000000000000>
+20:54:17.743 - httpx.dispatch.connection_pool - release_connection connection=HTTPConnection(origin=Origin(scheme='https' host='www.google.com' port=443))
+```
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -5,6 +5,7 @@ import h11
 from ..concurrency.base import BaseStream, ConcurrencyBackend, TimeoutFlag
 from ..config import TimeoutConfig, TimeoutTypes
 from ..models import AsyncRequest, AsyncResponse
+from ..utils import get_logger
 
 H11Event = typing.Union[
     h11.Request,
@@ -20,6 +21,9 @@ H11Event = typing.Union[
 # In practice the callback will be a functools partial, which binds
 # the `ConnectionPool.release_connection(conn: HTTPConnection)` method.
 OnReleaseCallback = typing.Callable[[], typing.Awaitable[None]]
+
+
+logger = get_logger(__name__)
 
 
 class HTTP11Connection:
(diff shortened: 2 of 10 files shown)


florimondmanca:
This is a great idea. I'll take a closer look later, but do you know about [structlog](https://github.com/hynek/structlog)? Your `key=value` logging style here made me think it could be worth considering it (or another structured logging library). :-)

sethmlarson:
I initially thought to use structlog, mostly didn't want to add a dependency just so we can do debug logging on a client library. Are there other structured logging libraries out there?

florimondmanca:
I don't know of other structured logging libraries. But I think the real question here is, as you said: is adding another dependency for debug logs worth it? I'm not sure it is either. :-)

It's probably possible to build our own formatting helper, starting with something really simple.

There's a structured logging section in the [Logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#implementing-structured-logging), which could helper us build a helper to be used like this:

```python
from httpx.logging import _
logger.debug(_("log message", key="value"))
```

The cookbook recipe dumps the output as JSON, but we can take inspiration from structlog's rendering implementation [here](https://github.com/hynek/structlog/blob/e8b0218cd73eccf1fdb80b49977b801fdac39110/src/structlog/processors.py) too to achieve the style you've got here.

florimondmanca:
I think all of this can be figured out later as a refactoring step for this PR, though — just jotting down ideas here, but it's not a blocker at all. :-)

docs/environment_variables.md
```diff
@@ -0,0 +1,47 @@
+Environment Variables
+=====================
+
+The HTTPX library can be configured via environment variables.
+Here are a list of environment variables that HTTPX recognizes
```
florimondmanca:
```suggestion
Here is a list of environment variables that HTTPX recognizes
```

docs/environment_variables.md
```diff
@@ -0,0 +1,47 @@
...
+and what function they serve:
+
+HTTPX_DEBUG
```
florimondmanca:
```suggestion
`HTTPX_DEBUG`
```

docs/environment_variables.md
```diff
@@ -0,0 +1,47 @@
...
+-----------
+
+Valid values: `1`, `true`
+
+If this environment variable is set to `1` or `true` then
+logging will be turned on by default to stderr about low-level
+details of the HTTP request and responses being sent and received.
```
florimondmanca: I think the phrasing is not the clearest here — may I suggest…

> If this environment variable is set then low-level details about the execution of HTTP requests will be logged to `stderr`.

httpx/utils.py
```diff
@@ -140,3 +142,31 @@ def parse_header_links(value: str) -> typing.List[typing.Dict[str, str]]:
...
+    """Gets a `logging.Logger` instance and optionally
+    sets up debug logging if the user requests it via
+    the `HTTPX_DEBUG=1` environment variable.
+    """
+    global _LOGGER_INITIALIZED
+
+    # If the user wants debug logging on we set it up for them.
+    if not _LOGGER_INITIALIZED:
+        _LOGGER_INITIALIZED = True
+        if os.environ.get("HTTPX_DEBUG", "").lower() in ("1", "true"):
```
florimondmanca: Do we really need do validate what's in the environment variable, so long as it's non-empty? Just asking — I don't know if there are best practices about this.

```suggestion
        if os.environ.get("HTTPX_DEBUG"):
```

docs/environment_variables.md
```diff
@@ -0,0 +1,47 @@
...
+over the wire and to which location.
+
+Example:
+
+```python
+# test_script.py
+
+import httpx
+client = httpx.Client()
+client.request("GET", "https://google.com")
```
florimondmanca: I think we showcase `.get()` in other examples in the docs. Since `.request()` is not the focus here how about using `.get()` here as well for consistency?

```suggestion
client.get("https://google.com")
```

florimondmanca:
Also, will this fix https://github.com/encode/httpx/issues/33?

docs/environment_variables.md
florimondmanca: Or really? I looked at the "Developer Interface" page and e.g. `Client` is rendered in a monospace font. Perhaps it's the use of `===` / `---` instead of `#` and `##` for headers?

httpx/utils.py
florimondmanca: Oh right, good point. :+1:

docs/environment_variables.md
florimondmanca: I checked, and regardless of the header markup style (which I think we could/should convert to using shebangs since that's what we use everywhere else?) I do get monospace when using code fences:

Screenshot 2019-08-26 at 23 43 38

httpx/utils.py
```diff
@@ -140,3 +142,31 @@ def parse_header_links(value: str) -> typing.List[typing.Dict[str, str]]:
...
+
+
+def get_logger(name: str) -> logging.Logger:
+    """Gets a `logging.Logger` instance and optionally
+    sets up debug logging if the user requests it via
+    the `HTTPX_DEBUG=1` environment variable.
+    """
+    global _LOGGER_INITIALIZED
+
+    # If the user wants debug logging on we set it up for them.
```
florimondmanca: To be fair, I think this comment isn't necessary — we already have the information in the docstring and the code is very easy to follow.

docs/environment_variables.md
```diff
@@ -0,0 +1,46 @@
...
+The HTTPX library can be configured via environment variables.
+Here is a list of environment variables that HTTPX recognizes
+and what function they serve:
+
+`HTTPX_DEBUG`
+-----------
+
+Valid values: `1`, `true`
+
+If this environment variable is set to `1` or `true` then low-level
```
florimondmanca: Nitpick: let's not duplicate the information about valid values in prose? 🙏

tests/dispatch/utils.py
```diff
@@ -4,9 +4,9 @@
 import h2.config
 import h2.connection
 import h2.events
+from tests.concurrency import sleep
 
 from httpx import AsyncioBackend, BaseStream, Request, TimeoutConfig
-from tests.concurrency import sleep
```
florimondmanca: Could you run lint to undo this?

httpx/dispatch/http11.py
```diff
@@ -150,6 +166,12 @@ def __init__(
         """
         while True:
             event = self.h11_state.next_event()
+
+            if isinstance(event, h11.Data):
+                logger.debug(f"receive_event event=Data({len(event.data)})")
```
tomchristie: Should we format this as something like `Data(<5 bytes>)` to make it clear that it's not a literal?

httpx/dispatch/http11.py
```diff
@@ -73,6 +77,12 @@ def __init__(
         """
         Send the request method, URL, and headers to the network.
         """
+        logger.debug(
+            f"send_request method={request.method!r} "
+            f"target={request.url.full_path!r} "
+            f"headers={request.headers!r}"
+        )
```
tomchristie: We probably also want to log the sending of `h11.Data` and `h11.EndOfMessage` events, right?

tomchristie:
Fantastic, I'm loving this!

Couple of comments worth taking a look at.

tomchristie:
Possible later additions here:

* `INFO` level - log responses in an access log like style - method, URL, status code. We'd probably want to include all redirect requests there too.
* `WARNING` level - Dropped connections, invalid responses, timeouts etc. Basically anything where a request was started but no response was available to log.

```shell
INFO httpx.client GET https://example.com/redirect-to-home 304 <Location: https://example.com/>
INFO httpx.client GET https://example.com/ 200
WARN http.client GET https://example.com/ Failed with ReadTimeout(...)
```

httpx/utils.py
```diff
@@ -140,3 +142,30 @@ def parse_header_links(value: str) -> typing.List[typing.Dict[str, str]]:
...
+
+
+def get_logger(name: str) -> logging.Logger:
+    """Gets a `logging.Logger` instance and optionally
+    sets up debug logging if the user requests it via
+    the `HTTPX_DEBUG=1` environment variable.
+    """
+    global _LOGGER_INITIALIZED
+
+    if not _LOGGER_INITIALIZED:  # pragma: no cover
```
florimondmanca: The no cover here has got me wondering — how about adding a simple test case to check that something is output to stderr if HTTPX_DEBUG is on?

PR #284, opened 2019-08-27
Force HTTP/1.1 on short-lived connections

sethmlarson:
Makes all calls to `httpx.request()`, `httpx.get()`, etc use HTTP/1.1 since there's no reason to open an HTTP/2 connection if almost always going to have a lifetime of 1 request.


tomchristie:
Good one. The README.md and index.md examples probably needs updating alongside this.

PR #292, opened 2019-08-29
Serve test server in thread

florimondmanca:
Following https://github.com/encode/httpx/pull/276#issuecomment-525709778 cc @tomchristie

### DIFF
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -20,12 +20,17 @@ async def test_start_tls_on_socket_stream(https_server):
     timeout = TimeoutConfig(5)
 
     stream = await backend.connect("127.0.0.1", 8001, None, timeout)
-    assert stream.is_connection_dropped() is False
-    assert stream.stream_writer.get_extra_info("cipher", default=None) is None
 
-    stream = await backend.start_tls(stream, "127.0.0.1", ctx, timeout)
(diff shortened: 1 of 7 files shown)


florimondmanca:
Thanks @sethmlarson, I'll merge once the build passes. :-)

tests/test_concurrency.py
```diff
@@ -20,12 +20,17 @@
...
-    assert (await stream.read(8192, timeout)).startswith(b"HTTP/1.1 200 OK\r\n")
+        stream = await backend.start_tls(stream, "127.0.0.1", ctx, timeout)
+        assert stream.is_connection_dropped() is False
+        assert stream.stream_writer.get_extra_info("cipher", default=None) is not None
+
+        await stream.write(b"GET / HTTP/1.1\r\n\r\n")
+        assert (await stream.read(8192, timeout)).startswith(b"HTTP/1.1 200 OK\r\n")
+
+    finally:
+        await stream.close()
```
florimondmanca: For the record, I had to add this `.close()` call otherwise uvicorn would wait for the connection to terminate indefinitely and not shut down. In general I think it's best to make sure we do use the `async with` or `.close()` API on the objects manipulated in tests when possible, as we do in the core package.

PR #297, opened 2019-08-29
Tighten up response.next()

### LINKED ISSUE #37: Tighten up `.next()` behavior
Currently `.next` is set as an attribute on the response.

Really it'd be good to have:

* An explicit `async def next(self):` method
* A `.call_next` attribute that the RedirectAdapter can set.

This'll mean we can have more neatly guarded behavior...

```python
async def call_next(self):
    if not self.is_redirect:
        raise NotRedirectResponse(...)
    if self.call_next is None:
        raise CallNextNotSet("")
    return await self.call_next()
```

florimondmanca:
Fixes #37 :+1:

By the way @tomchristie, am I correct saying that `response.next()` is not currently documented?


florimondmanca:
Thanks for the review @sethmlarson, should I go ahead and merge this?

sethmlarson:
Yep looks great! :)

tomchristie:
>  am I correct saying that response.next() is not currently documented?

Yup.

Nice work here!

Issue #293, opened 2019-08-29
Requests migration story


florimondmanca:
I know that Requests compatibility is an important feature that HTTPX aims at providing. I was thinking about this and had a couple of ideas relative to having people migrating from Requests to HTTPX.

1 - **Migration guide**. If I was a user thinking of doing the migration, I'd be looking for a migration guide. It should be prominently placed on the project home page, explain the various steps to follow, and highlight key points to pay attention to (e.g. API differences and missing/replaced features). I think we already capture part of this [here](https://www.encode.io/httpx/compatibility/).

2 - **Migration QA**. To ensure that our API is *actually* Requests-compatible (at least on a basic level), we could just *test it*. For example, ensure that a simple requests-based test suite keeps passing if we search/replace `requests` with `httpx`.

3 - **Automated migration**. Pushing the idea further, we could provide some kind of tool to *automatically* upgrade code from Requests to HTTPX. (Being familiar with the Angular ecosystem, I remember that [rxjs-5-to-6-migrate](https://github.com/ReactiveX/rxjs-tslint#migration-to-rxjs-6) has made the switch from RxJS v5 to RxJS v6 much easier, and I'm sure it resulted in increased adoption overall.) I'm not sure how relevant and how much impact this will have, but I'd be interested in experimenting with this (seems like an interesting opportunity to play with the AST module or a tool like [Bowler](https://pybowler.io/)? 😄).

Mostly opening this to capture these ideas, not sure where this will go. Any thoughts much welcome!

florimondmanca:
Lemme drop this here @sethmlarson 🙃 https://twitter.com/sethmlarson/status/1174396408276955137?s=21

sethmlarson:
That'd be a true 💪 flex if I ever saw one.

yeraydiazdiaz:
Big 👍 from me.

I'm still unclear on how `httpx` would work on multithreading/multiprocessing situations though given that we use `run_in_executor` for our sync story. If I recall correctly libraries like [elasticsearch-py](https://elasticsearch-py.readthedocs.io/en/master/helpers.html) are heavily based on threading/processing solutions. Maybe it's worth investigating separately?

florimondmanca:
Closing this as I think we've got a fairly comprehensive Requests compatibility guide now, and anything more definitely won't fit in anyone's time table I believe. :-)

PR #298, opened 2019-08-30
Requests compatibility study

florimondmanca:
Refs #293 

Approach:

- Copy-paste [test_requests.py#TestRequests](https://github.com/psf/requests/blob/master/tests/test_requests.py), which tests the main Requests(2) API.
- Run it through Black and fix things so that the tests "compile" (i.e. pytest can run them although they fail/error) - this is the 1st commit.
- Review tests (see diff in the 2nd commit):
  - `skip` tests that aren't relevant for HTTPX (e.g. request building).
  - Edit tests with the HTTPX equivalent APIs if we offer the same functionality (this provides hints on migration steps from Requests to HTTPX and things that could break. Hint: *most of what's not not in the high-level `requests.get/post/etc` API*).
  - `xfail` tests that showcase a functional incompatibility.

Notes:

- A passing test does **not** mean that we're compatible out of the box — I may have edited the test to either suit the HTTPX API, or get rid of irrelevant Requests-only testing details. Either way, the **diff** of these tests is what provides information, more than the fact it's passing.
- The main `httpx.[...]` API seems top-notch compatible so far.
- There's a lot of test edition associated to the fact we `.prepare()` the request by default, while Requests does not.
- Compatibility with the redirect API (`response.history` in particular) is pretty good.
- Cookies behavior is very close too!
- HTTP Basic Auth is *almost* 100% compatible, only the validation of `username` and `password` remains.
- Our handling of POST `data` is mostly right, but there are a few bugs.
- The `files` attribute isn't validated (favors #122).
- Requests tested netrc via monkey-patching — HTTPX doesn't allow that so I replaced the test to basically do what we do in our own tests (setting the environment variable pointing at a local (but this time generated) `.netrc` file).
- This suite includes tests for HTTP proxies and Digest Auth (which have been xfailed for now).
- I'm having issues with some SSL-related tests not being able to find a certificate bundle. Maybe Requests has some special test setup to cover this — will need to investigate a bit.
- Seems we aren't supporting `response.ok`. I remember seeing a conversation explaining the decision but I couldn't find it.
- HTTPX uses `request.content` instead of `request.body` — should we consider providing a `body` alias for compatibility?
- Should `URL` be comparable with plain strings, as in `URL("http://example.com") == "http://example.com"`, to ease comparing `request.url`/`response.url` without having to `str()` them?


tests/test_requests_compat.py
```diff
@@ -0,0 +1,2049 @@
...
+    def test_params_original_order_is_preserved_by_default(self):
+        param_ordered_dict = collections.OrderedDict(
+            (("z", 1), ("a", 1), ("k", 1), ("d", 1))
+        )
+        request = requests.Request(
+            "GET", "http://example.com/", params=param_ordered_dict
+        )
+        assert request.url == "http://example.com/?z=1&a=1&k=1&d=1"
+
+    @pytest.mark.xfail(reason="HTTPX doesn't support bytes-typed 'params' yet")
```
sethmlarson: Interesting! I didn't even know this was a supported case for Requests :) Should we support it too?

tests/test_requests_compat.py
```diff
@@ -0,0 +1,2049 @@
...
+            httpbin("redirect-to"),
+            data=io.BytesIO(byte_str),
+            params={"url": "post", "status_code": 307},
+        )
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+        assert r.json()["data"] == byte_str.decode("utf-8")
+
+    @pytest.mark.xfail(
```
sethmlarson: Was there a reason this number was chosen over 20? Both are huge :) @tomchristie

tests/test_requests_compat.py
```diff
@@ -0,0 +1,2049 @@
...
+        # Session cookie should not be modified
+        assert s.cookies["foo"] == "bar"
+
+    def test_request_cookies_not_persisted(self, httpbin):
+        s = requests.session()
+        s.get(httpbin("cookies"), cookies={"foo": "baz"})
+        # Sending a request with cookies should not add cookies to the session
+        assert not s.cookies
+
+    @pytest.mark.xfail(reason="cookies object are different instances (but == would work)")
```
sethmlarson: Hmm... is there any value in the cookiejars being the same instance?

florimondmanca: Dunno. For now I’m doing this work pretty systematically. We can then go over the xfails and decide whether it’s a use case we want to support to/if it has value at all. :)

tests/test_requests_compat.py
tomchristie: I'm not clear from the context what we're talking about here?

tomchristie:
*Really* neat idea - nice one!

tests/test_requests_compat.py
sethmlarson: Sorry, was meant to be placed on a comment about the maximum number of redirects allowed being 30 instead of 20.

tomchristie: Not sure. 20 seems reasonable given this: https://stackoverflow.com/questions/9384474/in-chrome-how-many-redirects-are-too-many

florimondmanca:
Second pass is up — I'm at 50% in — and I updated the notes in the PR description.

It seems the amount of bugs or missing bits of API we're going to uncover with this is tremendous. It will definitely be worthwhile to go through each test and see whether we need to take action on the HTTPX side. :)

tests/test_requests_compat.py
florimondmanca: Refactored this to skip the `is` check — I figured what we really tested there was whether cookies were sent correctly.

StephenBrown2:
This is great work! Could we add `-rxXs` to the [noxfile](https://github.com/encode/httpx/blob/master/noxfile.py#L41) to see the [summary info on xfailed, xpassed, and skipped tests](https://docs.pytest.org/en/latest/skipping.html)?

sethmlarson:
I've been wanting to revamp our certificate bundles with the "bad" TLS certificate / errors but you might be running into an issue with not pinning pytest-httpbin. They checked their cert with the latest version.

florimondmanca:
@sethmlarson Thanks for the heads up. I can pin `pytest-httpbin` to >=1.0 — would you recommend any upper bound other than 2.0?

florimondmanca:
Alright, I'm done reworking all these tests. :tada:

Based on your available time, feel free to go through each test one by one, and please comment on whether the comments I've made make sense to you, and whether there's any action that might be worth taking (e.g. opening new bug reports or feature requests). 😄

tests/compat/test_requests.py
```diff
@@ -0,0 +1,2285 @@
...
+        exception. Otherwise, multiple headers are created from single string.
+        """
+        headers_ret = {"foo": "bar\r\nbaz: qux"}
+        headers_lf = {"foo": "bar\nbaz: qux"}
+        headers_cr = {"foo": "bar\rbaz: qux"}
+
+        # Test for newline
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin("get"), headers=headers_ret)
+        # Test
```
sethmlarson: There isn't a .stream() on sync responses?

sethmlarson: This is a bug, since it's possible to know the size of the request body via .tell() we should be setting a content length header.

tests/compat/test_requests.py
```diff
@@ -0,0 +1,2285 @@
...
+        r.raw = io.BytesIO(b"the content")
+        chunks = r.iter_content(None)
+        assert list(chunks) == [b"the content"]
+
+        r = requests.Response()
+        r.raw = io.BytesIO(b"the content")
+        with pytest.raises(TypeError):
+            chunks = r.iter_content("1024")
+
+    @pytest.mark.xfail(
```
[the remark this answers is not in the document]
tomchristie: The condition for this is presuamably that we only want to support pickling responses that have been closed, right? (We can't serialize stuff if it's stull attached to some I/O.)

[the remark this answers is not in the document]
sethmlarson: Yup! We should be able to pickle responses that are all loaded into memory.

tests/compat/test_requests.py
```diff
@@ -0,0 +1,2285 @@
...
+        ses.headers["First"] = "1"
+        ses.headers["Second"] = "2"
+        headers = collections.OrderedDict([("Third", "3"), ("Fourth", "4")])
+        headers["Fifth"] = "5"
+        headers["Second"] = "222"
+        resp = ses.get(httpbin("get"), headers=headers)
+        items = list(resp.request.headers.items())
+
+        # Remove headers automatically added by HTTPX.
+        while items[0][0] != "accept-encoding":
```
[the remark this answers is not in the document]
florimondmanca: Not sure — *if* the test case should pass (e.g. order is preserved), then all headers in the list before the first one we added (`Accept-Encoding`) can be removed, so that we're only left with the ones we want to check the order of. Besides I don't think we want to couple this with other HTTPX's behavior.

tests/compat/test_requests.py
```diff
@@ -0,0 +1,2285 @@
...
+            requests.get(httpbin_secure(), cert=(".", INVALID_PATH))
+        assert str(
+            e.value
+        ) == "Could not find the TLS key file, invalid path: {}".format(INVALID_PATH)
+
+    def test_http_with_certificate(self, httpbin):
+        r = requests.get(httpbin(), cert=".")
+        assert r.status_code == 200
+
+    @pytest.mark.xfail(reason="HTTPX does not emit such warnings - probably should?")
```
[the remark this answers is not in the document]
florimondmanca: So should we `.skip()` this test then?

tests/compat/test_requests.py
```diff
@@ -0,0 +1,2285 @@
...
+            "status": "200 OK",
+            "vary": "Accept",
+            "x-content-type-options": "nosniff",
+            "x-github-media-type": "github.beta",
+            "x-ratelimit-limit": "60",
+            "x-ratelimit-remaining": "57",
+        }
+        assert r.links["next"]["rel"] == "next"
+
+    @pytest.mark.xfail(reason="Cookies.set() doesn't have 'secure' and 'rest' options")
```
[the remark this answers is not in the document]
florimondmanca: Not a bug, but probably a missing feature? I'll mark it as a bug though.

tests/compat/test_requests.py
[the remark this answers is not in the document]
florimondmanca: I'm not sure GitHub is showing me the line you initially intended this comment to be at? (Same for the other 2 comments below)

florimondmanca: Yup, didn't realize there was one. Thinking of `test_manual_redirect_with_partial_body_read`,  I'm not sure how to deal with the `resolve_redirects()` API that we're missing. The `call_next()` that we set on responses is async, and I don't know if we can propagate the `stream` option there (or if it does so automatically).

tests/compat/test_requests.py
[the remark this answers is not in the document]
sethmlarson: Definitely!

tests/compat/test_requests.py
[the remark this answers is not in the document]
sethmlarson: You're right, let's leave as-is.

florimondmanca:
So, I think I need to take a more bite-sized approach to this PR. This PR is just too big. I don't even want to rework it to add all the nice edit comments and all, and I'm pretty sure it's a pain to review as well.

The organization of tests within Requests' own test suite is not easy to work with, because it puts everything in a single `TestRequests` class. Yet, the tests *are* grouped in logical groups (e.g. testing cookies, testing POST data, etc).

So I'm thinking of starting with identifying some groups of tests, add the first one in an isolated PR, and work my way up from there. Does that sound reasonable?

I'll close this and move on to a `transfer-encoding: chunked` approach. :wink:

StephenBrown2:
Sounds good. Definitely keep the `tests/requests_compat/test_*` structure, and group the tests in classes or not, but definitely [split out by file](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code). Godspeed!

sethmlarson:
I'm surprised it took this long for an HTTP pun to show up in this issue tracker. ;)

PR #300, opened 2019-08-31
Add server.url property in tests

### LINKED ISSUE #288: Add server.url property in tests
We are defining the URL to the uvicorn server that's running during our tests in lots of place, e.g.:

```python
# tests/client/test_async_client.py

async def test_get(server, backend):
    url = "http://127.0.0.1:8000/"
    async with httpx.AsyncClient(backend=backend) as client:
        response = await client.get(url)
    # ...

async def test_raise_for_status(server, backend):
    async with httpx.AsyncClient(backend=backend) as client:
        for status_code in (200, 400, 404, 500, 505):
            response = await client.request(
                "GET", f"http://127.0.0.1:8000/status/{status_code}"
            )
        # ...
```

As suggested by @sethmlarson in https://github.com/encode/httpx/pull/276#issuecomment-525445494, it would be better to use a `URL` property on the `server` provided by the `server` and `https_server` fixtures instead. This way, we can change the root URL in one place (close to wherre the server is actually defined), while being able to build child URLs using `server.url.copy_with(path="/echo")`.

This can probably be done by creating and using a uvicorn `Server` subclass with the `URL` instance set as a class attribute.

theruziev:
Fixes #288

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,7 +12,7 @@ from cryptography.hazmat.primitives.serialization import (
 from uvicorn.config import Config
 from uvicorn.main import Server
 
-from httpx import AsyncioBackend
+from httpx import URL, AsyncioBackend
 
 
 @pytest.fixture(params=[pytest.param(AsyncioBackend, marks=pytest.mark.asyncio)])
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -145,9 +145,13 @@ def install_signal_handlers(self) -> None:
             loop.create_task(super().serve(sockets=sockets)),
             loop.create_task(self.watch_restarts()),
         }
-
         await asyncio.wait(tasks)
 
+    @property
+    def url(self) -> URL:
+        protocol = "https" if self.config.is_ssl else "http"
+        return URL(f"{protocol}://{self.config.host}:{self.config.port}/")
```
florimondmanca: Just a nit: could you consider placing this property at the top of the class? The `serve`,  `restart` and `watch_restarts()` methods are all kind of related in a non-immediately obvious way, so I think it's best to keep them together.

theruziev: Done

tomchristie:
Good stuff!
We could also accept a pull request adding this directly to uvicorn's API itself.

PR #301, opened 2019-08-31
Add support for SSLKEYLOGFILE

### LINKED ISSUE #299: Add support for SSLKEYLOGFILE for Python 3.8b4+
Python 3.8b4 (with OpenSSL 1.1.1) added support for the `SSLKEYLOGFILE` environment variable which will be very useful for using Wireshark to do analysis of HTTP/2 requests for example.

This should be a small change within `SSLConfig._create_default_ssl_context()` see [this commit](https://github.com/python/cpython/commit/c7f7069e77c58e83b847c0bfe4d5aadf6add2e68) for the gist of how to implement.

Would also require that we document within the Environment Variables section that we support this environment variable and under what circumstances.

tomasfarias:
Addresses the following issue:

Closes #299


tomasfarias:
Tests are passing with my local installation of Python 3.8.0b4, the error suggests travis-ci may not be running the same version (as the `keylog_filename` attribute was added in 3.8.0b4), but it seems to be correct given the logs:

```
$ python --version
Python 3.8.0b4+
``` 

Does anybody have any idea why?

tomasfarias:
I started an interactive shell using the same Python 3.8 travis uses and got:

```
Python 3.8.0b4+ (heads/3.8:6922b9e, Aug 31 2019, 09:14:24)
```

Which is not quite what I get with my own Python 3.8 venv:

```
Python 3.8.0b4 (default, Aug 31 2019, 17:23:49)
```

So there might be a difference in versions after all.

florimondmanca:
@tomasfarias Could this be related to the installed OpenSSL version on Travis? The build reads:

```console
openssl version
OpenSSL 1.0.2g  1 Mar 2016
```

But the new keylog file requires OpenSSL 1.1+, right? Perhaps the `ssl` API depends on the available OpenSSL version, which could explain the error we see in the build.

florimondmanca:
I did a bit of research anticipating the need to use OpenSSL 1.1, and there doesn't seem to be an easy way to get it on Travis other than install it ourselves (or move to Docker-based testing…).

There's an [open issue](https://bugs.launchpad.net/xenial-backports/+bug/1785412) on upgrading it in Xenial though — which I assume might end up in Travis's `dist: xenial` environment if resolved?

For manual installation, I found [this commit](https://github.com/ruby/ruby/pull/1980/files) adding OpenSSL 1.1 to Ruby's TravisCI setup. I don't know yet which way we'd like to go it does happen that the OpenSSL version is causing the error here.

tomasfarias:
Thanks for pointing out the issue. We could also skip the test, mimicking what's being done [here],(https://github.com/python/cpython/commit/c7f7069e77c58e83b847c0bfe4d5aadf6add2e68#diff-e144a9cacd10921d4dae0aeac0300a6fR4410) but with `pytest.mark`.

florimondmanca:
> We could also skip the test, mimicking what's being done [here](https://github.com/python/cpython/commit/c7f7069e77c58e83b847c0bfe4d5aadf6add2e68#diff-e144a9cacd10921d4dae0aeac0300a6fR4410), but with `pytest.mark`.

Yes — that would definitely be useful for contributors who may not have OpenSSL 1.1 installed.

I think we do want to test this on CI, though.

Can you try adding `dist: bionic` on the 3.8 matrix job? It should make the build run on Ubuntu 18.04 (see [here](https://docs.travis-ci.com/user/reference/bionic)), which I believe comes with OpenSSL 1.1 ([source](https://launchpad.net/ubuntu/bionic/+source/openssl)). If it works, let's add a comment and celebrate. :-)

httpx/config.py
```diff
@@ -166,6 +168,11 @@ def _create_default_ssl_context(
         if ssl.HAS_NPN:  # pragma: no cover
             context.set_npn_protocols(http_versions.alpn_identifiers)
 
+        if hasattr(context, "keylog_filename"):
+            keylogfile = os.environ.get("SSLKEYLOGFILE")
+            if keylogfile and not sys.flags.ignore_environment:
+                context.keylog_filename = keylogfile  # type: ignore
```
florimondmanca: What happens if the specified file name points at a non-existing location? Will Python create it for us?

httpx/config.py
```diff
@@ -166,6 +168,11 @@ def _create_default_ssl_context(
         if ssl.HAS_NPN:  # pragma: no cover
             context.set_npn_protocols(http_versions.alpn_identifiers)
 
+        if hasattr(context, "keylog_filename"):
+            keylogfile = os.environ.get("SSLKEYLOGFILE")
+            if keylogfile and not sys.flags.ignore_environment:
```
florimondmanca: I believe there’s a trust_env option on the client for this kind of check — is it available here/would it make sense to use it here as well?

tests/test_config.py
```diff
@@ -171,3 +173,32 @@ def test_timeout_from_tuple():
...
+
+    ssl_config = httpx.SSLConfig()
+    ssl_config.load_ssl_context()
+
+    assert ssl_config.ssl_context.keylog_filename == filename
+
+    if old is None:
+        del os.environ["SSLKEYLOGFILE"]
+    else:
+        os.environ["SSLKEYLOGFILE"] = old
```
florimondmanca: Note that if the test fails we never reach this part of the code. We’d probably want to extract this logic as an override_environ() context manager helper (if pytest doesn’t provide it yet?). Not sure if we aren’t already doing this kind of override in other tests though — might be worth seeing if we can reuse any existing utility first?

.travis.yml
```diff
@@ -20,6 +20,7 @@ matrix:
       env: NOX_SESSION=test-3.7
     - python: 3.8-dev
       env: NOX_SESSION=test-3.8
+      dist: bionic
```
florimondmanca: Also, can we just mention here that we’re mostly setting this to get OpenSSL 1.1.1+?

httpx/config.py
tomasfarias: If it's a non-existing file, Python creates it. However, if it's a path that contains a non-existing directory it will raise `FileNotFoundError`.

.travis.yml
tomasfarias: Will add a comment

tests/test_config.py
tomasfarias: Couldn't find any override similar to this one in other tests: all other tests that set an environment variable do so without resetting the environment after the fact. Checked pytest and it uses [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html#monkeypatching-environment-variables) for it. Another alternative is using [this](https://github.com/depop/django-concurrent-test-helper/blob/master/django_concurrent_tests/utils.py#L151).

httpx/config.py
sethmlarson: I think it makes sense for us to ignore this variable when we use `trust_env=False`. I guess that'll have to be piped into `SSLConfig`.

tomasfarias: Gave it a shot in the latest commit. I'm a bit unsure about the default behavior of `trust_env` in `SSLConfig`: currently it defaults to `trust_env=None`, but I wonder whether it should imitate the behavior of `BaseClient` when it's set to `None`.

tests/test_config.py
florimondmanca: The monkeypatch way looks good. :)

httpx/config.py
florimondmanca: On re-reading yes, we probably want to have the same default in both places.

What would we say about setting the default to True though, instead of None and then defaulting that to True? It would be simpler, and I don’t think we need to support the idea of « I passed a None value to get the default behavior » here. Thoughts @sethmlarson?

httpx/dispatch/connection.py
```diff
@@ -37,9 +37,10 @@ def __init__(
         http_versions: HTTPVersionTypes = None,
         backend: ConcurrencyBackend = None,
         release_func: typing.Optional[ReleaseCallback] = None,
+        trust_env: bool = None,
```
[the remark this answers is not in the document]
tomasfarias: Sounds good, I'll do that. Thanks for your comments

docs/environment_variables.md
```diff
@@ -44,3 +44,38 @@ user@host:~$ HTTPX_DEBUG=1 python test_script.py
 20:54:17.742 - httpx.dispatch.http2 - receive_event stream_id=0 event=<PingReceived ping_data:0000000000000000>
 20:54:17.743 - httpx.dispatch.connection_pool - release_connection connection=HTTPConnection(origin=Origin(scheme='https' host='www.google.com' port=443))
 ```
+
+`SSLKEYLOGFILE`
+-----------
+
+Valid values: a filename
+
+If this environment variable is set, TLS keys will be appended to the specified file, creating it if it doesn't exist, whenever key material is generated or received. The keylog file is designed for debugging purposes only.
```
tomchristie: It might be helpful to explain the motivation a bit more, and/or link to some resource(s) here giving an example of why & how you can use this. Right now it's couched quite technically, so I don't get much sesnse of the intent, without a bit of researching.

(eg. https://jimshaver.net/2015/02/11/decrypting-tls-browser-traffic-with-wireshark-the-easy-way/ or whatever else)

tomchristie:
Ace bit of improvement, this. Thank you!

Issue #174, opened 2019-07-29
Create a Client.build_request() method


sethmlarson:
See: https://github.com/encode/httpx/issues#issuecomment-515950955

VeNoMouS:
@sethmlarson bad reference link

florimondmanca:
@sethmlarson I second @VeNoMouS — the link leads to the issues board. Any rememberance of where it should have gone to?

sethmlarson:
I don't recall actually. :( That's a shame!

VeNoMouS:
Waaaah Waaaah waaaaaaaaaah

florimondmanca:
@sethmlarson Okay! Do you recall any more context on what `Client.build_request()` should do?

sethmlarson:
I think it was meant to combine all the `merge_*()` functions? Do you remember @tomchristie?

tomchristie:
> I think it was meant to combine all the merge_*() functions?

Yup - we'd like to be able to do something like...

```python
request = client.build_request()
response = await client.send(request)
```

PR #303, opened 2019-09-01
narrow client request methods down

cansarigol:
#174 
Hi, I tried to narrow `Client` methods down with `TypeDict`. However, does not seem possible until [the issue](https://github.com/python/mypy/issues/4441) fixed. This code may be useful.

### DIFF
--- a/httpx/decorators.py
+++ b/httpx/decorators.py
@@ -1,0 +1,17 @@
+import functools
+import inspect
+import typing
+
+
+def share_signature(source: typing.Callable) -> typing.Callable:
+    sig = inspect.signature(source)
+
+    def wrap(func):  # type: ignore
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):  # type: ignore
+            return func(*args, **kwargs)
+
+        wrapped.__signature__ = sig
+        return wrapped
+
+    return wrap
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -17,6 +17,7 @@ from .config import (
     TimeoutTypes,
     VerifyTypes,
 )
+from .decorators import share_signature
 from .dispatch.asgi import ASGIDispatch
 from .dispatch.base import AsyncDispatcher, Dispatcher
 from .dispatch.connection_pool import ConnectionPool
@@ -47,6 +48,7 @@ from .models import (
     ResponseContent,
     URLTypes,
 )
+from .types import RequestType
 from .utils import get_netrc_login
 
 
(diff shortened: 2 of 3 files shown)


florimondmanca:
Interesting take, thanks! It’s great to see a possible solution to better typing support of **kwargs.

How well does this work with autocompletion, though? I suppose this was the main motivation for repeating the keyword arguments in the client methods.

I think there’s a big blocker, though — TypeDict is a new, 3.8-only feature of the typing module, but we support 3.6 and up. So I don’t think this will make it but I really appreciate you giving this a shot.

(Though, I’m not sure the approach demonstrated here was the idea behind #174. We definitely need to discuss the proposed scope and API of build_request() there.)

cansarigol:
Thanks for review.
> How well does this work with autocompletion, though? I suppose this was the main motivation for repeating the keyword arguments in the client methods.

I've been trying to fix two things. First, use one line client functions and decorate them with a generic function to handle their signature. Second, when `mypy` runs, it introspects generic arguments from `kwargs`. I don't know how IDEs handle moving autocomplete arguments.

>I think there’s a big blocker, though — TypeDict is a new, 3.8-only feature of the typing module, but we support 3.6 and up. So I don’t think this will make it but I really appreciate you giving this a shot.

Are you sure about this? because it comes from `mypy_extensions` module and I didn't get any error.

florimondmanca:
> Are you sure about this? because it comes from mypy_extensions module and I didn't get any error.

Oh, [indeed](https://mypy.readthedocs.io/en/stable/more_types.html?highlight=mypy_extensions#typeddict). Cool!

> First, use one line client functions and decorate them with a generic function to handle their signature.

From experience, signature-altering decorators don't play that well with IDEs (because those generally rely on the initial signature of the function). I'll give it a go in VSCode when I have some time, and we'll see. :)

cansarigol:
by the way, I came across [this issue](https://github.com/python/typing/issues/270) I'm watching it.

florimondmanca:
VSCode isn't giving me any autocompletions, and IPython can't see the signature either:

Screenshot 2019-09-02 at 09 21 44

Compare with `httpx.get`:

Screenshot 2019-09-02 at 09 22 16

httpx/decorators.py
```diff
@@ -0,0 +1,13 @@
...
+def request_signature(func):  # type: ignore
+    def wrapped(self, url, **kwargs):  # type: ignore
+        new_signature = inspect.signature(self.request)
+        new_signature.parameters["method"].replace(
+            kind=inspect.Parameter.POSITIONAL_ONLY
+        )
+        func.__signature__ = new_signature
+        return func(self, url, **kwargs)
+
+    return wrapped
```
florimondmanca: I don't think this is doing what you'd like it to do: if we say `client = httpx.Client()` then `client.get` will refer to `wrapped`, not the original function. We should assign `wrapped.__signature__` instead of `func.__signature__` — but there's no way to get access to `request()` from outside `wrapped`.

We *could* get it to work using:

```python
def share_signature(source: typing.Callable):
    sig = inspect.signature(source)

    def wrap(func):
        @functools.wraps(func)
        def wrapped(*args, **kwargs):
            return func(*args, **kwargs)

        wrapped.__signature__ = sig
        return wrapped

    return wrap
```

and then

```python
def request(self, ...): ...

@share_signature(request)
def get(self, ...): ...
```

At least, it would allow `help()` to display information about the signature. But autocompletion in the terminal (IPython) and VSCode still don't work.

cansarigol:
at least I added `share_signature` 😄 

```
>>> import httpx
>>> httpx.Client.get.__signature__
<Signature (self, method: str, url: Union[ForwardRef('URL'), str], *, data: Union[dict, str, bytes, Iterator[bytes]] = None, files: Dict[str, Union[IO[
~AnyStr], Tuple[str, IO[~AnyStr]], Tuple[str, IO[~AnyStr], str]]] = None, json: Any = None, params: Union[ForwardRef('QueryParams'), Mapping[str, Union
[str, int, float, bool, NoneType]], List[Tuple[str, Union[str, int, float, bool, NoneType]]], str] = None, headers: Union[ForwardRef('Headers'), Dict[~
AnyStr, ~AnyStr], List[Tuple[~AnyStr, ~AnyStr]]] = None, cookies: Union[ForwardRef('Cookies'), http.cookiejar.CookieJar, Dict[str, str]] = None, stream
: bool = False, auth: Union[Tuple[Union[str, bytes], Union[str, bytes]], Callable[[ForwardRef('AsyncRequest')], ForwardRef('AsyncRequest')]] = None, al
low_redirects: bool = True, cert: Union[str, Tuple[str, str], Tuple[str, str, str]] = None, verify: Union[str, bool, ssl.SSLContext] = None, timeout: U
nion[float, Tuple[float, float, float], ForwardRef('TimeoutConfig')] = None, trust_env: bool = None) -> httpx.models.Response>
>>> 
```

florimondmanca:
So, I’m going to close this because I think it’s not something we need, esp with respect to maintaining the readability, ease of maintenance and editor support of the client. But thanks again for trying it out! (In the future, feel free to open issues to discuss this kind of idea, if you’d like a quick feedback on whether it’s something worth working on.)

PR #305, opened 2019-09-02
Digest auth middleware

yeraydiazdiaz:
This PR implements Digest authentication as a middleware.

Digest is mentioned as pending work in the README.

Tested against [httpbin.org's digest auth endpoint](https://httpbin.org/#/Auth/get_digest_auth__qop___user___passwd___algorithm_) and mimicking [request's implementation](https://github.com/kennethreitz/requests/blob/master/requests/auth.py#L108).

Note this PR does not include handling the `auth-int` Quality of Protection (qop) option in the RFC mainly because requests does not do it either. I'm happy to add it in a separate PR if it is deemed necessary though.


httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+            return hash_func(data).hexdigest().encode("utf-8")
+
+        def keyed_digest(secret: bytes, data: bytes) -> bytes:
+            return digest(b":".join((secret, data)))
+
+        A1 = b":".join((username, realm, password))
+        HA1 = digest(A1)
+
+        path = request.url.full_path.encode("utf-8")
+        A2 = b":".join((request.method.encode("utf-8"), path))
```
sethmlarson: method is guaranteed to be ASCII, you can use .encode()

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Construct Authenticate header string
+        if nonce != self._previous_nonce:
+            self._nonce_count = 1
+        else:
+            self._nonce_count += 1
+        self._previous_nonce = nonce
+        nc_value = b"%08x" % self._nonce_count
+
+        s = str(self._nonce_count).encode("utf-8")
```
sethmlarson: ASCII

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+        s += nonce
+        s += time.ctime().encode("utf-8")
```
sethmlarson: ASCII

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+        s += os.urandom(8)
+
+        cnonce = hashlib.sha1(s).hexdigest()[:16].encode("utf-8")
```
sethmlarson: ASCII

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+        if algorithm.lower().endswith("-sess"):
+            A1 += b":".join((nonce, cnonce))
+
+        if algorithm == "MD5-SESS":
+            HA1 = digest(b":".join((HA1, nonce, cnonce)))
+
+        if qop == b"auth" or b"auth" in qop.split(b","):
```
sethmlarson: Is this a space-separated list potentially?

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+            format_args["algorithm"] = algorithm
+        if qop:
+            format_args["qop"] = b"auth"
+            format_args["nc"] = nc_value
+            format_args["cnonce"] = cnonce
+
+        header_value = ", ".join(
+            [
+                '{}="{}"'.format(
+                    key, value if isinstance(value, str) else value.decode("utf-8")
```
sethmlarson: Maybe we need a `to_str()` function in `utils`?

httpx/utils.py
```diff
@@ -169,3 +169,12 @@ def get_logger(name: str) -> logging.Logger:
             logger.addHandler(handler)
 
     return logging.getLogger(name)
+
+
+def safe_encode(str_or_bytes: typing.Union[str, bytes]) -> bytes:
```
sethmlarson: Can we change this to `to_bytes()` to match `six`? Also this is almost certainly applicable in other places in the codebase, let's open a separate small PR to apply that change? :)

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+        return "Digest " + header_value
+
+    def _parse_header(self, header: str) -> dict:
+        result = {}
+        for item in header[7:].split(","):
+            key, value = item.strip().split("=")
+            value = value[1:-1] if value[0] == value[-1] == '"' else value
+            result[key] = value
+
+        return result
```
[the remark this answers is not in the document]
florimondmanca: On reflection, I’m not sure how stable mypy_extensions is, so it’s probably better to not explicitly depend on it. We can still build a quick NamedTuple instead — I do think there are benefits to having mypy help us type-check this parsing result.

yeraydiazdiaz:
Thanks for the reviews @sethmlarson and @florimondmanca, brace for incoming commits 😄

httpx/middleware/auth.py
```diff
@@ -0,0 +1,177 @@
...
+        if self._should_return_digest_auth(response):
+            request.headers["Authorization"] = self._build_auth_header(
+                request, response
+            )
+            return await self(request, get_response)
+
+        return response
+
+    def _should_return_digest_auth(self, response: AsyncResponse) -> bool:
+        auth_header = response.headers.get("www-authenticate")
```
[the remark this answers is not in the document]
yeraydiazdiaz: I'm not particularly happy forcing the user to instantiate and pass `HTTPDigestAuth` ahead of time while it's perfectly possible to derive such need from the header. If we can live with it for now I'd say it's probably worth a separate PR.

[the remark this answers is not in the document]
florimondmanca: @yeraydiazdiaz But how would we get the username/password if we only react to the WWW-Authenticate header? This is how we do for basic auth already, and we need to assume that the user knows what kind of authentication the endpoint they’re calling requires. Or maybe I’m missing the point?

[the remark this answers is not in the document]
yeraydiazdiaz: No, you're right. I think having to instantiate an `HTTPDigestAuth` is an indication of what you're referring to as the user knowing the type of authentication required beforehand, which it perfectly acceptable and what requests does as well.

That being said it does feel like we should create other `HTTP<type>Auth` classes as well.

Getting back to @sethmlarson's original point, are we happy with having the `WWW-Authenticate` header issue in a different PR?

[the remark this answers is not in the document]
florimondmanca: Yes, I’m okay with refining the processing of WWW-Authenticate later too. :)

yeraydiazdiaz:
This is now ready for re-review after implementing most of the comments.

@florimondmanca I couldn't find a clean way of applying your suggestion of laying the code out as a hint for the general algorithm since values produced in the calculation of the response are also used in the final digest value. I felt like after it was already cleaner after the minor refactorings but if you have suggestions let me know.

httpx/middleware/auth.py
```diff
@@ -0,0 +1,200 @@
...
+        return StatusCode.is_client_error(response.status_code) and (
+            auth_header is None or "digest" in auth_header.lower()
+        )
```
florimondmanca: What happens if we send the authorization header after receiving the first 401, but it is wrong (eg wrong username and password)?

The ‘is None’ check part makes me think that we’d go in an infinite loop: we would build and send the authorization header again, and fail, build and send it, and so on...

I suppose that the 401 if we pass wrong credentials won’t contain the WWW-Authenticate header, so should we instead check that it is present and digest is in there?

httpx/middleware/auth.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @classmethod
+    def from_header(cls, header: str) -> "DigestAuthChallenge":
+        """Returns a challenge from a Digest WWW-Authenticate header.
+
+        These take the form of:
+        `Digest realm="realm@host.com",qop="auth,auth-int",nonce="abc",opaque="xyz"`
+        """
+        scheme, _, fields = header.partition(" ")
+        assert scheme.lower() == "digest"
```
florimondmanca: Should we raise a ProtocolError here as well?

httpx/middleware/auth.py
```diff
@@ -0,0 +1,200 @@
...
+    @classmethod
+    def from_header_dict(cls, header_dict: dict) -> "DigestAuthChallenge":
+        realm = header_dict["realm"].encode()
+        nonce = header_dict["nonce"].encode()
+        qop = header_dict["qop"].encode() if "qop" in header_dict else None
+        opaque = header_dict["opaque"].encode() if "opaque" in header_dict else None
+        algorithm = header_dict.get("algorithm")
+        return cls(
+            realm=realm, nonce=nonce, qop=qop, opaque=opaque, algorithm=algorithm
+        )
```
florimondmanca: I have the feeling we’re over complicating this part a bit — a plain namedtuple and a function that builds it out of parsed data would be enough, in particular since we’re not really using these two class methods anywhere else. But I think the code is understandable enough, so perhaps you can call me on nitpicking. :)

httpx/middleware/auth.py
yeraydiazdiaz: That's a really good point, as it stands we enter an infinite loop since the server keeps returning 401 with a different nonce which we are not handling. 👍

httpx/middleware/auth.py
yeraydiazdiaz: Not sure, I don't particularly like namedtuples (might be `attrs` bias) and I believe we cannot use dataclasses so I went for a class since we're supporting 3.6.

Is there anything particularly complicated about this approach?

httpx/middleware/auth.py
```diff
@@ -0,0 +1,200 @@
...
+    def from_header(cls, header: str) -> "DigestAuthChallenge":
+        """Returns a challenge from a Digest WWW-Authenticate header.
+
+        These take the form of:
+        `Digest realm="realm@host.com",qop="auth,auth-int",nonce="abc",opaque="xyz"`
+        """
+        scheme, _, fields = header.partition(" ")
+        assert scheme.lower() == "digest"
+        header_dict: typing.Dict[str, str] = {}
+        for field in fields.split(","):
```
tomchristie: This doesn't look right on first sight. Eg. `Digest realm="realm@host.com",qop="auth,auth-int",nonce="abc",opaque="xyz"` contains commas inside the quotes which looks like it'd break the split.

florimondmanca: Good point @tomchristie. Maybe we should be looking at using regexes at this point?

httpx/middleware/auth.py
florimondmanca: > Is there anything particularly complicated about this approach?

Not really — I’m fine with the current approach. :) (Though I don’t see what’s not to like about namedtuples, in particular when using the typing.NamedTuple style.)

httpx/middleware/auth.py
tomchristie: Best prior art to consider would be:

* What does requests do here?
* What does werkzeug do here?

httpx/middleware/auth.py
yeraydiazdiaz: I mostly think they obscure what would otherwise be a lovely simple class 😄 

https://www.attrs.org/en/stable/why.html#namedtuples sums other issues as well, though not directly relevant here

httpx/middleware/auth.py
yeraydiazdiaz: Looks like they're using the [same helper function](https://github.com/psf/requests/blob/3d968ff34dd9945f2a86396001408ba69c1ab3fd/requests/utils.py#L349)

httpx/middleware/auth.py
florimondmanca: Cool, thanks for sharing the context on why attrs thinks namedtuples are bad. :)

I think they do have their use as a way of structuring/type-hinting immutable return values — as is properly noted in one of the footnotes. I don’t like the collections version of it (the one displayed in the attrs docs), but the type hint-capable version in the typing module is pretty neat for this use case.

httpx/middleware/auth.py
florimondmanca: @yeraydiazdiaz Actually it seems to me they use a urllib helper, see [here](https://github.com/psf/requests/blob/3d968ff34dd9945f2a86396001408ba69c1ab3fd/requests/compat.py#L59) and [here](https://github.com/psf/requests/blob/3d968ff34dd9945f2a86396001408ba69c1ab3fd/requests/utils.py#L372).

tomchristie: Fabulous, that's a great pointer.

yeraydiazdiaz:
@encode/httpx-maintainers this is ready for re-review. Thanks for all the helpful comments and links. 🌟

httpx/middleware/auth.py
```diff
@@ -0,0 +1,219 @@
...
+from ..utils import to_bytes, to_str, unquote
+from .base import BaseMiddleware
+
+
+class BasicAuthMiddleware(BaseMiddleware):
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ):
+        if isinstance(username, str):
+            username = username.encode("latin1")
```
sethmlarson: We should probably use UTF-8 here given this comment: https://github.com/encode/httpx/pull/259#discussion_r320645771

Eventually we should merge the two implementations, I can do that as a part of my PR once this lands. :)

tests/client/test_auth.py
```diff
@@ -127,3 +200,142 @@ def test_auth_invalid_type():
...
+def test_digest_auth(algorithm, expected_hash_length, expected_response_length):
+    url = "https://example.org/"
+    auth = HTTPDigestAuth(username="tomchristie", password="password123")
+
+    with Client(dispatch=MockDigestAuthDispatch(algorithm=algorithm)) as client:
+        response = client.get(url, auth=auth)
+
+    assert response.status_code == 200
+    auth = response.json()["auth"]
+    assert auth.startswith("Digest")
```
sethmlarson: Super nit-picky but we could do `auth.startswith("Digest ")` to ensure that the header is being formed correctly. Currently returning `WWW-Authenticate: DigestZ realm=...` would pass this test and below.

yeraydiazdiaz: That's not nit-picky at all :)

That scenario is covered in the malformed header test below. The `partition` call would return a `scheme` equal to `DigestZ` which the `if scheme.lower() != "digest"` guard would catch.

docs/quickstart.md
```diff
@@ -379,3 +379,25 @@ value to be more or less strict:
...
+>>> httpx.get("https://example.com", auth=("my_user", "password123"))
+```
+
+To provide credentials for Digest authentication you need to instantiate
+an `HTTPDigestAuth` object providing, again, the plaintext username and
+password as arguments. This object can be then passed as the `auth` argument
+to the request methods as above:
+
+```python
+>>> auth = httpx.DigestAuth("my_user", "password123")
```
tomchristie: I think there's a bit of inconsistency here, right?
The class is currently `HTTPDigestAuth`.
I'd suggest that the class should be simply `DigestAuth`, as in this example, which scans more nicely to me.

tests/client/test_auth.py
yeraydiazdiaz: I've added some more test cases including yours for sanity. :)

tomchristie:
This is fabulous stuff!

I've got one suggestion here, to consider...

* Currently `auth=` accepts either a tuple, or a `request->request` callable.
* I think we should change that so that `auth=` accepts either a tuple, or a `request->request` callable, *or an instance of BaseMiddleware*.

Then we can:

* Drop the `DigestAuth` model.
* Just name the basic auth middleware `BasicAuth`.
* Just name the digest auth middleware `DigestAuth`.
* Import both `BasicAuth` and `DigestAuth` from the top level of the package.

That way we're not introducing an extra model, and we have a nice consistency where `auth=httpx.BasicAuth(username=..., password=...)` is also a valid expression.

Make sense?

docs/quickstart.md
sethmlarson: Requests names theirs `HTTPDigestAuth`.

httpx/client.py
```diff
@@ -217,6 +219,11 @@ def _get_auth_middleware(
         if callable(auth):
             return CustomAuthMiddleware(auth=auth)
 
+        if isinstance(auth, HTTPDigestAuth):
+            return HTTPDigestAuthMiddleware(
+                username=auth.username, password=auth.password
+            )
```
tomchristie: With my "support `auth=<some middleware instance>`" suggest, this would look more like...

```python
if isinstance(auth, tuple):
    return BasicAuth(username=auth[0], password=auth[1])

elif isinstance(auth, BaseMiddleware):
    return auth

elif callable(auth):
    return CustomAuth(auth=auth)

else:
    # Bail!
```

docs/quickstart.md
sethmlarson: But we don't have an `auth` module anymore so it'll have to be renamed regardless. This is probably fine then. :) I do like the name without the HTTP prefix better.

tomchristie:
Plus, if a user is implementing a custom auth style that needs to be introduced as middleware, then they can still install it using `auth=...`.

That'll alsp give us better ability to control any "only run auth if we're still on the initial request origin". (Because if a user installed auth as middleware instead, we don't exactly *know* that it's an authenticating thing.)

yeraydiazdiaz:
Sounds good to me! The only thing that comes to mind is the check for `BaseMiddleware` may be a bit broad, but given that we're only allowing users to pass them as `auth` I think it should be fine. 👍

yeraydiazdiaz:
@tomchristie this is ready for review with your suggested changes

httpx/middleware/auth.py
```diff
@@ -0,0 +1,219 @@
...
+
+class BasicAuth(BaseMiddleware):
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ):
+        if isinstance(username, str):
+            username = username.encode("utf-8")
+
+        if isinstance(password, str):
+            password = password.encode("utf-8")
```
florimondmanca: Thoughts on using to_bytes() here as well? :)

httpx/client.py
```diff
@@ -212,28 +209,30 @@ def _get_auth_middleware(
...
+        elif isinstance(auth, BaseMiddleware):
+            return auth
+        elif callable(auth):
+            return CustomAuth(auth=auth)
 
         if auth is not None:
             raise TypeError(
                 'When specified, "auth" must be a (username, password) tuple or '
                 "a callable with signature (AsyncRequest) -> AsyncRequest "
+                "or an subclass of BaseMiddleware "
```
florimondmanca: Typo: “*a* subclass”?

httpx/middleware/auth.py
tomchristie: Seems like that’d just be an obfuscation here.

florimondmanca: How is it different from what we’re doing in DigestAuth, though?

tomchristie: Oh sorry, I missed that it was str_or_bytes. Yeah seems fair then.

florimondmanca:
One question: have we validated that this works when making a *real* request to a server exposing digest auth? :)

httpx/models.py
```diff
@@ -61,6 +61,7 @@
 AuthTypes = typing.Union[
     typing.Tuple[typing.Union[str, bytes], typing.Union[str, bytes]],
     typing.Callable[["AsyncRequest"], "AsyncRequest"],
+    "BaseMiddleware",  # type: ignore
```
florimondmanca: If we ignore this to prevent circular imports, consider adding this import:

```python
if typing.TYPE_CHECKING:
    from .middleware.base import BaseMiddleware
```

so that mypy can resolve "BaseMiddleware".

httpx/middleware/auth.py
```diff
@@ -0,0 +1,219 @@
...
+
+class BasicAuth(BaseMiddleware):
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ):
+        if isinstance(username, str):
+            username = to_bytes(username)
+
+        if isinstance(password, str):
+            password = to_bytes(password)
```
florimondmanca: Sorry to nitpick but I think the instance check is already made by to_bytes(), isn’t it?

If mypy complains then perhaps we can consider converting to bytes directly when building the `userpass` below?

httpx/models.py
yeraydiazdiaz: Yes, that's why it's there. I did consider that option but seemed odd to add an `if` statement for it, but if that is The Right Thing™️  I'm happy to add it 🙂

yeraydiazdiaz:
Glad you asked! I have been testing this against httpbin.org which has been quite useful. So much in fact that I thought we should consider having a "integration" test suite using [its Docker image](https://hub.docker.com/r/kennethreitz/httpbin/).

Not sure how we feel about that, but just now I noticed that my test script fails for some of the algorithms 😓

httpx/models.py
yeraydiazdiaz: Hmm, that's annoying, adding that if upsets flake8 becuase `BaseMiddleware` is not being used 😕

florimondmanca: I think it’s actually a pretty standard pattern. I’ve been using it in various places where I’d only need to import something for typing purposes. Mypy recommends it here: https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

yeraydiazdiaz: Doesn't look like it does, I've ignored that error on that line for now. Thanks 👍

florimondmanca:
@yeraydiazdiaz Did you know about [pytest-httpbin](https://github.com/kevin1024/pytest-httpbin)? It spins up a local httpbin instance (Flask server) for requesting during tests. I provide an example in #298 since the Requests test suite use it and I didn’t want to change the test setup too much.

We’re already live-testing most of the time via the uvicorn server, but I agree providing integration tests for cases when we use a mock dispatcher (here and in HTTP/2) would be nice. Not sure if httpbin supports HTTP/2 though. Anyway, I think if we agree that we’ll end up using pytest-httpbin via #298, you could setup an integration test with it here?

yeraydiazdiaz:
Oh cool! I don't know how I missed that. I'll definitely add it to the PR.

Any thoughts on whether we should keep the mock dispatch tests?

florimondmanca:
> Any thoughts on whether we should keep the mock dispatch tests?

I think we should keep those that can’t be addressed via httpbin (eg special error cases or situations).

yeraydiazdiaz:
So turns out it wasn't a bug in the implementation but rather a fairly subtle usability bug.

Since the `DigestAuth` holds state you cannot pass an instance to the client and reuse it for subsequent requests; you must pass a new instance to each request for it to work correctly. I've pushed a check to disallow passing `DigestAuth` instances on client initialization.

Side note about tests with pytest-httpbin, I think I'll leave it out of this PR for clarity. If this PR is merged before #298 I'm happy to add the tests there.

sethmlarson:
I'd think we want people passing in auth as a part of client init. This will be a problem when people implement custom middleware as well.

florimondmanca:
> Since the DigestAuth holds state you cannot pass an instance to the client and reuse it for subsequent requests; you must pass a new instance to each request for it to work correctly.

Ah, good point. Since DigestAuth() only accepts static arguments (username and password), I think we should be able to rework the implementation to hide this detail from the user.

For example, the whole content of DigestAuth.__call__() could be moved into a separate (stateful) helper class which DigestAuth would instanciate and defer request handling to — if that makes sense?

(One learning point of this is that middleware shouldn’t hold per-request state themselves.)

tomchristie:
> you must pass a new instance to each request for it to work correctly. I've pushed a check to disallow passing DigestAuth instances on client initialization.

I don't think that's the right thing to do. Broadly we *shouldn't* store state on middleware instances full-stop, because as you identify, they're shared across requests. It's slightly different in the case of the digest nonce, because essentially that's a single globally shared counter.

The `previous_nonce_check` should be *local* state, created in __call__ or wherever else that calls into. We *probably* want to have the `nonce_count` itself be a global, since it's a bit clearer then that what it's relationship is.

httpx/middleware/auth.py
```diff
@@ -0,0 +1,249 @@
...
+
+    async def __call__(
+        self, request: AsyncRequest, get_response: typing.Callable
+    ) -> AsyncResponse:
+        response = await get_response(request)
+        if not (
+            StatusCode.is_client_error(response.status_code)
+            and "www-authenticate" in response.headers
+        ):
+            self._num_401_responses = 0
```
sethmlarson: Difference in `self.num_401_responses` and `self._num_401_responses`

httpx/middleware/auth.py
```diff
@@ -0,0 +1,249 @@
...
+        cnonce = self._get_client_nonce(nonce_count, challenge.nonce)
+
+        HA1 = digest(A1)
+        if challenge.algorithm.lower().endswith("-sess"):
+            HA1 = digest(b":".join((HA1, challenge.nonce, cnonce)))
+
+        qop = challenge.qop
+        if not qop:
+            digest_data = [HA1, challenge.nonce, HA2]
+        elif qop == b"auth" or b"auth" in qop.split(b",") or b"auth" in qop.split(b" "):
```
sethmlarson: So my comment about being able to handle spaces was we have to be able to handle `qop` values like the one in [RFC 7616 Section 3.9.1](https://tools.ietf.org/html/rfc7616#section-3.9.1) so we should pull all values via something like:

`qop = re.split(b", ?", challenge.qop)`

and then check for inclusiveness (`"auth" in qop`). We should only raise the `NotImplementedError` if `auth` isn't in the list.

httpx/middleware/auth.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def __init__(
+        self, username: bytes, password: bytes, per_nonce_count: typing.Dict[bytes, int]
+    ) -> None:
+        self.username = username
+        self.password = password
+        self.per_nonce_count = per_nonce_count
+        self.num_401_responses = 0
+
+    async def __call__(
```
sethmlarson: I'm still worried about this middleware not being able to handle multiple requests concurrently. Unless we mitigate somehow we're going to have to warn users implementing middleware that hold onto state about how to properly implement them and the pitfalls.

yeraydiazdiaz: > I'm still worried about this middleware not being able to handle multiple requests concurrently

Not sure what you mean? It is capable of handling concurrent requests, it's just that there's global state to keep track of. Or am I missing something?

httpx/middleware/auth.py
```diff
@@ -0,0 +1,257 @@
...
+        self, request: AsyncRequest, get_response: typing.Callable
+    ) -> AsyncResponse:
+        request = self.auth(request)
+        return await get_response(request)
+
+
+class DigestAuth(BaseMiddleware):
+    per_nonce_count: typing.Dict[bytes, int] = defaultdict(lambda: 0)
```
yeraydiazdiaz: A note on this global variable. Requests has a much pragmatic approach to nonce counting, it simply keeps track of the *last* nonce and increases the nonce count if it is the same. However I chose to adhere a bit more closely to the RFC at the potential risk of this structure growing uncontrollably.

I'm not sure what the best solution for this would be. Maybe keep the las N nonce values? Reset on successful authentication?

tomchristie: 1. Just make it a global. It's actually a benefit for it to be shared across clients, across threads, across tasks.
2. Make it essentially be an LRU dict, storing N items. (1000, 10000, or whatever). Whenever it's more than N items, delete the first entry. (Easy now that python dicts are ordered)
3. We don't need the `_RequestDigestAuth` class. Nonce state stuff can just be a dumb global, and everything can just be implemented on the middleware class.

yeraydiazdiaz: > Just make it a global

It is a mutable class variable so it should qualify as global, do you mean moving it to the module?

> We don't need the _RequestDigestAuth class. Nonce state stuff can just be a dumb global, and everything can just be implemented on the middleware class.

I had this intuition as well, but the `num_401_requests` can't really be global since it would prevent concurrent non-authenticated requests from going through the flow correctly. For instance if I fire 3 concurrent requests the only the first one would go through the flow and the rest would assume the legitimate 401 responses are due to credentials being invalid and return the 401 as the final response.

Following closely the RFC the ideal solution would be to prevent the additional requests to be sent until the first authentication succeeds and then reuse the auth header (which the RFC suggests we should).

However this logic can become even more complex as the server might decide to reject one of the requests and ask for reauthentication so we'd have to block concurrent requests again.

It seemed like quite a complex solution, though there are some benefits, like potentially saving a round trip for authentication but the RFC says the server may or may not take advantage of it. In the end I decided to take the pragmatic approach requests seems to take.

Of course if we decide otherwise I'm happy to implement the full fledged solution.

> Make it essentially be an LRU dict, storing N items. (1000, 10000, or whatever). Whenever it's more than N items, delete the first entry. (Easy now that python dicts are ordered)

Yep, that's what I was thinking as well 👍

yeraydiazdiaz:
@sethmlarson @florimondmanca any opinions on https://github.com/encode/httpx/pull/305#discussion_r321966898?

Personally I'm torn between having a more complete implementation and introducing much more complexity in this middleware.

For reference [request's implementation](https://github.com/psf/requests/blob/master/requests/auth.py#L108) takes a much more pragmatic approach that has been valuable for many years (the fact that `auth-int` has not been implemented in that time is probably a testament on how many servers actually require it).

sethmlarson:
@yeraydiazdiaz Hmm requests uses thread locals to hold onto state... I'm a lot more comfortable with how they're implementing it than with putting so much state into globals / class variables.  Could we do something with contextvars potentially?

florimondmanca:
@yeraydiazdiaz In retrospect, I agree with @tomchristie that the request middleware is not strictly necessary. I think we should be able to convert all utility methods into regular functions, and end up with the typical `__init__()`/`__call__()` pair of methods, with the logic in `__call__` delegating as much as possible to the helper functions.

I also think @sethmlarson's idea about context variables is interesting. I have the intuition we can use them to ditch the whole nonce counting idea, and only track whether the server has already rejected our credentials. It's hard to explain in words without implementing the whole thing, so I might draft a PR to illustrate?

Lastly, to help reduce the scope of this PR I'm going to submit one for refactoring the middleware into separate modules. This way we'll be able to solely focus on `middleware/digest_auth.py`, if that sounds alright?

yeraydiazdiaz:
Thanks all.

> Hmm requests uses thread locals to hold onto state... I'm a lot more comfortable with how they're implementing it than with putting so much state into globals / class variables. Could we do something with contextvars potentially?

This is something I suggested in a TODO comment on my first attempt on introducing Digest auth, [@tomchristie was not too keen on them and suggested using class variables](https://github.com/encode/httpx/pull/136#discussion_r306179205). Personally I agree with him but I'm not particularly familiar with contextvars and their benefits, maybe someone can break them down for me?

> I have the intuition we can use them to ditch num_401_responses, and only track whether the server has already rejected our credentials. It's hard to explain in words without implementing the whole thing, so I might draft a PR to illustrate?

Sure, actually I'd be happy to close this PR in favor of yours if that's preferable. Frankly this one has gotten unwieldy and there's just too much information for me to figure out what the correct approach would be. Maybe it's best if someone with a fresh pair of eyes approaches the problem.

I would suggest scoping out how much we want to stick to the RFC, there's quite a bit of SHOULDs and MAYs in there that requests does not implement on top of them not having to support concurrency.

florimondmanca:
> Personally I agree with him but I'm not particularly familiar with contextvars and their benefits, maybe someone can break them down for me?

Not a contextvars expert either, but the way I see them context variables are attached to a particular *async context*, i.e. a stack of coroutines that `await` one another. (Put differently, coroutines started concurrently via `loop.create_task()` or a trio `nursery` won't share the same async context.)

One typical use case I can think of is setting a context-local `REQUEST_ID` of some kind, e.g. for application monitoring. [Gist](https://gist.github.com/florimondmanca/6352a9cbcc670c7677a351e9bf8a987d#file-task_id_contextvars_demo-py)

That said, when I wrote…

> …and only track whether the server has already rejected our credentials

we don't *have* to use context variables to track that kind of information. A dirty trick would be to add `_credentials_rejected=False` to the middleware signature, and set the flag to `True` when we make the recursive call.

tomchristie:
> Lastly, to help reduce the scope of this PR I'm going to submit one for refactoring the middleware into separate modules. This way we'll be able to solely focus on middleware/digest_auth.py, if that sounds alright?

Fantastic, yes!

florimondmanca:
@yeraydiazdiaz 

> Sure, actually I'd be happy to close this PR in favor of yours if that's preferable.

I think we're nearly there, though! If this PR has too much old/now irrelevant content (which I'd definitely agree with), we can perfectly close it and re-open one to start from a clean slate. I can also open a PR to your fork with a refactoring proposal, if you'd like? :-)

tomchristie:
I'd suggest that the best tack onto this would be to start with a PR for Digest Auth that *simply doesn't include any nonce checking at all*.

We can then issue another PR on top of that that adds the nonce checking and nothing else. I think I have a clear handle onto that, but it will make it far easier to discuss and tackle if we're treating it in isolation. 👍

httpx/utils.py
```diff
@@ -192,3 +192,44 @@ def to_str(str_or_bytes: typing.Union[str, bytes]) -> str:
...
+        super().__setitem__(key, value)
+        if len(self) > self.max_size:
+            del self[list(self)[0]]
+
+
+class DefaultLRUDict(LRUDict):
+    def __init__(
+        self,
+        max_size: int,
+        default_factory: typing.Callable = None,
```
StephenBrown2: I've wondered this when typing my own functions, should this be `typing.Optional[typing.Callable] = None,`?

httpx/utils.py
```diff
@@ -192,3 +192,44 @@ def to_str(str_or_bytes: typing.Union[str, bytes]) -> str:
...
+        *args: typing.Iterable[typing.Tuple[typing.Any, typing.Any]],
+        **kwargs: typing.Any,
```
StephenBrown2: Should this be `typing.Dict[typing.Any, typing.Any]`, or `typing.Mapping[typing.Any, typing.Any]`?

florimondmanca: Actually, this annotation is correct, see [PEP 484](https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values). :)

StephenBrown2: Ahhh, there is deeper magic. 👍

httpx/utils.py
florimondmanca: The Optional declaration is inferred by mypy, I believe. Here’s a [nice blog post](https://marc.khouri.ca/posts/2018/mypy-optional-modifier-for-parameters-with-default-none.html) on this particular topic. 👍

yeraydiazdiaz:
Thanks all, I'll close this one and pick off bits on a new one [as Tom suggests](https://github.com/encode/httpx/pull/305#issuecomment-529524099) 🌟 🌟 🌟

PR #307, opened 2019-09-02
Added support ssl cert file environment

### LINKED ISSUE #306: Add support for SSL_CERT_FILE, REQUESTS_CA_BUNDLE, and CURL_CA_BUNDLE
Part of #33 these three environment variables all have the same behavior, to specify where a CA bundle is if none are specified by default and `trust_env` is True.

We should check the three environment variables in the order above, as SSL_CERT_FILE is a PEP and a standard whereas the other are products of other projects.

If multiple are set and that file doesn't exist we should skip it and try the next environment variable.

We'll also have to document these environment variables within `docs/environment.md`.

cansarigol:
Closes #306


httpx/utils.py
```diff
@@ -107,6 +107,14 @@ def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
     return netrc_info.authenticators(host)  # type: ignore
 
 
+def get_ssl_cert_file() -> typing.Optional[str]:
+    for env_name in ("REQUESTS_CA_BUNDLE", "SSL_CERT_FILE", "CURL_CA_BUNDLE"):
```
StephenBrown2: > We should check the three environment variables in the order above
> `(SSL_CERT_FILE, REQUESTS_CA_BUNDLE, CURL_CA_BUNDLE)`,
> as SSL_CERT_FILE is a PEP and a standard whereas the other are products of other projects.
- @sethmlarson 	https://github.com/encode/httpx/issues/306#issue-488260040

This also affects [the](https://github.com/encode/httpx/pull/307/files#diff-c73723c2d8a7c6dfecc80f229ebcef55R83) [docs](https://github.com/encode/httpx/pull/307/files#diff-c73723c2d8a7c6dfecc80f229ebcef55R88)

tests/test_utils.py
```diff
@@ -111,3 +116,19 @@ def test_parse_header_links(value, expected):
 
     # Reset the logger so we don't have verbose output in all unit tests
     logging.getLogger("httpx").handlers = []
+
+
+def test_get_ssl_cert_file():
```
StephenBrown2: This should also test if all three are set and if none are valid files, and verify the order tested.

docs/environment_variables.md
```diff
@@ -79,3 +79,24 @@ SERVER_TRAFFIC_SECRET_0 XXXX
...
+```python
+# test_script.py
+
+import httpx
+client = httpx.Client(verify=None, trust_env=True)
+client.get("https://google.com")
+```
+
+```console
+REQUESTS_CA_BUNDLE=test.pem python test_script.py
```
StephenBrown2: We should probably use `SSL_CERT_FILE` as the example, since it is [PEP 476](https://www.python.org/dev/peps/pep-0476/#trust-database).

sethmlarson:
@StephenBrown2 You took the review comments right out of my mouth! :tada: Thanks @cansarigol for tackling this issue. :)

sethmlarson:
Also this would benefit from #308 landing. :)

sethmlarson:
Btw #308 landed so I'd rebase your branch on master so that testing this is easier. :)

cansarigol:
@StephenBrown2  @sethmlarson thanks review.

docs/environment_variables.md
```diff
@@ -79,3 +79,24 @@ SERVER_TRAFFIC_SECRET_0 XXXX
...
+-----------
+
+Valid values: a filename
+
+If one of these environment variables is set and `trust_env=True` and `verify=True` or `verify is None`, `verify` will be set by respectively `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE` environments as file.
+
+Example:
+
+```python
+# test_script.py
```
tomchristie: We probably don't need this comment line.

docs/environment_variables.md
```diff
@@ -79,3 +79,24 @@ SERVER_TRAFFIC_SECRET_0 XXXX
 CLIENT_HANDSHAKE_TRAFFIC_SECRET XXXX
 CLIENT_TRAFFIC_SECRET_0 XXXX
 ```
+
+`SSL_CERT_FILE` `REQUESTS_CA_BUNDLE` `CURL_CA_BUNDLE`
+-----------
+
+Valid values: a filename
+
+If one of these environment variables is set and `trust_env=True` and `verify=True` or `verify is None`, `verify` will be set by respectively `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE` environments as file.
```
tomchristie: I reckon we ought to think about rephrasing this slightly. For one thing we should omit the `trust_env` note here.

Instead the opening paragraphs of the "Environment Variables" documentation should mention `trust_env`, noting that users can disable all variables here by setting `trust_env=False`

docs/environment_variables.md
```diff
@@ -79,3 +79,24 @@ SERVER_TRAFFIC_SECRET_0 XXXX
...
+
+Example:
+
+```python
+# test_script.py
+
+import httpx
+client = httpx.Client(verify=None, trust_env=True)
```
tomchristie: I think we should just document this as `client = httpx.Client()`, since we'll anyways default to supporting `SSL_CERT_FILE` *unless* verify/trust_env has been overridden.

httpx/config.py
```diff
@@ -61,6 +62,10 @@ def __init__(
             verify = True
             self._load_client_certs(ssl_context)
 
+        if trust_env:
+            if verify is True or verify is None:
+                verify = get_ssl_cert_file()  # type: ignore
```
tomchristie: This doesn't look quite right to me, but perhaps I've got myself in a muddle.

If verify is True, but `SSL_CERT_FILE` is not set, then `get_ssl_cert_file` will return `None`, which looks like a behavioral change from what we've got right now.

tomchristie: I also think we should *probably* inspect the environment at the point of calling `load_ssl_context`, rather than on `__init__`.

cansarigol:
Thanks @tomchristie , I applied your comments. I got the order from [this pr](https://github.com/psf/requests/pull/2903) Bwt it wasn't merged

sethmlarson:
Shouldn't we be supporting (but I agree not recommending) REQUESTS_CA_BUNDLE for the sake of requests compatibility? I'm fine with dropping curls.

httpx/utils.py
```diff
@@ -107,6 +107,14 @@ def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
     return netrc_info.authenticators(host)  # type: ignore
 
 
+def get_ssl_cert_file() -> typing.Optional[str]:
+    for env_name in ("SSL_CERT_FILE", "SSL_CERT_DIR"):
+        ssl_file = Path(os.getenv(env_name, ""))
+        if ssl_file and ssl_file.is_file():
```
StephenBrown2: This won't work for `SSL_CERT_DIR`, for hopefully obvious reasons.

cansarigol:
Hi, I dropped `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` envs support. is it ok?

httpx/utils.py
cansarigol: ahh sorry, it's path. I'm fixing

tomchristie:
> Shouldn't we be supporting (but I agree not recommending) REQUESTS_CA_BUNDLE for the sake of requests compatibility? I'm fine with dropping curls.

I don't know. I think we should possibly make that call seperately, once we're closing in on 1.0, when we're assessing and drawing up our requests compat. guide.

My rough feel at this point, is that this is probably one of the points where we ought to allow ourselves to diverge, but it depends on a more general question of just how closely we want to align.

In summary: I'd probably rather we make that specific call within the context of a more general review of our requests compat & divergences, rather than as part of this "add the functionality" PR.

Seem reasonable?

httpx/utils.py
```diff
@@ -107,6 +107,16 @@ def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
     return netrc_info.authenticators(host)  # type: ignore
 
 
+def get_ssl_cert_file() -> typing.Optional[str]:
```
sethmlarson: Let's change the title of this since "cert" is what we use for client cert. Really this should be `get_ca_bundle_from_env()` or something like it?

tests/test_config.py
```diff
@@ -26,6 +27,19 @@ def test_load_ssl_config_verify_existing_file():
     assert context.check_hostname is True
 
 
+@pytest.mark.parametrize("config", ("SSL_CERT_FILE", "SSL_CERT_DIR"))
+def test_load_ssl_config_verify_env_file(config):
+    default_path = httpx.config.DEFAULT_CA_BUNDLE_PATH
```
sethmlarson: We should use a non-default and verify that it's loaded. I think there's a way to see how many certificates are loaded in an SSLContext?

docs/environment_variables.md
```diff
@@ -79,3 +86,27 @@ SERVER_TRAFFIC_SECRET_0 XXXX
...
+
+import httpx
+client = httpx.Client()
+client.get("https://google.com")
+```
+
+```console
+SSL_CERT_FILE=tests/test.pem python test_script.py
+SSL_CERT_DIR=tests/ python test_script.py
+
```
sethmlarson: Remove whitespace

docs/environment_variables.md
```diff
@@ -2,6 +2,13 @@ Environment Variables
 =====================
 
 The HTTPX library can be configured via environment variables.
+To activate the environment variables, `trust_env` has to be set `true`.
+`trust_env` can set three ways:
+
+* for client trust_env's default value is true, httpx.Client()
```
sethmlarson: Let's capitalize the first letter of each entry in this list

docs/environment_variables.md
```diff
@@ -2,6 +2,13 @@ Environment Variables
 =====================
 
 The HTTPX library can be configured via environment variables.
+To activate the environment variables, `trust_env` has to be set `true`.
```
sethmlarson: We should use `True` to match Python's sentinel

sethmlarson: Environment variables are used by default, we should mention that it can be disabled via trust_env=False

cansarigol:
Hi, @sethmlarson thanks for your review.

tests/test_config.py
cansarigol: I couldn't understand this part. Could you give an example roughly?

sethmlarson:
@cansarigol by the way we squash all commits when we merge so you don't have to amend and force push, you can just add a commits. Easier to review so we can see the incremental changes you're making.

cansarigol:
Ah sorry, you're right. this is my habit. I will pay attention.

tests/test_config.py
sethmlarson: Basically if we don't set any `SSL_CERT_...` environment variable we choose `httpx.config.DEFAULT_CA_BUNDLE_PATH` so we can't test the behavior of `SSL_CERT_...` using that path as our expected value because there's no way to tell if the environment variable worked or we just picked the default anyways.

I suggest you use one of the auto-generated cert fixtures to test this functionality. :)

cansarigol: thanks. in addition, added ssl_config.verify validation to the end of the test to make sure.

httpx/config.py
```diff
@@ -90,6 +91,12 @@ def load_ssl_context(
     ) -> ssl.SSLContext:
         http_versions = HTTPVersionConfig() if http_versions is None else http_versions
 
+        if self.trust_env:
```
sethmlarson: Can we move this section into the `load_ssl_context_no_verify()`? Since we're only doing the loading if we've got verify active anyways.

Also is there any harm in having `SSLConfig.verify` default to `True` in the constructor? It defaults to `True` on the client anyways.

tests/test_config.py
```diff
@@ -26,6 +28,18 @@ def test_load_ssl_config_verify_existing_file():
...
+@pytest.mark.parametrize("config", ("SSL_CERT_FILE", "SSL_CERT_DIR"))
+def test_load_ssl_config_verify_env_file(cert_pem_file, config):
+    os.environ[config] = (
+        cert_pem_file if config.endswith("_FILE") else str(Path(cert_pem_file).parent)
+    )
+    ssl_config = httpx.SSLConfig(trust_env=True)
+    context = ssl_config.load_ssl_context()
+    assert context.verify_mode == ssl.VerifyMode.CERT_REQUIRED
+    assert context.check_hostname is True
+    assert ssl_config.verify == os.environ[config]
```
sethmlarson: Verify that certs are loaded via `assert len(context.get_ca_certs) > 0`

cansarigol: `get_ca_certs` returns empty if use `context.load_verify_locations(capath=str(ca_bundle_path))`. 
To get a result, we should create an `SSLSocket` and call [getpeercert](https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert) like below

```
if config == "SSL_CERT_DIR":
        HOST = "example.org"
        PORT = 443
        conn: ssl.SSLSocket = ssl_config.ssl_context.wrap_socket(
            socket.socket(socket.AF_INET, socket.SOCK_STREAM), server_hostname=HOST
        )
        conn.connect((HOST, PORT))
        conn.getpeercert()
```

httpx/config.py
cansarigol: Did you mean `load_ssl_context_verify`? when `verify` active, `load_ssl_context_verify` is called.

sethmlarson: Yep! Big typo there haha

tests/test_config.py
sethmlarson: Maybe we can use https_server.url.host and https_server.url.port to accomplish that. Then our certificate will verify as well :)

httpx/config.py
cansarigol: :) i'm moving

docs/environment_variables.md
tomchristie: With environment variables I guess we want to treat them as case insensitive.

I'd probably provide examples as `"true"`, in order to make it more visually distinct to the Python `True` keyword/singleton instance.

sethmlarson: But this is a Python instance of True?

sethmlarson:
After we get this merged I think we're ready to tag 0.7.3! @yeraydiazdiaz or @florimondmanca should do the release PR so we all get practice :)

tests/test_config.py
cansarigol: I couldn't verify with `cert_pem_file` for `https_server.url`. Please help :)

docs/environment_variables.md
```diff
@@ -79,3 +86,19 @@ SERVER_TRAFFIC_SECRET_0 XXXX
 CLIENT_HANDSHAKE_TRAFFIC_SECRET XXXX
 CLIENT_TRAFFIC_SECRET_0 XXXX
 ```
+
+`SSL_CERT_FILE` `SSL_CERT_DIR`
```
tomchristie: I guess let's have a comma here, so: "`SSL_CERT_FILE`, `SSL_CERT_DIR`"

tests/test_utils.py
tomchristie: I'd probably tend to be okay either way on this sort of thing.
We can't ever test every possible permutation, and test code does have it's own cost.

httpx/utils.py
```diff
@@ -107,6 +107,16 @@ def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
     return netrc_info.authenticators(host)  # type: ignore
 
 
+def get_ca_bundle_from_env() -> typing.Optional[str]:
+    ssl_path = Path(os.getenv("SSL_CERT_FILE", ""))
+    if ssl_path and ssl_path.is_file():
+        return str(ssl_path)
+    ssl_path = Path(os.getenv("SSL_CERT_DIR", ""))
+    if ssl_path and ssl_path.is_dir():
```
florimondmanca: I don't think this is doing what we want: `bool(ssl_path)` is always going to be `True` because it refers to a `Path` object.

As a result, we'll probably end up using the current directory as the `SSL_CERT_DIR` in all cases. 😕 

We should probably instead write something like…

```python
if "SSL_CERT_DIR" in os.environ:
    ssl_path = Path(os.environ["SSL_CERT_DIR"])
    if ssl_path.is_dir():
        return str(ssl_path)
```

and do the same kind of procesing for `SSL_CERT_FILE` above (although the bug wouldn't happen because the current directory is never a file).

florimondmanca: (Trivia: I think this is typically a case where the Python 3.8 walrus operator would shine 😄)

```python
if (
    ssl_cert_dir := os.getenv("SSL_CERT_DIR") is not None
    and (ssl_path := Path(ssl_cert_dir)).is_dir()
):
    return str(ssl_path)
```

tests/test_utils.py
```diff
@@ -111,3 +116,15 @@ def test_parse_header_links(value, expected):
...
```
florimondmanca: I'd say add tests at the top here for the cases when `SSL_CERT_FILE` and `SSL_CERT_DIR` are not set. Ideally we should be able to catch the potential bugs I mention in my other comment. :)

docs/environment_variables.md
```diff
@@ -2,6 +2,13 @@ Environment Variables
 =====================
 
 The HTTPX library can be configured via environment variables.
+Environment variables are used by default. To ignore environment variables, `trust_env` has to be set `False`.
+There are three ways to set `trust_env` to disable environment variables:
+
+* On the client via `httpx.Client(trust_env=False)`
+* Per request via `client.get("<url>", trust_env=False)`
+* On the `SSLConfig` via `httpx.SSLConfig(trust_env=False)`
```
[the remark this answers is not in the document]
cansarigol: I think you're right. In the past, we accepted SSLConfig as an argument like below.
```
async def request(
    ...
    ssl: SSLConfig = DEFAULT_SSL_CONFIG,
```
Now we accept `verify` and `cert` and override them.

tests/test_config.py
florimondmanca: @cansarigol Is there anything we can do to help unblock this?

cansarigol: Would be great. Thanks.  I'm stuck about create a sslsocket because of ssl validation error. I don't know if it is related trustme scope

florimondmanca: @cansarigol Can you share a traceback so we know what part is failing exactly? :) Thanks

cansarigol: I'm very sorry that I didn't do anything but i will asap

sethmlarson: This is blocked by #354, just gave it a spin myself. I think we should maybe merge this now?

sethmlarson: #354 is closed now, I tried changing the test to look like this right now:

```python
    os.environ[config] = (
        ca_cert_pem_file
        if config.endswith("_FILE")
        else str(Path(ca_cert_pem_file).parent)
    )
    ssl_config = httpx.SSLConfig(trust_env=True)
    context = ssl_config.load_ssl_context()
    assert context.verify_mode == ssl.VerifyMode.CERT_REQUIRED
    assert context.check_hostname is True
    assert ssl_config.verify == os.environ[config]

    host = https_server.url.host
    port = https_server.url.port
    conn = socket.create_connection((host, port))
    context.wrap_socket(conn, server_hostname=host)
    assert len(context.get_ca_certs()) == 1
```

But I'm still failing on the `SSL_CERT_DIR` test case and I don't know why, pretty annoying!

cansarigol: I added a test to be able to compare with below traceback.

```
tests/test_config.py ...FSSL error in data received
protocol: <asyncio.sslproto.SSLProtocol object at 0x103c044a8>
transport: <_SelectorSocketTransport closing fd=19 read=idle write=<idle, bufsize=0>>
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/sslproto.py", line 526, in data_received
    ssldata, appdata = self._sslpipe.feed_ssldata(data)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/sslproto.py", line 189, in feed_ssldata
    self._sslobj.do_handshake()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 763, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:1045)
```

sethmlarson: Spoke with @florimondmanca and I think we're just going to skip the `SSL_CERT_DIR` case here and if we run into issues with it we'll fix them then. Until then this PR is good to go. Thank you so much @cansarigol for following this one all the way through. :)

rehevkor5:
FYI, this was removed in https://github.com/encode/httpx/pull/3022

PR #308, opened 2019-09-02
Add fixture for ensuring os.environ is not persisted past a test

sethmlarson:
We've got lots of tests for environment variable features and it'd be good if we didn't have to worry about setting something in one test and it affecting a different test.


tomchristie:
Neato

PR #310, opened 2019-09-03
Drop Content-Length headers on GET redirects

### LINKED ISSUE #309: h11._util.LocalProtocolError: Too little data for declared Content-Length
client POC:

```python
import asyncio
import logging
import httpx

logging.basicConfig(level=logging.DEBUG)


async def test():
    client = httpx.AsyncClient()
    url = 'http://127.0.0.1:8000/debug'
    resp = await client.post(url, data='debug', allow_redirects=True)
    print(resp.content)


if __name__ == '__main__':
    asyncio.get_event_loop().run_until_complete(test())
```

app POC:

```python
from starlette.responses import RedirectResponse
from starlette.applications import Starlette
import uvicorn

app = Starlette(debug=True)

@app.route('/debug', methods=['POST', 'GET'])
async def debug(request):
    return RedirectResponse(url='https://httpbin.org/headers', status_code=302)

if __name__ == '__main__':
    uvicorn.run(app)
```

log:
```
DEBUG:asyncio:Using selector: KqueueSelector
DEBUG:httpx.dispatch.connection_pool:new_connection connection=HTTPConnection(origin=Origin(scheme='http' host='127.0.0.1' port=8000))
DEBUG:httpx.dispatch.connection:start_connect host='127.0.0.1' port=8000 timeout=TimeoutConfig(timeout=5.0)
DEBUG:httpx.dispatch.connection:connected http_version='HTTP/1.1'
DEBUG:httpx.dispatch.http11:send_headers method='POST' target='/debug' headers=Headers({'host': '127.0.0.1:8000', 'user-agent': 'python-httpx/0.7.2', 'accept': '*/*', 'content-length': '5', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive'})
DEBUG:httpx.dispatch.http11:receive_event event=NEED_DATA
DEBUG:httpx.dispatch.http11:send_data data=Data(<5 bytes>)
DEBUG:httpx.dispatch.http11:receive_event event=Response(status_code=302, headers=[(b'date', b'Tue, 03 Sep 2019 03:50:53 GMT'), (b'server', b'uvicorn'), (b'location', b'https://httpbin.org/headers'), (b'transfer-encoding', b'chunked')], http_version=b'1.1', reason=b'Found')
DEBUG:httpx.dispatch.http11:receive_event event=EndOfMessage(headers=[])
DEBUG:httpx.dispatch.http11:response_closed our_state=DONE their_state=DONE
DEBUG:httpx.dispatch.connection_pool:release_connection connection=HTTPConnection(origin=Origin(scheme='http' host='127.0.0.1' port=8000))
DEBUG:httpx.dispatch.connection_pool:new_connection connection=HTTPConnection(origin=Origin(scheme='https' host='httpbin.org' port=443))
DEBUG:httpx.dispatch.connection:start_connect host='httpbin.org' port=443 timeout=TimeoutConfig(timeout=5.0)
DEBUG:httpx.dispatch.connection:connected http_version='HTTP/1.1'
DEBUG:httpx.dispatch.http11:send_headers method='GET' target='/headers' headers=Headers({'host': 'httpbin.org', 'user-agent': 'python-httpx/0.7.2', 'accept': '*/*', 'content-length': '5', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive'})
DEBUG:httpx.dispatch.http11:receive_event event=NEED_DATA


  File "http3/httpx/dispatch/http11.py", line 53, in send
    http_version, status_code, headers = await self._receive_response(timeout)
  File "http3/httpx/dispatch/http11.py", line 133, in _receive_response
  File "http3/httpx/dispatch/http11.py", line 174, in _receive_event
    self.READ_NUM_BYTES, timeout, flag=self.timeout_flag
  File "http3/httpx/concurrency/asyncio.py", line 92, in read
    raise ReadTimeout() from None
http3.httpx.exceptions.ReadTimeout

  File "test_proto.py", line 16, in <module>
    asyncio.get_event_loop().run_until_complete(test())
  File "/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "test_proto.py", line 11, in test
    resp = await client.post(url, data='debug', allow_redirects=True)
  File "http3/httpx/client.py", line 415, in post
    trust_env=trust_env,
  File "http3/httpx/client.py", line 566, in request
    trust_env=trust_env,
  File "http3/httpx/client.py", line 237, in send
    return await get_response(request)
  File "http3/httpx/middleware.py", line 72, in __call__
    return await self(next_request, get_response)
  File "http3/httpx/middleware.py", line 62, in __call__
    response = await get_response(request)
  File "http3/httpx/client.py", line 202, in get_response
  File "http3/httpx/dispatch/connection_pool.py", line 126, in send
    raise exc
  File "http3/httpx/dispatch/connection_pool.py", line 121, in send
  File "http3/httpx/dispatch/connection.py", line 65, in send
  File "http3/httpx/dispatch/http11.py", line 53, in send
    http_version, status_code, headers = await self._receive_response(timeout)
  File "http3/httpx/concurrency/asyncio.py", line 285, in __aexit__
    await self.task
  File "http3/httpx/dispatch/http11.py", line 108, in _send_request_data
    await self._send_event(event, timeout)
  File "http3/httpx/dispatch/http11.py", line 123, in _send_event
    bytes_to_send = self.h11_state.send(event)
  File "site-packages/h11/_connection.py", line 469, in send
    data_list = self.send_with_data_passthrough(event)
  File "site-packages/h11/_connection.py", line 502, in send_with_data_passthrough
    writer(event, data_list.append)
  File "site-packages/h11/_writers.py", line 79, in __call__
    self.send_eom(event.headers, write)
  File "site-packages/h11/_writers.py", line 102, in send_eom
    raise LocalProtocolError("Too little data for declared Content-Length")
h11._util.LocalProtocolError: Too little data for declared Content-Length
```

tomchristie:
Closes #309.

Having a look at this highlighted a couple of things to me, tho...

* `del headers['DOES_NOT_EXIST']` doesn't raise a KeyError. I assume that really it should, and we should be using `.pop('WHATEVER', None)` to remove stuff that may or may not be there in the `redirect_headers` method.
* On away-from-origin redirects we're dropping the host header, but not including a new one. I think we ought to be doing `headers['Host'] = url.authority.encode("ascii")` instead?


florimondmanca:
There was an `example.py` script that seemed like a debug script, so I went ahead and pushed a removal commit. :+1:

Overall this looks great — I'll wait for the test case to submit a review.

> On away-from-origin redirects we're dropping the host header, but not including a new one.

Actually, we are auto-adding it when preparing the request:

```python
# models.py
if not has_host:
    auto_headers.append((b"host", self.url.authority.encode("ascii")))
```

But certainly setting it explicitly in the redirect middleware is the best way to go.

> del headers['DOES_NOT_EXIST'] doesn't raise a KeyError

I think that's a problem too — it kind of violates the expectations we'd have from a standard `dict`-like interface.

httpx/middleware.py
```diff
@@ -138,15 +138,22 @@ def redirect_url(self, request: AsyncRequest, response: AsyncResponse) -> URL:
...
         if url.origin != request.url.origin:
+            # Strip Authorization headers when responses are redirected away from
+            # the origin.
             del headers["Authorization"]
-            del headers["host"]
+            del headers["Host"]
+
+        if method != request.method and method == "GET":
+            # Strip Content-Length headers when we've switch to a 'GET' request.
+            del headers["Content-Length"]
```
sethmlarson: Should we also be stripping `Transfer-Encoding`? What other headers are only used when a request has a body?

tomchristie: Ah good call, yeah.

Those two are at any rate the only *automatic* ones that we need to concern ourselves with. https://github.com/encode/httpx/blob/fa74aaa6dbd9578b094d2dffb2c9f409e2bd8f34/httpx/models.py#L549

tests/client/test_redirects.py
```diff
@@ -243,12 +249,29 @@ class MockDispatch(AsyncDispatcher):
...
     response = await client.post(url, data=data)
     assert response.url == URL("https://example.org/redirect_body_target")
-    assert response.json() == {"body": "Example request body"}
+    assert response.json()["body"] == "Example request body"
+    assert "content-length" in response.json()["headers"]
+
+
+async def test_no_body_redirect(backend):
+    """
+    A 303 redirect should remove the request body.
```
florimondmanca: Should we test we have the same behavior for 302 as well?

tests/client/test_redirects.py
```diff
@@ -243,12 +249,29 @@ class MockDispatch(AsyncDispatcher):
...
+    """
+    client = AsyncClient(dispatch=MockDispatch(), backend=backend)
+    url = "https://example.org/redirect_no_body"
+    data = b"Example request body"
+    response = await client.post(url, data=data)
+    assert response.url == URL("https://example.org/redirect_body_target")
+    assert response.json()["body"] == ""
+    assert "content-length" not in response.json()["headers"]
```
florimondmanca: Looks like an assert for Transfer-Encoding is missing. :)

tests/client/test_redirects.py
tomchristie: No need to over-egg it. 😀
Test Cases have a maintenance overhead too, and it’s not functionally different right?

tests/client/test_redirects.py
tomchristie: True yes, tho it’s not present in this case anyways, since it’s not a streaming request.

tests/client/test_redirects.py
florimondmanca: I mean, it kind of is another functional path from my point of view, but I agree it might not be worth it considering the things we’d have to setup (another extremely similar handler on the test app, and a copy of this test with 302 instead). Let’s leave it. :)

tests/client/test_redirects.py
florimondmanca: Right! Not sure we should assert it, unless with set stream=True, but then this test is really about the request body being stripped out anyway. I’m good with the current state.

tomchristie:
Cool. I’ll merge this tomoz, tho if someone wants to jump in first, then they should add issues for the two aspects described above.

tests/client/test_redirects.py
sethmlarson: Is it not possible to parametrize the test case?

florimondmanca:
@tomchristie Has this been merged? I can’t see the commit in master and I don’t think this has been closed in favor of another PR.

tomchristie:
Thanks @florimondmanca! Banging on the wrong buttons. 😃

PR #312, opened 2019-09-04
Add test for origin from URL string

yeraydiazdiaz:
Bringing coverage back to 100%

This PR may or may not have been triggered by my OCD 😄

### DIFF
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -1,6 +1,6 @@
 import pytest
 
-from httpx import URL
+from httpx import URL, Origin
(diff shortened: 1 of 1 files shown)


tomchristie:
How do we want to go about turning on 100% enforcement?

florimondmanca:
@tomchristie I remember there were already bribes of discussion about this somewhere but I can’t remember. Should we open a dedicated issue?

tomchristie:
Same here yes, but I don't recall where exactly.
Let's have an issue for it yeah.

PR #314, opened 2019-09-04
Drop NPN support

### LINKED ISSUE #280: Drop NPN support
@pgjones suggested in https://github.com/encode/httpx/pull/276#discussion_r317585211 to stop using NPN to get the HTTP protocol in `.get_http_version()` on concurrency backends, and switch to ALPN-only.

(Technical notes of mine: [NPN](https://tools.ietf.org/html/draft-agl-tls-nextprotoneg-04) (Next Protocol Negociation) is a TLS extension that was added as part of [SPDY](https://en.wikipedia.org/wiki/SPDY) a Google protocol whose spec eventually got merged into the HTTP/2 standard. Google announced this in 2016 in [Hello HTTP/2, good bye SPDY](https://blog.chromium.org/2015/02/hello-http2-goodbye-spdy.html), and [ALPN now replaces NPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation#History).)

I did a bit of research, and NPN removal does seem to have been on its way across the interwebs:

- [Chrome 51, 2016](https://www.chromestatus.com/feature/5767920709795840)
- [Firefox 53, 2017](https://www.fxsitecompat.dev/en-CA/docs/2017/npn-support-has-been-removed/)
- [Cloudflare, 2018](https://blog.cloudflare.com/deprecating-spdy/)
- [Safari, 2019](https://webkit.org/blog/8569/removing-legacy-spdy-protocol-support/) (macOS Mojave and iOS 12)
- [Hypercorn, 2019](https://gitlab.com/pgjones/hypercorn/merge_requests/16) (😄)

Anyone more experienced than me on HTTP protocols negociation got any thoughts on this?

dobrovolsky:
Closes #280

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -68,8 +68,9 @@ class Stream(BaseStream):
             return "HTTP/1.1"
 
         ident = ssl_object.selected_alpn_protocol()
+
         if ident is None:
-            ident = ssl_object.selected_npn_protocol()
+            return "HTTP/1.1"
 
         return "HTTP/2" if ident == "h2" else "HTTP/1.1"
 
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -171,8 +171,6 @@ class SSLConfig:
 
         if ssl.HAS_ALPN:
             context.set_alpn_protocols(http_versions.alpn_identifiers)
-        if ssl.HAS_NPN:  # pragma: no cover
-            context.set_npn_protocols(http_versions.alpn_identifiers)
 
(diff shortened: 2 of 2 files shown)


tomchristie:
Welcome! Nice little improvement here, thank you. 🎉

We *could* drop the `if ident is None:` block here now, since the `return "HTTP/2" if ident == "h2" else "HTTP/1.1"` line would still return the right result there. However I don't think it's neccessarily a bad thing to handle the `None` case explicitly, so let's get this in as it currently stands.

PR #319, opened 2019-09-05
Add `build_request` for `Client`

### LINKED ISSUE #174: Create a Client.build_request() method

dobrovolsky:
Using a Client instance to make build requests
```python
>>> client = httpx.Client()
>>> r = client.build_request("GET", "https://example.com")
>>> r
<AsyncRequest('GET', 'https://example.com')>
>>> client.send(r)
<Response [200 OK]>
```
And async client
```python
>>> import httpx
>>> client = httpx.AsyncClient()
>>> r = client.build_request("GET", "https://example.com")
>>> r
<AsyncRequest('GET', 'https://example.com')>
>>> client.send(r)
<coroutine object AsyncClient.send at 0x7fbc588f7560>
```

Closes #174

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -63,8 +63,7 @@ Using a Client instance to make build requests
 ```python
 >>> client = httpx.Client()
 >>> r = client.build_request("GET", "https://example.com")
->>> r
-<AsyncRequest('GET', 'https://example.com')>
+>>> req.url.full_path = "*"
 >>> client.send(r)
 <Response [200 OK]>
 ```
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -68,16 +68,6 @@ Using a Client instance to make build requests
 >>> client.send(r)
 <Response [200 OK]>
 ```
-And async client
-```python
->>> import httpx
->>> client = httpx.AsyncClient()
->>> r = client.build_request("GET", "https://example.com")
->>> r
-<AsyncRequest('GET', 'https://example.com')>
->>> client.send(r)
-<coroutine object AsyncClient.send at 0x7fbc588f7560>
-```
 
 ## .netrc Support
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -59,11 +59,13 @@ client = httpx.Client(dispatch=dispatch)
 
 ## Build Request
 
-Using a Client instance to make build requests
+You can use `Client.build_request()` to build a request and
+make modifications before sending the request.
+
 ```python
 >>> client = httpx.Client()
->>> r = client.build_request("GET", "https://example.com")
->>> req.url.full_path = "*"
+>>> r = client.build_request("OPTIONS", "https://example.com")
+>>> req.url.full_path = "*"  # Build an 'OPTIONS *' request for CORS
 >>> client.send(r)
 <Response [200 OK]>
 ```
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -57,6 +57,28 @@ dispatch = httpx.WSGIDispatch(app=app, remote_addr="1.2.3.4")
 client = httpx.Client(dispatch=dispatch)
 ```
 
+## Build Request
+
+Using a Client instance to make build requests
+```python
+>>> client = httpx.Client()
+>>> r = client.build_request("GET", "https://example.com")
+>>> r
+<AsyncRequest('GET', 'https://example.com')>
+>>> client.send(r)
+<Response [200 OK]>
+```
+And async client
+```python
+>>> import httpx
+>>> client = httpx.AsyncClient()
+>>> r = client.build_request("GET", "https://example.com")
+>>> r
+<AsyncRequest('GET', 'https://example.com')>
+>>> client.send(r)
+<coroutine object AsyncClient.send at 0x7fbc588f7560>
+```
+
 ## .netrc Support
 
 HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -64,7 +64,7 @@ make modifications before sending the request.
 
 ```python
 >>> client = httpx.Client()
->>> r = client.build_request("OPTIONS", "https://example.com")
+>>> req = client.build_request("OPTIONS", "https://example.com")
 >>> req.url.full_path = "*"  # Build an 'OPTIONS *' request for CORS
 >>> client.send(r)
 <Response [200 OK]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -42,30 +42,6 @@
 * `def .build_request(method, url, [data], [files], [json], [params], [headers], [cookies])`
 * `def .close()`
 
-## `AsyncClient`
-
-*An HTTP client, with connection pooling, HTTP/2, redirects, cookie persistence, etc.*
-
-```python
->>> client = httpx.AsyncClient()
->>> response = client.get('https://example.org')
-```
-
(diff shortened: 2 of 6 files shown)


httpx/client.py
```diff
@@ -237,6 +237,33 @@ def _get_auth_middleware(
 
         return None
 
+    def build_request(
```
cansarigol: is `build_request` func's position correct in BaseClient ?

cansarigol: I asked to understand if BaseClient is an abstract class for all custom client classes. so far it is used to narrow `Client` and `AsyncClient` code down.

docs/advanced.md
```diff
@@ -57,6 +57,28 @@ dispatch = httpx.WSGIDispatch(app=app, remote_addr="1.2.3.4")
 client = httpx.Client(dispatch=dispatch)
 ```
 
+## Build Request
+
+Using a Client instance to make build requests
```
sethmlarson: We should probably show the use-case here such as being able to modify the request in ways you usually can't with our standard interface. An example of this is making an `OPTIONS *` request for CORS:

```python
client = httpx.Client()
req = client.build_request("OPTIONS", "https://example.com")
req.url.full_path = "*"
resp = client.send(req)
```

(I don't think you can *currently* do this but you'll be able to for sure after my proxy PR lands)

docs/advanced.md
```diff
@@ -57,6 +57,28 @@ dispatch = httpx.WSGIDispatch(app=app, remote_addr="1.2.3.4")
...
+```python
+>>> client = httpx.Client()
+>>> r = client.build_request("GET", "https://example.com")
+>>> r
+<AsyncRequest('GET', 'https://example.com')>
+>>> client.send(r)
+<Response [200 OK]>
+```
+And async client
```
sethmlarson: Let's not have a duplicate demonstration for the AsyncClient, client and async client have the same interface, only show where they differ.

httpx/client.py
```diff
@@ -153,7 +153,7 @@ def merge_headers(
             return merged_headers
         return headers
 
-    async def send(
+    async def _send(
```
sethmlarson: I'm not sure I understand the reasoning of this change, can we not keep the `send()` methods the same as they were previously?

tests/client/test_async_client.py
```diff
@@ -14,6 +14,18 @@
     assert repr(response) == "<Response [200 OK]>"
 
 
+async def test_build_request(server, backend):
+    url = server.url
+    async with httpx.AsyncClient(backend=backend) as client:
+        request = client.build_request("GET", url)
+        response = await client.send(request)
```
sethmlarson: Should probably modify the request in some way before sending it and then verify the request is modified in the result.

tests/client/test_client.py
```diff
@@ -19,6 +19,23 @@ def test_get(server):
     assert repr(response) == "<Response [200 OK]>"
 
 
+def test_build_request(server):
+    url = server.url
+    with httpx.Client() as http:
+        request = http.build_request("GET", url)
```
sethmlarson: Same as above review comment.

httpx/client.py
dobrovolsky: I think so
`Client` and `AsyncClient` build Request in same way

httpx/client.py
dobrovolsky: we need some internal method for sending request
also we need different method with different signature 
`send() -> Response` and `send() -> AsyncResponse`

tests/client/test_async_client.py
dobrovolsky: do you have any suggestion for testing?

docs/advanced.md
dobrovolsky: In your example we can't change `full_path`  because it's an attribute
Do you have any suggestions?

sethmlarson: You can't change it now but once another PR is merged you will be able to.
You shouldn't create a test case with it, but you can add it to documentation. :)

httpx/client.py
sethmlarson: You're right! I take back this comment, although this method might need a better name?

tests/client/test_async_client.py
sethmlarson: Maybe adding a header, just something that wouldn't be done by default.

florimondmanca:
I think it's a shame that `.send()` and `.dispatch` are already taken on the client, to be honest. In terms of public API, `response = client.send(request)` is what makes most sense even from a Requests compatibility point of view.

I'm thinking we could very well define a synchronous `def .send()` on `Client` that calls the asynchronous `super().send` coroutine, no?

Mypy would complain, but maybe that's what I would expect as a user: `Client` has a sync `.send()`, while `AsyncClient` has an `async .send()`.

See:

```python
    request = client.build_request("GET", "https://example.com")
    response = await client.send(request)

    request = client.build_request("GET", "https://example.com")
```

That way we don't have to throw away the existing `.send()` verb that's used throughout the client, dispatchers and connections.

(Also it's actually the interface we already advertise in the [Developer Interface](https://www.encode.io/httpx/api/#client), but currently `.send()` is always async. Which means the docs are currently wrong, actually.)

Thoughts?

tests/conftest.py
```diff
@@ -111,6 +113,25 @@ def backend(request):
     await send({"type": "http.response.body", "body": body})
 
 
+async def echo_headers(scope, receive, send):
+    body: bytes = b""
+
+    more_body = scope.get("headers", [])
+    for h in more_body:
```
florimondmanca: Can we rephrase this to…?

```python
    for name, value in scope.get("headers", []):
```

docs/api.md
```diff
@@ -38,6 +39,7 @@
 * `def .delete(url, [data], [json], [params], [headers], [cookies], [auth], [stream], [allow_redirects], [verify], [cert], [timeout])`
 * `def .request(method, url, [data], [params], [headers], [cookies], [auth], [stream], [allow_redirects], [verify], [cert], [timeout])`
 * `def .send(request, [stream], [allow_redirects], [verify], [cert], [timeout])`
+* `def .build_request(method, url, [data], [files], [json], [params], [headers], [cookies])`
```
florimondmanca: For consistency, let's move this to below `.request()` and above `.send()`?

tests/conftest.py
```diff
@@ -111,6 +113,23 @@ def backend(request):
     await send({"type": "http.response.body", "body": body})
 
 
+async def echo_headers(scope, receive, send):
+    body: bytes = b""
+
+    for name, value in scope.get("headers", []):
+        value = f"{name.capitalize().decode()}: {value.decode()}\n"
+        body += value.encode()
```
yeraydiazdiaz: Nit: Maybe it'd be more ergonomic to return the headers as a JSON object? I think it would make the assertions more readable.

dobrovolsky: Nice hint, thanks

tests/client/test_client.py
```diff
@@ -19,6 +19,24 @@ def test_get(server):
...
+
+    with httpx.Client() as http:
+        request = http.build_request("GET", url)
+        request.headers.update(headers)
+        response = http.send(request)
+
+    assert response.status_code == 200
+    assert response.url == url
+
+    response_data = response.json()
```
sethmlarson: Since we're only asserting one value we can simplify our asserts like this: `assert response.json()["custom-header"] == "value"`

sethmlarson:
Thanks @dobrovolsky! :clap:

PR #321, opened 2019-09-05
Setting HOST header to url.authority.encode('ascii') in redirect middleware

### LINKED ISSUE #318: Away-from-origin redirects should add a new host header
Prompted by #310

Away-from-origin redirects currently remove the `Host` header, without adding a new one.
I think we ought to be using `headers['Host'] = url.authority.encode("ascii")` instead of simply deleting the header.

gdhameeja:
NOTE:
__setitem__ in class Headers already encodes with its encoding attribute
Added check to skip encoding the value if it is already bytes.

Closes #318

### DIFF
--- a/httpx/middleware.py
+++ b/httpx/middleware.py
@@ -148,7 +148,7 @@ class RedirectMiddleware(BaseMiddleware):
             # Strip Authorization headers when responses are redirected away from
             # the origin.
             del headers["Authorization"]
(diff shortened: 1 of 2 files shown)


gdhameeja:
#318

sethmlarson:
FYI @gdhameeja I edited your original pull request to add `Closes #318` so that once we merge this PR the corresponding issue will be closed as well :)

Issue #35, opened 2019-05-02
Close all outstanding responses on client close


tomchristie:
Client close currently [closes off any keep-alive connections](https://github.com/encode/httpx/blob/master/httpx/dispatch/connection_pool.py#L171).
Actually the behavior would be more neatly bounded if we forcibly close off *all* connections.

```python
with SyncClient() as client:
    response = client.get(..., stream=True)
response.read()  # This should raise an error.
```

jhtimmins:
@florimondmanca is this still up for grabs? And if so, would it make a good first issue?

florimondmanca:
Hi @jhtimmins! There's already a PR for this over at https://github.com/encode/httpx/pull/322, but unfortunately it seems to have slipped through.

So feel free to take it up. I don't think the fix will be trivial, but you can take a crack at it and see where you end up. :)

jhtimmins:
Thanks @florimondmanca! Sounds great. I'll do that.

tomchristie:
Obsoleted by the more tightly bounded `stream()` API. (Which ensures that connections can only be held open within a `stream()` context block)

PR #322, opened 2019-09-06
Close all outstanding responses on client close

dobrovolsky:
related to #35 
```python
with Client() as client:
    response = client.get(..., stream=True)
response.read()  # This should raise an error.
```

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1,4 +1,5 @@
 import pytest
+from h11 import RemoteProtocolError
 
 import httpx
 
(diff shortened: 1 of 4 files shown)


sethmlarson:
Thanks for opening this @dobrovolsky! :)

tests/client/test_client.py
```diff
@@ -33,33 +34,44 @@ def test_post_json(server):
...
 
+def test_stream_manger_release_resources(server):
+    with httpx.Client() as http:
+        response = http.post(
+            server.url.copy_with(path="/echo_body"), stream=True, data=b"*" * 409600
+        )
+
+    assert response.status_code == 200
+    with pytest.raises(RemoteProtocolError):
+        response.read()
```
florimondmanca: I think we should not let an `h11` exception bubble up to our users; we need to raise an HTTPX-specific exception here. I believe `ResponseClosed` is the most adequate.

This made me realize that we are not force-closing responses yet here on client exit. You can see this by turning on `HTTPX_DEBUG` and seeing `receive_event NEED_DATA` log lines because the response object is still trying to receive response data, and that is because we did not call `.close()` on it to send a `ResponseClosed` event to h11.

I guess implementing response closing on client exit involves some form of storing "pending responses" on the client. We'd add the response there in the `.request()` methods, remove it when the response's `on_close()` callback is normally called, and iterate over it in client's `__exit__()` or `__aexit__()` to close any pending responses.

Would you like to look into this?

tomchristie:
Thanks for having put a bunch of time into this.
As it happens, it's now obsoleted by our more tightly bounded `stream()` API, which ensures that connections can only be held open within a `stream()` context block.

PR #324, opened 2019-09-07
Fix #317

### LINKED ISSUE #317: Deletion of non-existent keys from `Headers` should raise KeyError.
Prompted by  #310

To replicate:

```python
headers = httpx.Headers()
del headers['DOES_NOT_EXIST']  # Should raise a KeyError, but doesn't
```

To resolve this issue we'll need to:

* Ensure that deletion of non-existant keys raises a KeyError.
* Ensure that we have a `.pop(key, default)` that we can use instead for removal without KeyErrors.
* Replace instances of `del headers[key]` in our codebase with `headers.pop(key, None)` where appropriate.

gdhameeja:
Fixes issue #317 
KeyError is raised if a header that was to be deleted did not exist in Headers
`RedirectMiddleware` now can't directly delete a header without checking.
Hence added checks for stripping off request body headers(Content-Length, Transfer-Encoding)
and Authorization headers in case of redirects
Also added a test to check if KeyError is raised for a header to be deleted which does not exist

### DIFF
--- a/httpx/middleware.py
+++ b/httpx/middleware.py
@@ -147,16 +147,37 @@ class RedirectMiddleware(BaseMiddleware):
         if url.origin != request.url.origin:
             # Strip Authorization headers when responses are redirected away from
             # the origin.
-            del headers["Authorization"]
-            del headers["Host"]
+            self._strip_headers(headers)
 
         if method != request.method and method == "GET":
-            # If we've switch to a 'GET' request, then strip any headers which
-            # are only relevant to the request body.
-            del headers["Content-Length"]
-            del headers["Transfer-Encoding"]
+         self._delete_request_body_headers(headers)
 
         return headers
+
+    def _strip_headers(self, headers: Headers) -> None:
+        """
+        Strip headers in case of redirect
+        """
+        headers_to_strip = [
+            'Authorization',
+            'Host'
+        ]
+        for header in headers_to_strip:
+            if header in headers:
+                del headers[header]
+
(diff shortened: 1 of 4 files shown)


httpx/middleware.py
```diff
@@ -147,17 +147,33 @@ def redirect_headers(self, request: AsyncRequest, url: URL, method: str) -> Head
...
         if method != request.method and method == "GET":
-            # If we've switch to a 'GET' request, then strip any headers which
-            # are only relevant to the request body.
-            del headers["Content-Length"]
-            del headers["Transfer-Encoding"]
+            self._delete_request_body_headers(headers)
 
         return headers
 
+    def _strip_headers(self, headers: Headers) -> None:
```
sethmlarson: Let's revert the changes moving the `del` logic into their own methods until the logic is needed in multiple places :)

httpx/middleware.py
```diff
@@ -147,17 +147,33 @@ def redirect_headers(self, request: AsyncRequest, url: URL, method: str) -> Head
...
+        """
+        Strip headers in case of redirect
+        """
+        headers_to_strip = ["Authorization", "Host"]
+        for header in headers_to_strip:
```
sethmlarson: Let's revert these changes moving the logic into methods until it's needed in more than one location. :) Instead we can do `headers.pop("Authorization", None)`

gdhameeja: Done.

PR #325, opened 2019-09-07
Refactor middleware

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/305#issuecomment-529144881

Moves the existing middleware in `middleware.py` into their own module in a `middleware` package.

Next up, a basic auth refactor including:
- Add and use `to_bytes()`
- Extract a reusable `build_basic_auth_header()` helper.

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -169,3 +169,7 @@ def get_logger(name: str) -> logging.Logger:
             logger.addHandler(handler)
 
     return logging.getLogger(name)
+
+
+def to_bytes(value: typing.Union[str, bytes]) -> bytes:
+    return value.encode() if isinstance(value, str) else value
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -171,5 +171,5 @@ def get_logger(name: str) -> logging.Logger:
     return logging.getLogger(name)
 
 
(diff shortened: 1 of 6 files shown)


sethmlarson:
Could we do those extras in this PR too under the cover of "refactor middlewares"?

florimondmanca:
> Could we do those extras in this PR too under the cover of "refactor middlewares"?

Uh, I guess I'm OK with that. I have another idea for refactoring the `RedirectMiddleware` to use contextvars, but that will *definitely* be in another PR. ;-) I'll push the extras.

sethmlarson:
Totally agree. You might have some minor conflicts over the PR I just merged.

florimondmanca:
Already force-pushed a rebase. :+1:

httpx/utils.py
```diff
@@ -169,3 +169,7 @@ def get_logger(name: str) -> logging.Logger:
             logger.addHandler(handler)
 
     return logging.getLogger(name)
+
+
+def to_bytes(value: typing.Union[str, bytes]) -> bytes:
```
sethmlarson: We should take an encoding parameter here too, default to UTF-8.

florimondmanca: Fixed!

PR #326, opened 2019-09-07
Refactor redirect middleware using contextvars

florimondmanca:
While #325, and having https://github.com/encode/httpx/pull/305#issuecomment-529135254 in mind, I noticed that the `history` was a middleware instance variable on `RedirectMiddleware`, and that surely it could be refactored to use context variables instead. So, here we are! I think the result is quite pleasing.

The docs for contextvars, if need be: [contextvars](https://docs.python.org/3.7/library/contextvars.html).

@yeraydiazdiaz The second commit here illustrates what I suggested in https://github.com/encode/httpx/pull/305#issuecomment-529144881:

> I think we should be able to convert all utility methods into regular functions, and end up with the typical __init__()/__call__() pair of methods, with the logic in __call__ delegating as much as possible to the helper functions.

### DIFF
--- a/httpx/compat/contextvars.py
+++ b/httpx/compat/contextvars.py
@@ -1,0 +1,8 @@
+try:
+    # Python 3.6 backport w/ asyncio support.
+    # Try to import it first since it installs the contextvars backport too.
+    import aiocontextvars as _contextvars
+except ImportError:
+    import contextvars as _contextvars  # type: ignore
+
+ContextVar = _contextvars.ContextVar
--- a/httpx/middleware/redirect.py
+++ b/httpx/middleware/redirect.py
@@ -1,8 +1,8 @@
 import functools
 import typing
-from contextvars import ContextVar
 
 from ..config import DEFAULT_MAX_REDIRECTS
+from ..compat.contextvars import ContextVar
 from ..exceptions import RedirectBodyUnavailable, RedirectLoop, TooManyRedirects
 from ..models import URL, AsyncRequest, AsyncResponse, Cookies, Headers
 from ..status_codes import codes
--- a/httpx/middleware/redirect.py
+++ b/httpx/middleware/redirect.py
@@ -1,11 +1,14 @@
 import functools
 import typing
+from contextvars import ContextVar
 
 from ..config import DEFAULT_MAX_REDIRECTS
 from ..exceptions import RedirectBodyUnavailable, RedirectLoop, TooManyRedirects
 from ..models import URL, AsyncRequest, AsyncResponse, Cookies, Headers
 from ..status_codes import codes
 from .base import BaseMiddleware
+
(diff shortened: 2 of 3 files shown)


florimondmanca:
On 3.6, I used the [aiocontextvars](https://github.com/fantix/aiocontextvars) backport which includes asyncio support (while the `contextvars` backport [does not](https://github.com/MagicStack/contextvars/issues/2) (yet)).

We *may* be running into issues when extending this to add trio support, though. Trio has built-in support for context variables, and it installs the `contextvars` backport on < 3.7. This means that users that have trio installed *and* try to use the asyncio backend (e.g. with the sync `Client`) will run into issues because they'll end up using a backport that lacks asyncio support.

One solution I see is to build a custom `ContextVar` class that holds both versions, and relies on [sniffio](https://github.com/python-trio/sniffio) to detect the async environment and defer calls to the correct implementation.

Thoughts?

sethmlarson:
You're gonna have to install the contextvars backport on Python <3.7

florimondmanca:
A non-asyncio-specific refactor could look something like this:

```python
import sys
import typing
import sniffio

# Also available on Python 3.6+ via aiocontextvars.
import contextvars as _contextvars

try:
    import aiocontextvars as _aiocontextvars
except ImportError:  # pragma: no cover
    _aiocontextvars = None  # type: ignore

T = typing.TypeVar("T")


class ContextVar(typing.Generic[T]):
    """Stripped-down version of contextvars.ContextVar.

    Compatible with asyncio / Python 3.6.
    """

    def __init__(self, name: str) -> None:
        self._default_var: _contextvars.ContextVar[T] = _contextvars.ContextVar(name)
        self._asyncio_var = (
            _aiocontextvars.ContextVar(name) if sys.version_info < (3, 7) else None
        )

    def _get_variable(self) -> typing.Any:
        if (
            sniffio.current_async_library() == "asyncio"
            and self._asyncio_var is not None
        ):
            return self._asyncio_var
        return self._default_var

    def get(self) -> T:
        variable = self._get_variable()
        return variable.get()

    def set(self, value: T) -> None:
        variable = self._get_variable()
        variable.set(value)
```

Should we include it in here directly?

tomchristie:
Hold up there. 😃
We don't want to be using task local or thread local state in *any* of these cases. If we think we need task local or thread local state for something here, then we're probably doing it wrong.

What we *do* want is plain old local variables (passed around if needed), rather than state that's being persisted on the class.

florimondmanca:
@tomchristie I hear you. I let this PR up to let the idea decay, and in retrospect I can see how context variables are definitely not what we need to use in this particular situation.

I guess what we want is for RedirectMiddleware.__call__ to define the recursive function and return a first call to it, with history defined as a local variable enclosed within the recursive function, right?

Anyway I’ll close this and if the above sounds about right I can tackle that refactor instead.

PR #330, opened 2019-09-09
[Requests] Test entrypoints

florimondmanca:
Refs #293, prompted by https://github.com/encode/httpx/pull/298#issuecomment-529574573

So, here's a very small PR that adds *one* test from Requests.

---

Things I'm unsure about, that don't necessarily need to be sorted out right now:

**How should we keep track of changes made to the test cases?**

I introduced a `copied_from` pytest marker for this purpose. The test here links to a line in the test code from Requests v2.22 (we shouldn't link to master as line numbers will change over time).

Alternatively we could have special comments, e.g. `NOTE: modified xyz because of abc` but I figured it would be less structured than resorting to markers. Also we have a nice summary of the changes at the top of the test.

**How should we deal with API incompatibilities?**

I see two outcomes:

- HTTPX is functionally compatible, but the API is different.

In this case, we can monkeypatch `requests` (i.e. `httpx`) to expose the modified API, which allows us to keep track of migration steps *in code*.

- HTTPX is functionally incompatible, i.e. Requests allows to do something that HTTPX doesn't or it exposes a concept that HTTPX doesn't have an equivalent for (e.g. Requests' adapters).

In this case, the basic approach would be to `xfail` the test with a well-defined `reason`, and then we can see if there are patterns that would enable us to record the exact reasons in a more structured way.

I'd be happy to have anyone's thoughts on this, as well as how rigorous we want this whole compatibility test suite to be. I think the good approach for now is to ensure the main APIs are somewhat compatible, and track the migration steps that lie in that "somewhat".


setup.cfg
```diff
@@ -17,3 +17,5 @@ multi_line_output = 3
 
 [tool:pytest]
 addopts = --cov=httpx --cov=tests --cov-report=term-missing
```
StephenBrown2: Here might be a better place to put `-rsXx` than before, actually.

tests/requests_compat/test_api.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@pytest.mark.copied_from(
+    "https://github.com/psf/requests/blob/v2.22.0/tests/test_requests.py#L72",
+    changes=["added noqa comment to silence flake8"],
+)
+@pytest.mark.xfail(
+    reason="PoolManager has no obvious equivalent in HTTPX", raises=ImportError
+)
+def test_poolmanager_entrypoint():
```
sethmlarson: What are your thoughts on straight up removing tests that have zero chance of running with HTTPX (like this test)?

sethmlarson: Besides this question this looks good to me :+1:

florimondmanca: So this relates to the “rigor” I mentioned in the PR description.

I think including these makes sure we track *important* Requests APIs that we don’t have. (In a sense, adding them and xfailing them is a form of testing too, in contrast with not including or skipping them?)

I don’t think it will be worth adding every single test from Requests though, as a lot of them test regressions or API details that may not be relevant to us.

PR #332, opened 2019-09-10
DigestAuth as middleware - No nonce count

yeraydiazdiaz:
Split from https://github.com/encode/httpx/pull/305

This PR introduces a `DigestAuth` middleware that can be used as an `auth` parameter to `Client` and the request methods.

This works for the synchronous Client situations, for example:

```python
import httpx

URL = "https://httpbin.org/digest-auth/auth/yeray/{password}/{algorithm}"

ALGORITHMS = [
    "MD5",
    "MD5-SESS",
    "SHA",
    "SHA-SESS",
    "SHA-256",
    "SHA-256-SESS",
    "SHA-512",
    "SHA-512-SESS",
]


def main():
    password = "foo"
    with httpx.Client(auth=httpx.DigestAuth("yeray", password)) as client:
        for algorithm in ALGORITHMS:
            response = client.get(URL.format(password=password, algorithm=algorithm))
            print("FAILED!" if response.status_code != 200 else "", algorithm, response)

        print("\nThis should fail...")
        print(client.get(URL.format(password="bar", algorithm=ALGORITHMS[0])))


if __name__ == "__main__":
    main()
```

Prints:

```
 MD5 <Response [200 OK]>
 MD5-SESS <Response [200 OK]>
 SHA <Response [200 OK]>
 SHA-SESS <Response [200 OK]>
 SHA-256 <Response [200 OK]>
 SHA-256-SESS <Response [200 OK]>
 SHA-512 <Response [200 OK]>
 SHA-512-SESS <Response [200 OK]>

This should fail...
<Response [401 Unauthorized]>
```

But fails on concurrent AsyncClient such as:

```python
iimport asyncio

import httpx

URL = "https://httpbin.org/digest-auth/auth/yeray/{password}/{algorithm}"

ALGORITHMS = [
    "MD5",
    "MD5-SESS",
    "SHA",
    "SHA-SESS",
    "SHA-256",
    "SHA-256-SESS",
    "SHA-512",
    "SHA-512-SESS",
]


async def main():
    password = "foo"
    client = httpx.AsyncClient(auth=httpx.DigestAuth("yeray", password))
    fs = [
        client.get(URL.format(password=password, algorithm=algorithm))
        for algorithm in ALGORITHMS
    ]
    responses = await asyncio.gather(*fs)
    for algorithm, response in zip(ALGORITHMS, responses):
        print("FAILED!" if response.status_code != 200 else "", algorithm, response)

    print("\nThis should fail...")
    print(await client.get(URL.format(password="bar", algorithm=ALGORITHMS[0])))


if __name__ == "__main__":
    asyncio.run(main())
```

Printing inconsistent results with only some of the requests succeeding:

```
 MD5 <Response [200 OK]>
FAILED! MD5-SESS <Response [401 Unauthorized]>
FAILED! SHA <Response [401 Unauthorized]>
FAILED! SHA-SESS <Response [401 Unauthorized]>
FAILED! SHA-256 <Response [401 Unauthorized]>
 SHA-256-SESS <Response [200 OK]>
FAILED! SHA-512 <Response [401 Unauthorized]>
 SHA-512-SESS <Response [200 OK]>

This should fail...
<Response [401 Unauthorized]>
```

There's already a fair bit of code in the PR so I opted not to include the discussed solution in the previous PR (having a per request instance of the middleware) until I get feedback.

### DIFF
--- a/httpx/middleware/digest_auth.py
+++ b/httpx/middleware/digest_auth.py
@@ -28,7 +28,6 @@ class DigestAuth(BaseMiddleware):
     ) -> None:
         self.username = to_bytes(username)
         self.password = to_bytes(password)
-        self._num_401_responses = 0
 
     async def __call__(
         self, request: AsyncRequest, get_response: typing.Callable
@@ -38,10 +37,7 @@ class DigestAuth(BaseMiddleware):
             StatusCode.is_client_error(response.status_code)
             and "www-authenticate" in response.headers
         ):
-            self._num_401_responses = 0
             return response
-        else:
-            self._num_401_responses += 1
 
         header = response.headers["www-authenticate"]
         try:
@@ -49,11 +45,8 @@ class DigestAuth(BaseMiddleware):
         except ValueError:
             raise ProtocolError("Malformed Digest authentication header")
 
-        if self._num_401_responses > 1:
-            return response
-
         request.headers["Authorization"] = self._build_auth_header(request, challenge)
-        return await self(request, get_response)
+        return await get_response(request)
 
     def _build_auth_header(
         self, request: AsyncRequest, challenge: "DigestAuthChallenge"
--- a/httpx/middleware/digest_auth.py
+++ b/httpx/middleware/digest_auth.py
@@ -161,7 +161,7 @@ class DigestAuthChallenge:
 
         header_dict: typing.Dict[str, str] = {}
         for field in parse_http_list(fields):
-            key, value = field.strip().split("=")
+            key, value = field.strip().split("=", 1)
             header_dict[key] = unquote(value)
 
         try:
--- a/httpx/middleware/digest_auth.py
+++ b/httpx/middleware/digest_auth.py
@@ -1,0 +1,188 @@
+import hashlib
+import os
+import re
+import time
+import typing
+from urllib.request import parse_http_list
+
+from ..exceptions import ProtocolError
+from ..models import AsyncRequest, AsyncResponse, StatusCode
+from ..utils import to_bytes, to_str, unquote
+from .base import BaseMiddleware
+
+
+class DigestAuth(BaseMiddleware):
+    ALGORITHM_TO_HASH_FUNCTION: typing.Dict[str, typing.Callable] = {
+        "MD5": hashlib.md5,
+        "MD5-SESS": hashlib.md5,
+        "SHA": hashlib.sha1,
+        "SHA-SESS": hashlib.sha1,
+        "SHA-256": hashlib.sha256,
+        "SHA-256-SESS": hashlib.sha256,
+        "SHA-512": hashlib.sha512,
+        "SHA-512-SESS": hashlib.sha512,
+    }
+
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ) -> None:
+        self.username = to_bytes(username)
+        self.password = to_bytes(password)
+        self._num_401_responses = 0
+
+    async def __call__(
+        self, request: AsyncRequest, get_response: typing.Callable
+    ) -> AsyncResponse:
+        response = await get_response(request)
+        if not (
+            StatusCode.is_client_error(response.status_code)
+            and "www-authenticate" in response.headers
+        ):
+            self._num_401_responses = 0
+            return response
+        else:
+            self._num_401_responses += 1
+
+        header = response.headers["www-authenticate"]
+        try:
+            challenge = DigestAuthChallenge.from_header(header)
+        except ValueError:
+            raise ProtocolError("Malformed Digest authentication header")
+
+        if self._num_401_responses > 1:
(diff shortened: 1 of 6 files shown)


httpx/middleware/digest_auth.py
```diff
@@ -0,0 +1,188 @@
...
+        These take the form of:
+        `Digest realm="realm@host.com",qop="auth,auth-int",nonce="abc",opaque="xyz"`
+        """
+        scheme, _, fields = header.partition(" ")
+        if scheme.lower() != "digest":
+            raise ValueError("Header does not start with 'Digest'")
+
+        header_dict: typing.Dict[str, str] = {}
+        for field in parse_http_list(fields):
+            key, value = field.strip().split("=")
```
sethmlarson: Should do .split(=, 1) so we can't get unpacking errors.

tests/client/test_auth.py
```diff
@@ -127,3 +203,136 @@ def test_auth_invalid_type():
...
+
+def test_digest_auth_returns_no_auth_if_no_digest_header_in_response():
+    url = "https://example.org/"
+    auth = DigestAuth(username="tomchristie", password="password123")
+
+    with Client(dispatch=MockDispatch()) as client:
+        response = client.get(url, auth=auth)
+
+    assert response.status_code == 200
+    assert response.json() == {"auth": None}
```
sethmlarson: Can we add a test case where only 1 of the 2 conditions are met like 200+WWW-Authenticate or 401 without the header?

tests/client/test_auth.py
```diff
@@ -127,3 +203,136 @@ def test_auth_invalid_type():
...
+def test_digest_auth(algorithm, expected_hash_length, expected_response_length):
+    url = "https://example.org/"
+    auth = DigestAuth(username="tomchristie", password="password123")
+
+    with Client(dispatch=MockDigestAuthDispatch(algorithm=algorithm)) as client:
+        response = client.get(url, auth=auth)
+
+    assert response.status_code == 200
+    auth = response.json()["auth"]
+    assert auth.startswith("Digest")
```
sethmlarson: startswith("Digest ")

tests/client/test_auth.py
```diff
@@ -127,3 +203,136 @@ def test_auth_invalid_type():
...
+def test_digest_auth_no_specified_qop():
+    url = "https://example.org/"
+    auth = DigestAuth(username="tomchristie", password="password123")
+
+    with Client(dispatch=MockDigestAuthDispatch(qop=None)) as client:
+        response = client.get(url, auth=auth)
+
+    assert response.status_code == 200
+    auth = response.json()["auth"]
+    assert auth.startswith("Digest")
```
sethmlarson: Same as above

tests/client/test_auth.py
```diff
@@ -127,3 +203,136 @@ def test_auth_invalid_type():
...
+    assert digest_data["username"] == '"tomchristie"'
+    assert digest_data["realm"] == '"httpx@example.org"'
+    assert len(digest_data["nonce"]) == 64 + 2  # extra quotes
+    assert digest_data["uri"] == '"/"'
+    assert len(digest_data["response"]) == 64 + 2
+    assert len(digest_data["opaque"]) == 64 + 2
+    assert digest_data["algorithm"] == "SHA-256"
+
+
+def test_digest_auth_qop_including_spaces_and_auth_returns_auth():
```
sethmlarson: Could we parametrize this with: (current input), "auth-int,auth", and "unknown, auth"?

httpx/middleware/digest_auth.py
```diff
@@ -0,0 +1,188 @@
...
+        "SHA-512": hashlib.sha512,
+        "SHA-512-SESS": hashlib.sha512,
+    }
+
+    def __init__(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ) -> None:
+        self.username = to_bytes(username)
+        self.password = to_bytes(password)
+        self._num_401_responses = 0
```
tomchristie: Okay, so we can't have *mutable* state on our middleware classes.

I assume the usage of `self._num_401_responses` here is why the async case is currently failing (and multi-threaded likely would do too.)

I would probably start by removing this property entirely, and removing the associated branching in `__call__` (If the very first response we get back is OK, then return it, otherwise build a request with the auth header, and return whatever response we get back from that.)

Once that's in line, I'll give this another review.

Great stuff. 👍

sethmlarson: Inevitably people will want to store mutable state on middleware, should I open a design discussion issue for us to nail those details and document them for others as well?

StephenBrown2: I'd say go for it, even if the stance is "No mutable state, just act on it and pass it through"

sethmlarson:
Thanks @yeraydiazdiaz :)

Issue #275, opened 2019-08-25
Adding HTTP/3 support using aioquic


jlaine:
It would be nice to start planning support for HTTP/3. aioquic provides a sans-I/O API for HTTP/3 similar to h2 which would make such an integration possible.

The main hurdle is that the connection model is very different to HTTP/1.1 and HTTP/2 : we cannot simply establish a (reader, writer) pair over a TCP connection and pass this to the protocol layer.

In order to provide a first demonstration, @tomchristie suggested I write a dispatcher which makes use of aioquic's HTTP/3 support.  The code is still a bit rough, but it works:

https://github.com/aiortc/aioquic/blob/master/examples/httpx_client.py

This can be run as:

`python examples/httpx_client.py https://cloudflare-quic.com/`

florimondmanca:
I don't know enough about HTTP/3 to tell (for example, I'm not sure what the differences between QUIC and HTTP/3 and the associated aioquic APIs are?), but the approach highlighted by @sethmlarson sounds sensible to me. One thing I'm sure of is that we'll need to cut this work into multiple PRs to help with review and incremental improvements.

> We'll need to add UDP support to our backend interface, that's definitely step 1.

Yup, and UDP support is definitely going to be a big chunk.

HTTPX is currently very much coupled to TCP. In particular, right now a given `HTTPConnection` doesn't know about other connections, but from what you've said it seems like a QUIC "connection" needs to know about the surrounding connections to do packet routing (though it might be a defining property of UDP?), correct? Anyway, there'll be some refactoring needed before supporting UDP to clarify what is TCP-specific and what is not.

> After that we need to see how we want to implement client storage so we can remember things like Alt-Svc headers between requests.

I *think* this should be an improvement over a basic design, right? (Under the hypothesis that remembering Alt-Svc headers will only allow us to skip the protocol switching phase of establishing an HTTP/3 connection.)

So, in terms of planning, I could see something like:

1. Rename `BaseStream` / `Stream` / `ConcurrencyBackend.connect()` to `BaseTCPStream` / `TCPStream` / `ConcurrencyBackend.connect_tcp()`.
1. Add UDP support. This will most likely include building an UDP version of `TCPStream` and the associated methods on `ConcurrencyBackend`. At this point, the resulting updated `ConcurrencyBackend` (with an asyncio implementation) can be unit-tested against a very basic asyncio UDP server.
[...]

PR #339, opened 2019-09-13
Rename Stream to TCPStream

florimondmanca:
Prompted by https://github.com/encode/httpx/issues/275#issuecomment-531349234, this is a probably valuable refactoring anyway and a good first step towards UDP support.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -1,13 +1,3 @@
-"""
-The `Stream` class here provides a lightweight layer over
-`asyncio.StreamReader` and `asyncio.StreamWriter`.
-
-Similarly `PoolSemaphore` is a lightweight layer over `BoundedSemaphore`.
-
-These classes help encapsulate the timeout logic, make it easier to unit-test
-protocols, and help keep the rest of the package more `async`/`await`
-based, and less strictly `asyncio`-specific.
-"""
 import asyncio
 import functools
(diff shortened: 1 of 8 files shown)


httpx/concurrency/asyncio.py
```diff
@@ -176,13 +166,13 @@ def loop(self) -> asyncio.AbstractEventLoop:
                 self._loop = asyncio.new_event_loop()
         return self._loop
 
-    async def connect(
+    async def open_tcp_stream(
```
sethmlarson: I don't think the symmetry of this method name with the backend serves us well. Can we revert to `connect()`?

florimondmanca: I guess you’re referring to the symmetry with TCPStream?

I’m considering what will happen when we want to connect via UDP here. Instead of having a broad connect() method and a more precise connect_udp() or similar, I thought we could name this method after what it *actually* does: open a Stream that happens to represent a TCP connection. And then we can have open_udp_stream(), and all things will be crystal clear.

Would you reconsider it under this angle?

sethmlarson: For some reason I thought this change was on the stream interface. Disregard my comment. :)

florimondmanca:
Thanks! Looks like this could have used a rebase — the master build is broken. 😕 I’ll send a fix tomorrow morning.

PR #340, opened 2019-09-13
AnyIO-based UDP PoC

florimondmanca:
Prompted by https://github.com/encode/httpx/issues/275#issuecomment-531369129

This demonstrates how we could build an `UDPStream` interface to support the needs of HTTP/3 support via aioquic.

@jlaine This PoC only provides single-target read/write operations. Will we actually need to send data to arbitrary hosts to implement QUIC+HTTP/3?

@sethmlarson While there would be a clear benefit to using AnyIO in terms of slimming down the burden of maintaining I/O operations for various libraries, I think this PR shows we'd still need our stream and concurrency backend interfaces, if only for the timeout/retry policies.

Example server:

```python
"""Hello world UDP server."""

import anyio


async def main() -> None:
    async with await anyio.create_udp_socket(port=5678) as socket:
        print("Connected")
        async for packet, (host, port) in socket.receive_packets(1024):
            response = b", ".join((b"Hello", packet))
            print(f"Response(data={response!r}, host={host!r}, port={port!r})")
            await socket.send(response, host, port)


anyio.run(main)
```

Example client:

```python
"""Hello world UDP client."""
import anyio

from httpx.concurrency.anyio import AnyioBackend
from httpx.config import DEFAULT_TIMEOUT_CONFIG


async def main() -> None:
    backend = AnyioBackend()
    stream = await backend.open_udp_stream(
        "127.0.0.1", 5678, timeout=DEFAULT_TIMEOUT_CONFIG
    )
    await stream.write(b"Bob")
    datagram = await stream.read(1024)
    assert datagram.data == b"Hello, Bob", datagram.data
    assert datagram.address == "127.0.0.1", datagram.address
    assert datagram.port == 5678, datagram.port

    await stream.close()


anyio.run(main)
```

### DIFF
--- a/httpx/concurrency/anyio.py
+++ b/httpx/concurrency/anyio.py
@@ -1,0 +1,73 @@
+import typing
+
+import anyio
+
+from ..config import TimeoutConfig
+from ..exceptions import ConnectTimeout, ReadTimeout
+
+
+class Datagram:
+    def __init__(self, data: bytes, address: str, port: int):
+        self.data = data
+        self.address = address
+        self.port = port
+
+
+class UDPStream:
+    def __init__(self, socket: anyio.UDPSocket, timeout: TimeoutConfig):
+        self.socket = socket
+        self.timeout = timeout
+
+    async def read(
+        self, n: int, timeout: TimeoutConfig = None, flag: typing.Any = None
+    ) -> Datagram:
+        if timeout is None:
+            timeout = self.timeout
+
+        read_timeout = (
+            timeout.read_timeout if timeout.read_timeout is not None else float("inf")
+        )
+
+        while True:
+            # Check our flag at the first possible moment, and use a fine
+            # grained retry loop if we're not yet in read-timeout mode.
+            should_raise = flag is None or flag.raise_on_read_timeout
+
+            if should_raise:
+                read_timeout = (
+                    timeout.read_timeout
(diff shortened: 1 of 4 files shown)


sethmlarson:
Yep, definitely will need our stream and backend interfaces. And to answer for @jlaine based on my understanding of QUIC we'd need to be able to support sending to multiple different hosts. The return type for `datagrams_to_send()` is `typing.List[typing.Tuple[NetworkAddress, bytes]]`

florimondmanca:
Yes, sounds good. I'm actually going to close this, not expecting us to merge this anyway, but I think it was a useful experiment.

jlaine:
> Yep, definitely will need our stream and backend interfaces. And to answer for @jlaine based on my understanding of QUIC we'd need to be able to support sending to multiple different hosts. The return type for `datagrams_to_send()` is `typing.List[typing.Tuple[NetworkAddress, bytes]]`

You are correct concerning the return type of `datagrams_to_send`, but bear in mind this covers both the client and server case, which are not identical. A client may change IP address any number of times as it switches between networks. In the current QUIC/HTTP3 drafts, the only case in which a server is expected to change address is if it announces a `PreferredAddress` (in the transport parameters which are conveyed during the TLS handshake) and the client switches to it.

PR #341, opened 2019-09-13
Move timeout logic into BaseStream

florimondmanca:
In both https://github.com/encode/httpx/pull/276/files and https://github.com/encode/httpx/pull/340, I ended up re-implementing the timeout logic twice, with a lot of copy-pasting. Since timeout policy is quite sensitive, I figured we should define that once and for all in the base implementation, and only define how to read/write *and maybe timeout* in concurrency backends.

There might be an extra refactoring opportunity in that #276 and #340 use the timeout logic in different places (`Stream` vs `UDPStream`), and what we *really* need there is to *wait for an async call to terminate or timeout*. So perhaps that should be an extra primitive on `ConcurrencyBackend`?

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -13,7 +13,6 @@ from .base import (
     BaseQueue,
     BaseTCPStream,
     ConcurrencyBackend,
-    TimeoutFlag,
 )
 
 SSL_MONKEY_PATCH_APPLIED = False
@@ -47,9 +46,9 @@ class TCPStream(BaseTCPStream):
         stream_writer: asyncio.StreamWriter,
         timeout: TimeoutConfig,
     ):
+        super().__init__(timeout)
         self.stream_reader = stream_reader
         self.stream_writer = stream_writer
-        self.timeout = timeout
 
     def get_http_version(self) -> str:
         ssl_object = self.stream_writer.get_extra_info("ssl_object")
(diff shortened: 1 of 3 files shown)


tomchristie:
I think this makes sense to me yeah.

I reckon we should have the docstrings for the various methods make it more clear which bits are expected to be implemented, and which bits are expected to be actually called into by the dispatch implementations.

florimondmanca:
> I reckon we should have the docstrings for the various methods make it more clear which bits are expected to be implemented, and which bits are expected to be actually called into by the dispatch implementations.

Sure, that’d make it clearer when maintaining concurrency backends. I can add docstrings for the proposed abstract methods added here (read_or_timeout(), write_or_timeout()), but I guess other parts of the base backend should be documented separately.

florimondmanca:
@tomchristie This should be ready for re-review. :+1:

httpx/concurrency/base.py
```diff
@@ -44,20 +45,98 @@ class BaseTCPStream:
...
+                break
+            except ReadTimeout:
+                if should_raise:
+                    raise
+
+        return data
 
     def write_no_block(self, data: bytes) -> None:
-        raise NotImplementedError()  # pragma: no cover
+        self.write_buffer += data
```
florimondmanca: I find this a bit misleading, because the `write_no_block` operation was influenced by asyncio's `.write()` / `.drain()` API, but what we're doing here doesn't *actually* write any data, and it won't until we make the next `.write()` call.

Should we repurpose this into `.queue_for_next_write()` or similar?

httpx/concurrency/base.py
```diff
@@ -44,20 +45,98 @@ class BaseTCPStream:
     backends, or for stand-alone test cases.
     """
 
+    def __init__(self, timeout: TimeoutConfig):
+        self.timeout = timeout
+        self.write_buffer = b""
```
JayH5: Slightly confusing thing here is that now both the `BaseTCPStream` and the Trio `TCPStream` have a `write_buffer`. Is the logic just duplicated in the Trio `TCPStream`?

florimondmanca: Ah, very good point! I seem to have forgotten to clean up the implementation from the trio backend. Will do. Thanks :)

httpx/concurrency/asyncio.py
```diff
@@ -60,58 +59,27 @@ def get_http_version(self) -> str:
...
 
-        if timeout is None:
-            timeout = self.timeout
+            raise ReadTimeout() from None
 
+    async def write_or_timeout(
+        self, data: bytes, timeout: typing.Optional[float]
+    ) -> None:
         self.stream_writer.write(data)
-        while True:
```
JayH5: This is subtly different now and I'm not sure I understand how looping on the write is safe. Previously the `while True` retry loop encapsulated just the `drain()` call, but with this change both the `write` and `drain` are inside the loop.

The asyncio docs say of the `write()` method:
> The method attempts to write the data to the underlying socket immediately. If that fails, the data is queued in an internal write buffer until it can be sent.

If we retry on write how can we be sure we aren't writing duplicated data to that internal write buffer?

httpx/concurrency/trio.py
```diff
@@ -41,23 +39,19 @@ def get_http_version(self) -> str:
...
-                return await self.stream.receive_some(max_bytes=n)
-
-            if should_raise:
-                raise ReadTimeout() from None
+    async def write_or_timeout(
+        self, data: bytes, timeout: typing.Optional[float]
+    ) -> None:
+        with trio.move_on_after(_or_inf(timeout)):
+            async with self.write_lock:
+                await self.stream.send_all(data)
```
JayH5: Similarly to the asyncio write method, I don't understand how looping on calls to `send_all` is safe?

From the Trio docs for `send_all()`:
> Most low-level operations in Trio provide a guarantee: if they raise trio.Cancelled, this means that they had no effect, so the system remains in a known state. This is not true for send_all(). If this operation raises trio.Cancelled (or any other exception for that matter), then it may have sent some, all, or none of the requested data, and there is no way to know which.

How can we loop on this while guaranteeing that we don't send duplicate data?

florimondmanca: So, while asyncio has a write/drain pair that *could* allow us to safely retry waiting for the data to be sent, it seems trio has no such semantics, right? Should we just *never* retry on write?

JayH5: >  it seems trio has no such semantics, right?

_Technically_ what I think you _could_ do (although it might be horrible/complicated) is to shield the future returned by the write operation so that the timeout doesn't cancel it and then either cancel the future if you're in write timeout mode or await it if not. I think that would essentially mean _not_ doing this PR and keeping the retry logic in the specific backends.

> Should we just never retry on write?

I don't understand well enough the original intention of this read/write timeout logic to say.

florimondmanca: As it stands, even in the current state the trio backend is affected by this potential catch because we're looping on `send_all()`.

I'm going to cc @tomchristie because I know he's been wanting to review the timeout logic for a while.

florimondmanca:
Because of the very valid caveats raised by @JayH5 here, I'm thinking of abandoning this PR altogether, because the final benefits if we try to get it right don't seem to be worth it anymore. Besides, the current state of things isn't *that* bad since we only have two places where the timeout configuration is defined.

I feel like we should still review the timeout flag switch logic taking into account the write retry edge cases, and *then* maybe come back and refactor.

So, closing. :)

PR #347, opened 2019-09-16
Fix remaining references to BaseStream in tests

florimondmanca:
Refs https://github.com/encode/httpx/pull/339#issuecomment-531601806


florimondmanca:
The fix is really minor and potentially affecting other branches so I'll merge once the build finishes. :)

Issue #350, opened 2019-09-16
Test client usage question: how to trigger lifespan startup and shutdown ?


euri10:
I have an issue switching from the `TestClient` from [async_asgi_testclient](https://github.com/vinissimus/async-asgi-testclient) to the `httpx.AsyncClient` :  the startup and shutdown events of my app aren't triggered by the latter.

It's probably a bad usage on my end, but if anyone could pinpoint what I'm doing wrong it would be of tremendous help, 

I wrote a small set of 2 tests below to better describe the situation, the one with `httpx.AsyncClient` fails while the other passes fine:

```
import logging

import pytest
from async_asgi_testclient import TestClient
from httpx import AsyncClient
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import JSONResponse

logger = logging.getLogger(__name__)
logger.setLevel(logging.DEBUG)

app = Starlette()


@app.on_event("startup")
async def startup():
    logger.debug("starting the app")


@app.on_event("shutdown")
async def shutdown():
    logger.debug("shutdown the app")


@app.route("/")
def homepage(request: Request):
    logger.debug("HOME")
    return JSONResponse({"hello": "world"})


@pytest.fixture
async def asyncasgitestclient():
    yield TestClient(app)


@pytest.fixture
async def httpxclient():
    yield AsyncClient(app=app, base_url="https://domain.tld")


@pytest.mark.asyncio
async def test_homepage_with_httpxclient(httpxclient, caplog):
    async with httpxclient:
        url = app.url_path_for("homepage")
        resp = await httpxclient.get(url)
        assert resp.status_code == 200
        assert resp.json() == {"hello": "world"}
    logger.debug(caplog.messages)
    assert all(
        x in caplog.messages for x in ["starting the app", "HOME", "shutdown the app"]
    )


@pytest.mark.asyncio
async def test_homepage_with_async_asgi_testclient(asyncasgitestclient, caplog):
    async with asyncasgitestclient:
        url = app.url_path_for("homepage")
        resp = await asyncasgitestclient.get(url)
        assert resp.status_code == 200
        assert resp.json() == {"hello": "world"}
    logger.debug(caplog.messages)
    assert all(
        x in caplog.messages for x in ["starting the app", "HOME", "shutdown the app"]
    )
```

florimondmanca:
[...]

PR #352, opened 2019-09-17
Add support for lifespan in ASGIDispatch

florimondmanca:
Refs #350 

Don't know yet whether this should be considered or not, but I thought I'd give it a shot. :)

Random thoughts: handling errors correctly for this turned out to be *way* more complicated than I initially expected. The `background_manager` interface hasn't helped, but I think it can't easily be improved because we use class-based async context managers a lot, instead of e.g. using `@asynccontextmanager`, which would have the benefit of handling exceptions the right way all the time.

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -641,6 +641,7 @@ class AsyncClient(BaseClient):
         await self.dispatch.close()
 
     async def __aenter__(self) -> "AsyncClient":
+        await self.dispatch.__aenter__()
         return self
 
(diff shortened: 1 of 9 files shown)


tomchristie:
As per #350 I'm kinda keen to punt on this. Whaddya reckon?

florimondmanca:
> As per #350 I'm kinda keen to punt on this. Whaddya reckon?

Yep, sounds good.

PR #353, opened 2019-09-18
Proxy Configurations

### LINKED ISSUE #33: Environment support

sethmlarson:
Here's part 2 of the HTTP proxy PRs, configuring on the client. I also worked out some issues with re-using the TCP tunnel connection.

Closes #33

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -115,15 +115,19 @@ password example-password
 ## HTTP Proxying
 
 HTTPX supports setting up proxies the same way that Requests does via the `proxies` parameter.
+For example to forward all HTTP traffic to `http://127.0.0.1:3080` and all HTTPS traffic
+to `http://127.0.0.1:3081` your `proxies` config would look like this:
 
 ```python
 >>> client = httpx.Client(proxies={
   "http": "http://127.0.0.1:3080",
-  "https": "http://127.0.0.1:3080"
+  "https": "http://127.0.0.1:3081"
 })
 ```
 
-Proxies can be configured per scheme, scheme+host, for all schemes of a host, or for all requests:
+Proxies can be configured for a specific scheme and host, all schemes of a host,
+all hosts for a scheme, or for all requests. When determining which proxy configuration
+to use for a given request this same order is used.
 
 ```python
 >>> client = httpx.Client(proxies={
@@ -136,9 +140,9 @@ Proxies can be configured per scheme, scheme+host, for all schemes of a host, or
 ```
 
 !!! warning
-    It is recommended for privacy reasons to use HTTPS and tunnel requests if possible.
-    Proxies can read your traffic even if the `proxy_url` parameter uses HTTPS
-    unless you make requests to the target domain with HTTPS through a `CONNECT` TCP tunnel.
+    To make sure that proxies cannot read your traffic,
+    and even if the proxy_url uses HTTPS, it is recommended to
+    use HTTPS and tunnel requests if possible.
 
 By default `HTTPProxy` will operate as a forwarding proxy for `http://...` requests
 and will establish a `CONNECT` TCP tunnel for `https://` requests. This doesn't change
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -111,3 +111,48 @@ password example-password
 
 ...
 ```
+
+## HTTP Proxying
+
+HTTPX supports setting up proxies the same way that Requests does via the `proxies` parameter.
+
+```python
+>>> client = httpx.Client(proxies={
+  "http": "http://127.0.0.1:3080",
+  "https": "http://127.0.0.1:3080"
+})
+```
+
+Proxies can be configured per scheme, scheme+host, for all schemes of a host, or for all requests:
+
+```python
+>>> client = httpx.Client(proxies={
+    "http://example.com":  "...",  # Host+Scheme
+    "all://example.com":  "...",  # Host
+    "http": "...",  # Scheme
+    "all": "...",  # All
+})
+>>> client = httpx.Client(proxies="...")  # Shortcut for 'all'
+```
+
+!!! warning
+    It is recommended for privacy reasons to use HTTPS and tunnel requests if possible.
+    Proxies can read your traffic even if the `proxy_url` parameter uses HTTPS
+    unless you make requests to the target domain with HTTPS through a `CONNECT` TCP tunnel.
+
+By default `HTTPProxy` will operate as a forwarding proxy for `http://...` requests
+and will establish a `CONNECT` TCP tunnel for `https://` requests. This doesn't change
+regardless of the `proxy_url` being `http` or `https`.
+
+Proxies can be configured to have different behavior such as forwarding or tunneling all requests:
+
+```python
+proxy = httpx.HTTPProxy(
+    proxy_url="https://127.0.0.1",
+    proxy_mode=httpx.HTTPProxyMode.TUNNEL_ONLY
+)
+client = httpx.Client(proxies=proxy)
+
+# This request will be tunnelled instead of forwarded.
+client.get("http://example.com")
+```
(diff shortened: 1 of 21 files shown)


tests/dispatch/test_proxy_http.py
```diff
@@ -105,23 +116,36 @@
...
+        assert response1.status_code == 404
+        assert response1.headers["Server"] == "origin-server"
 
-        assert response.request.method == "GET"
-        assert response.request.url == "https://example.com"
-        assert response.request.headers["Host"] == "example.com"
+        assert response1.request.method == "GET"
+        assert response1.request.url == "https://example.com"
+        assert response1.request.headers["Host"] == "example.com"
+        await response1.read()
```
sethmlarson: Had a question here, is there a way that we can detect that a response doesn't need data (ie `Content-Length: 0` or a status code that must have no content) and to immediately "release" the response?

docs/advanced.md
```diff
@@ -111,3 +111,48 @@ password example-password
...
+    "http": "...",  # Scheme
+    "all": "...",  # All
+})
+>>> client = httpx.Client(proxies="...")  # Shortcut for 'all'
+```
+
+!!! warning
+    It is recommended for privacy reasons to use HTTPS and tunnel requests if possible.
+    Proxies can read your traffic even if the `proxy_url` parameter uses HTTPS
+    unless you make requests to the target domain with HTTPS through a `CONNECT` TCP tunnel.
```
florimondmanca: It seems to me both sentences here convey the same tl;dr. So, may I suggest…?

> To make sure that proxies cannot read your traffic, and even if the `proxy_url` uses HTTPS, it is recommended to use HTTPS and tunnel requests if possible.

Also, I'm now wondering what prevents us from applying this recommendation by default? Does it depend on proxies capabilities?

docs/environment_variables.md
```diff
@@ -79,3 +79,17 @@ SERVER_TRAFFIC_SECRET_0 XXXX
...
 ```
+
+`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`
+----------------------------------------
+
+Valid values: A URL to a proxy
+
+Sets the proxy to be used for `http`, `https`, or all requests respectively.
+
+```console
```
florimondmanca: I think the bash syntax highlighting would look pretty nice here :)

```suggestion
```bash
```

Screenshot 2019-09-18 at 21 12 38

tests/dispatch/test_proxy_http.py
florimondmanca: As in, is there a way to make calling `await .read()` optional if we *know* it has no content, for the sake of convenience? I think checking for `Content-Length: 0` would be feasible/sensible, but not sure it's worth breaking the consistency.

To be rigorous I think we should really be using an `async with`-based API on async dispatchers, something like…

```python
async with proxy.request("GET", "https://example.com") as response:
    # response already .read().
    # It wouldn't be if we passed `stream=True`.
    # It is automatically `.close()`d on aexit.
```

(Right now this design is hard to implement because nothing follows this pattern internally or in our public API… Anyway, just jotting down thoughts here, kind of related to thoughts I had after #352.)

docs/advanced.md
```diff
@@ -111,3 +111,48 @@ password example-password
...
+
+Proxies can be configured per scheme, scheme+host, for all schemes of a host, or for all requests:
+
+```python
+>>> client = httpx.Client(proxies={
+    "http://example.com":  "...",  # Host+Scheme
+    "all://example.com":  "...",  # Host
+    "http": "...",  # Scheme
+    "all": "...",  # All
+})
```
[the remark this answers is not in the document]
sethmlarson: I'm actually going to do the opposite and match the wording to the example.
Also need to mention that we scan for a matching proxy based on those values.

docs/advanced.md
sethmlarson: Many proxies (especially public/free) don't offer HTTPS for both hops unfortunately.
Users will have to determine the capabilities of their proxy and hopefully configure properly, this warning will hopefully encourage people to try HTTPS first?

docs/advanced.md
```diff
@@ -111,3 +111,48 @@ password example-password
 
 ...
 ```
+
+## HTTP Proxying
+
+HTTPX supports setting up proxies the same way that Requests does via the `proxies` parameter.
```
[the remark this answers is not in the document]
sethmlarson: Done!

tests/dispatch/test_proxy_http.py
sethmlarson: Yeah I don't know, I left it as `.read()` for now. Understandable why we don't want the special cases.

docs/advanced.md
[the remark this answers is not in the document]
florimondmanca: > Also need to mention that we scan for a matching proxy based on those values.

I was going to merge but has this been adressed already?

florimondmanca:
> I also think we should update the Developer Interface docs with the proxies parameter where relevant.

I've just looked at that part of the docs and honestly it's a bit of a mess unfortunately — there are already multiple parameters missing, and I think the overall presentation could be improved. We can add the `proxies` parameter there, but we could also include that in a PR that cleans up this section of the docs. What do you prefer?

Besides, I think we forgot to add the `proxies` parameter to the high-level `httpx` API as well.

(On this note, I know we don't just use `**kwargs` on the various requesting methods because of [the khwarg problem](https://www.youtube.com/watch?v=MQMbnhSthZQ), but I have the feeling we could write some tests to make sure all those entrypoints are consistent in terms of exposed parameters. I may look into this. :))

docs/advanced.md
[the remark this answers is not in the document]
sethmlarson: Yep!

sethmlarson:
I agree that part of the docs needs a revamp, all that spam about parameters in get, post, options, etc isn't helping anyone. :)

I'm going to add the parameter in this PR then we can cleanup separately.

sethmlarson:
@florimondmanca How do these latest changes look?

httpx/client.py
```diff
@@ -238,6 +259,31 @@ def _get_auth_middleware(
...
+                f"all://{url.host}" if is_default_port else None,
+                url.scheme,
+                "all",
+            )
+            for proxy_key in proxy_keys:
+                if proxy_key and proxy_key in proxies:
+                    dispatcher = proxies[proxy_key]
+                    return dispatcher
+
+        return self.dispatch
```
yeraydiazdiaz: Is this logic covered in the tests? Seems like there should be some for the prioritization of `all` over specific IPs and the host defaulting.

sethmlarson: All kinda just means "catch-all" in this scenario. We follow requests lead here.

httpx/client.py
```diff
@@ -162,13 +173,20 @@ def merge_headers(
...
         trust_env: bool = None,
+        proxies: ProxiesTypes = None,
     ) -> AsyncResponse:
         if request.url.scheme not in ("http", "https"):
             raise InvalidURL('URL scheme must be "http" or "https".')
 
+        if proxies is not None:
+            dispatch_proxies = _proxies_to_dispatchers(proxies)
+        else:
+            dispatch_proxies = proxies or self.proxies
```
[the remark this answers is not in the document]
sethmlarson: Good catch!

sethmlarson:
@florimondmanca @yeraydiazdiaz Am I good to merge this?

PR #357, opened 2019-09-18
Use a certificate issued from the trustme CA in tests

### LINKED ISSUE #354: Change trustme.CA() usage to mimic real-life
We need to change our usage of `trustme.CA()` to allow doing TLS verifications within the unit test suite.

What this would look like:
- Global level `CA` object
- Fixture for an issued cert
- The private key and public PEM `.tempfile()` calls for these two components
- Switch the `https_server` instance to use the issued private key and public PEM.
- Switch all HTTPS tests to use the CA's PEM (not the issued PEM) as `verify=...`

We shouldn't be using verify=False within the test suite unless we're testing that specific behavior.

JayH5:
Fixes #354

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,9 @@ import typing
 
 import pytest
 import trustme
+from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.serialization import (
+    load_pem_private_key,
     BestAvailableEncryption,
     Encoding,
     PrivateFormat,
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -270,6 +276,7 @@ def https_server(cert_pem_file, cert_private_key_file):
         lifespan="off",
         ssl_certfile=cert_pem_file,
         ssl_keyfile=cert_private_key_file,
+        host="localhost",
```
JayH5: This is a bit of a less-than-ideal workaround for cert validation 😕. Since the server URL that the client uses is built using this host value (maybe it shouldn't be?) it needs to match the certificate name.

tests/conftest.py
```diff
@@ -139,47 +141,51 @@ def backend(request):
...
-    return ca
+def localhost_cert(cert_authority):
+    return cert_authority.issue_cert("localhost")
 
 
 @pytest.fixture(scope=SERVER_SCOPE)
-def cert_pem_file(example_cert):
-    with example_cert.cert_pem.tempfile() as tmp:
+def cert_pem_file(localhost_cert):
+    with localhost_cert.cert_chain_pems[0].tempfile() as tmp:
```
JayH5: From the trustme docs: "The zeroth entry in this list is the actual PEM-encoded certificate..."

Just including this cert and not the whole chain is fine, since there are no intermediary CA certs.

Issue #346, opened 2019-09-15
Does detection of lost connections implicitly depend on asyncio?


florimondmanca:
I'm mostly opening this to track the issue — I'll try to tackle this based on the trio backend in #276, and see what could work, e.g. setting up a background task to check for socket readability.

Worth mentioning though that [anyio already seems to handle this quite well](https://anyio.readthedocs.io/en/latest/api.html#anyio.wait_socket_readable).

---

The `Stream` interface exposes a `is_connection_dropped()` method which the `ConnectionPool` uses to know whether it can reuse a connection when acquiring a new one.

https://github.com/encode/httpx/blob/f1dde95d358a5a707d75e6d61f912224ed65ee8e/httpx/dispatch/connection_pool.py#L130-L139

Also noted by @njsmith in https://github.com/encode/httpx/pull/276#discussion_r322000402:

> asyncio always keeps reading from the socket in the background, whether you call read or not, as long as the loop is running. Trio and sync code don't. Unfortunately, httpx currently relies on the asyncio behavior to detect when an idle connection has gone stale. IMO this is a bug in httpx's idle connection handling, but that's the current state [...].

Also in https://github.com/encode/httpx/pull/276#discussion_r322004378:

>  [A]rguably it's a bug that when we want to detect if a connection is reusable, then the asyncio backend checks for EOF, when ideally it should be checking for readability. 

Indeed, that's what we do:

https://github.com/encode/httpx/blob/f1dde95d358a5a707d75e6d61f912224ed65ee8e/httpx/concurrency/asyncio.py#L124-L125

There was an entire discussion about readable vs closed before, see for example https://github.com/encode/httpx/pull/143#issuecomment-515181778. Counter-intuitively, it seems that when the other end has closed the conection then a socket is said to be *readable* (`.recv()` would return immediately with `b""`).

I guess the tl;dr here is that the way we implement `.is_connection_closed()` for asyncio is *good enough* (if the socket was closed, asyncio's background socket checker will mark at as having reached EOF), but we *might* need to change the way we check for dropped connections.

florimondmanca:
It seems I tangled myself up on the previous discussions and the counter-intuitive notion of *readability*. As I said, in the asyncio case checking for EOF is *good enough*, because a socket that reached EOF in the sense of asyncio is also readable (right?). So the code might just need some comments, which I added in https://github.com/encode/httpx/pull/276/commits/c5427357fa568e1571ec2d17a4d16298adf3de6e.

[...]

PR #358, opened 2019-09-18
Mark `BaseTCPStream.is_connection_dropped` as async

### LINKED ISSUE #356: Should `BaseTCPStream.is_connection_dropped` be async?
Looking at the urllib3 implementation of `is_connection_dropped` (https://github.com/urllib3/urllib3/blob/f7a4bed04085975918b3469bf94e195df9faf29a/src/urllib3/util/connection.py#L7-L26) it looks like their (pretty robust and battle tested 😄) implementation will sometimes block when testing if a connection is dropped (eventually calls into https://github.com/urllib3/urllib3/blob/f7a4bed04085975918b3469bf94e195df9faf29a/src/urllib3/util/wait.py#L105).

It looks like https://github.com/encode/httpx/issues/346 was closed, deciding the current implementation is good enough for now with some great research, but it sounds like you may want to improve this API in the future in a way that might require blocking.  Given you're pre-1.0 it would be good to get the APIs solid on what can eventually be async before 1.0 so you can more easily use a blocking implementation in the future.

I'm new here and just poking around so please LMK if there's any context I'm missing!

rowillia:
The current implementation doesn't require blocking, but given
https://github.com/encode/httpx/issues/346 discussed some possible future improvements,
[trio.hazmat.wait_readable](https://trio.readthedocs.io/en/stable/reference-hazmat.html#trio.hazmat.wait_readable) is async,
and [urllib3 relies on blocking APIs to detect dropped connections](https://github.com/urllib3/urllib3/blob/f7a4bed04085975918b3469bf94e195df9faf29a/src/urllib3/util/wait.py#L105) it is worth marking this API as async pre-1.0 to avoid future breaking changes.

Closes #356

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -111,7 +111,7 @@ class TCPStream(BaseTCPStream):
                 if should_raise:
                     raise WriteTimeout() from None
 
-    def is_connection_dropped(self) -> bool:
+    async def is_connection_dropped(self) -> bool:
         return self.stream_reader.at_eof()
 
     async def close(self) -> None:
--- a/httpx/concurrency/base.py
+++ b/httpx/concurrency/base.py
@@ -61,7 +61,7 @@ class BaseTCPStream:
     async def close(self) -> None:
         raise NotImplementedError()  # pragma: no cover
 
-    def is_connection_dropped(self) -> bool:
+    async def is_connection_dropped(self) -> bool:
         raise NotImplementedError()  # pragma: no cover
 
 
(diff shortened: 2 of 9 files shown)


rowillia:
@sethmlarson sorry forgot to link the ticket!  Context is here: https://github.com/encode/httpx/issues/356

florimondmanca:
While I think this is a great proposal, I can't help but feel like this might be a case of YAGNI — none of the *working* asyncio and trio (#276) implementations of `.is_connection_dropped()` use async/await. Also considering that this change forces a bunch of other areas in the code to be async when they did not need to be otherwise, I'm not sure we really need to make it *now*.

I'm pretty sure there's going to be some refactoring needed before we reach 1.0 anyway, so this might as well be part of it once we hit that point, but until then I'd rather keep things simple as they are today.

sethmlarson:
Maybe we should defer this until we know we need it?

florimondmanca:
I second you, @sethmlarson. Thank you again @rowillia, once we know we need this we can come back here and just grab the code! 😄

Issue #239, opened 2019-08-19
scripts as a single top-level point of project automation.


tomchristie:
Generally across the `encode` projects I've been using [GitHub's "Scripts to Rule Them All" pattern](https://github.blog/2015-06-30-scripts-to-rule-them-all/) as a consistent style for project testing, linting, publishing etc...

It's benenficial from my POV, because it ends up that we can always just run `scripts/install`, or `scripts/test` from *any* project and expect that to perform the same kind of thing, even if we're using different tooling between projects. (eg. started adopting `nox` in some places, but not yet there accross the board.)

I'd like to do the same with `httpx`, while still keeping the niceness of `nox` (So eg. reintroduce a `scripts/test`, `scripts/lint` and `scripts/publish`.) but figure it's worth raising this for discussion first.

I'd also be quite keen on having a fast-feedback test suite being the default, with nox's run against multiple environments as an explicit case?

Any general thoughts here or should I just dive in and take a crack at this?

florimondmanca:
Sharing my personal workflow here:

I'd say the `scripts/install` and `scripts/lint` scripts are rather straightforward and would be very helpful to have.

As for tests, I've been running `$ pytest` instead of nox to run tests against my virtual environment. Nox is too heavy for quick feedback, but running a complete pytest suite is moderately cheap (~15s). I also sometimes filter tests based on which ones I'm trying to get to pass, e.g. `pytest -k "test_something"` or `pytest tests/test_foo.py`.

So not sure whether `scripts/test` should default to `pytest`, or if it should also forward extra args for finer control?

sethmlarson:
We can add `session.posargs` to the `test` nox session so that you can be narrower in your test executions. :)

I have no issue with having the extra scripts, I didn't realize it was an encode org-wide thing.

tomchristie:
> I have no issue with having the extra scripts, I didn't realize it was an encode org-wide thing.

Totally not a problem - I think `nox` is a fantastic addition here, so not a big issue having to do a bit of rejigging to get there.

PR #363, opened 2019-09-19
Document release process

### LINKED ISSUE #313: Contributing guide - Release process
Inspired by #311, I had been thinking that it would be good for maintainers (at least personally) to document the versioning and release process: what operations to perform, in what order, with what requirements/permissions etc.

In other projects of mine I usually add a “Notes to maintainers” section to the contributing guide (maintainers are contributors as well, right?) with this kind of information.

Of course we wouldn’t include any sensitive information there, but rather document the steps so that anyone with the appropriate permissions can create a new version, push it to PyPI, and deploy the docs.

Thoughts on this @encode/httpx-maintainers?

florimondmanca:
Fixes #313

I may update this again once 0.7.3 is fully released. :)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -121,13 +121,13 @@ $ mkdocs serve
 
 Before releasing a new version, create a pull request that includes:
 
-- **An update to the changelog**.
-  - We follow the format from [keepachangelog](https://keepachangelog.com/en/1.0.0/).
-  - [Compare](https://github.com/encode/httpx/compare/) `master` with the tag of the latest release, and list all entries that are of interest to our users.
-  - Things that **must** go in the changelog: added, changed, deprecated or removed features, and bug fixes.
-  - Things that **should not** go in the changelog: changes to documentation, tests or tooling.
-  - Try sorting entries in descending order of impact / importance.
-  - Keep it concise and to-the-point. 🎯
+- **An update to the changelog**:
+    - We follow the format from [keepachangelog](https://keepachangelog.com/en/1.0.0/).
+    - [Compare](https://github.com/encode/httpx/compare/) `master` with the tag of the latest release, and list all entries that are of interest to our users:
+        - Things that **must** go in the changelog: added, changed, deprecated or removed features, and bug fixes.
+        - Things that **should not** go in the changelog: changes to documentation, tests or tooling.
+        - Try sorting entries in descending order of impact / importance.
+        - Keep it concise and to-the-point. 🎯
 - **A version bump**: see `__version__.py`.
 
 For an example, see [#362](https://github.com/encode/httpx/pull/362).
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -114,3 +114,21 @@ To spawn the docs server run:
 ```shell
 $ mkdocs serve
 ```
+
+## Releasing
+
+*This section is targeted at HTTPX maintainers.*
+
+Before releasing a new version, create a pull request that includes:
+
+- **An update to the changelog**.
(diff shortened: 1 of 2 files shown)


docs/contributing.md
```diff
@@ -114,3 +114,21 @@ To spawn the docs server run:
...
+
+## Releasing
+
+*This section is targeted at HTTPX maintainers.*
+
+Before releasing a new version, create a pull request that includes:
+
+- **An update to the changelog**.
+  - Visit `https://github.com/encode/httpx/compare/PREVIOUS_VERSION_TAG...HEAD` and list all entries that are of interest to our users. This typically includes bug fixes and new/changed/removed features. Documentation updates *may* make it to the changelog if they're of particular interest to our users. Updates to our tooling or test setup typically don't go in the changelog.
+  - Keep it concise and to-the-point.
```
sethmlarson: Shouldt we mention the keepachangelog format here?

florimondmanca: Happy to add it here, yes!

florimondmanca:
The `scripts/release` + `mkdocs gh-deploy` combo went as smooth as butter, and all the mandatory steps should be there now.

I'm wondering if we should add `mkdocs gh-deploy` to the `publish` script, though? I know Starlette and Uvicorn have that, and it's probably a good idea to always sync the docs after a PyPI release.

sethmlarson:
Can we make a "release" nox task? 🤩

florimondmanca:
> Can we make a "release" nox task? 🤩

In the light of https://github.com/encode/httpx/issues/239, I think `$ scripts/publish` serves its purpose well enough, doesn't it? 😄

sethmlarson:
Makes sense to me, disregard the nox task

docs/contributing.md
```diff
@@ -114,3 +114,24 @@ To spawn the docs server run:
...
+        - Things that **should not** go in the changelog: changes to documentation, tests or tooling.
+        - Try sorting entries in descending order of impact / importance.
+        - Keep it concise and to-the-point. 🎯
+- **A version bump**: see `__version__.py`.
+
+For an example, see [#362](https://github.com/encode/httpx/pull/362).
+
+Once the release PR is merged, run `$ scripts/publish` to publish the new release to PyPI.
+
+To release the docs site, run `$ mkdocs gh-deploy`.
```
sethmlarson: Do we need this now that docs publishing is in the publish script?

florimondmanca: I guess not, good catch :)

scripts/publish
```diff
@@ -15,11 +15,18 @@ if ! command -v "${PREFIX}twine" &>/dev/null ; then
     exit 1
 fi
 
+if ! command -v "${PREFIX}mkdocs" &>/dev/null ; then
+    echo "Unable to find the 'mkdocs' command."
+    echo "Install from PyPI, using '${PREFIX}pip install mkdocs'."
+    exit 1
+fi
```
yeraydiazdiaz: Should we add a `dev-requirements.txt` that builds on top of `test-requirements.txt` and adds `mkdocs` and `twine`?

florimondmanca: Ah, good point. *Or*… we could add a nox `release` task that installs those for us? 😅 @sethmlarson

florimondmanca: I tried to do that, and the fact that the `publish` script currently defaults to using `venv/bin/` as a command prefix is there is a `venv` directory means that it can't use the `mkdocs` and `twine` commands we'd install so long as we have a local virtualenv.

sethmlarson: @florimondmanca That's why I wanted a release task :)

StephenBrown2: > Should we add a `dev-requirements.txt` that builds on top of `test-requirements.txt` and adds `mkdocs` and `twine`?

Personally I would prefer all the `requirements.txt` files to be grouped and match "requirements-*.txt", i.e.
```
requirements-dev.txt
requirements-test.txt
```

Issue #262, opened 2019-08-21
HTTP/2 download speeds / flow control settings


tomchristie:
Which having an initial look at HTTP/3, @jlaine dug out that we seem to have signiifcantly slow uploading right now.

First thing to do would be to investigate and replicate, by eg. compare and contrast a simple upload from `httpx` vs `requests/urllib3` - does it replicate trivially and reliably? This really *shouldn't* be an issue for us so will need some looking at.

Initial thoughts on where issues could be:

* We're sending too-small chunks, and always waiting for the network to fully drain on each chunk is negatively impacting performance.
* We're sending too-small chunks, and ending up with largish computation overhead relative to data transfer as a result.
* We've got some big unseen overhead in the sync-to-async marshalling.
* Something else?

jlaine:
Actually the issue I reported was *download* speeds, and it does seem to be linked to HTTP/2.

Here is a script to reproduce the issue:

```python
import httpx
import time
import requests

url = "https://quic.aiortc.org/50000000"

start = time.time()
r = requests.get(url)
print("requests %.1f s" % (time.time() - start))

for http_version in ["HTTP/1.1", "HTTP/2"]:
    client = httpx.Client(http_versions=http_version)

    start = time.time()
    r = client.get(url)
    print("httpx %s %.1f s" % (http_version, time.time() - start))
```

The results (the first two results vary somewhat, but are always close to each other):

```
requests 5.6 s
httpx HTTP/1.1 5.8 s
httpx HTTP/2 45.7 s
```

tomchristie:
Okay, at this point it's a bit more complicated, since we've also got different performance on the server side too.

A sensible first thing to do might be to try to replicate against a non-python HTTP/1.1 and HTTP/2 server in order to try to help isolate if we've for a client issue or a server issue here (or both)?

tomchristie:
Tried this against `httpbin`, because that'll be implemented as a WSGI HTTP/1.1 - I was expecting it to be fronted by something that supported both HTTP/1.1 and HTTP/2 (I think it's on heroku?)

That didn't actually progress things tho, since it turns out whatever it's hosted on is HTTP/1.1 only...

```python
import httpx
import time
import requests

url = "https://httpbin.org/stream-bytes/50000000"

start = time.time()
r = requests.get(url)
print("requests %.1f s" % (time.time() - start))

for http_version in ["HTTP/1.1", "HTTP/2"]:
    client = httpx.Client(http_versions=http_version)

    start = time.time()
[...]

PR #364, opened 2019-09-20
Add profiler tool

florimondmanca:
I created https://github.com/florimondmanca/httpx-profile and @sethmlarson suggested I bring it in here. :)

Running it has uncovered some interesting insights already. For example, inside a `client.get()`:

- We're spending 30% (!) of the time looking up the `.netrc` login — we should probably cache that info once we have it?
- 13% is spent in `Client.build_request()` (the blue rectangle on the right), mostly in `merge_url()` — our `rfc3986` seems to be slowing us down here.

Screenshot 2019-09-20 at 23 40 02

We have some performance issues up such as #262, so this tool could be useful for all sorts of performance things.


yeraydiazdiaz:
This is awesome @florimondmanca, thanks for putting this together. 🌟 

As a somewhat related node, in the past I've used [`pytest-benchmark`](https://pytest-benchmark.readthedocs.io/en/stable/) and wrote a different test suite to monitor the performance of the code. It first occured to me we could combine it with profiling but this solution seems more focused for finding what is causing the performance problem is rather than detecting changes on it.

I might take a stab at benchmark testing if it sounds useful.

florimondmanca:
That does sound interesting @yeraydiazdiaz!

This tool is indeed more geared towards identifying performance issues rather than measuring execution speed. I'll reword things here a bit to move uses of the word "benchmark" altogether.

florimondmanca:
Hmm, so `pytest-benchmark` made me think that maybe there were pytest plugins to benchmark the test code as it runs, potentially resulting in a much simpler setup for us.

Turns out, there is one: [`pytest-profiling`](https://pypi.org/project/pytest-profiling/). I'm having issues having it generate the SVG file though, and the project hasn't seen a commit since last May. I guess this custom tool is simple enough that we can afford to maintain it ourselves.

florimondmanca:
I also pushed some polishing commits — the supporting surver now runs automatically in the background.

Example output:

```console
$ httpxprof run async                                    
Server started at 127.0.0.1:8123.
100%|███████████████████████████████████████████████████████████████████████████| 1000/1000 [00:03<00:00, 277.89it/s]
Stopping server...
```

sethmlarson:
@florimondmanca What are your thoughts on pointing our linting (maybe not mypy) at the `tools/` directory as well? :)

florimondmanca:
@sethmlarson I added that in the latest commit. :) Is this good to merge?

PR #369, opened 2019-09-21
Use Python 3.8 asyncio.Stream where possible

### LINKED ISSUE #256: AsyncioBackend: DeprecationWarning for StreamReader/StreamWriter in Py3.8
Python 3.8 builds are filled with warnings about asyncio API changes: https://travis-ci.org/encode/httpx/jobs/574559125

Looks like as as of Python 3.8, `asyncio.open_connection()` is deprecated in favor of `asyncio.connect()` — see the [asyncio 3.8+ docs](https://docs.python.org/3.8/library/asyncio-stream.html).

```python
# < 3.8
reader, writer = await asyncio.open_connection('127.0.0.1', 8888)

# 3.8+
async with asyncio.connect('127.0.0.1', 8888) as stream:
    ...
# or:
stream = await asyncio.connect('127.0.0.1', 8888)
```

This is not a trivial API change, and I don't know if there are plans to back-port `asyncio.connect()` to 3.7 and below.

So at first glance resolving this would need to adapt the `.connect()` method on `AsyncioBackend` to use one or the other depending on the Python version we're running on — and also make `Reader`/`Writer` pair (and soon `Stream`) handle both possibilities ?

Originally raised in: https://github.com/encode/httpx/pull/255#discussion_r315904773

JayH5:
Fixes #256 (although there is still one warning somewhere in `websockets`).

On Python < 3.8, this wraps the `asyncio.StreamReader`/`asyncio.StreamWriter` objects in a `StreamCompat` type with a similar interface to the new `asyncio.Stream`. On Python >= 3.8, `asyncio.Stream` is used instead.

Similarly, on Python < 3.8 `loop.open_connection()` is used, but on >=3.8 `loop.connect()` is used.


httpx/concurrency/asyncio/backend.py
```diff
@@ -112,10 +107,10 @@ def write_no_block(self, data: bytes) -> None:
                     raise WriteTimeout() from None
 
     def is_connection_dropped(self) -> bool:
-        return self.stream_reader.at_eof()
+        return self.stream.at_eof()
 
     async def close(self) -> None:
-        self.stream_writer.close()
+        await self.stream.close()
```
JayH5: This `await` should be the only real functional change (at least for Python < 3.8). I get spurious errors in tests from this, though, that look a lot like [this](https://github.com/aio-libs/aiohttp/issues/3535) bug that is supposed to be fixed in Python 3.7.4 which I'm running 🤔

httpx/concurrency/asyncio/compat.py
```diff
@@ -0,0 +1,138 @@
...
+
+            reader.set_transport(tls_transport)
+            self.reader = reader
+            self.writer = asyncio.StreamWriter(
+                transport=tls_transport, protocol=protocol, reader=reader, loop=loop
+            )
+
+    def write(self, data: bytes) -> typing.Awaitable[None]:
+        self.writer.write(data)
+        return _OptionalAwait(self.drain)
```
JayH5: `await`-ing on the `drain` call is not _exactly_ the same as how things work in Python 3.8 when you `await` on `Stream.write()` it first tries a fast-path to flush data and then falls back to a full drain.

httpx/concurrency/asyncio/backend.py
```diff
@@ -4,9 +4,9 @@
 import typing
 from types import TracebackType
 
-from ..config import PoolLimits, TimeoutConfig
-from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
-from .base import (
+from ...config import PoolLimits, TimeoutConfig
+from ...exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
```
JayH5: How many dots is too many dots? 🤔

florimondmanca: 3 or more is definitely too much I'd say. We can switch from `from httpx import ...` I think. :)

httpx/concurrency/asyncio/compat.py
florimondmanca: Apparently [drain() was deprecated in 3.8 too](https://docs.python.org/3.8/library/asyncio-stream.html#asyncio.Stream.drain), and it is now recommended to `await stream.write()`. Should we switch to this latter API? This means we'd need to refactor `AsyncioBackend.write()` a bit, and potentially use a manual buffering mechanism for `.write_no_block()`. (There's an example of this in the trio backend: #276.)

httpx/concurrency/asyncio/backend.py
```diff
@@ -126,10 +122,12 @@ def is_connection_dropped(self) -> bool:
         # (For a solution that uses private asyncio APIs, see:
         # https://github.com/encode/httpx/pull/143#issuecomment-515202982)
 
-        return self.stream_reader.at_eof()
+        return self.stream.at_eof()
 
     async def close(self) -> None:
-        self.stream_writer.close()
+        # FIXME: We should await on this call, but need a workaround for this first:
+        # https://github.com/aio-libs/aiohttp/issues/3535
```
JayH5: With the `await`, the `test_start_tls_on_socket_stream` fails transiently with the error mentioned in this thread in about 1 out of 5-30 test runs (thanks, `pytest-repeat`) for me on Python 3.7.2, 3.7.4, and 3.8.0b4.

There is a possible fix (a custom exception handler set on the event loop) in the thread, but it's fairly complicated and I'm not sure how to integrate it nicely. This is no _worse_ than we had so I think it should be fixed in a separate PR.

httpx/concurrency/asyncio/compat.py
JayH5: I was hoping this could tie into #341 quite nicely maybe?

florimondmanca: Yes, actually it does. :)

httpx/concurrency/asyncio/backend.py
sethmlarson: Wow, that thread doesn't look good. :fearful:

sethmlarson:
If you're good with this @florimondmanca we can merge?

httpx/concurrency/asyncio/backend.py
florimondmanca: Is it OK to not await this `.close()` call, though?

florimondmanca: Thanks, all these details are very helpful. :)

Since the 3.8 docs it’s only *possible* to await on .close() (which I find kind of weird, having in mind all those « coroutine was never awaited » exceptions that usually causes), I’m okay with keeping it as it is currently. As you said, we actually never waited for the socket to close before (and if we did, we’d have probably only encountered this issue earlier?).

florimondmanca:
Note to myself that I’ll have to deal with the changes in #341 once this get merged. :)

httpx/concurrency/asyncio/backend.py
JayH5: One (not very good) reason to include this `try`/`except` is that without it I don't think this change has 100% test coverage since `wait_closed` is never called.

tomchristie:
Crackin work, thank you! ✨

To the team in general - can we peddle just a wee bit slower on getting this one in?

We need to unify `StreamReader` and `StreamWriter` into a single `Stream` interface in any case, and I'm a bit uncomfortable with some basics such as having such a deeply nested file layout.

I'd quite like to approach this, by tackling the `StreamReader`+`StreamWriter` -> a single interface *first*, and then looking at how 3.8 support changes that.

Also in cases where we need to have version-specific branches I typically try to isolate those *strictly* all into a `compat.py` module, that can then be excluded from coverage reporting. I think we may well want to have a bit of a review of if we can/should be doing the same with `httpx`.

florimondmanca:
@tomchristie The reader/writer pair is very much specific to asyncio — our stream interface which concurrency backends manipulate is already a single interface. As a matter of fact, I don’t think this PR changes any of that. Instead, it basically reimplements asyncio.Stream for < 3.8 so that we can use *that* to implement the existing Stream interface in the asyncio backend. So is there anything I’m missing about existing reader/writer interfaces?

tomchristie:
Ah okay makes sense - I’m getting a bit back up to speed and probably need to take another look over.

sethmlarson:
@JayH5 This looks like it needs some conflict resolution with master before it can land. Mind taking a look? :)

JayH5:
git didn't have any problem merging master. I guess GitHub must use a simpler merging algorithm 🤷‍♂

sethmlarson:
Thanks for much for this contribution @JayH5! :raised_hands: Hope to see more work in our complex async machinery from you. :)

tomchristie:
Not a big problem, but from my POV we've moved too fast on merging this. 🏎🌬
I'd *really like us to slow down* on substantial changes like this. 😇

(Even without the change in Python 3.8 I was expecting to take a fuller review onto it, first.)

Given https://github.com/encode/httpx/issues/256#issuecomment-536274181 should we raise an issue to just plain revert this, or something else?

PR #371, opened 2019-09-21
Add debug logs to ASGIDispatch

florimondmanca:
Working on and debugging #352, I noticed we didn't have debug logs for the `ASGIDispatch`, so here they are. :)

Example output:

```python
import httpx

    assert scope["type"] == "http"
    await send(
        {
            "status": 200,
        }
    )

with httpx.Client(app=app) as client:
    r = client.get("http://app")
    print(r)
```

```console
20:18:22.096 - httpx.dispatch.asgi - started
20:18:22.096 - httpx.dispatch.asgi - received message={'type': 'http.response.start', 'status': 200, 'headers': '<...>'}
20:18:22.096 - httpx.dispatch.asgi - received message={'type': 'http.response.body', 'body': '<13 bytes>'}
20:18:22.097 - httpx.dispatch.asgi - completed
<Response [200 OK]>
```

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -187,6 +187,15 @@ def get_logger(name: str) -> logging.Logger:
     return logging.getLogger(name)
 
 
+def kv_format(**kwargs: typing.Any) -> str:
+    """Format arguments into a key=value line.
+
+    >>> formatkv(x=1, name="Bob")
+    "x=1 name='Bob'"
+    """
+    return " ".join(f"{key}={value!r}" for key, value in kwargs.items())
+
+
 def get_environment_proxies() -> typing.Dict[str, str]:
     """Gets proxy information from the environment"""
 
@@ -259,7 +268,6 @@ ASGI_PLACEHOLDER_FORMAT = {
     "body": "<{length} bytes>",
     "bytes": "<{length} bytes>",
     "text": "<{length} chars>",
-    "headers": "<...>",
 }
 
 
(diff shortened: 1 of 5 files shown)


sethmlarson:
Since we show headers for the non-ASGI loggers, should we consider showing them here as well?

florimondmanca:
Good point @sethmlarson, which made me realize there were much more data we'd be interested in seeing. We now dump all of the ASGI scope and events, which yields something like this:

```console
22:35:40.591 - httpx.dispatch.asgi - started type='http' asgi={'version': '3.0'} http_version='1.1' method='GET' headers=[(b'host', b'app'), (b'user-agent', b'python-httpx/0.7.3'), (b'accept', b'*/*'), (b'accept-encoding', b'gzip, deflate, br'), (b'connection', b'keep-alive')] scheme='https' path='/' query_string=b'' server='app' client=('127.0.0.1', 123) root_path=''
22:35:40.591 - httpx.dispatch.asgi - received type='http.response.start' status=200 headers=[[b'content-type', b'text/plain']]
22:35:40.591 - httpx.dispatch.asgi - received type='http.response.body' body='<13 bytes>'
22:35:40.594 - httpx.dispatch.asgi - completed
```

httpx/utils.py
```diff
@@ -241,3 +250,54 @@ def elapsed(self) -> timedelta:
...
+}
+
+
+def asgi_message_with_placeholders(message: dict) -> dict:
+    """
+    Return an ASGI message, with any body-type content omitted and replaced
+    with a placeholder.
+    """
+    new_message = message.copy()
+    for attr in ASGI_PLACEHOLDER_FORMAT:
```
sethmlarson: I think we need to have a function that can take something with an `.items()` method and filter out headers like `Authorization` and `Proxy-Authorization` like `Headers.__repr__()`.  Then we can use it in both places.

StephenBrown2:
Can this logger middleware be made more generic for requests, or is is specific to ASGI apps only?

florimondmanca:
@StephenBrown2 Its very much specific to ASGI. Our other loggers typically log h11/h2 events, I don’t see a way we can reconcile them with this. I hope having « ASGI » present in the name of the logger middleware (because it *is* an ASGI middleware, ie it implements the ASGI interface) conveys this intent well enough.

StephenBrown2:
Aye! `HTTPX_DEBUG` is good enough for everything, this just adds that logging to ASGI.

httpx/utils.py
```diff
@@ -160,6 +160,18 @@ def parse_header_links(value: str) -> typing.List[typing.Dict[str, str]]:
...
 
 
+SENSITIVE_HEADERS = {"authorization", "proxy-authorization"}
+
+
+def obfuscate_sensitive_headers(
+    items: typing.Iterable[typing.Tuple[typing.AnyStr, typing.AnyStr]]
+) -> typing.Iterator[typing.Tuple[typing.AnyStr, typing.AnyStr]]:
+    for k, v in items:
+        if to_str(k) in SENSITIVE_HEADERS:
```
sethmlarson: `k.lower()`

Issue #368, opened 2019-09-21
`params` is missing from `Client`/`AsyncClient`


florimondmanca:
From #365:

> `requests.Session` has a `params` attribute that attaches those params to each session request. This is not present in httpx `Client` classes like with headers/cookies. This is a useful feature for this and other use cases of mine and a divergence from requests.

Put differently, we would like to be able to define default query parameters on the client so that they are included by default in requests:

```python
import httpx

client = httpx.Client(params={"api_key": "..."})
r = client.get("https://api.app.io")  # `api_key` query parameter was sent
```

Solution hints:

- Update `BaseClient` constructor to accept the `params` parameter.
- Create a `.params` property similar to `.headers` and `.cookies`.
- Add a `.merge_query_params()` method similar to `.merge_headers()`.
- Update `.build_request()` to use this new method.

Issue #366, opened 2019-09-21
Query params should support list-of-strings values.


Senpos:
Hi,

I recently tried to use `httpx` instead of `requests` in my simple app that calls few API endpoints and  suddenly it did not work.

The problem here is that `httpx` and `requests` process query parameters differently.

Here's a simple script to see how both libraries process `params`:
```python
from httpx.models import QueryParams
from requests.models import RequestEncodingMixin

dict_params = {
    "documentIds": ["a55b2f894b90dd1213201471_25", "b54b2f894b90dd1213201471_9"]
}

httpx_params = str(QueryParams(dict_params))
print("httpx", httpx_params)

requests_params = RequestEncodingMixin._encode_params(dict_params)
print("requests", requests_params)
```

Results:
```
httpx documentIds=%5B%27a55b2f894b90dd1213201471_25%27%2C+%27b54b2f894b90dd1213201471_9%27%5D
requests documentIds=a55b2f894b90dd1213201471_25&documentIds=b54b2f894b90dd1213201471_9
```

If we try to parse the query string `httpx` produced, here's what we get:
```python
from urllib.parse import parse_qs

httpx_params_parsed = parse_qs('documentIds=%5B%27a55b2f894b90dd1213201471_25%27%2C+%27b54b2f894b90dd1213201471_9%27%5D')
print(httpx_params_parsed )
```
Result:
```
{'documentIds': ["['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']"]}
```
Which is not what you would expect, I guess. At least, it works differently than in `requests`.

If it is an intended behavior, can you please recommend me the proper way of passing a list into the query params?

Thanks!

Senpos:
Here's a bit more "live" example:
```python
import httpx
import requests

params = {"documentIds": ["a55b2f894b90dd1213201471_25", "b54b2f894b90dd1213201471_9"]}

URL = "https://httpbin.org/get"
r = requests.get(URL, params=params)
print("requests", r.json()["args"])

r = httpx.get(URL, params=params)
print("httpx", r.json()["args"])


```

Results:
```
requests {'documentIds': ['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']}
httpx {'documentIds': "['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']"}
```

florimondmanca:
Hi @Senpos! Thank you for reporting this.

[...]

PR #372, opened 2019-09-21
Adding params to Client

tyrelsouza:
References https://github.com/encode/httpx/issues/368

I models how I did the `__getitem__`  and `__setitem__` after the Headers one, so I chose to normalize the keys to lower. 

First time doing Python in _a while_ so figured I'd try to contribute.


httpx/models.py
```diff
@@ -316,8 +316,22 @@ def get(self, key: typing.Any, default: typing.Any = None) -> typing.Any:
...
 
+    def update(self, params: QueryParamTypes = None) -> None:  # type: ignore
+        if params:
+            params = QueryParams(params)
+        else:
+            params = QueryParams()
+        for param in params:
+            self[param] = params[param]
+
     def __getitem__(self, key: typing.Any) -> str:
```
sethmlarson: I don't think params are case insensitive per Requests implementation, it looks like they're just using a normal dictionary? I wasn't able to replicate the behavior being tested below:

```python
import requests

sess = requests.Session()
sess.params["a"] = "b"
assert session.params["A"] == "b"  # This fails
sess.params  # {"a": "b"}
```

tests/client/test_query_params.py
```diff
@@ -0,0 +1,13 @@
+from httpx import Client, QueryParams
+
+
+def test_client_query_params():
+    client = Client(params={"a": "b"})
+    assert isinstance(client.params, QueryParams)
+    assert client.params["A"] == "b"
+
+
+def test_client_query_params_string():
```
sethmlarson: We'll additionally want some functional tests to make sure the params are being sent over the wire as we expect.

httpx/models.py
tyrelsouza: Ah great, that makes it easier then!

tests/client/test_query_params.py
tyrelsouza: Ooh yeah good call, also noticed that queryparams is both `_list` and `_dict`. so I need to do the same indexing as `Headers` does

httpx/client.py
```diff
@@ -162,6 +175,15 @@ def merge_headers(
             return merged_headers
         return headers
 
+    def merge_query_params(
```
tyrelsouza: It seems there's more instances in the code of `queryparam` vs `query_param`, the Issue suggested to use `merge_query_params` -> but I don't see the similar methods being used elsewhere besides in `build_request`, I'm going to make a judgement call and change my `query_params` to `queryparams` to match the rest of the codebase.

tyrelsouza:
Looking at [this comment](https://github.com/encode/httpx/issues/366#issuecomment-533788781) from @florimondmanca, my `__setitem__` doesn't contribute any fixes towards multi-valued queries, wonder if that should be addressed first, or if there's a deeper issue if I keep this `__setitem__`.

florimondmanca:
@tyrelsouza I think #366 will require us to switch the query Paramus data structure to a [multidict](https://github.com/aio-libs/multidict) of some sort. IMO this can (should?) be addressed later — when we make the switch, we’ll come back here and merge multidicts instead of regular dictionaries.

florimondmanca:
Forget about what I said earlier about multidicts — that's obviously already what we're using.

I played around with the bug from #366 and it turns out it can be fixed independently from this. I'll file a fix in a sec. :-)

httpx/client.py
```diff
@@ -69,6 +70,7 @@ def __init__(
         app: typing.Callable = None,
         backend: ConcurrencyBackend = None,
         trust_env: bool = True,
+        params: QueryParamTypes = None,
```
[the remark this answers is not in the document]
tyrelsouza: Sure! I tossed it after auth so that it's in the `params,headers,cookies` order

httpx/client.py
```diff
@@ -112,6 +114,10 @@ def __init__(
             proxies
         )
 
+        self._params = QueryParams()
+        if params:
+            self._params.update(params)
```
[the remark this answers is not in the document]
tyrelsouza: Can do!

tyrelsouza:
Sure! I can get to that in two hours. If you don't want to have to wait that long I won't be hurt if I don't do it!

PR #374, opened 2019-09-22
Query parameter encoding fixes

### LINKED ISSUE #366: Query params should support list-of-strings values.

florimondmanca:
Fixes #366 

Re-running the example from https://github.com/encode/httpx/issues/366#issuecomment-533786466:

```python
import httpx
import requests

URL = "https://httpbin.org/get"
PARAMS = {"documentIds": ["a55b2f894b90dd1213201471_25", "b54b2f894b90dd1213201471_9"]}

resp = requests.get(URL, params=PARAMS)
print("requests", resp.json()["args"])

r = httpx.get(URL, params=PARAMS)
json = r.json()
assert isinstance(json, dict)
print("httpx", json["args"])

print("httpx", str(httpx.QueryParams(PARAMS)))
```

```console
requests {'documentIds': ['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']}
httpx {'documentIds': ['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']}
httpx documentIds=a55b2f894b90dd1213201471_25&documentIds=b54b2f894b90dd1213201471_9
```

cc @Senpos

This also fixes encoding of bytes-type query params:

```python
import httpx

q = httpx.QueryParams({"data": b"hello"})
print(q)  # 'data=hello'
# Previously: 'b%27hello%27'
```

I dared to include a refactoring of the tests to make use for parametrization — the main motivation is to not stop at the first failing `assert`.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -52,6 +52,7 @@ QueryParamTypes = typing.Union[
     "QueryParams",
     typing.Mapping[str, PrimitiveData],
     typing.List[typing.Tuple[str, PrimitiveData]],
+    typing.Mapping[str, typing.Sequence[PrimitiveData]],
     str,
 ]
 
@@ -298,7 +299,7 @@ class QueryParams(typing.Mapping[str, str]):
                 for u in (
                     v  # type: ignore
                     if hasattr(v, "__iter__") and not isinstance(v, (str, bytes))
-                    else [v]
+                    else [v]  # type: ignore
                 )
             ]
 
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -291,7 +291,16 @@ class QueryParams(typing.Mapping[str, str]):
         elif isinstance(value, list):
             items = value  # type: ignore
         else:
-            items = value.items()  # type: ignore
+            # A dict-like object whose values may be lists.
+            items = [
+                (k, u)  # type: ignore
+                for k, v in value.items()
+                for u in (
+                    v  # type: ignore
+                    if hasattr(v, "__iter__") and not isinstance(v, (str, bytes))
+                    else [v]
+                )
+            ]
 
         self._list = [(str(k), str_query_param(v)) for k, v in items]
         self._dict = {str(k): str_query_param(v) for k, v in items}
@@ -312,9 +321,7 @@ class QueryParams(typing.Mapping[str, str]):
         return list(self._list)
 
     def get(self, key: typing.Any, default: typing.Any = None) -> typing.Any:
-        if key in self._dict:
-            return self._dict[key]
-        return default
+        return self._dict.get(key, default)
 
     def update(self, params: QueryParamTypes = None) -> None:  # type: ignore
         if not params:
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -51,8 +51,8 @@ URLTypes = typing.Union["URL", str]
 QueryParamTypes = typing.Union[
     "QueryParams",
     typing.Mapping[str, PrimitiveData],
-    typing.List[typing.Tuple[str, PrimitiveData]],
     typing.Mapping[str, typing.Sequence[PrimitiveData]],
+    typing.List[typing.Tuple[str, PrimitiveData]],
     str,
 ]
 
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -11,6 +11,9 @@ from time import perf_counter
 from types import TracebackType
 from urllib.request import getproxies
 
+if typing.TYPE_CHECKING:  # pragma: no cover
+    from .models import PrimitiveData
+
 
 def normalize_header_key(value: typing.AnyStr, encoding: str = None) -> bytes:
     """
@@ -30,7 +33,7 @@ def normalize_header_value(value: typing.AnyStr, encoding: str = None) -> bytes:
     return value.encode(encoding or "ascii")
 
 
-def str_query_param(value: typing.Optional[typing.Union[str, int, float, bool]]) -> str:
+def str_query_param(value: "PrimitiveData") -> str:
     """
     Coerce a primitive data type into a string value for query params.
 
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -42,6 +42,8 @@ def str_query_param(value: typing.Optional[typing.Union[str, int, float, bool]])
         return "false"
     elif value is None:
         return ""
+    elif isinstance(value, bytes):
+        return value.decode()
     return str(value)
 
 
--- a/tests/models/test_queryparams.py
+++ b/tests/models/test_queryparams.py
@@ -1,8 +1,18 @@
+import pytest
+
 from httpx import QueryParams
 
 
-def test_queryparams():
-    q = QueryParams("a=123&a=456&b=789")
+@pytest.mark.parametrize(
+    "source",
+    [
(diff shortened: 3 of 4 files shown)


tests/models/test_queryparams.py
```diff
@@ -18,36 +28,41 @@ def test_queryparams():
...
 
-    q = QueryParams({"a": 123})
-    assert str(q) == "a=123"
+@pytest.mark.parametrize(
+    "source, output",
+    [
+        ({"a": True}, "a=true"),
+        ({"a": False}, "a=false"),
+        ({"a": ""}, "a="),
+        ({"a": None}, "a="),
```
sethmlarson: I still don't agree with this behavior. There's no way to represent just `a` in a query string without `=` and it's one of the [most commented issues for Requests](https://github.com/psf/requests/issues/2651) there's definitely a need.

httpx/utils.py
```diff
@@ -42,6 +42,8 @@ def str_query_param(value: typing.Optional[typing.Union[str, int, float, bool]])
         return "false"
     elif value is None:
         return ""
+    elif isinstance(value, bytes):
```
sethmlarson: The signature for `value` isn't `PrimitiveData`, shame mypy didn't catch this.

tests/models/test_queryparams.py
florimondmanca: I see the need in there, yes. My safest bet is that we could use `None` for this.

But Requests also uses it to allow "unsetting" query params from the client for a specific request. Do we want to support that use case too? Manually omitting the key by rebuilding the `params` dict without the target value wouldn't work since we merge query params with the client's.

(I rarely wish we had an `undefined` value as in JavaScript, but this is one of those moments.)

florimondmanca: What I think would make sense: https://github.com/sindresorhus/query-string#falsy-values

```js
queryString.stringify({foo: false});
//=> 'foo=false'

queryString.stringify({foo: null});
//=> 'foo'

queryString.stringify({foo: undefined});
//=> ''
```

httpx/utils.py
florimondmanca: Probably because both actually refer to the same type, just with a different name. I updated the code so we use `PrimitiveData` everywhere. :)

tests/models/test_queryparams.py
sethmlarson: Hmm, I wonder if the `httpx.UNSET` falls into the YAGNI category?

BTW Requests didn't allow setting session-level params from the constructor either (it should have though!) and only allows editing as a property. We can document that the params property has different behavior when it comes to unsetting a predefined value?

florimondmanca: > We can document that the params property has different behavior when it comes to unsetting a predefined value?

Our behavior being that we don't support unsetting? :)

sethmlarson: Hah, my wording was bad there but yes. Document that None adds a no-value entry to the query string instead of unsetting.

sethmlarson: We can leave this for the future. Maybe we should create an issue for it.

httpx/utils.py
sethmlarson: Oh I thought `PrimitiveData` included bytes (hence your isinstance() call). Do we want to include bytes here?

florimondmanca: Ah I see! I also thought PrimitiveData included bytes from the beginning, but it never did. So this is technically a new feature then (bytes support in query params) — probably worth adding right? If that’s reasonable I’ll see if it makes sense to add bytes to PrimitiveData, or include bytes as an extra option here as you initially mentioned.

tests/models/test_queryparams.py
florimondmanca: Yes, I think that can be discussed and tackle separately. Feel fee to open the issue, or I’ll do it as part of post merge cleanup. :)

httpx/utils.py
sethmlarson: It seems reasonable to me. :)

httpx/models.py
```diff
@@ -291,7 +292,16 @@ def __init__(self, *args: QueryParamTypes, **kwargs: typing.Any) -> None:
...
+            # A dict-like object whose values may be lists.
+            items = [
+                (k, u)  # type: ignore
+                for k, v in value.items()
+                for u in (
+                    v  # type: ignore
+                    if hasattr(v, "__iter__") and not isinstance(v, (str, bytes))
+                    else [v]  # type: ignore
+                )
+            ]
```
tomchristie: I think double comprehensions are a bit incomprehensible. Any other tack we could take here maybe?

tomchristie:
So, I'd probably rather we *didn't* also roll in the "allow bytes values" alongside the "allow list of primitive data values".

My tendency is that we ought to try to be really tight on ensuring that PRs really do deal with one issue only, and that we're following a hard maxim of pull requests being "The absolute minimum possible coherant change that makes a functional improvement".

It's possible that I'm coming in harder on that than neccessary, but from my POV it's a really important way to ensure sustained quality over the long-term. I find it's far too eaiser to otherwise accept larger-than-needed PRs.

tomchristie:
I know my viewpoint there is probably a whole lot stricter than many folks' might be, but I really do think it'd be a good guideline, even if it sometimes puts a little bit of a brake on things.

florimondmanca:
Sure thing, I fully understand your point of view @tomchristie. As you said, it’s easier to not be as strict, but the discipline definitely has benefits. I’ll split the PR in two. :)

florimondmanca:
OK, I repurposed this in https://github.com/encode/httpx/pull/386 — no bytes handling, no not-strictly-needed refactoring. Closing this one, will add an issue for allowing bytes data.

PR #375, opened 2019-09-23
fix invalid f-string

Hanaasagi:
`f` prefix only affect the following string.

```Python
var = 0
string1 = (
    f"test "
    "{var}"
)
string2 = (
    "test "
    f"{var}"
)

assert string1 == "test {var}"
assert string2 == "test 0"

```

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -269,8 +269,8 @@ class HTTPVersionConfig:
             }
         else:
             raise TypeError(
-                f"HTTP version should be a string or list of strings, "
-                "but got {type(http_versions)}"
+                "HTTP version should be a string or list of strings, "
+                f"but got {type(http_versions)}"
             )
 
         for version in self.http_versions:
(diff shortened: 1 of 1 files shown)


sethmlarson:
Nice catch! Thanks @Hanaasagi

StephenBrown2:
Might we add https://pypi.org/project/flake8-pie/ to the [test-requirements](https://github.com/encode/httpx/blob/master/test-requirements.txt#L9) and [check](https://github.com/encode/httpx/blob/master/noxfile.py#L23) session?
Adds the following lints, PIE782 will specifically catch this issue:

- PIE781: You are assigning to a variable and then returning. Instead remove the assignment and return.
- **PIE782: Unnecessary f-string. You can safely remove the f prefix.**
- PIE783: Celery tasks should have explicit names.
- PIE784: Celery crontab is missing explicit arguments.
- PIE785: Celery tasks should have expirations.

sethmlarson:
@StephenBrown2 I'd accept a PR with that change. :)

Hanaasagi:
I think the next release of flake8 will detect this problem.

StephenBrown2:
That would be great! Can you link the GitLab issue so we can track it?

pquentin:
The functionality was added by pyflakes in https://github.com/PyCQA/pyflakes/pull/443, but no pyflakes release includes this yet. I don't think there's a GitLab issue to track on the flake8 side.

PR #379, opened 2019-09-24
Run isort in the nox check session

JayH5:
So that it is checked during CI. Currently it is only run in the lint session which is not run during CI.


sethmlarson:
This is great, thank you @JayH5!

Issue #316, opened 2019-09-05
Enforce 100% test coverage


yeraydiazdiaz:
As [previously commented](https://github.com/encode/httpx/pull/312#issuecomment-527869994) we should enfoce a 100% test coverage while testing locally and in CI.

tomchristie:
I'm *really*, *really* keen on us getting this back.
Dropping the 1.0 milestone tho, since it's not *strictly* required for us to get to a point where we consider HTTPX to be API-stable.

yeraydiazdiaz:
I'd like us to revisit this, IMO we should at least aim to have 100% and if in some situations we can't then we mark as `nocover` and explain why.

Thoughts?

tomchristie:
Yup, v keen on this.

yeraydiazdiaz:
In order to get this going I think we should start enforcing the coverage not to drop on each commit.

I seem to remember we used to have a codecov but I'm not sure when and why we dropped that?

tomchristie:
> I seem to remember we used to have a codecov but I'm not sure when and why we dropped that?

We *used* to have 100% enforced, and somewhere *way* back that seemed to get accidentally dropped.
After that we didn't have any enforcement, and I probably dropped codecov when switching across to GitHub actions, as a potential "follow up" action, that I didn't want to block us being able to move to GitHub actions.

I'd *probably* be quite keen on instead enforcing it ourselves instead. It's not ideal while we're at <100% because we can probably only enforce "this percent and no less", but it does mean that we can make sure that running the tests locally ensure compliance, rather than having to wait for the cycle of "submit PR", "test suite fails", "maintainer points out why", "contributor adds coverage".

Also it's one less moving part, and I've found codecov to just end up being a bit invasive compared to what I'd like to see.

Not absolute on this, so *could* consider it, or *could* consider having both local compliance plus "send the results to codecov on CI".

yeraydiazdiaz:
I agree it would be best to have it locally.

I guess it's not too bad to enforce it via pytest-cov's `--cov-fail-under=$CURRENT_COV`? We could bump the value manually ourselves as the coverage increases. A quick search hasn't yielded any obvious way of doing this automatically.

tomchristie:
A quick note on how I'd recommend we loop this all in...

Our GitHub actions have the lint stage nicely separated from the test stage, so we have `scripts/lint` and `scripts/test` steps, which helps us see at a glance which of the linting or the tests have failed.

[...]

PR #380, opened 2019-09-24
Enforce 100% test coverage

jtmiclat:
Resolves: #316 
Caveats: Currently failing as test coverage is not 100%

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -38,4 +38,4 @@ def docs(session):
 @nox.session(python=["3.6", "3.7", "3.8"])
 def test(session):
     session.install("-r", "test-requirements.txt")
-    session.run("python", "-m", "pytest", *session.posargs)
+    session.run("python", "-m", "pytest", "--cov-fail-under=100", *session.posargs)
--- a/noxfile.py
+++ b/noxfile.py
@@ -38,4 +38,8 @@ def docs(session):
 @nox.session(python=["3.6", "3.7", "3.8"])
 def test(session):
     session.install("-r", "test-requirements.txt")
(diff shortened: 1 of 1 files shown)


sethmlarson:
@jtmiclat Could you make coverage only apply to Python 3.7+? 3.6 is deliberately missing a large chunk of coverage.

JayH5:
Is it not possible to get Codecov to enforce coverage? In the past when I have used it it shows up as a CI check next to Travis but maybe it's not configured for that at the moment?

sethmlarson:
Might require @tomchristie to add a codecov to the org somehow, that's how I'd prefer to handle this as well.

jtmiclat:
@sethmlarson Just added skipping test coverage for python 3.6!

tomchristie:
I've generally switched to enforcing it alongside the tests because I found that otherwise we ended up with a large number of pull requests that didn't keep 100% coverage because devs were missing it at the point of commit.

Ideally I'd have `./scripts/test` be a comprehensive "this also oughta pass CI", so that we don't end up with lots of churn of "commit, PR, CI fail, try again".

florimondmanca:
I personally enjoy Codecov reports as well for their detailed output and visualization, but agree with Tom that we should detect insufficient coverage as early as possible. Having the tests fail locally in that case is a good way to get that feedback.

So... Can we just use both?

sethmlarson:
Btw I'm working on improving code coverage currently, will submit that change then we can merge this. :)

florimondmanca:
Doing some cleanup, and I think we're not just ready to turn on 100% coverage *yet* — still some small additions needed to reach it. I'll close this but we can revisit and see how we'd like to turn on 100% checks later, yes?

PR #386, opened 2019-09-25
Allow lists in query params

### LINKED ISSUE #366: Query params should support list-of-strings values.

florimondmanca:
Fixes #366 

Re-running the example from https://github.com/encode/httpx/issues/366#issuecomment-533786466:

```python
import httpx
import requests

URL = "https://httpbin.org/get"
PARAMS = {"documentIds": ["a55b2f894b90dd1213201471_25", "b54b2f894b90dd1213201471_9"]}

resp = requests.get(URL, params=PARAMS)
print("requests", resp.json()["args"])

r = httpx.get(URL, params=PARAMS)
json = r.json()
assert isinstance(json, dict)
print("httpx", json["args"])

print("httpx", str(httpx.QueryParams(PARAMS)))
```

```console
requests {'documentIds': ['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']}
httpx {'documentIds': ['a55b2f894b90dd1213201471_25', 'b54b2f894b90dd1213201471_9']}
httpx documentIds=a55b2f894b90dd1213201471_25&documentIds=b54b2f894b90dd1213201471_9
```

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -32,6 +32,7 @@ from .exceptions import (
 from .multipart import multipart_encode
 from .status_codes import StatusCode
 from .utils import (
+    flatten_queryparams,
     guess_json_utf,
     is_known_encoding,
     normalize_header_key,
@@ -51,7 +52,7 @@ URLTypes = typing.Union["URL", str]
 
 QueryParamTypes = typing.Union[
     "QueryParams",
-    typing.Mapping[str, PrimitiveData],
+    typing.Mapping[str, typing.Union[PrimitiveData, typing.Sequence[PrimitiveData]]],
     typing.List[typing.Tuple[str, PrimitiveData]],
     str,
 ]
@@ -311,14 +312,15 @@ class QueryParams(typing.Mapping[str, str]):
 
         value = args[0] if args else kwargs
 
+        items: typing.Sequence[typing.Tuple[str, PrimitiveData]]
         if isinstance(value, str):
             items = parse_qsl(value)
         elif isinstance(value, QueryParams):
             items = value.multi_items()
         elif isinstance(value, list):
-            items = value  # type: ignore
+            items = value
         else:
-            items = value.items()  # type: ignore
+            items = flatten_queryparams(value)
 
         self._list = [(str(k), str_query_param(v)) for k, v in items]
         self._dict = {str(k): str_query_param(v) for k, v in items}
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -1,4 +1,5 @@
 import codecs
+import collections
 import logging
 import netrc
 import os
@@ -10,6 +11,9 @@ from pathlib import Path
 from time import perf_counter
 from types import TracebackType
 from urllib.request import getproxies
+
+if typing.TYPE_CHECKING:  # pragma: no cover
+    from .models import PrimitiveData
 
 
 def normalize_header_key(value: typing.AnyStr, encoding: str = None) -> bytes:
@@ -30,7 +34,7 @@ def normalize_header_value(value: typing.AnyStr, encoding: str = None) -> bytes:
     return value.encode(encoding or "ascii")
 
 
-def str_query_param(value: typing.Optional[typing.Union[str, int, float, bool]]) -> str:
+def str_query_param(value: "PrimitiveData") -> str:
     """
     Coerce a primitive data type into a string value for query params.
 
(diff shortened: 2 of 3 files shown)


httpx/utils.py
```diff
@@ -256,6 +260,23 @@ def unquote(value: str) -> str:
...
+    items: typing.Iterable[
+        typing.Union[typing.Tuple[K, typing.Union[V, typing.Sequence[V]]]]
+    ]
+) -> typing.Iterator[typing.Tuple[K, V]]:
+    for k, v in items:
+        if isinstance(v, collections.abc.Sequence) and not isinstance(v, (str, bytes)):
+            for u in v:
+                yield (k, u)
+        else:
+            yield (k, typing.cast(V, v))
```
florimondmanca: @tomchristie Here's what the double list comprehension turned into. Let me know if this sounds reasonable to keep in this firm.

sethmlarson: I'll also note that I like the `(str, bytes)` defensiveness here, otherwise you'd get the actual integer values of the bytes which would be quite strange behavior if a user were to accidentally put bytes into a param.

httpx/models.py
```diff
@@ -51,7 +52,7 @@
 
 QueryParamTypes = typing.Union[
     "QueryParams",
-    typing.Mapping[str, PrimitiveData],
+    typing.Mapping[str, typing.Union[PrimitiveData, typing.Sequence[PrimitiveData]]],
```
tomchristie: I think we can probably be tighter on this an restrict our expected type to explicitly be `typing.List` rather than genericly be `Sequence`. There's arguments either way, but I reckon we should be okay with forcing a narrower subset where possible.

Certainly that'd be more in line with our `typing.List[typing.Tuple[str, PrimitiveData]]` case below, at any reate.

httpx/utils.py
```diff
@@ -256,6 +260,23 @@ def unquote(value: str) -> str:
...
 
 
+K = typing.TypeVar("K")
+V = typing.TypeVar("V")
+
+
+def flatten_on_values(
+    items: typing.Iterable[
+        typing.Union[typing.Tuple[K, typing.Union[V, typing.Sequence[V]]]]
+    ]
```
tomchristie: I'd probably prefer for this not to be genericly typed, and to instead explicitly be `flatten_queryparam_values`.
Right now we're only using it in one place, and it'd be clearly to me what it does if it's more explcitly focused.

I'd also rather we switch the function around slightly so that it returns a list, rather than a generator. (Marginally less elgant in the implementation, but slightly more helpful at the point-of-use)

tomchristie: A brief docstring too would make sense. As with `str_query_param` I'd *probably* prefer not to have the indirection to `PrimitiveData`, but I'm not absolute on that one.

httpx/models.py
tomchristie: Actually, having had a look at a code example with a tuple usage here, I'm unsure about Sequence vs. List. Maybe already good as it currently is.

florimondmanca: Yup, allowing for tuples as well as lists was the main reason behind using Sequence here.

httpx/utils.py
florimondmanca: Agreed that specializing this helper more is a neat idea.

florimondmanca:
@tomchristie @sethmlarson Ready for re-review! :-)

httpx/utils.py
tomchristie: Wouldn't mind a docstring on this, just "Given queryparams as a mapping, return a flat list of two-tuples representing each item, eg... *insert simple example case*"

Otherwise all look *ace* here!

florimondmanca: Docstring's in!

sethmlarson:
This is a great change, thanks @florimondmanca! :)

PR #387, opened 2019-09-25
Document supported async libraries

florimondmanca:
Following the addition of trio support via #276.

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -74,7 +74,7 @@ You can use `AsyncClient` with any of the following async libraries.
 
 ### [asyncio](https://docs.python.org/3/library/asyncio.html) (Default)
 
-By default, `AsyncClient` uses `asyncio` to perform asynchronous operations and I/O calls. It's a safe default: `asyncio` is built into the standard library and has widespread usage.
+By default, `AsyncClient` uses `asyncio` to perform asynchronous operations and I/O calls.
 
 ```python
 import asyncio
--- a/docs/async.md
+++ b/docs/async.md
@@ -64,3 +64,44 @@ The async parallel methods are:
 * `.parallel()` *Used as an "async with" context manager.*
 * `.get_response()`
 * `.next_response()`
+
+## Supported async libraries
+
+You can use `AsyncClient` with any of the following async libraries.
+
+!!! tip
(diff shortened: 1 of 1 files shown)


docs/async.md
```diff
@@ -64,3 +64,41 @@ The async parallel methods are:
...
+    async with httpx.AsyncClient() as client:
+        ...
+
+asyncio.run(main())
+```
+
+### [trio](https://github.com/python-trio/trio)
+
+To make asynchronous requests in `trio` programs, pass a `TrioBackend` to the `AsyncClient`:
+
```
sethmlarson: Should we mention anything about `python -m pip install httpx[trio]`?

florimondmanca:
@sethmlarson I added an extra admonition with installation instructions for trio; I think your comment (which it seems you removed?) was relevant.

sethmlarson:
Eh, I'm wondering if it's needed cuz it's not possible to run trio without trio installed.

florimondmanca:
Yeah, fair point. Does that mean we don't actually need the `[trio]` extra?

sethmlarson:
Probably not!

docs/async.md
```diff
@@ -64,3 +64,44 @@ The async parallel methods are:
...
+## Supported async libraries
+
+You can use `AsyncClient` with any of the following async libraries.
+
+!!! tip
+    You will typically be using `AsyncClient` in async programs that run on `asyncio`. If that's the case, or if you're not sure what this is all about, you can safely ignore this section.
+
+### [asyncio](https://docs.python.org/3/library/asyncio.html) (Default)
+
+By default, `AsyncClient` uses `asyncio` to perform asynchronous operations and I/O calls. It's a safe default: `asyncio` is built into the standard library and has widespread usage.
```
pquentin: Do we really need the "safe default" part? It's like you're telling people not to use your work on the trio backend :) My humble opinion is that the tip above is enough, and that people already using asyncio will stick to asyncio anyways.

florimondmanca: Good point, this is a relic from me trying to find a phrasing for « if you don’t care about all this, just use the default », but the introduction paragraph fits this purpose now. I’ll remove this bit.

PR #400, opened 2019-09-27
added lru cache for get_netrc_login function

### LINKED ISSUE #384: Cache NETRC logins after first lookup
As noted in #364, looking up NETRC logins is eating 25% of request time (!) — see `get_netrc_login()` below (times are absolute for a series of 1000 requests).

The `.netrc` file typically won't change across a session. So while we can't avoid paying that performance cost on the first request, we can definitely cache the NETRC credentials on the client once we looked them up (even if they don't exist).

This improvement would give us a 25% performance boost (on localhost, so ignoring network latency), getting us closer to aiohttp speeds [according to this benchmark](https://github.com/sonic182/aiosonic#aiosonic).

cansarigol:
Closes #384 

previous:
Screen Shot 2019-09-27 at 09 23 43

After lru_cache:
Screen Shot 2019-09-27 at 09 23 51

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -89,8 +89,10 @@ h2_response = h2_client.get("https://myserver.com")
 HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
 not defined, HTTPX tries to add auth into request's header from .netrc file.
 
-HTTPX has been loaded that content within the .netrc file is only checked once.
-The recommended way to refresh the cache would be to restart the interpreter.
+!!! note
+    The NETRC file is cached across requests made by a client.
+    If you need to refresh the cache (e.g. because the NETRC file has changed),
+    you should create a new client or restart the interpreter.
 
 As default `trust_env` is true. To set false:
 ```python
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -89,6 +89,9 @@ h2_response = h2_client.get("https://myserver.com")
 HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
 not defined, HTTPX tries to add auth into request's header from .netrc file.
 
+HTTPX has been loaded that content within the .netrc file is only checked once.
+The recommended way to refresh the cache would be to restart the interpreter.
+
 As default `trust_env` is true. To set false:
 ```python
 >>> httpx.get('https://example.org/', trust_env=False)
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -277,12 +277,16 @@ class BaseClient:
             )
 
         if trust_env:
-            netrc_login = get_netrc_login(request.url.authority)
+            netrc_login = self._get_netrc_login(request.url.authority)
             if netrc_login:
                 username, _, password = netrc_login
                 return BasicAuthMiddleware(username=username, password=password)
 
         return None
+
+    @functools.lru_cache(1)
+    def _get_netrc_login(self, host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
+        return get_netrc_login(host)
 
     def _dispatcher_for_request(
         self, request: AsyncRequest, proxies: typing.Dict[str, AsyncDispatcher]
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -1,6 +1,5 @@
 import functools
 import inspect
-import os
 import typing
 from types import TracebackType
 
@@ -49,7 +48,7 @@ from .models import (
     ResponseContent,
     URLTypes,
 )
-from .utils import ElapsedTimer, get_environment_proxies, get_netrc
+from .utils import ElapsedTimer, get_environment_proxies, get_netrc_login
 
 
 class BaseClient:
@@ -258,13 +257,6 @@ class BaseClient:
     def _get_auth_middleware(
         self, request: AsyncRequest, trust_env: bool, auth: AuthTypes = None
     ) -> typing.Optional[BaseMiddleware]:
-        _NETRC = None
-
-        def _get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
-            nonlocal _NETRC
-            if _NETRC is None:
-                _NETRC = get_netrc(os.getenv("NETRC", ""))
-            return _NETRC.authenticators(host)  # type: ignore
 
         if isinstance(auth, tuple):
             return BasicAuthMiddleware(username=auth[0], password=auth[1])
@@ -286,7 +278,7 @@ class BaseClient:
             )
 
         if trust_env:
-            netrc_login = _get_netrc_login(request.url.authority)
+            netrc_login = get_netrc_login(request.url.authority)
             if netrc_login:
                 username, _, password = netrc_login
                 return BasicAuthMiddleware(username=username, password=password)
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -257,7 +257,6 @@ class BaseClient:
     def _get_auth_middleware(
         self, request: AsyncRequest, trust_env: bool, auth: AuthTypes = None
     ) -> typing.Optional[BaseMiddleware]:
-
         if isinstance(auth, tuple):
             return BasicAuthMiddleware(username=auth[0], password=auth[1])
         elif isinstance(auth, BaseMiddleware):
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import os
 import typing
 from types import TracebackType
 
@@ -48,7 +49,7 @@ from .models import (
     ResponseContent,
     URLTypes,
 )
-from .utils import ElapsedTimer, get_environment_proxies, get_netrc_login
+from .utils import ElapsedTimer, get_environment_proxies, get_netrc
 
 
 class BaseClient:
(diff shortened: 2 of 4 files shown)


httpx/utils.py
```diff
@@ -94,6 +95,7 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
 NETRC_STATIC_FILES = (Path("~/.netrc"), Path("~/_netrc"))
 
 
+@lru_cache()
```
florimondmanca: Because of the `host` parameter, we're now still looking up NETRC logins once per host. Since `netrc_info` doesn't depend on `host` at all, how about moving all logic associated to looking it up in a separate function, and apply `@functools.lru_cache()` to *that* instead?

```python
import functools

@functools.lru_cache
def get_netrc():
    ...
    return netrc_info

def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
    netrc = get_netrc()  # This is only fetched on 1st call.
    return netrc.authenticators(host)  # type: ignore
```

sethmlarson: Does it make sense to use lru_cache if the values won't change?

florimondmanca: Well um, I guess the alternative is manually caching the result into a module-level variable:

```python
_NETRC = None

def get_netrc_login(host: str)
    nolocal _NETRC
    if _NETRC is None:
        _NETRC = get_netrc()
    return _NETRC.authenticators(host)
```

Pushing this idea further, what we could do instead is to do caching at the `Client` level: have a `self.netrc = None` attribute, populate it in `.get_auth_middleware()` if/when we need to, and then use `self.netrc.authenticators(host)` to get the logins.

That way, we don't magically cache anything in the utilities, and it makes everything simpler and more testable.

What do you think @cansarigol?

cansarigol:
Hi, I added with the host to cache python for each host. you don't want it because of a memory leaks?
I noticed that the same need applies to netrc env.

cansarigol:
this new get_netrc_login
Screen Shot 2019-09-28 at 09 11 04

get_netrc
Screen Shot 2019-09-28 at 09 11 24

httpx/utils.py
cansarigol: I pushed a commit about this. Is it what you want? Actually I agree with you because of get_netrc_login is used just in get_auth_middleware. More clear for me too. if it is appropriate, I will fix the tests.

httpx/utils.py
```diff
@@ -94,8 +95,9 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
 NETRC_STATIC_FILES = (Path("~/.netrc"), Path("~/_netrc"))
 
 
-def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
-    NETRC_FILES = (Path(os.getenv("NETRC", "")),) + NETRC_STATIC_FILES
+@functools.lru_cache()
+def get_netrc(netrc_env: str) -> typing.Optional[typing.Any]:
```
sethmlarson: Instead of `typing.Any` let's have `typing.Optional[netrc.netrc]` because `netrc` is actually a class.

httpx/utils.py
```diff
@@ -94,8 +95,9 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
 NETRC_STATIC_FILES = (Path("~/.netrc"), Path("~/_netrc"))
 
 
-def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
-    NETRC_FILES = (Path(os.getenv("NETRC", "")),) + NETRC_STATIC_FILES
+@functools.lru_cache()
```
sethmlarson: Let's limit this with `@functions.lru_cache(1)` so that whenever the environment variable changes we throw away the old one.

httpx/client.py
```diff
@@ -257,6 +258,14 @@ def wrap(
     def _get_auth_middleware(
         self, request: AsyncRequest, trust_env: bool, auth: AuthTypes = None
     ) -> typing.Optional[BaseMiddleware]:
+        _NETRC = None
+
+        def _get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
```
sethmlarson: Could we keep `get_netrc_login(host)` in `utils` and use `get_netrc()` within that function? That way the implementation of `get_netrc_login()` doesn't have to leak into the Client like this.

httpx/utils.py
cansarigol: Actually I tried as netrc but I couldn't :) I will try again

cansarigol:
Thanks, @sethmlarson I applied your comments.

httpx/client.py
```diff
@@ -257,6 +257,7 @@ def wrap(
     def _get_auth_middleware(
         self, request: AsyncRequest, trust_env: bool, auth: AuthTypes = None
     ) -> typing.Optional[BaseMiddleware]:
+
```
florimondmanca:
```suggestion
```

httpx/utils.py
```diff
@@ -94,8 +95,9 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
 NETRC_STATIC_FILES = (Path("~/.netrc"), Path("~/_netrc"))
 
 
-def get_netrc_login(host: str) -> typing.Optional[typing.Tuple[str, str, str]]:
-    NETRC_FILES = (Path(os.getenv("NETRC", "")),) + NETRC_STATIC_FILES
+@functools.lru_cache(1)
```
yeraydiazdiaz: More of a general question about the approach to caching. If the *contents* of the netrc file change wouldn't we be returning stale crendentials?

I guess it's not very common but it could be quite confusing for the user.

pquentin: :+1: this change probably needs to be mentioned in the documentation

sethmlarson: Totally agree we need to document this within the netrc authentication section. :) Then this looks good to merge for me.

cansarigol: ".netrc file is cached after first use by env file (not by host)"

cansarigol: Should we mentione about reset the cache if the file content is changed

sethmlarson: We should mention that once HTTPX has been loaded that content within the `netrc` file is only checked once. Recommended way to refresh the cache would be to restart the interpreter.

tomchristie: Behaviorally the *ideal* would be to have the netrc file loaded and cached on the Client instance, once and once only. (Either at instantiation, or one the first request.)

LRU caching on the host lookup, is an okay light-touch approach tho.

httpx/client.py
```diff
@@ -277,13 +277,17 @@ def _get_auth_middleware(
...
         if trust_env:
-            netrc_login = get_netrc_login(request.url.authority)
+            netrc_login = self._get_netrc_login(request.url.authority)
             if netrc_login:
                 username, _, password = netrc_login
                 return BasicAuthMiddleware(username=username, password=password)
 
         return None
 
+    @functools.lru_cache(1)
```
tomchristie: Oh neat, I'd not considered that just applying it against a method like this.

One thing to consider - we *might* want to cache `get_netrc` rather than `get_netrc_login`, so that we're only looking up the netrc info once, rather than once per host.

httpx/utils.py
```diff
@@ -95,7 +94,6 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
 NETRC_STATIC_FILES = (Path("~/.netrc"), Path("~/_netrc"))
 
 
-@functools.lru_cache(1)
 def get_netrc(netrc_env: str) -> typing.Optional[netrc.netrc]:
```
tomchristie: I'd push `netrc_env = os.getenv("NETRC", "")` inside this function, and drop the `netrc_env` argument.

cansarigol:
Hi, @tomchristie thanks for corrections.

docs/advanced.md
```diff
@@ -89,6 +89,9 @@ h2_response = h2_client.get("https://myserver.com")
 HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
 not defined, HTTPX tries to add auth into request's header from .netrc file.
 
+HTTPX has been loaded that content within the .netrc file is only checked once.
+The recommended way to refresh the cache would be to restart the interpreter.
```
florimondmanca: This needs updating now that we cache per-client; also suggesting putting this in an admonition:

```markdown
!!! note
    The NETRC file is cached across requests made by a client. If you need to refresh the cache (e.g. because the NETRC file has changed), you should create a new client or restart the interpreter.
```

httpx/client.py
```diff
@@ -277,13 +278,21 @@ def _get_auth_middleware(
...
-                username, _, password = netrc_login
-                return BasicAuthMiddleware(username=username, password=password)
+            netrc_info = self._get_netrc()
+            if netrc_info:
+                netrc_login = netrc_info.authenticators(request.url.authority)
+                if netrc_login:
+                    username, _, password = netrc_login
+                    return BasicAuthMiddleware(  # type: ignore
+                        username=username, password=password
+                    )
```
florimondmanca: I *think* the check stage in CI is failing because `.authenticators()` actually returns `[str, str, Optional[str]]` (i.e. the `password` is not guaranteed to be non-`None`).

The `# type: ignore` comment is probably misplaced, but this can also be fixed with an assert:

```python
username, _, password = netrc_login
assert password is not None
return BasicAuthMiddleware(username=username, password=password)
```

tests/test_utils.py
```diff
@@ -54,28 +54,25 @@ def test_guess_by_bom(encoding, expected):
...
 def test_get_netrc_login():
     os.environ["NETRC"] = "tests/.netrc"
-    assert get_netrc_login("netrcexample.org") == (
-        "example-username",
-        None,
-        "example-password",
-    )
+    netrc_content = str(get_netrc())
+    assert "example-username" in netrc_content
+    assert "example-password" in netrc_content
```
florimondmanca: The previous test was more precise in the assertion, I think we can stick to this style:

```python
netrc = get_netrc()
assert netrc.authenticators("netrcexample.org") == (
    "example-username", None, "example-password",
)
```

cansarigol:
@florimondmanca Thanks for the advice

sethmlarson:
Thanks @cansarigol for seeing this PR through! :)

PR #402, opened 2019-09-27
fixed for windows tests

### LINKED ISSUE #191: Test case `test_write_timeout` fails
### Summary

- Test case `test_write_timeout` fails on my environment
  + case did not raised `WriteTimeout` Exception

### Tested Environment

```
Windows 10 Pro 1809 (Build 17763.615)
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 22:20:52) [MSC v.1916 32 bit (Intel)] on win32
```

```
Package               Version
--------------------- ---------
appdirs               1.4.3
asn1crypto            0.24.0
atomicwrites          1.3.0
attrs                 19.1.0
autoflake             1.3
black                 19.3b0
brotlipy              0.7.0
certifi               2019.6.16
cffi                  1.12.3
chardet               3.0.4
Click                 7.0
colorama              0.4.1
coverage              4.5.4
cryptography          2.7
entrypoints           0.3
flake8                3.7.8
flake8-bugbear        19.3.0
flake8-comprehensions 2.1.0
h11                   0.8.1
h2                    3.1.1
hpack                 3.0.0
hstspreload           2019.7.31
htmlmin               0.1.12
hyperframe            5.2.0
idna                  2.8
importlib-metadata    0.19
isort                 4.3.21
Jinja2                2.10.1
jsmin                 2.2.2
livereload            2.6.1
Markdown              3.1.1
MarkupSafe            1.1.1
mccabe                0.6.1
mkdocs                1.0.4
mkdocs-material       4.4.0
mkdocs-minify-plugin  0.2.1
more-itertools        7.2.0
mypy                  0.720
mypy-extensions       0.4.1
packaging             19.1
pip                   19.0.3
pluggy                0.12.0
py                    1.8.0
pycodestyle           2.5.0
pycparser             2.19
pyflakes              2.1.1
Pygments              2.4.2
pymdown-extensions    6.0
pyparsing             2.4.2
pytest                5.0.1
pytest-asyncio        0.10.0
pytest-cov            2.7.1
PyYAML                5.1.2
rfc3986               1.3.2
setuptools            40.8.0
six                   1.12.0
toml                  0.10.0
tornado               6.0.3
trustme               0.5.2
typed-ast             1.4.0
typing-extensions     3.7.4
uvicorn               0.8.6
wcwidth               0.1.7
websockets            7.0
zipp                  0.5.2
```

### Logs

```
___ test_write_timeout ___

server = <uvicorn.main.Server object at 0x03FD6290>

    @pytest.mark.asyncio
    async def test_write_timeout(server):
        timeout = TimeoutConfig(write_timeout=0.000001)
    
        async with AsyncClient(timeout=timeout) as client:
            with pytest.raises(WriteTimeout):
                data = b"*" * 1024 * 1024 * 100
>               await client.put("http://127.0.0.1:8000/slow_response", data=data)
E               Failed: DID NOT RAISE <class 'httpx.exceptions.WriteTimeout'>

tests\test_timeouts.py:30: Failed
--- Captured log setup ---
INFO     uvicorn:main.py:321 Started server process [2180]
INFO     uvicorn:main.py:381 Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
--- Captured log call ---
INFO     uvicorn:h11_impl.py:443 ('127.0.0.1', 59595) - "PUT /slow_response HTTP/1.1" 200
--- Captured log teardown ---
INFO     uvicorn:main.py:418 Shutting down
INFO     uvicorn:main.py:327 Finished server process [2180]
=== 1 failed, 236 passed in 18.31 seconds ===
```

cansarigol:
Closes #191


.travis.yml
```diff
@@ -9,19 +9,74 @@ branches:
...
+    - os: windows
+      language: shell
+      python: 3.8-dev
+      env:
+        PATH=/c/Python38:/c/Python38/Scripts:$PATH
+        NOX_SESSION=test-3.8
+      before_install:
+        - choco install python --version 3.8-dev
+        - python -m pip install --upgrade pip
+      dist: bionic # Required to get OpenSSL 1.1.1+
```
tomchristie: Not wild about this amount of overhead. Any thoughts on how we could keep things more slimline?

tests/test_utils.py
```diff
@@ -179,11 +180,11 @@ def test_get_ssl_cert_file():
             {"https": "http://127.0.0.1", "http": "https://127.0.0.1"},
         ),
         (
-            {"all_proxy": "http://127.0.0.1", "ALL_PROXY": "https://1.1.1.1"},
+            {"all_proxy": "http://127.0.0.1", "_ALL_PROXY": "https://1.1.1.1"},
```
tomchristie: This change isn't clear to me?

tests/test_utils.py
```diff
@@ -179,11 +180,11 @@ def test_get_ssl_cert_file():
...
             {"all": "http://127.0.0.1"},
         ),
         (
-            {"https_proxy": "http://127.0.0.1", "HTTPS_PROXY": "https://1.1.1.1"},
+            {"https_proxy": "http://127.0.0.1", "_HTTPS_PROXY": "https://1.1.1.1"},
```
tomchristie: Same with `_ALL_PROXY` - it's not obvious what's going on here?

tests/test_utils.py
cansarigol: the reason is case insensitive problem. `all_proxy` env is affected by upper case `ALL_PROXY`. Because of that, I added an underscore.

tests/test_utils.py
cansarigol: due to same reason.

tests/test_utils.py
sethmlarson: That's the reason this test is here, to verify that lowercase is preferred over uppercase.  On Windows is that not true?

cansarigol: os.environ is case-insensitive on windows os.  in `os.environ`, there is just `ALL_PROXY`
linux:
```
>>>os.environ['all_proxy']
'http://127.0.0.1'
>>>os.environ['ALL_PROXY']
'https://1.1.1.1'
```

windows:
```
>>>os.environ['all_proxy']
'https://1.1.1.1'
>>>os.environ['ALL_PROXY']
'https://1.1.1.1'
```

cansarigol: We can use `nt.envrion` but have to change `urllib.request.getproxies`

tomchristie: > to verify that lowercase is preferred over uppercase

Do we want that behavior in any case? I'm slightly surprised that we support lowercased envvars (on case sensitive platforms) since the convention in these cases is usually UPPERCASED_NAMES, no?

sethmlarson: @tomchristie This is another case of "we support it because Requests (and curl) supports it".

tomchristie: Okay, that's fair enough. I don't think we *neccessarily* need to specify or test which is prefered, tho I'd probably tend to uppercase first.

(Setting both variants is a broken behaviour that's going to lead to uncertainty regardless of anything we specify)

cansarigol:
After merge this (if it suits) I will create an issue that contains fix `xfail`s about `AsyncioBackend` on `win32`

florimondmanca:
@cansarigol It seems to me this PR does at least three different things:

- Add Windows to the build matrix.
- Skip tests that fail on Windows.
- Remove lowercased environment variables test cases from the proxies tests.

For this reason, while I appreciate the initiative, I think this PR is not very workable right now, and we should definitely split things up more.

Here’s my intuition for a plan:

I think the last point should be submitted separately.

As for Windows, I think we should first add *one* Windows job (3.7?) to the build matrix, and allow it to fail. (There’s an example of how to do this in the config file; were already doing it for 3.8.) Then we can incrementally submit PRs that fix tests that are broken on Windows (ideally one at a time?).

What do you think @cansarigol?

cansarigol:
very nice and clear for me. I'm separating these as per this pattern.

tests/test_utils.py
```diff
@@ -131,9 +132,9 @@ def test_get_ssl_cert_file():
     assert get_ca_bundle_from_env() == "tests"
 
     del os.environ["SSL_CERT_DIR"]
-    os.environ["SSL_CERT_FILE"] = "tests/test_utils.py"
+    os.environ["SSL_CERT_FILE"] = str(Path("tests", "test_utils.py"))
     # SSL_CERT_FILE is correctly set, SSL_CERT_DIR is not set.
-    assert get_ca_bundle_from_env() == "tests/test_utils.py"
+    assert get_ca_bundle_from_env() == str(Path("tests", "test_utils.py"))
```
florimondmanca: Nice catch on these ones. I think the slightly more idiomatic usage would be:

```suggestion
    assert get_ca_bundle_from_env() == str(Path("tests") / "test_utils.py")
```

tests/test_timeouts.py
```diff
@@ -21,6 +24,11 @@
 
 
 async def test_write_timeout(server, backend):
+    # FIX: win32 doesn't raise asyncio.TimeoutError
+    # in 'asyncio.wait_for(writer.drain(), timeout.write_timeout)'
+    if sys.platform == "win32":
+        if isinstance(backend, AsyncioBackend):
+            pytest.skip()
```
florimondmanca: If this test is expected to fail, would `xfail` be better suited here?

We can also add the comment content into the `reason` so that it is reported at the end of the test session by pytest.

```suggestion
            pytest.xfail(
                "'asyncio.wait_for(writer.drain())' does not raise a 'TimeoutError' on Windows."
            )
```

tests/test_concurrency.py
```diff
@@ -9,6 +9,10 @@
     sys.version_info < (3, 7),
     reason="Requires Python 3.7+ for AbstractEventLoop.start_tls()",
 )
+#  FIX: Why doesn't return result
+@pytest.mark.xfail(
+    sys.platform == "win32", reason="StreamReader.read doesn't return result on Win32"
+)
```
florimondmanca: A few questions:

- What is happening exactly? Is `.read()` just hanging forever? We should try to make the reason as explicit as possible.
- Is this affecting all Python versions, or e.g. just 3.8?

Also, I don't think `xfail` supports conditions, so we should move this inside the test

```python
async def test_start_tls_on_socket_stream(https_server):
    if sys.platform == "win32":
        pytest.xfail("StreamReader.read doesn't return result on win32")
```

cansarigol: the connection is dropped and doesn't raise an error. return data is `b''`. I added these as comment and removed `xfail` for entire test.

cansarigol: > Is this affecting all Python versions, or e.g. just 3.8?

I will test this

florimondmanca: > the connection is dropped and doesn't raise an error. return data is b''.

Actually, that sounds like the bug that's affecting us and making the build fail on 3.8 currently: https://github.com/encode/httpx/pull/423#discussion_r330871311

cansarigol: btw I removed 3.8 on Windows in other PR #457

cansarigol: > > the connection is dropped and doesn't raise an error. return data is b''.
> 
> Actually, that sounds like the bug that's affecting us and making the build fail on 3.8 currently: [#423 (comment)](https://github.com/encode/httpx/pull/423#discussion_r330871311)

I tested this on 3.7 on Windows.

tomchristie:
Closing this as I'm assuming it's rather out of date at this point.
It'd be great to have our tests all sorted out for Windows tho, so would def. welcome any further pull requests here.

PR #406, opened 2019-09-27
Get test suite back to 100% line coverage

sethmlarson:
Getting our test suite back up to 100% so that #380 can land.


sethmlarson:
We're still not at 100%, especially because we have the Stream compatibility for Python 3.8 now.

So unless we want to mark all that code as `pragma: no cover` and only enforce coverage on Python 3.8 it looks like Codecov may be our only option.

tests/dispatch/test_http2.py
```diff
@@ -204,3 +204,11 @@ def test_http2_reconnect():
     for setting_code, changed_setting in settings.changed_settings.items():
         assert isinstance(changed_setting, h2.settings.ChangedSetting)
         assert changed_setting.new_value == expected_settings[setting_code]
+
+
+async def test_http2_live_request(backend):
+    async with AsyncClient(backend=backend) as client:
+        resp = await client.get("https://nghttp2.org/httpbin/anything")
```
yeraydiazdiaz: Could https://github.com/encode/httpx/pull/412 help with test?

tests/client/test_proxies.py
```diff
@@ -46,3 +53,66 @@ def test_proxies_has_same_properties_as_dispatch():
...
+    client = httpx.AsyncClient(proxies=proxies)
+    request = httpx.AsyncRequest("GET", url)
+    dispatcher = client._dispatcher_for_request(request, client.proxies)
+
+    if expected is None:
+        assert isinstance(dispatcher, httpx.ConnectionPool)
+        assert dispatcher is client.dispatch
+    else:
+        assert isinstance(dispatcher, httpx.HTTPProxy)
+        assert dispatcher.proxy_url == expected
```
yeraydiazdiaz: Nice 👍

tests/client/test_auth.py
```diff
@@ -91,9 +91,6 @@ def challenge_send(self, request: AsyncRequest) -> AsyncResponse:
         ]
         return AsyncResponse(401, headers=headers, content=b"", request=request)
 
-    def reset(self) -> None:
-        self._response_count = 0
-
```
yeraydiazdiaz: That one's my fault, thanks 👍

tests/dispatch/test_http2.py
sethmlarson: Definitely, it's a lot further out than we want this work completed though. Hopefully soon!

florimondmanca:
Great work on this one @sethmlarson 👏

PR #407, opened 2019-09-27
Expose ASGIDispatch & WSGIDispatch in the 'dispatch' namespace.

### LINKED ISSUE #398: ASGIDispatch and WSGIDispatch should be importable from the top-level httpx package
From #396:

> ``ASGIDispatch`` and ``WSGIDispatch`` are documented as top-level but aren't exposed at the top level. This is definitely an issue, I'd recommend the route of making both available top-level.

mxahmed:
Resolves #398

`ASGIDispatch` and `WSGIDispatch` were added to `dispatch` namespace first, then imported from there in the top-level package.
This way they're available at both `httpx.ASGIDispatch` and `httpx.dispatch.ASGIDispatch` which might be handy.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -53,7 +53,7 @@ For example:
 
 ```python
 # Instantiate a client that makes WSGI requests with a client IP of "1.2.3.4".
-dispatch = httpx.WSGIDispatch(app=app, remote_addr="1.2.3.4")
+dispatch = httpx.dispatch.WSGIDispatch(app=app, remote_addr="1.2.3.4")
 client = httpx.Client(dispatch=dispatch)
 ```
 
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -19,6 +19,7 @@ from .config import (
     TimeoutTypes,
     VerifyTypes,
 )
+from .dispatch import ASGIDispatch, WSGIDispatch
 from .dispatch.base import AsyncDispatcher, Dispatcher
 from .dispatch.connection import HTTPConnection
 from .dispatch.connection_pool import ConnectionPool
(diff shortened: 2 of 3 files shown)


florimondmanca:
Thanks @mxahmed!

> both `httpx.ASGIDispatch` and `httpx.dispatch.ASGIDispatch`

In the light of "there should be one — and preferably only one — obvious way to do it", I think we should choose between those two possibilities so that our users don't need to.

I think the preferred option suggested in #398 is `httpx.dispatch.<...>`? (I know @sethmlarson suggested to add top-level import but on second thought is it worth adding a top-level import we don't currently have anyway, knowing we'll remove it when refactoring to using namespaces?)

mxahmed:
> I think the preferred option suggested in #398 is httpx.dispatch.<...>? (I know @sethmlarson suggested to add top-level import but on second thought is it worth adding a top-level import we don't currently have anyway, knowing we'll remove it when refactoring to using namespaces?)

I believe this is a valid point and to choose the `httpx.dispatch.<...>` route will be more consistent and avoids any breaking changes for the user if a move to a more specific namespaces approach in the future is planned.

it's also clearer for the user to initiate a custom dispatcher form `httpx.dispatch.<...>` than `httpx.<....>`

sethmlarson:
If we go that route let's fix the examples in the documentation as well. :)

tomchristie:
> I believe this is a valid point and to choose the httpx.dispatch.<...> route will be more consistent and avoids any breaking changes for the user if a move to a more specific namespaces approach in the future is planned.

This seems like a good call to me too. Particularly given that `Client(app=...)` is the more common usage for pointing at an ASGI/WSGI app, and that the API of *explicitly* setting the dispatcher is an advanced usage.

PR #410, opened 2019-09-28
Fix wrongly deleted params introduced #408

jtmiclat:
I deleted the wrong parameters in #408 . Sorry about that!


florimondmanca:
Indeed, and CI tests in #408 were failing because of that. Thanks for sending a fix!

sethmlarson:
Thanks for catching this @jtmiclat! I should have seen it during review :)

Issue #290, opened 2019-08-28
Hypercorn test server


pgjones:
Hello,

Would you be receptive to me adding Hypercorn as a test server (alongside Uvicorn)? I think for this project it would be good to test HTTP/2 against a production server. Whilst for Hypercorn it would be great to know that it doesn't upset this client.

florimondmanca:
We might benefit from this in #276 as well.

Let’s first refactor the current server setup as mentioned in https://github.com/encode/httpx/pull/276#issuecomment-525724384, and then perhaps consider this?

pgjones:
ok, sounds good.

florimondmanca:
Hmm... Sorry that my previous message didn’t sound as excited as I am about this idea!

If you want to get together a PR demonstrating the idea it would be great, no need to wait for the refactor I mentioned to be done — you may have to deal with conflicts, but that’s about it. Thanks for suggesting this!

florimondmanca:
@pgjones FYI, the test server refactoring I mentioned earlier is in. :+1:

tomchristie:
If we're able to run the tests against both hypercorn & uvicorn, then that seems like it might be a nice bit of progress yeah.

tomchristie:
It's possible that I've been overzealous wrt. "it'd be easier to manage for `encode` if we're using an `encode` server". I'm wondering if there's actually blockers there, of if that's just me being closed-minded.

* Would there be any issues wrt. differing supported Python versions that could make this problematic for us?
* If we *did* want to do this, I'd rather we just switched away from uvicorn completely.
* I guess if possible it might be sensible for the majority of tests to just run against a server instance that's setup in a seperate thread and that is available *throughout* the entire test suite lifetime. (?)

florimondmanca:
> a server instance that's setup in a seperate thread and that is available throughout the entire test suite lifetime

FWIW that's already what we've got — up to 4 servers, i.e. one for {HTTP, HTTPS} x {TCP, UDS}, running in their own thread and available throughout the test `session`. (We *might* be able to reduce to 2 by having servers support both HTTP *and* HTTPS, though?)

pgjones:
If you do wish to use Hypercorn you can have a single instance serve HTTP, HTTPS, TCP, UDP, h11, h2, h3 at once.

tomchristie:
@pgjones Good to know, thanks.

I guess the main criteria we'd want for figuring if this'd be a good idea would be:

[...]

PR #412, opened 2019-09-29
Introduce Hypercorn as a test HTTP/2 server

pgjones:
This allows the HTTP/2 tests to be directed at an actual HTTP/2 server
rather than a mock server. This is beneficial as it is closer to the
actual usage by users.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,7 +14,7 @@ combine_as_imports = True
 force_grid_wrap = 0
 include_trailing_comma = True
 known_first_party = httpx,httpxprof,tests
-known_third_party = brotli,certifi,chardet,click,cryptography,h11,h2,hstspreload,nox,pytest,requests,rfc3986,setuptools,tqdm,trio,trustme,uvicorn
+known_third_party = brotli,certifi,chardet,click,cryptography,h11,h2,hstspreload,hypercorn,nox,pytest,requests,rfc3986,setuptools,tqdm,trio,trustme,uvicorn
 line_length = 88
 multi_line_output = 3
 
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -8,6 +8,7 @@ flake8
 flake8-bugbear
 flake8-comprehensions
 flake8-pie
+hypercorn
 isort
 mypy
 pytest
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,14 @@
 import asyncio
+import functools
 import json
 import os
 import threading
 import time
 import typing
+import sys
 
 import pytest
+import trio
 import trustme
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.serialization import (
@@ -14,6 +17,8 @@ from cryptography.hazmat.primitives.serialization import (
     PrivateFormat,
     load_pem_private_key,
 )
+import hypercorn.config
+import hypercorn.trio
 from uvicorn.config import Config
 from uvicorn.main import Server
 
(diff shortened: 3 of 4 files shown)


tests/dispatch/test_http2.py
```diff
@@ -1,51 +1,108 @@
...
+        headers = [(b"content-length", b"%d" % len(content))]
+        await send({"type": "http.response.start", "status": 200, "headers": headers})
+        await send({"type": "http.response.body", "body": content, "more_body": False})
+    else:
+        raise Exception()
+
+
+@pytest.fixture(name="h2_server", scope="module")
+def _h2_server(cert_pem_file, cert_private_key_file):
+    bind = "0.0.0.0:8543"
```
pgjones: This is one problem - whilst Hypercorn can bind to a random available port (0.0.0.0:0) there isn't an easy way (unless you use Trio) to figure out which port it has bound to. Actually I think I can just switch this to Trio, right?

tests/dispatch/test_http2.py
```diff
@@ -1,51 +1,108 @@
...
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+        loop.run_until_complete(serve(app, config, shutdown_trigger=_shutdown_trigger))
+
+    thread = threading.Thread(target=_run)
+    thread.start()
+    import time
+
+    time.sleep(0.5)
```
pgjones: Related to the random port problem - will try Trio.

tests/dispatch/test_http2.py
```diff
@@ -1,51 +1,108 @@
...
+    yield f"https://{bind}"
+    shutdown_event.set()
+    thread.join()
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 7), reason="Hypercorn requires python3.7 or higher"
```
pgjones: Not sure if you think this must run in 3.6... Hypercorn dropped support when the SSL handshake vulnerability was found.

tests/dispatch/test_http2.py
```diff
@@ -133,16 +201,27 @@ def test_http2_multiple_requests():
...
     """
     If a connection has been dropped between requests, then we should
     be seemlessly reconnected.
     """
-    backend = MockHTTP2Backend(app=app)
+    backend = MockHTTP2Backend(app=mock_app)
 
     with Client(backend=backend) as client:
         response_1 = client.get("http://example.org/1")
-        backend.server.close_connection = True
```
pgjones: Woops, will restore.

pgjones: I think using the mock backend here is probably better than Hypercorn as it is hard to make the latter drop a connection from the outside.

pgjones:
I think it is at a good stage for feedback.

setup.cfg
```diff
@@ -11,7 +11,7 @@ combine_as_imports = True
 force_grid_wrap = 0
 include_trailing_comma = True
 known_first_party = httpx,httpxprof,tests
-known_third_party = brotli,certifi,chardet,click,cryptography,h11,h2,hstspreload,nox,pytest,requests,rfc3986,setuptools,tqdm,trio,trustme,uvicorn
+known_third_party = brotli,certifi,chardet,click,cryptography,h11,h2,hstspreload,hypercorn,nox,pytest,requests,rfc3986,setuptools,tqdm,trio,trustme,uvicorn
```
sethmlarson: We can remove uvicorn from this list?

test-requirements.txt
```diff
@@ -7,6 +7,7 @@ cryptography
 flake8
 flake8-bugbear
 flake8-comprehensions
+hypercorn
 isort
 mypy
```
sethmlarson: Can we remove uvicorn from this list too?

tests/dispatch/test_http2.py
```diff
@@ -1,51 +1,108 @@
...
     assert response.status_code == 200
     assert json.loads(response.content) == {"method": "GET", "path": "/", "body": ""}
 
 
-async def test_async_http2_get_request(backend):
-    backend = MockHTTP2Backend(app=app, backend=backend)
-
-    async with AsyncClient(backend=backend) as client:
-        response = await client.get("http://example.org")
+@pytest.mark.skipif(
```
sethmlarson: Can the `h2_server` raise the skip within the fixture?

tomchristie:
From a POV of managing encode as a whole this one is a mixed bag.

I'd personally much rather look at adding HTTP/2 support to uvicorn, than switch to hypercorn as the test server. It's also problematic for us to switch to something that has a different set of Python requirements, (3.7+) rather than being able to keep roughly in line across a suite of projects.

I'm *way* more invested in promoting ecosystem-as-a-whole rather than individual projects, but just wrt. logisitics and management overhead, this change probably comes out as a net negative to me.

setup.cfg
pgjones: Uvicorn is still used as a test server - for the HTTP/1.1 parts. Is the expectation that Hypercorn replaces Uvicorn?

pgjones:
> ...than switch to hypercorn as the test server. 

Hi @tomchristie this adds Hypercorn as **a** test server rather than replacing uvicorn as **the** test server, as per #290. In its current form it only uses Hypercorn for the HTTP/2 tests.

pgjones:
@florimondmanca thanks. I'm happy to tidy this up - I just got it into a workable state to get initial thoughts.

pgjones:
Thanks @florimondmanca I've pulled in your changes and rebased against master.

tests/conftest.py
```diff
@@ -288,3 +293,83 @@ def https_server(cert_pem_file, cert_private_key_file):
...
+
+        trio.run(main)
+
+
+@pytest.fixture(scope=SERVER_SCOPE)
+def h2_server(
+    cert_pem_file: str, cert_private_key_file: str
+) -> typing.Iterator[H2Server]:
+    if sys.version_info < (3, 7):
+        pytest.skip(reason="Hypercorn requires Python 3.7 or higher")
```
JayH5: The reason kwarg is for `pytest.mark.skip`/`xfail`. For `pytest.skip` the message is just the first argument for some reason 😕

florimondmanca:
While I appreciate a lot the effort put into getting something (almost!) working for this one, this PR has gone a bit stale because opinions are mixed about eventually integrating it into master. I'll close it, and maybe we can take inspiration from this for a future implementation (maybe when Uvicorn gets HTTP/2?). Thanks @pgjones! ❤️

PR #416, opened 2019-10-01
setup.py: add AsyncIO and Trio frameworks

pquentin:
This should help to discover httpx from PyPI. We can always remove the Trio framework when moving its support to a third-party package.


florimondmanca:
The build failed because the streams API reversion on 3.8 seems to have landed. :)

tomchristie:
I think as a rule of thumb we need to have a hard stop on accepting PRs if the tests are failing (even if it's an unrelated cause) 😇

Issue #391, opened 2019-09-26
Host header contains userinfo component


sethmlarson:
`BaseRequest.prepare()` incorrectly sets the default `Host` header to `userinfo@example.com:123` if given the following URL: `http://userinfo@example.com:123`.

We should be removing the userinfo component from the authority before assigning to `Host`.

tomchristie:
Do we want a `URL.host_with_port` property, or something similar?

Are there any prior art examples of URL libraries giving an explicit naming to the host+explicit-port-if-it-exists portion of the authority string?

sethmlarson:
host_with_port sounds more relevant for the CONNECT case.

I've seen "hostname" used for host and port if it's non-default.

tomchristie:
Reviewing [stdlib](https://docs.python.org/3/library/urllib.parse.html), [furl](https://github.com/gruns/furl), [purl](https://github.com/codeinthehole/purl), [hyperlink](https://hyperlink.readthedocs.io/en/latest/index.html), I don't see any prior art (and stdlib uses `hostname` as just the `host` component.)

We might want to switch our `port` property to return an optional port, rather than a port-or-default, which would be helpful here.

Potentially also some thinking to do here wrt. bytewise vs. str representations. Eg. I'm not sure we've completely nailed down some very definitive "this URL property returns an hostname string" vs. "this URL property returns a bytewise encoded hostname".

(In this case we want the plain bytewise representation)

Similarly wrt. path representations and escaping.

sethmlarson:
You're right, was misremembering the stdlib. Totally agree with the port not returning a value via inference and then have a method/property to resolve the port with defaults applied.

PR #417, opened 2019-10-01
Don't include username/password components in `Host` header

cansarigol:
Closes: #391


httpx/models.py
```diff
@@ -613,7 +614,32 @@ def prepare(self) -> None:
         has_connection = "connection" in self.headers
 
         if not has_host:
-            auto_headers.append((b"host", self.url.authority.encode("ascii")))
+            _params: typing.Dict[str, None] = {}
+            if self.url._uri_reference.userinfo:
```
tomchristie: We shouldn't need to reach into the private `_uri_reference` from anywhere outside of the `URL` class. If we're doing that anywhere then we ought to treat it as a signal that we're missing some required bit of interface on the `URL` class.

cansarigol: yes but I didn't want to perform it every time regarding performance.

tomchristie: Gotcha, so...

* We should probably add a `.userinfo` property to the `URL` class.
* We should then use something like...

```python
url = self.url
if url.userinfo:
    url = url.copy_with(username=None, password=None)
...
```

Right?

tomchristie:
I don't think we want this issue to be "remove auth and port from host header". It's just the auth (username/password) that we want to remove, right? If there's a port component then I think we still want that included.

cansarigol:
as per issue, I thought that we don't want to port in the host. Besides that, I checked `stdlib` as you mentioned.
```
>>> from urllib.parse import urlparse
>>> o = urlparse('http://www.cwi.nl:80/%7Eguido/Python.html')
>>> o.hostname
'www.cwi.nl'
```

httpx/models.py
cansarigol: to copy `userinfo`, added `copy_with_urlparse` Is it appropriate?

httpx/models.py
```diff
@@ -613,7 +620,16 @@ def prepare(self) -> None:
         has_connection = "connection" in self.headers
 
         if not has_host:
-            auto_headers.append((b"host", self.url.authority.encode("ascii")))
+            url = self.url
+            if url.userinfo:
+                url = url.copy_with_urlparse(userinfo=None)
+                for userinfo in ("username", "password"):
```
tomchristie: We don't want to add "username" and "password" headers to the request.
I think we can drop all of this "for" block.

httpx/models.py
```diff
@@ -613,7 +620,16 @@ def prepare(self) -> None:
         has_connection = "connection" in self.headers
 
         if not has_host:
-            auto_headers.append((b"host", self.url.authority.encode("ascii")))
+            url = self.url
+            if url.userinfo:
+                url = url.copy_with_urlparse(userinfo=None)
```
tomchristie: I'd rather we use the existing `.copy_with(username=None, password=None)`, instead of adding a new method.

I *do also* think we should also open a seperate issue to ensure that we support the *full* range of URL keyword arguments on `copy_with()`, so that eg. we support `.copy_with(userinfo=...)`, `copy_with(host=...)`, and other variants that `rfc3986` doesn't support directly.

httpx/models.py
cansarigol: Ah sorry, I applied like stdlib
>>> o = urlparse('http://uname@www.cwi.nl:80/%7Eguido/Python.html')
>>> o.username
'uname'

httpx/models.py
cansarigol: > I'd rather we use the existing `.copy_with(username=None, password=None)`, instead of adding a new method.
`URIReference` copy_with doesn't allow us to use `username` and `password` arguments. we can only use these: - scheme, authority, path, query, fragment

tomchristie: Good point, yeah. Okay in that case I'd suggest that we've got a neccessary precursor to this pull request...

A *seperate pull request* that adds support for using `username, `password`, `hostname`, and `port` in our `copy_with(...)` method. There's an example of how we'd do that [in the Starlette codebase, here](https://github.com/encode/starlette/blob/6fa0038c699d2ac06f240d438263a4ae94e8996d/starlette/datastructures.py#L98-L119). (Tho replace "netloc" with "authority" in our case.)

Once we've got that in, then we can tackle this PR without having to add a new method to the API, which we don't really want to do.

Does that make sense / seem reasonable?

Thanks for digging through all this!

cansarigol: Hi, @tomchristie I've pushed a new PR. After merge I will pursue from here

tomchristie:
Retitled slightly to better reflect the requirements for this one. (Because we don't want to remove the port)

tomchristie:
Fantastic, yup.

cansarigol:
Thanks for your help

Issue #395, opened 2019-09-26
Same interface for for all http methods in api.py


jtmiclat:
When using elasticsearch with `httpx`, I quickly realized that `httpx.get` does not take in `json` as a function argument.  A quick look at `api.py`  showed that not all methods have the same interface. Specifically, `get`/`options`/`head have fewer arguments than the others.

There are some arguments about that these methods should not send data. But I think as an HTTP client, users should not be restricted this way.  I suggest that all methods in `api.py` should have the same interface. This also would solve compatibility with `requests` as it accepts json parameter in all its methods.

sethmlarson:
Thanks for opening this issue, definitely an oversight!

We should fix the discrepancies and write a unit test using `inspect.Signature` to ensure this issue doesn't come up again.

StephenBrown2:
I would counter that this is not particularly an oversight, but a deliberate choice based on the capabilities of HTTP, and also matching the `requests` [api](https://github.com/psf/requests/blob/master/requests/api.py)

Though the underlying `request` function does allow for arbitrary `data`/`json`/etc parameters, the others are more restricted based on the common expected use cases.

I would recommend referring to [httpx.request](https://github.com/encode/httpx/blob/master/httpx/api.py#L18) for similar uses, that go outside the standard expectations for [HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) [methods](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods).

thebigmunch:
> and also matching the `requests` [api](https://github.com/psf/requests/blob/master/requests/api.py)

Actually, as the OP pointed out, httpx's behavior does not match the requests API. The requests API uses ``:param **kwargs: Optional arguments that ``request`` takes.`` for all the top level API functions, so, in turn, they all accept ``data``/``json``/etc.

That being said, I actually agree with you that those functions should, in fact, not have params for things they don't actually support. And I agree with your suggestion of using ``httpx.request`` for those situations.

sethmlarson:
@StephenBrown2 @thebigmunch You've both got good points, keeping our API trimmed in areas where parameters typically aren't found or have no meaning is a good way to steer our users to doing the right thing.

It looks like the `.delete()` function has the `data`, `files`, and `json` parameters, are those appropriate for the `DELETE` method?

[...]

PR #418, opened 2019-10-01
Add compatibility differences with get, delete, head, options

jtmiclat:
Add compatibility differences with get, delete, head, and options.  

Relevant discussion: #395


tomchristie:
I think(?) we actually have the same API as requests here, except in the `delete` case.
Where we disallow `files`/`data`/`json`, but `requests does not.

jtmiclat:
@tomchristie requests actually allowed you to pass in the `json`, `files`, and `data` in all methods. It passed them as part of  `kwargs` to `api.request`. While for httpx, it was intentionallity not allowed.

Relevant link:
https://github.com/psf/requests/blob/master/requests/api.py

tomchristie:
Ah yeah, so it does. Thanks for confirming!

sethmlarson:
:tada: Thanks @jtmiclat! :tada:

Issue #420, opened 2019-10-01
Revert PR #369


sethmlarson:
See: https://github.com/encode/httpx/pull/369

This is currently failing our CI on Python 3.8 which was just released. While we're at it we can switch to 3.8 on Travis instead of 3.8-dev

florimondmanca:
Funny — I thought I opened an issue for this this morning. Guess I forgot to press the send button? 😅 Thanks for tracking this!

tomchristie:
Thought for the future: We ought mark any `-dev` Python builds as "allowed failure" given that API can change under our feet, until there's a fixed release to work against.

Somewhat surprised to see such a big API change land without warning, immediatly prior to the release candidate release, but there we go.

PR #421, opened 2019-10-01
Make HTTPError importable from the top-level

### LINKED ISSUE #367: HTTPError should be importable frop the top-level httpx package
From #365:

> `HTTPError` is not available at the top level like the other exceptions and like it was in requests. This is a somewhat common exception to catch raise_for_status, so having to add another import statement for it is a bit odd if intentional.

Put differently, `HTTPError` is missing from `httpx/__init__.py`. Adding it would allow us to do:

```python
import httpx

try:
    r = httpx.get("https://example.org")
    r.raise_for_status()
except httpx.HTTPError:
    ...
```

jcugat:
Fixes #367

I've updated the only two tests that used `HTTPError` to use the top-level package too. I did not find any reference in the documentation.


jcugat:
The travis failure seems unrelated to my changes, might be due to #420

sethmlarson:
Yep this looks good, thanks for opening this! :)

PR #423, opened 2019-10-01
Revert PR #369

### LINKED ISSUE #420: Revert PR #369

jcugat:
In my previous PR I found out that tests are failing in travis with python 3.8 due to #369.

This PR reverts it, and also updates the python version in travis to use `3.8` instead of `3.8-dev`.

Fixes #420


jcugat:
Ooops, it seems `3.8` is not available yet in travis, keeping the `-dev` for now: https://docs.travis-ci.com/user/languages/python/#python-versions

florimondmanca:
The build initially failed because of the `stream.read()` output was not the one expected, we should now expect it to fail because of a bug in `importlib.metadata` (used by pytest) that seems to have passed through 3.8.0.rc1. See: https://bugs.python.org/issue38342

(Also, is it me or 3.8 Travis builds are super slow? Probably due to the latest release not being pre-built on their side yet.)

tomchristie:
I'd also okay with us dropping 3.8 from the build matrix / marking as allowed failure until the *actual* final release is made, if that helps.

tests/test_concurrency.py
```diff
@@ -25,11 +25,11 @@
...
-        assert stream.stream.get_extra_info("cipher", default=None) is None
+        assert stream.stream_writer.get_extra_info("cipher", default=None) is None
 
         stream = await backend.start_tls(stream, https_server.url.host, ctx, timeout)
         assert stream.is_connection_dropped() is False
-        assert stream.stream.get_extra_info("cipher", default=None) is not None
+        assert stream.stream_writer.get_extra_info("cipher", default=None) is not None
 
         await stream.write(b"GET / HTTP/1.1\r\n\r\n")
         assert (await stream.read(8192, timeout)).startswith(b"HTTP/1.1 200 OK\r\n")
```
florimondmanca: I debugged the failing test on 3.8, which was related to receiving empty bytes at this point. This seems to indicate the connection was dropped after the wirte, and by expanding assertions into…

```python
        await stream.write(b"GET / HTTP/1.1\r\n\r\n")
        assert stream.is_connection_dropped() is False
        data = await stream.read(8192, timeout)
        assert data
        assert data.startswith(b"HTTP/1.1 200 OK\r\n")
```

the first assertion is indeed failing: `stream.is_connection_dropped()` is actually `True`.

florimondmanca: Commenting all the other lines above in the `try:` block at least passes that assert, but then we still receive `b""` on the `.read()` call.

No idea what's happening. Might be a bug in 3.8rc1?

tomchristie:
Great - thanks!
We've at least got the history of this work, so we can reassess when Python 3.9 is in the works, or whatever.

PR #425, opened 2019-10-02
Skip Python 3.8-dev builds on Travis

sethmlarson:
Taken from [this comment](https://github.com/encode/httpx/pull/423#issuecomment-537455052).
Until we get a stable 3.8 release we'll say passing 3.8 is optional.


sethmlarson:
Yeah that was a big change at the last minute. We'll re-enable 3.8 after the release candidate. :)

PR #427, opened 2019-10-02
Create laba4

herusann:


tomchristie:
Thanks @sethmlarson - I’ve started blocking users if they:

* Make a completely incoherent drive-by PR
* Don’t have any history that validates them as a genuine user account.

I’ve seen a few like this and they’re indistinguishable from how an attempted bot account setup might look.

PR #436, opened 2019-10-03
added authority copy feature in URL.copy_with

cansarigol:
Closes: #437 

As mentioned in #417 from now on we can copy `username`,  `password`, `host`, `port` and create a new URL from old url instance. After merging this PR, we can apply this to the source PR.


tomchristie:
Great, yup!

PR #440, opened 2019-10-04
Fix link to parallel page from async page

nwalsh1995:
Current page 404s since the href is incorrect.

https://www.encode.io/httpx/async/ contains a link to 'paralell requests' that points to https://www.encode.io/parallel/


sethmlarson:
Thanks @nwalsh1995! :tada:

PR #442, opened 2019-10-04
use declarative metadata

graingert:


setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+
+[options]
+python_requires = >=3.6
+packages = find:
+include_package_data = True
+zip_safe = False
```
graingert: pretty sure this package is zip_safe. (it's not hard to be, you just use importlib.resources for package data)

setup.cfg
```diff
@@ -1,3 +1,50 @@
+[metadata]
+name = httpx
+version = attr: src.get_version
+url = https://github.com/encode/httpx
+license = BSD
+description = The next generation HTTP client.
+long_description = file: README.md, file: CHANGELOG.md
```
graingert: I believe this means you don't need a MANIFEST.in anymore

src.py
```diff
@@ -0,0 +1,9 @@
+import pathlib
+import re
+
+PACKAGE = "httpx"
+
+
+def get_version(package=PACKAGE):
+    version = pathlib.Path(package, "__version__.py").read_text()
+    return re.search("__version__ = ['\"]([^'\"]+)['\"]", version).group(1)
```
cansarigol: Should we move this part into `httpx/httpx/__version__.py`

graingert: No because you can't import httpx until after setup

setup.py
```diff
@@ -1,76 +1,2 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-
-import re
-from pathlib import Path
-
 from setuptools import setup
```
graingert:
```suggestion
from setuptools import setup

```

setup.cfg
florimondmanca: Unfortunately we need this flag — required by PEP 561 to include the py.typed marker used by type checkers for detecting type annotations.

graingert: https://github.com/ethanhs/pep-561/blob/master/setup.py this example doesn't define zip safe

setup.cfg
```diff
@@ -1,3 +1,50 @@
+[metadata]
+name = httpx
+version = attr: src.get_version
```
graingert: Imho this should be a file (httpx/version.txt) and we use importlib_resources in `__version__` to grab it

setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+url = https://github.com/encode/httpx
+license = BSD
```
graingert: Strictly you doing need license if there's a trove available

setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+description = The next generation HTTP client.
```
graingert: This is duplicated from `httpx.__version__` because this key doesn't support `attr:` if people are ok with me adding importlib_resources I'll change it to a `file:`

setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+    Framework :: Trio
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+
+[options]
+python_requires = >=3.6
+packages = find:
+include_package_data = True
```
graingert: From my reading of the spec you don't need this if you specify `[options.package_data]`

setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+description = The next generation HTTP client.
+long_description = file: README.md, file: CHANGELOG.md
+long_description_content_type = text/markdown
+author = Tom Christie
+author_email = tom@tomchristie.com
+classifiers =
+    Development Status :: 3 - Alpha
+    Environment :: Web Environment
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
```
graingert: This license is wrong. This package is licensed under the BSD 3 clause

graingert: https://github.com/pypa/warehouse/issues/2996

src.py
sethmlarson: Honestly this one edge-case has me not wanting to merge this. :(

graingert:
This also fixes #193 

Here it is showing that py.typed ends up in the sdist and whl https://asciinema.org/a/6HXD5BEhcKtetzesXuCY8Fkex

src.py
graingert: `setup.py` already does not import `http.__version__.__version__` because of this limitation. It's not an edge case

setup.cfg
graingert: https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages

> If you use setuptools, you must pass the option zip_safe=False to setup(), or mypy will not be able to find the installed package.

setup.cfg
graingert:
```suggestion
license = BSD-3-Clause
```

setup.cfg
graingert:
```suggestion
```

src.py
sethmlarson: I meant having this file to accomplish only this one task.

sethmlarson:
So I don't think we should do this unless we want to move to flit+pyproject.toml (which I'd be a fan of, idk about what others think @florimondmanca, @tomchristie?)

florimondmanca:
Although this is a very interesting change (I didn't know setuptools supported file-based declaration), I think that's also the exact reason why this should *not* get in.

All other Encode projects use `setup.py` to define package metadata. It's a straight-forward, battle-tested approach, and I think changing the way we do things *just in HTTPX* breaks the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and tends to increase maintenance overhead.

So, I agree with @sethmlarson and would be in favor of closing this.

I'd be open to see how switching to a `pyproject.toml`-based tool (flit, poetry?) would look like as well, but honestly I think the current setup works *just fine*.

sethmlarson:
Gonna go ahead and close this, thank you for putting this together! Would love if you went through all the setuptools options we're missing like issue tracker, docs, etc.

PR #446, opened 2019-10-04
Fix flaky elapsed delay test

florimondmanca:
These test were spurriously failing on my machine, e.g. giving "0.063" which failed 0.05 ± 0.01.

I tweaked them so that repeating the tests 30 times does not fail:

```console
$ pip install pytest-repeat
$ pytest --count=30 -k test_elapsed_delay
```


cansarigol:
I got this error when I tried to run tests on Windows os. I had changed it as `abs=0.2`. I will try this.

PR #447, opened 2019-10-04
Don't check trio import in definition of backend fixture params

florimondmanca:
Small refactoring based on the observation that trio is always installed during tests.


sethmlarson:
Lint issue then you can merge :)

sethmlarson:
:tada: Thanks @florimondmanca!

PR #448, opened 2019-10-05
Remove wheels package from test-requirements.txt

jcugat:
As [discussed here](https://github.com/encode/httpx/pull/439#discussion_r331390607) removed `wheel` package from `test-requirements.txt` and moved the check in `./scripts/publish`.

### DIFF
--- a/scripts/publish
+++ b/scripts/publish
@@ -15,6 +15,12 @@ if ! command -v "${PREFIX}twine" &>/dev/null ; then
     exit 1
 fi
 
+if ! ${PREFIX}pip show wheel &>/dev/null ; then
+    echo "Unable to find the 'wheel' command."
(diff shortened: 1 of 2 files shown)


scripts/publish
```diff
@@ -15,6 +15,12 @@ if ! command -v "${PREFIX}twine" &>/dev/null ; then
     exit 1
 fi
 
+if ! ${PREFIX}pip show wheel &>/dev/null ; then
```
jcugat: The check is slightly different from the rest since in MacOS for example `/usr/local/bin/wheel` exists, but the build process was failing because the virtualenv did not have the package installed.

sethmlarson: Thanks for the clarification!

PR #450, opened 2019-10-05
Allow serving docs via nox

florimondmanca:
Quick addition to our `nox` setup to allow serving docs using `nox` instead of having to install `mkdocs` and `mkdocs-material` locally.

(Not a fan of the snake_cased command name, but both "docs:watch" and "docs-watch" messed with the unselection of the task when running `$ nox`.)

(Edit: edited to `$ nox -s serve`.)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -102,17 +102,12 @@ $ nox -s test -- tests/test_multipart.py
 
 ## Documenting
 
-To work with the documentation, make sure you have `mkdocs` and
-`mkdocs-material` installed on your environment:
+Documentation pages are located under the `docs/` folder.
+
+To run the documentation site locally (useful for previewing changes), use:
 
 ```shell
-$ pip install mkdocs mkdocs-material
-```
-
-To spawn the docs server run:
-
-```shell
-$ mkdocs serve
+$ nox -s serve
 ```
 
 ## Releasing
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,6 +2,7 @@ import nox
 
 nox.options.stop_on_first_error = True
 nox.options.reuse_existing_virtualenvs = True
+nox.options.keywords = "not serve"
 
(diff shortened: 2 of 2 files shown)


florimondmanca:
So, I tried `name="docs-watch"` but for some reason I can't make `nox` ignore it correctly when running `$ nox`:

```console
$ nox -l
Sessions defined in /Users/florimond/Developer/python-projects/httpx/noxfile.py:

* lint
* check
- docs
* docs-watch
* test-3.6
* test-3.7
* test-3.8

sessions marked with * are selected, sessions marked with - are skipped.
```

I would have expected `docs-watch` to be skipped instead of `docs`.

This is probably a bug in `nox` worth reporting?

Anyway, I can switch to `serve` as a workaround (and probably a nicer, shorter session name after all). :)

sethmlarson:
Hmm I don't know enough about what's going on here to see if the above is a bug or not.. cc: @theacodes @dhermes

sethmlarson:
Hmm I don't know enough about what's going on here to see if the above is a bug or not.. cc: @theacodes @dhermes

theacodes:
keywords actually work by parsing expressions as python, same as `pytest`. So `snake_case` should be used.

There's actually a slighty different way you can accomplish this. Use one `docs` session and branch on `session.interactive`:

```python
@nox.session(reuse_venv=True)
def docs(session):
    session.install("mkdocs", "mkdocs-material")

    if session.interactive:
        session.run("mkdocs", "serve")
    else:
        session.run("mkdocs", "build")
```

When run locally, it'll serve the docs. When run on your CI, it'll run just "build".

See http://nox.thea.codes/en/stable/config.html#nox.sessions.Session.interactive

florimondmanca:
Nice tip @theacodes, thanks. Unfortunately what I was aiming for here is to skip the `serve` by default, and `interactive` isn't `False` when e.g. running `$ nox` locally.

I updated the PR to use `nox -s serve` instead. @sethmlarson Looking good?

PR #399, opened 2019-09-27
Add hints for debugging CI issues

sethmlarson:
Closes #311

PR #451, opened 2019-10-05
Add test script

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/399#discussion_r329471921

Running

```bash
$ scripts/test
```

runs the test suite and runs the `check` session.

I made it so that the test suite only for the current Python version is run (e.g. 3.7), but we could also run `$ nox -s test` so that all available Python versions are tested (e.g. all of 3.6, 3.7 and 3.8 if they're detected as available by `nox`). This is mostly because running tests against all Python versions can get very long, esp. because virtualenvs aren't reused. Thoughts?

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -100,6 +100,14 @@ You can also run a single test script like this:
 $ nox -s test -- tests/test_multipart.py
 ```
 
+Lastly, to ensure you're on track to pass the CI build, run:
+
+```shell
+$ scripts/test
+```
+
+This command is a light wrapper around `nox` that will run code style checks and test the code against all installed Python versions.
+
 ## Documenting
 
 Documentation pages are located under the `docs/` folder.
--- a/scripts/test
+++ b/scripts/test
@@ -1,0 +1,6 @@
+#!/bin/sh -e
+
+set -x
+
+nox -s check
+nox -s test


florimondmanca:
I'm also wondering — what should we do about the contributing docs, since they currently recommend using `$ nox -s test`?

pquentin:
Would it help to configure nox to reuse virtual environments? There are various places where it can be done: https://nox.thea.codes/en/stable/config.html

florimondmanca:
@pquentin We actually reuse virtualenvs in all other nox sessions, but not in `test`. I think reusing would be OK in that case too though. We can always `--no-reuse-existing-virtualenvs` on the command line. I'll submit a separate PR for that. :)

docs/contributing.md
```diff
@@ -100,6 +100,15 @@ You can also run a single test script like this:
 $ nox -s test -- tests/test_multipart.py
 ```
 
+Lastly, to ensure you're on track to pass the CI build, run:
+
+```shell
+# Note: you may need to run `$ chmod +x scripts/test` once beforehand.
```
tomchristie: We shouldn't need to do that.
I think git handles file permissions, so make sure they're set correctly in this PR.

florimondmanca: Oh, cool. I’ll try to check them in then. :)

florimondmanca: I didn't know Git checked in file permissions — turns out I had already got them into this PR! I removed this comment.

PR #453, opened 2019-10-06
Issue a new connection if HTTP/2 stream IDs expire.

### LINKED ISSUE #283: Open new HTTP/2 connections when stream space is exhausted.
http://http2.github.io/http2-spec/index.html#StreamIdentifiers "Long-lived connections can result in an endpoint exhausting the available range of stream identifiers. A client that is unable to establish a new stream identifier can establish a new connection for new streams. "

Somewhat related to #282

gdhameeja:
Fixes #283 
Create new HTTPConnection if stream ids expire. The making of new connection is done in ConnectionPool instead of in HTTPConnection.

### DIFF
--- a/httpx/dispatch/connection_pool.py
+++ b/httpx/dispatch/connection_pool.py
@@ -126,7 +126,6 @@ class ConnectionPool(AsyncDispatcher):
             self.max_connections.release()
             raise exc
         except NewConnectionRequired:
-            # raised when stream ids expire
             return await self.send(
                 request=request, verify=verify, cert=cert, timeout=timeout
             )
--- a/httpx/dispatch/connection_pool.py
+++ b/httpx/dispatch/connection_pool.py
@@ -1,7 +1,5 @@
 import typing
 
-from h2.exceptions import NoAvailableStreamIDError
-
 from ..concurrency.asyncio import AsyncioBackend
 from ..concurrency.base import ConcurrencyBackend
 from ..config import (
@@ -13,6 +11,7 @@ from ..config import (
     TimeoutTypes,
     VerifyTypes,
 )
+from ..exceptions import NewConnectionRequired
 from ..models import AsyncRequest, AsyncResponse, Origin
 from ..utils import get_logger
 from .base import AsyncDispatcher
@@ -126,13 +125,11 @@ class ConnectionPool(AsyncDispatcher):
             self.active_connections.remove(connection)
             self.max_connections.release()
             raise exc
-        except NoAvailableStreamIDError:
-            # NoAvailableStreamIDError propogates from http2.HTTP2Connection
+        except NewConnectionRequired:
             # raised when stream ids expire
-            logger.debug("HTTP2 connection streams maxed out")
-            connection = await self.acquire_connection(origin=request.url.origin)
-            response = await connection.send(
-                request, verify=verify, cert=cert, timeout=timeout)
+            return await self.send(
+                request=request, verify=verify, cert=cert, timeout=timeout
+            )
         return response
 
     async def acquire_connection(self, origin: Origin) -> HTTPConnection:
--- a/httpx/dispatch/connection_pool.py
+++ b/httpx/dispatch/connection_pool.py
@@ -1,5 +1,7 @@
 import typing
 
+from h2.exceptions import NoAvailableStreamIDError
+
 from ..concurrency.asyncio import AsyncioBackend
 from ..concurrency.base import ConcurrencyBackend
 from ..config import (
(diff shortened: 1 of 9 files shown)


gdhameeja:
@florimondmanca Here is the new PR for #283 where connections are being opened in the connection pool. Although the test doesn't work even now. After settings up the `MockHTTP2Server` how do I make requests to it with the `ConnectionPool`?

yeraydiazdiaz:
Hi @gdhameeja, thanks for your PR.

Your branch includes a very large `TAGS` file which I presume you accidentally commited in that last merging of master onto your branch.

I suggest creating a new branch from master and applying those changes to ensure a clean PR.

florimondmanca:
An alternative to revert without creating a new PR is to edit the commit:

```console
$ rm -rf TAGS
$ git add -A
$ git commit --amend --no-edit
$ git push --force
```

Seems like `TAGS` is an emacs/vim thing ([source](https://www.emacswiki.org/emacs/TagsFile)). Should we add it to `.gitignore`?

sethmlarson:
No, editor specific files should be in your own personal gitignore file.
@gdhameeja I recommend looking at this: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

httpx/dispatch/connection_pool.py
```diff
@@ -124,7 +126,13 @@ def num_connections(self) -> int:
             self.active_connections.remove(connection)
             self.max_connections.release()
             raise exc
-
+        except NoAvailableStreamIDError:
```
tomchristie: I think it'd be nice for us to *not* expose `h2.NoAvailableStreamIDError` to the connection pool,
but rather have an `httpx` exception, such as `ReconnectRequired`.

We're not exposing any other `h2` library implementation details at this layer, and it'd be nice to keep in line with that. (And I can imagine that there could be other circumstances where we'd want the connection to request a reconnect & retry.)

httpx/dispatch/connection_pool.py
```diff
@@ -124,7 +126,13 @@ def num_connections(self) -> int:
...
+            # NoAvailableStreamIDError propogates from http2.HTTP2Connection
+            # raised when stream ids expire
+            logger.debug("HTTP2 connection streams maxed out")
+            connection = await self.acquire_connection(origin=request.url.origin)
+            response = await connection.send(
```
tomchristie: This line needs to be wrapped with the same "except BaseException as exc" behavior that we've got around the initial `connection.send()`. It's not obvious-at-a-glance to me what the easiest way of doing that is.

tomchristie:
"An alternative to revert without creating a new PR is to edit the commit"

We squash and merge all PRs, so I don't think there's any need to remove it from the history. Just git delete it and we're good with keeping the rest of the pull request as it is.

httpx/dispatch/connection_pool.py
gdhameeja: @tomchristie Made the changes as requested. Could you please review if it looks okay? Also my test still doesn't work. I have no idea how to make request to the `MockHTTP2Server` I setup in my test. Thanks.

gdhameeja:
Awaiting review.

florimondmanca:
@gdhameeja The test is failing because `MockHTTP2Server` doesn't have an `url` — see the CI build logs.

This is looking okay to me *in principle*, but the mock test isn't convincing me that this will effectively allow HTTPX to handle the case *in the wild*. Have you thought of a way we could (manually?) test this against a live server?

httpx/dispatch/connection_pool.py
```diff
@@ -124,7 +125,11 @@ def num_connections(self) -> int:
             self.active_connections.remove(connection)
             self.max_connections.release()
             raise exc
-
+        except NewConnectionRequired:
+            # raised when stream ids expire
```
florimondmanca: We shouldn't let this HTTP/2 detail surface out here — it's probably OK to remove this comment.

```suggestion
```

gdhameeja:
@florimondmanca Currently I am unable to think of a way to test this manually. I might be wrong here but someone mentioned that we can have 2^31 labelled stream ids. With that condition, I think it'd be hard to test this manually.

gdhameeja:
@florimondmanca @sethmlarson @tomchristie Awaiting review.

httpx/dispatch/connection_pool.py
```diff
@@ -124,7 +125,10 @@ def num_connections(self) -> int:
             self.active_connections.remove(connection)
             self.max_connections.release()
             raise exc
-
+        except NewConnectionRequired:
+            return await self.send(
+                request=request, verify=verify, cert=cert, timeout=timeout
+            )
```
florimondmanca: It seems this block will actually never run, because it is shadowed by the catch-all block for `BaseException` — we need to move it above the one for `BaseException`.

tests/dispatch/test_http2.py
```diff
@@ -212,3 +213,18 @@ def test_http2_reconnect():
...
         assert resp.http_version == "HTTP/2"
+
+
+async def test_http2_connection_renewed(backend):
+    """Test new connection is formed when stream ids expire"""
+    server = MockHTTP2Server(app=app, backend=backend)
+    server.conn.HIGHEST_ALLOWED_STREAM_ID = 1
+
+    async with httpx.ConnectionPool(backend=backend) as http:
+        response = await http.request("GET", server.url)
```
florimondmanca: The mock server has no `.url` attribute, which makes the build fail. I think you can use anything here, since we don't make real requests, so e.g. `"/"`.

florimondmanca:
Sorry to see that this PR has be dragging a bit, I'll close it for housekeeping purposes but we can obviously revisit later with a clean slate. Thanks!

PR #456, opened 2019-10-06
Removed lowercased environment variables test cases from the proxies tests.

cansarigol:
as per comment in #402 , separated PR for lowercased environment variables

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -175,17 +175,10 @@ async def test_elapsed_timer():
         ({}, {}),
         ({"HTTP_PROXY": "http://127.0.0.1"}, {"http": "http://127.0.0.1"}),
         (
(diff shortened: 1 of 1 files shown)


sethmlarson:
These environment variables are standard to also accept the lowercase versions. See curl and requests. I think we should maintain the behavior.

tomchristie:
Given `curl` supports the lowercase variant I'm okay with us having both.
The only slight possible behavioral difference is that I'd also okay with us *underspecifying* a precedence. (Ie. If you're specifying both then we're not going to give you any definitive answer as to which you'd get.) The tests currently expect lowercase>uppercase, which I don't think is something we need to assert for.

florimondmanca:
So, what's the status on this? Should we re-add a test case that verifies lowercased environment variables are supported? (But not re-introduce the old tests which we asserting on which case is preferred.)

PR #457, opened 2019-10-06
Added win32 to the build matrix

cansarigol:
as per comment in #402 , separated PR for win32 tests in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ matrix:
   # Should still investigate failures periodically for new APIs.
   fast_finish: true
   allow_failures:
+    # FIX
+    # Some tests not yet resolved for Windows. (In progress)
     - os: windows
     # Allow builds to fail when using '-dev' Python versions.
     # We should still investigate failures periodically for new APIs.
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ matrix:
       env: NOX_SESSION=test-3.6
     - python: 3.7
       env: NOX_SESSION=test-3.7
-
     - python: 3.8-dev
       env: NOX_SESSION=test-3.8
       dist: bionic # Required to get OpenSSL 1.1.1+
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,6 @@ matrix:
         - choco install python --version 3.7
         - python -m pip install --upgrade pip
 
-  # Allow builds to fail when using '-dev' Python versions
-  # Should still investigate failures periodically for new APIs.
   fast_finish: true
   allow_failures:
     # FIX
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,17 @@ branches:
 
 matrix:
   include:
-    - os: linux
-      python: 3.7
+    - python: 3.7
       env: NOX_SESSION=check
-    - os: linux
-      python: 3.7
+    - python: 3.7
       env: NOX_SESSION=docs
 
-    - os: linux
-      python: 3.6
+    - python: 3.6
       env: NOX_SESSION=test-3.6
-    - os: linux
-      python: 3.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,24 +9,78 @@ branches:
...
+
+    - os: windows
+      language: shell
+      python: 3.8-dev
+      env:
+        PATH=/c/Python38:/c/Python38/Scripts:$PATH
+        NOX_SESSION=test-3.8
+      before_install:
+        - choco install python --pre
+        - python -m pip install --upgrade pip
```
florimondmanca: Could we restrict these to only the `test-3.7` session? Check and docs aren't really dependent on the OS, and not running 3.6 and 3.8-dev is a good compromise to keeping the build light enough while having basic Windows coverage.

.travis.yml
```diff
@@ -9,24 +9,78 @@ branches:
...
+    - os: linux
+      python: 3.6
       env: NOX_SESSION=test-3.6
-    - python: 3.7
+    - os: linux
+      python: 3.7
       env: NOX_SESSION=test-3.7
-    - python: 3.8-dev
+
+    - os: linux
```
florimondmanca: Is specifying `os: linux` actually required? I believe it is the default on Travis CI.

.travis.yml
```diff
@@ -9,24 +9,78 @@ branches:
...
+        NOX_SESSION=test-3.8
+      before_install:
+        - choco install python --pre
+        - python -m pip install --upgrade pip
   # Allow builds to fail when using '-dev' Python versions
   # Should still investigate failures periodically for new APIs.
   fast_finish: true
   allow_failures:
     - python: 3.8-dev
+    - os: windows
```
florimondmanca: Nice. :+1:

.travis.yml
```diff
@@ -9,24 +9,78 @@ branches:
...
+      language: shell
+      python: 3.8-dev
+      env:
+        PATH=/c/Python38:/c/Python38/Scripts:$PATH
+        NOX_SESSION=test-3.8
+      before_install:
+        - choco install python --pre
+        - python -m pip install --upgrade pip
   # Allow builds to fail when using '-dev' Python versions
   # Should still investigate failures periodically for new APIs.
```
florimondmanca: Should we move this comment so that it's clearer that it targets `- python: 3.8-dev`?

```yaml
fast_finish: true
allow_failures:
  - os: windows
  # Allow builds to fail when using '-dev' Python versions.
  # We should still investigate failures periodically for new APIs.
  - python: 3.8-dev
```

tomchristie:
So is the plan to get this in first, and then follow it up by getting the tests to pass on windows, and removing the "allow failure"?

florimondmanca:
@tomchristie Yup. Sounds reasonable?

tomchristie:
I'm okay with that, yeah. Generally a decent pattern. I guess we *could* include a comment in the `.travis.yml` file in the meantime. e.g. "# Some tests not yet resolved for Windows. In progress." or whatever.

.travis.yml
```diff
@@ -22,10 +22,25 @@ matrix:
...
+
   # Allow builds to fail when using '-dev' Python versions
   # Should still investigate failures periodically for new APIs.
   fast_finish: true
   allow_failures:
+    # FIX
+    # Some tests not yet resolved for Windows. (In progress)
+    - os: windows
+    # Allow builds to fail when using '-dev' Python versions.
+    # We should still investigate failures periodically for new APIs.
```
florimondmanca: Seems to me that this comment is now duplicated with the one just above `fast_finish: true`?

cansarigol: removed other lines

florimondmanca:
Thanks @cansarigol!

cansarigol:
Thanks

PR #460, opened 2019-10-06
Make nox always reuse virtualenvs by default

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/451#issuecomment-538735714

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -100,13 +100,6 @@ You can also run a single test script like this:
 $ nox -s test -- tests/test_multipart.py
 ```
 
-!!! tip
-    `nox` is configured to [reuse existing virtual environments](https://nox.thea.codes/en/stable/usage.html#re-using-virtualenvs). To force tests to run in a clean virtual environment (e.g. because dependencies have changed), use:
-
-    ```shell
-    $ nox -s test --no-reuse-existing-virtualenvs
-    ```
-
 ## Documenting
 
 To work with the documentation, make sure you have `mkdocs` and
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -22,3 +22,4 @@ nav:
 markdown_extensions:
   - admonition
   - codehilite
+  - pymdownx.superfences
```
florimondmanca: This bit and `pymdown-extensions` is required to be able to write code blocks in admonitions. :)

docs/contributing.md
```diff
@@ -100,6 +100,13 @@ You can also run a single test script like this:
 $ nox -s test -- tests/test_multipart.py
 ```
 
+!!! tip
+    `nox` is configured to [reuse existing virtual environments](https://nox.thea.codes/en/stable/usage.html#re-using-virtualenvs). To force tests to run in a clean virtual environment (e.g. because dependencies have changed), use:
```
sethmlarson: We can add `--upgrade` to `session.install()` steps to ensure we're getting latest dependencies even in resused venvs.

florimondmanca: Good idea, yup. I pushed a commit with this change. :)

mkdocs.yml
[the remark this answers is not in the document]
florimondmanca: Woops. I used to need it for the nested code block in the updated contributing guide, which in the end didn’t make it. So yep, this is technically superfluous wrt this PR. (I’ll argue it’s a nice addition though, as it was blocking us from writing code blocks in admonitions.)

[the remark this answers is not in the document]
tomchristie: Okay. Ideally let's defer until we've got somewhere we needed. (Tho it's a small issue so not overly fussed about it in the meantime)

Issue #360, opened 2019-09-19
Add support for the NO_PROXY environment variable


sethmlarson:
The `NO_PROXY` environment variable is described here:

- http://www.gnu.org/software/wget/manual/html_node/Proxies.html
- https://github.com/psf/requests/blob/840540b6b1f07ef87faab73392c03fbef0dcc9fe/requests/utils.py#L395

In order to accomplish this task we'd need:
- Implementing NO_PROXY, a list of hosts that will bypass the proxy if found within the list.
- Only enabled if `trust_env=True`
- Documentation updates for the new environment variable.
- Test cases for all code branches.

elwinmartin:
@sethmlarson Hi Seth! 

I would love to work on this if I could! 

I'm mostly familiar with this library as a user but I think the context provided should be clear enough to adapt this for httpx.

sethmlarson:
That would be awesome!! I'll assign the issue to you :)

If you have questions or doubts feel free to reach out here and we'll try our best to help.

tomchristie:
Looks good.
Related to #404. I wonder if [cURL would be a good baseline set of environment variables](https://curl.haxx.se/libcurl/c/libcurl-env.html) to adhere too? That's probably more robust and mature than any other tooling out there, so I'd be okay with us *explicitly* following their lead.

mariaS210:
@sethmlarson hello! is someone working on this? if not, i'd like to try it

florimondmanca:
Feel free to go ahead @mariaS210! 🎉

yeraydiazdiaz:
Closed via https://github.com/encode/httpx/pull/471

tomchristie:
Reopening this, since it's evidently been closed in error - #471 adds `NO_PROXY` handling utility functions, but doesn't tie it in to the client.

pawamoy:
Is there another way to bypass the proxies while waiting for `NO_PROXY` to be tied into the client?

tomchristie:
You can use `trust_env=False` to disable system proxies.

https://www.python-httpx.org/environment_variables/

evanlurvey:
I have addressed this in https://github.com/encode/httpx/pull/835

evanlurvey:
@florimondmanca Would you want to close this issue now?

PR #462, opened 2019-10-08
Add support for the NO_PROXY environment variable #360

mariaS210:
This PR is to implement issue #360 to  add support for NO_PROXY environment variable.

Added documentation string to environment_variables.md
Added option to pass no_proxy in Client
Added tests, hopefully for all branches (I might have been a bit lost in codecov)

Please let me know if I missed anything, and suggest improvements.
Thank you

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -129,3 +129,17 @@ export HTTP_PROXY=http://127.0.0.1:3080
 # This request will be sent through the proxy
 python -c "import httpx; httpx.get('http://example.com')"
 ```
+
+`NO_PROXY`
+----------------------------------------
+
+Valid values: A comma separated list of hosts.
+
+This variable should contain a comma-separated list of domain extensions proxy should not be used for.
+
+```bash
+export NO_PROXY=example.com,localhost
+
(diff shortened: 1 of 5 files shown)


sethmlarson:
This code is different if we were to move the proxy resolution into an `HTTPProxy` instance that wraps a `ConnectionPool` rather than doing it's own dispatch. I have some changes locally that might want to land before this actually, it'll be easier to implement.

sethmlarson:
Maybe you should change this PR into logic that takes a URL and then decides `True`/`False` whether that URL should be proxied. Then I can re-use that component. :)

mariaS210:
> This code is different if we were to move the proxy resolution into an `HTTPProxy` instance that wraps a `ConnectionPool` rather than doing it's own dispatch. I have some changes locally that might want to land before this actually, it'll be easier to implement.

Should I wait with this PR until you push your local changes, then see how to integrate the logic  into HTTPProxy?

sethmlarson:
@mariaS210 If you'd like you can open a separate PR implementing the component I've described above then come back to this PR later. :)

sethmlarson:
Closed in favor of #472.

PR #463, opened 2019-10-08
Unify timeout behaviour

### LINKED ISSUE #433: Passing 'timeout=None' should always result in no timeout

```python
import httpx

# (1)

# (2)

# (3)
client = httpx.Client()
```

---

```python
# models.py
UNSET = object()

# api.py
    ...
```

jcugat:
Fixes #433

Took a first stab at it. There are a few tests missing, but wanted to push it early to start getting some feedback.

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -8,9 +8,18 @@ import certifi
 from .__version__ import __version__
 from .utils import get_ca_bundle_from_env, get_logger
 
+
+class UnsetType:
+    pass
+
+
+UNSET = UnsetType()
+
 CertTypes = typing.Union[str, typing.Tuple[str, str], typing.Tuple[str, str, str]]
 VerifyTypes = typing.Union[str, bool, ssl.SSLContext]
-TimeoutTypes = typing.Union[float, typing.Tuple[float, float, float], "TimeoutConfig"]
+TimeoutTypes = typing.Union[
+    float, typing.Tuple[float, float, float], "TimeoutConfig", UnsetType
+]
 HTTPVersionTypes = typing.Union[
     str, typing.List[str], typing.Tuple[str], "HTTPVersionConfig"
 ]
@@ -241,6 +250,10 @@ class TimeoutConfig:
                 self.connect_timeout = timeout.connect_timeout
                 self.read_timeout = timeout.read_timeout
                 self.write_timeout = timeout.write_timeout
+            elif isinstance(timeout, UnsetType):
+                self.connect_timeout = DEFAULT_TIMEOUT_CONFIG.connect_timeout
+                self.read_timeout = DEFAULT_TIMEOUT_CONFIG.read_timeout
+                self.write_timeout = DEFAULT_TIMEOUT_CONFIG.write_timeout
             elif isinstance(timeout, tuple):
                 self.connect_timeout = timeout[0]
                 self.read_timeout = timeout[1]
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -233,19 +233,31 @@ class TimeoutConfig:
         self,
         timeout: TimeoutTypes = None,
         *,
-        connect_timeout: float = None,
-        read_timeout: float = None,
-        write_timeout: float = None,
+        connect_timeout: typing.Union[float, UnsetType, None] = UNSET,
+        read_timeout: typing.Union[float, UnsetType, None] = UNSET,
+        write_timeout: typing.Union[float, UnsetType, None] = UNSET,
     ):
         if timeout is None:
-            self.connect_timeout = connect_timeout
-            self.read_timeout = read_timeout
-            self.write_timeout = write_timeout
+            self.connect_timeout: typing.Optional[float] = (
+                connect_timeout
+                if not isinstance(connect_timeout, UnsetType)
+                else DEFAULT_TIMEOUT_CONFIG.connect_timeout
+            )
+            self.read_timeout: typing.Optional[float] = (
+                read_timeout
+                if not isinstance(read_timeout, UnsetType)
+                else DEFAULT_TIMEOUT_CONFIG.read_timeout
+            )
+            self.write_timeout: typing.Optional[float] = (
+                write_timeout
+                if not isinstance(write_timeout, UnsetType)
+                else DEFAULT_TIMEOUT_CONFIG.write_timeout
+            )
         else:
             # Specified as a single timeout value
-            assert connect_timeout is None
-            assert read_timeout is None
-            assert write_timeout is None
+            assert connect_timeout is UNSET
+            assert read_timeout is UNSET
+            assert write_timeout is UNSET
             if isinstance(timeout, TimeoutConfig):
                 self.connect_timeout = timeout.connect_timeout
                 self.read_timeout = timeout.read_timeout
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -75,7 +75,7 @@ class HTTPConnection(AsyncDispatcher):
         timeout: TimeoutTypes = UNSET,
     ) -> None:
         ssl = self.ssl.with_overrides(verify=verify, cert=cert)
-        timeout = TimeoutConfig(timeout if timeout is not UNSET else self.timeout)
+        timeout = self.timeout if timeout is UNSET else TimeoutConfig(timeout)
 
         host = self.origin.host
         port = self.origin.port
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -5,7 +5,7 @@ import typing
 from ..concurrency.asyncio import AsyncioBackend
 from ..concurrency.base import ConcurrencyBackend
 from ..config import (
-    DEFAULT_TIMEOUT_CONFIG,
+    UNSET,
     CertTypes,
     HTTPVersionConfig,
     HTTPVersionTypes,
@@ -34,7 +34,7 @@ class HTTPConnection(AsyncDispatcher):
         verify: VerifyTypes = True,
         cert: CertTypes = None,
         trust_env: bool = None,
-        timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
+        timeout: TimeoutTypes = UNSET,
         http_versions: HTTPVersionTypes = None,
         backend: ConcurrencyBackend = None,
         release_func: typing.Optional[ReleaseCallback] = None,
@@ -53,8 +53,10 @@ class HTTPConnection(AsyncDispatcher):
         request: AsyncRequest,
         verify: VerifyTypes = None,
         cert: CertTypes = None,
-        timeout: TimeoutTypes = None,
+        timeout: TimeoutTypes = UNSET,
     ) -> AsyncResponse:
+        timeout = timeout if timeout is not UNSET else self.timeout
+
         if self.h11_connection is None and self.h2_connection is None:
             await self.connect(verify=verify, cert=cert, timeout=timeout)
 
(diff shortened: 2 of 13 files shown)


httpx/config.py
```diff
@@ -8,9 +8,18 @@
 from .__version__ import __version__
 from .utils import get_ca_bundle_from_env, get_logger
 
+
+class UnsetType:
+    pass
+
+
+UNSET = UnsetType()
```
jcugat: Had to put this here instead of the model package due to a circular dependency.

httpx/config.py
```diff
@@ -8,9 +8,18 @@
 from .__version__ import __version__
 from .utils import get_ca_bundle_from_env, get_logger
 
+
+class UnsetType:
```
jcugat: Unfortunately can't use this pattern

```py
UnsetType = NewType('UnsetType', object)
UNSET = UnsetType(object())
```

because `NewType` does not allow `isinstance()` on it, and it was needed otherwise mypy was complaining in `TimeoutConfig`.

httpx/dispatch/connection_pool.py
```diff
@@ -113,9 +113,13 @@ def num_connections(self) -> int:
...
         verify: VerifyTypes = None,
         cert: CertTypes = None,
-        timeout: TimeoutTypes = None,
+        timeout: TimeoutTypes = UNSET,
     ) -> AsyncResponse:
-        connection = await self.acquire_connection(origin=request.url.origin)
+        timeout = timeout if timeout is not UNSET else self.timeout
+
+        connection = await self.acquire_connection(
+            origin=request.url.origin, timeout=timeout
```
jcugat: Slight change from the previous code, now we use the provided `timeout` for acquiring the connection too.

httpx/config.py
florimondmanca: Yup, well done here.

From Python 3.8, `typing.Literal` would allow to solve this quite elegantly:

```python
import typing

UNSET = object()
MyCustomType = typing.Union[bool, typing.Literal[UNSET]]
```

Not sure that there'll be a backport to 3.6 and 3.7 unfortunately, but just thought it was interesting to share. :)

httpx/config.py
```diff
@@ -241,6 +250,10 @@ def __init__(
                 self.connect_timeout = timeout.connect_timeout
                 self.read_timeout = timeout.read_timeout
                 self.write_timeout = timeout.write_timeout
+            elif isinstance(timeout, UnsetType):
```
florimondmanca: Is there any reason (maybe type hints-related?) we can't use `elif timeout is UNSET` here?

httpx/config.py
```diff
@@ -241,6 +250,10 @@ def __init__(
                 self.connect_timeout = timeout.connect_timeout
                 self.read_timeout = timeout.read_timeout
                 self.write_timeout = timeout.write_timeout
```
florimondmanca: I can't comment on the lines that aren't in the diff, but shouldn't the first `if timeout is None` branch also draw default values from `DEFAULT_TIMEOUT_CONFIG`? Which seems to hint like we should also use `UNSET` as defaults here… because we might want to explicitly pass e.g. `connect_timeout=None` while using the defaults for other values?

httpx/dispatch/connection.py
```diff
@@ -70,10 +72,10 @@ def __init__(
         self,
         verify: VerifyTypes = None,
         cert: CertTypes = None,
-        timeout: TimeoutTypes = None,
+        timeout: TimeoutTypes = UNSET,
     ) -> None:
         ssl = self.ssl.with_overrides(verify=verify, cert=cert)
-        timeout = self.timeout if timeout is None else TimeoutConfig(timeout)
+        timeout = TimeoutConfig(timeout if timeout is not UNSET else self.timeout)
```
florimondmanca: Shouldn't we keep the same logic as before?

```suggestion
        timeout = self.timeout if timeout is UNSET else TimeoutConfig(timeout)
```

httpx/dispatch/proxy_http.py
```diff
@@ -84,12 +84,14 @@ def __init__(
...
-    async def acquire_connection(self, origin: Origin) -> HTTPConnection:
+    async def acquire_connection(
+        self, origin: Origin, timeout: TimeoutTypes = UNSET
+    ) -> HTTPConnection:
         if self.should_forward_origin(origin):
             logger.debug(
                 f"forward_connection proxy_url={self.proxy_url!r} origin={origin!r}"
             )
-            return await super().acquire_connection(self.proxy_url.origin)
+            return await super().acquire_connection(self.proxy_url.origin, timeout)
```
florimondmanca: Should we pass `timeout` as a keyword argument for consistency?

```suggestion
            return await super().acquire_connection(self.proxy_url.origin, timeout=timeout)
```

tests/client/test_async_client.py
```diff
@@ -146,3 +146,13 @@
 
     assert response.status_code == 200
     assert response.content == data
+
+
+async def test_async_client_timeout_default(server, backend):
+    async with httpx.AsyncClient(backend=backend) as client:
+        assert client.dispatch.timeout == httpx.config.UNSET
```
florimondmanca: So, maybe this is an odd behavior. Shouldn't we expect accessing the timeout config at any point in the chain to return a definite `TimeoutConfig` object? cc @tomchristie

tests/client/test_async_client.py
```diff
@@ -146,3 +146,13 @@
...
+
+
+async def test_async_client_timeout_disabled(server, backend):
+    async with httpx.AsyncClient(backend=backend, timeout=None) as client:
+        assert client.dispatch.timeout is None
```
florimondmanca: *Ideally*, to fully test tis we should make a request to the `/slow_response` endpoint of the server with a long delay (e.g. 100ms), surround the call with a "raise a timeout after x seconds" (where e.g. x = 50ms) instruction, and ensure that the timeout error is raised.

The "raise timeout after x seconds" instruction depends on the underlying async library, but we can add an util in `tests/concurrency.py` (single-dispatch stuff not listed here):

```python
import asyncio

async def timeout_after(delay, coroutine):
    try:
        await asyncio.wait(coroutine, timeout=delay)
    except asyncio.TimeoutError:
        raise TimeoutError


import trio

async def timeout_after(delay, coroutine):
    async with trio.move_on_after(delay) as cancel_scope:
        await coroutine
    if cancel_scope.cancelled_caught:
        raise TimeoutError
```

And use it like this:

```python
async def test_async_client_timeout_disabled(server, backend):
    url = server.url.replace_with(path="/slow_response/100")
    async with httpx.AsyncClient(backend=backend, timeout=None) as client:
        assert client.dispatch.timeout is None
        with pytest.raises(TimeoutError):
            await timeout_after(50e-3, client.get(url))
```

Or would this be too much?

httpx/config.py
florimondmanca: `config.py` is probably a nicer home for this constant anyway. :)

httpx/config.py
jcugat: I tried that first but `mypy` was complaining, I will reproduce the actual error once I get home, maybe there is something I missed.

httpx/dispatch/proxy_http.py
jcugat: In `connection_pool.py` we pass both origin and timeout as keyword arguments, maybe we should change both then?

tests/client/test_async_client.py
jcugat: I was struggling a bit to change the default timeout of the client (didn't want to have a test that took the full 5s of the default timeout), but I really like this approach. I'll add a few tests with this pattern!

httpx/dispatch/proxy_http.py
florimondmanca: Yep, sounds good.

tomchristie:
Having seen this going all the way through the stack like this I'm a whole lot more cautious about it.
Am less sure now that it's definatly what we want to do.

httpx/config.py
jcugat: It's available in the `typing_extensions` package https://github.com/python/typing/tree/master/typing_extensions

But I guess it's not worth adding a new dependency just for this (since we have the class workaround anyway).

httpx/config.py
jcugat: This is the mypy error:

```
httpx/config.py:262: error: Incompatible types in assignment (expression has type "Union[float, UnsetType]", variable has type "Optional[float]")
httpx/config.py:263: error: Incompatible types in assignment (expression has type "Union[float, UnsetType]", variable has type "Optional[float]")
httpx/config.py:264: error: Incompatible types in assignment (expression has type "Union[float, UnsetType]", variable has type "Optional[float]")
```

The weird thing is that it complains about these lines:
```
else:
    self.connect_timeout = timeout
    self.read_timeout = timeout
    self.write_timeout = timeout
```

httpx/config.py
jcugat: Right, changed. This implied a few more changes in the class.

httpx/dispatch/connection.py
jcugat: Reverted

httpx/dispatch/proxy_http.py
jcugat: Changed

jcugat:
@tomchristie your concern is that now we have an `UNSET` value that is exposed in `client.dispatch.timeout` (instead of actual timeout values), or the amount of code changes it required?
I'm trying to understand your point of view so I can think about alternatives.

jcugat:
Is there something else I can do to help push this forward?

httpx/config.py
Hanaasagi: emm, this is because you declare the type of `timeout` is `TimeoutTypes`(in `__init__` method). It's a `Union` which includes `UnsetType`. And  you declare the type of  `self.connect_timeout` is `typing.Optional[float]` in line 241. So if you use `elif timeout is UNSET`, any instance of `UnsetType` will enter the `else` branch.

httpx/config.py
Hanaasagi: @florimondmanca I think `typing.Literal` is not valid in this usage. `UNSET` is not a `Literal`.

florimondmanca:
@jcugat I remember we discussed in #433 about the option of using `timeout=False` to disable timeouts. I explore this possibility in #490, feel free to take a look over there. :)

tomchristie:
Thanks for having put a bunch of time into this!

It's bit out of sync now, but a good basis for figure out what we need.
Here's what I think...

* Drop `DEFAULT_TIMEOUT_CONFIG` from `config.py`.
* Instead declare a `DEFAULT_TIMEOUT = 5` in `client.py` (Closer to the source where it's used is better.)
* Declare an `UNSET = object()` in `client.py`
* Use `timeout = DEFAULT_TIMEOUT` on the `Client.__init__`, and on the base API methods.
* Use `timeout = UNSET` on `Client.request`, `Client.get` & pals, and `Client.send()`
* Don't pass `timeout=...` to the connection pool on `__init__`
* `.send()` should use `timeout = self.timeout if timeout is UNSET else timeout` to select either the client timeout, or the explicitly passed override timeout. https://github.com/encode/httpx/blob/3f686012222ea34597c53d41d6b7e84d7b782038/httpx/client.py#L345
* Everywhere else (eg. the dispatch API) takes an optional `timeout: TimeoutConfig() = None` argument, and just uses `timeout = TimeoutConfig() if timeout is None else timeout`. It won't need to select between `self.timeout` and `timeout` anymore, because that decision has been taken at a client level instead.

tomchristie:
Superseeded by #592 - Thanks so much for your time on this, it's been really helpful in getting it squared away!

PR #464, opened 2019-10-09
First pass at autodoc support

tomchristie:
I've started working on [an autodoc extension](https://github.com/tomchristie/mkautodoc) for use with mkdocs.

This pull request demonstrates using it with the `httpx.request` function, just for visibility of where I think we need to go here.

Bits to do:

- Nested code blocks in the docstring are not rendering. (I tried installing the `pymdownx.superfences` extension, which didn't appear to resolve the issue, either. *Any help with resolving this would be much appreciated.**)
- Need to look at rendering classes, not just functions. More complicated. There's some limited support for this, but will need work.
- I'm using a `dd` element as a bit of a styling hack, as it gives me a nicely indented block for the docstring. Probably ought to be a div with an appropriate class, or some other appropriate block element, with styling applied ideally by mkdocs-material, rather than having to add custom styling for it.
- It'd be really nice to keep the typing information out of the main display, but have contextual information pop up against parameters. We can treat this as a nice-to-have-later.

Screen Shot 2019-10-09 at 11 45 39

### DIFF
--- a/httpx/api.py
+++ b/httpx/api.py
@@ -34,6 +34,43 @@ def request(
     trust_env: bool = None,
     proxies: ProxiesTypes = None,
 ) -> Response:
+    """
+    Constructs and sends a `Request`.
+
+    **Parameters:**
+
+    * **method** - HTTP method for the new `Request` object: `GET`, `OPTIONS`, `HEAD`,
+    `POST`, `PUT`, `PATCH`, or `DELETE`.
+    * **url** - URL for the new `Request` object.
+    * **params** - *(optional)* Query parameters to include in the URL, as a string, dictionary,
+    or list of two-tuples.
+    * **data** - *(optional)* Data to include in the body of the request, as a dictionary
+    * **files** - *(optional)* A dictionary of upload files to include in the body of the request.
+    * **json** - *(optional)* A JSON serializable object to include in the body of the request.
+    * **headers** - *(optional)* Dictionary of HTTP headers to include on the request.
+    * **cookies** - *(optional)* Dictionary of Cookie items to include in the request.
+    * **auth** - *(optional)* An authentication class to use when sending the request.
+    * **timeout** - *(optional)* The timeout configuration to use when sending the request.
+    * **allow_redirects** - *(optional)* Enables or disables HTTP redirects.
+    * **cert** - *(optional)* Either a path to an SSL certificate file, or two-tuple of
+    (certificate file, key file), or a three-tuple of (certificate file, key
+    file, password).
+    * **verify** - *(optional)* Enables or disables SSL verification.
+    * **trust_env** - *(optional)* Enables or disables usage of environment variables for
+    configuration.
+    * **proxies** - *(optional)* A dictionary mapping HTTP protocols to proxy URLs.
+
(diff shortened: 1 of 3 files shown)


tomchristie:
Here's how we're looking now...

aIAgAAjQpG

I think we're possibly(?) far along with this now that we could consider pulling it in, and then making incremental improvements as we go. It's a bit tricky trying to judge how much typing information to include in docstring descriptions when a parameter can accept a number of different types - I've not been terribly precise about that yet - would be happy to either try to clean anything up comprehensively in this PR, or take this as a starting point.

Things that would be nice in the future:

* Ensure that `mkautodocs` adds an `<em>async</em>` in front of any async methods/functions.
* Add *contextual* typing information to parameters. (Keep it hidden by default, but all hover or click popovers)
* Support documenting instance attributes. (Infer `self.<some_valid_variable_name_pattern =` from docstrings, and use any leading comment blocks prior to that as the docs.)

florimondmanca:
@tomchristie Do you think this is ready for review, and potentially merging in?

tomchristie:
@florimondmanca Yes I think so. Obvs we'll have some work to do on:

* Fleshing things out more.
* Figuring out a good approach to how much type information to talk about in the docstrings
* Figuring out a good approach to doc string styling.
* Improving functionality, eg. include contextual type information on hover/click or similar.

But from my POV I think it's probably worth us getting in, and then iterating on.

httpx/api.py
```diff
@@ -32,6 +32,51 @@ def request(
...
+    * **url** - URL for the new `Request` object.
+    * **params** - *(optional)* Query parameters to include in the URL, as a
+    string, dictionary, or list of two-tuples.
+    * **data** - *(optional)* Data to include in the body of the request, as a
+    dictionary
+    * **files** - *(optional)* A dictionary of upload files to include in the
+    body of the request.
+    * **json** - *(optional)* A JSON serializable object to include in the body
+    of the request.
+    * **headers** - *(optional)* Dictionary of HTTP headers to include on the
```
florimondmanca: Not sure if this is a typo?

```suggestion
    * **headers** - *(optional)* Dictionary of HTTP headers to include in the
```

httpx/client.py
```diff
@@ -654,6 +666,52 @@ class AsyncClient(BaseClient):
...
+    requests.
+    * **pool_limits** - *(optional)* The connection pool configuration to use
+    when determining the maximum number of concurrently open HTTP connections.
+    * **max_redirects** - *(optional)* The maximum number of redirect responses
+    that should be followed.
+    * **base_url** - *(optional)* A URL to use as the base when building
+    request URLs.
+    * **dispatch** - *(optional)* A dispatch class to use for sending requests
+    over the network.
+    * **app** - *(optional)* A WSGI or ASGI application to send requests too,
```
florimondmanca:
```suggestion
    * **app** - *(optional)* A WSGI or ASGI application to send requests to,
```

florimondmanca:
I fixed the few typos from my last review, merging in now! :tada:

tomchristie:
Thanks! 😃

PR #465, opened 2019-10-09
Updated publish script to be POSIX compliant

pereile:
Minor updates to the publish script to make it POSIX compliant.

### DIFF
--- a/scripts/publish
+++ b/scripts/publish
@@ -1,7 +1,13 @@
 #!/bin/sh -e
 
+err(){
+    echo "Unable to find the '$1' command."
+    echo "Install from PyPI, using '${PREFIX}pip install $1'."
+    exit 1
+}
+
 export PACKAGE="httpx"
-export VERSION=`cat ${PACKAGE}/__version__.py | grep __version__ | sed "s/__version__ = //" | sed "s/'//g"`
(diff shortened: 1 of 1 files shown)


florimondmanca:
Thanks for opening this @pereile.

While I appreciate the effort, t's not clear to me that this change is required — all HTTPX maintainers managed to make releases successfully so far. So as long as the current status works currently I'd advocate for keeping the script as it is now, similar to how it is in other Encode projects.

I'm going to close this as "not immediately necessary for our own use case", but if this was linked to a problem affecting you more directly feel free to continue discussion.

pereile:
These changes were meant to standardise the script and make it portable . 

Fortunately, the non-POSIX compliant bits in the current script won't cause it to fail even if you run it using a non-compliant shell. So you would still be able to publish releases - its just that the stuff that gets printed to stdout might be different.

It is a minor change, so of course you can choose not to use it. However, I think its still a good idea to standardise the script.

tomchristie:
I'm with @florimondmanca here - practicality over purity.
Curious tho: Any pointers on why the *particular* changes here make it POSIX compliant?

pereile:
https://wiki.ubuntu.com/DashAsBinSh#A.26.3E
http://mywiki.wooledge.org/BashFAQ/082

PR #470, opened 2019-10-10
Remove redundant proxy settings

camellia256:
1. Our `get_environment_proxies()` relies on the standard library function [urllib.request.getproxies_environment()](https://github.com/python/cpython/blob/f5ed41c1ae9a575e965d55c6a5e86fb59181eee8/Lib/urllib/request.py#L2456) to do the parsing work, which already prefers lower case envvars. So we only need to filter out protocols which we don't support.
2. As such, we only need to test if the filtering works as intended. (Resolves #456)
---
Additional thoughts:
1. I can't figure out the intention of `"://" in key`  in httpx/utils.py, L226. It looks unnecessary
2. [urllib.request.proxy_bypass](https://github.com/python/cpython/blob/f5ed41c1ae9a575e965d55c6a5e86fb59181eee8/Lib/urllib/request.py#L2737) looks like a robust implementation of 'NO_PROXY'。 #360
3. On Windows and macOS, `get_environment_proxies()` falls back on Registry and SystemConfiguration if no envvar ended with `_proxy` is specified. Should we document this behavior?

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -219,10 +219,11 @@ def get_environment_proxies() -> typing.Dict[str, str]:
     # Registry and Config for proxies on Windows and macOS.
     # We don't want to propagate non-HTTP proxies into
     # our configuration such as 'TRAVIS_APT_PROXY'.
+    SUPPORTED_PROXY_PROTOCOLS = ("http", "https", "all")
     return {
         key: val
         for key, val in getproxies().items()
-        if ("://" in key or key in ("http", "https"))
+        if ("://" in key or key in SUPPORTED_PROXY_PROTOCOLS)
     }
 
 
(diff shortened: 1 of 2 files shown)


sethmlarson:
>     1. Our `get_environment_proxies()` relies on the standard library function [urllib.request.getproxies_environment()](https://github.com/python/cpython/blob/f5ed41c1ae9a575e965d55c6a5e86fb59181eee8/Lib/urllib/request.py#L2456) to do the parsing work, which already prefers lower case envvars. So we only need to filter out protocols which we don't support.

Nice! 

>     1. I can't figure out the intention of `"://" in key`  in httpx/utils.py, L226. It looks unnecessary

This is in case `getproxies()` returns us a specific scheme+host from a System Registry or something. I'm not sure what possible values can come out of the system registries so I put that there to be defensive.

>     2. [urllib.request.proxy_bypass](https://github.com/python/cpython/blob/f5ed41c1ae9a575e965d55c6a5e86fb59181eee8/Lib/urllib/request.py#L2737) looks like a robust implementation of 'NO_PROXY'。 #360 

Ooooh! Maybe this can be used to close #360?  cc @mariaS210

>     3. On Windows and macOS, `get_environment_proxies()` falls back on Registry and SystemConfiguration if no envvar ended with `_proxy` is specified. Should we document this behavior?

Definitely worth documenting in case users are trying to figure out why requests are being proxied.

PR #472, opened 2019-10-12
should-b-proxied: logic whether url to be proxied

mariaS210:
This PR implements logic requested in https://github.com/encode/httpx/pull/462#issuecomment-539971761

This method could be moved if utils is not a good place, maybe in HTTPProxy.
If you wanted something else, please let me know.

### DIFF
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -213,20 +213,17 @@ def kv_format(**kwargs: typing.Any) -> str:
     return " ".join(f"{key}={value!r}" for key, value in kwargs.items())
 
 
-def get_environment_no_proxy() -> typing.List[str]:
-    """ Gets value of NO_PROXY environment value as a list"""
-    no_proxy = get_environ_lower_and_upper("NO_PROXY")
-    return [proxy.strip() for proxy in no_proxy.split(",")] if no_proxy else []
-
-
-def should_be_proxied(url: "URL") -> bool:
+def should_not_be_proxied(url: "URL") -> bool:
     """ Return True if url should be proxied,
     return False otherwise.
     """
-    no_proxy_list = get_environment_no_proxy()
-    if no_proxy_list and no_proxy_list[0] == "*":
+    no_proxy = getproxies().get("no")
+    if not no_proxy:
         return False
+    no_proxy_list = [host.strip() for host in no_proxy.split(",")]
     for name in no_proxy_list:
+        if name == "*":
+            return True
         if name:
             name = name.lstrip(".")  # ignore leading dots
             name = re.escape(name)
@@ -234,8 +231,8 @@ def should_be_proxied(url: "URL") -> bool:
             if re.match(pattern, url.host, re.I) or re.match(
                 pattern, url.authority, re.I
             ):
-                return False
-    return True
+                return True
+    return False
 
 
 def get_environment_proxies() -> typing.Dict[str, str]:
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -216,7 +216,7 @@ def kv_format(**kwargs: typing.Any) -> str:
 def get_environment_no_proxy() -> typing.List[str]:
     """ Gets value of NO_PROXY environment value as a list"""
     no_proxy = get_environ_lower_and_upper("NO_PROXY")
-    return no_proxy.split(",") if no_proxy else []
+    return [proxy.strip() for proxy in no_proxy.split(",")] if no_proxy else []
 
 
 def should_be_proxied(url: "URL") -> bool:
@@ -224,10 +224,17 @@ def should_be_proxied(url: "URL") -> bool:
     return False otherwise.
     """
     no_proxy_list = get_environment_no_proxy()
-    for exempt_url in no_proxy_list:
-        if exempt_url and url.host.endswith(exempt_url):
-            # This URL should be exempted from going through proxy
-            return False
+    if no_proxy_list and no_proxy_list[0] == "*":
+        return False
+    for name in no_proxy_list:
+        if name:
+            name = name.lstrip(".")  # ignore leading dots
+            name = re.escape(name)
+            pattern = r"(.+\.)?%s$" % name
+            if re.match(pattern, url.host, re.I) or re.match(
+                pattern, url.authority, re.I
+            ):
+                return False
     return True
 
 
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -14,6 +14,7 @@ from urllib.request import getproxies
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     from .models import PrimitiveData
+    from .models import URL
 
 
 def normalize_header_key(value: typing.AnyStr, encoding: str = None) -> bytes:
@@ -210,6 +211,24 @@ def kv_format(**kwargs: typing.Any) -> str:
     "x=1 name='Bob'"
     """
     return " ".join(f"{key}={value!r}" for key, value in kwargs.items())
+
+
+def get_environment_no_proxy() -> typing.List[str]:
+    """ Gets value of NO_PROXY environment value as a list"""
+    no_proxy = get_environ_lower_and_upper("NO_PROXY")
+    return no_proxy.split(",") if no_proxy else []
+
+
+def should_be_proxied(url: "URL") -> bool:
+    """ Return True if url should be proxied,
+    return False otherwise.
+    """
+    no_proxy_list = get_environment_no_proxy()
+    for exempt_url in no_proxy_list:
+        if exempt_url and url.host.endswith(exempt_url):
+            # This URL should be exempted from going through proxy
+            return False
+    return True
 
 
 def get_environment_proxies() -> typing.Dict[str, str]:
--- a/httpx/utils.py
+++ b/httpx/utils.py
@@ -214,7 +214,7 @@ def kv_format(**kwargs: typing.Any) -> str:
 
 
 def should_not_be_proxied(url: "URL") -> bool:
-    """ Return True if url should be proxied,
+    """ Return True if url should not be proxied,
     return False otherwise.
     """
     no_proxy = getproxies().get("no")
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -9,13 +9,12 @@ from httpx import utils
 from httpx.utils import (
     ElapsedTimer,
     get_ca_bundle_from_env,
-    get_environment_no_proxy,
     get_environment_proxies,
     get_netrc_login,
     guess_json_utf,
     obfuscate_sensitive_headers,
     parse_header_links,
-    should_be_proxied,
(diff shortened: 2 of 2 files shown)


tests/test_utils.py
```diff
@@ -169,6 +171,21 @@ def test_get_ssl_cert_file():
     assert timer.elapsed.total_seconds() == pytest.approx(0.1, abs=0.05)
 
 
+@pytest.mark.parametrize(
+    ["no_proxy", "exempt_list"],
+    [
+        ({}, []),
+        ({"NO_PROXY": ""}, []),
+        ({"NO_PROXY": "http://127.0.0.1"}, ["http://127.0.0.1"]),
```
yeraydiazdiaz: I've been searching for a formal definition of `NO_PROXY` and [the closest I could find](https://curl.haxx.se/docs/manual.html) suggests:

> A comma-separated list of host names that shouldn't go through any proxy is set in (only an asterisk, * matches all hosts)

So presumably the `https://` prefix should be invalid if we abide for that rule?

httpx/utils.py
```diff
@@ -212,6 +213,24 @@ def kv_format(**kwargs: typing.Any) -> str:
...
+    return no_proxy.split(",") if no_proxy else []
+
+
+def should_be_proxied(url: "URL") -> bool:
+    """ Return True if url should be proxied,
+    return False otherwise.
+    """
+    no_proxy_list = get_environment_no_proxy()
+    for exempt_url in no_proxy_list:
+        if exempt_url and url.host.endswith(exempt_url):
```
yeraydiazdiaz: Again following the definition from the curl manual:

> If the host name matches one of these strings, or the host is within the domain of one of these strings, transactions with that node will not be proxied. When a domain is used, it needs to start with a period. A user can specify that both www.example.com and foo.example.com should not use a proxy by setting NO_PROXY to .example.com. By including the full name you can exclude specific host names, so to make www.example.com not use a proxy but still have foo.example.com do it, set NO_PROXY to www.example.com.

Seems `url.host.endswith` is not quite enough?

httpx/utils.py
```diff
@@ -212,6 +213,24 @@ def kv_format(**kwargs: typing.Any) -> str:
     return " ".join(f"{key}={value!r}" for key, value in kwargs.items())
 
 
+def get_environment_no_proxy() -> typing.List[str]:
+    """ Gets value of NO_PROXY environment value as a list"""
+    no_proxy = get_environ_lower_and_upper("NO_PROXY")
+    return no_proxy.split(",") if no_proxy else []
```
cansarigol: I would add `strip` and empty control for results of `split`

httpx/utils.py
cansarigol: `exempt_url` can be converted to the `URL` instance. if it can, hosts match otherwise can be checked like this `if exempt_url in url.host`

tests/test_utils.py
```diff
@@ -208,3 +225,21 @@ def test_obfuscate_sensitive_headers(headers, output):
...
+@pytest.mark.parametrize(
+    ["url", "no_proxy", "expected"],
+    [
+        ("http://127.0.0.1", {"NO_PROXY": ""}, True),
+        ("http://127.0.0.1", {"NO_PROXY": "127.0.0.1"}, False),
+        ("http://127.0.0.1", {"NO_PROXY": "1.1.1.1"}, True),
+        ("http://courses.mit.edu", {"NO_PROXY": "mit.edu"}, False),
+        ("https://mit.edu.info", {"NO_PROXY": "mit.edu"}, True),
+        ("https://mit.edu.info", {"NO_PROXY": "mit.edu,edu.info"}, False),
+        ("https://mit.edu.info", {"NO_PROXY": "mit.edu,mit.info"}, True),
```
cansarigol: Should we include the `NO_PROXY` list case?

sethmlarson:
@mariaS210 Take a look at [this comment](https://github.com/encode/httpx/pull/470#issue-326960110) regarding using `urllib.request.proxy_bypass` to support NO_PROXY, might be a lot easier than writing our own! :)

mariaS210:
since `urllib.request.proxy_bypass_environment` does this exact thing, I should probably close this, right?

florimondmanca:
Just stumbled upon https://github.com/psf/requests/issues/3976 and https://github.com/psf/requests/pull/3979: `proxy_bypass` makes DNS lookups on Windows, which seems to slow things down dramatically.

florimondmanca:
Quick POC:

```console
$ NO_PROXY=myhost.com python
>>> import urllib.request
>>> urllib.request.proxy_bypass_environment("example.com")
1
>>> urllib.request.proxy_bypass_environment("google.com")
0
```

Also, just stumbled upon https://github.com/psf/requests/issues/3976 and https://github.com/psf/requests/pull/3979: `proxy_bypass` makes DNS lookups on Windows, which seems to slow things down dramatically on that platform. The implementation in this PR doesn't seem to account for that either, but just thought I'd link to it here.

@mariaS210 I like the idea of a `should_be_proxied` helper function though. We'd just need to remove `get_environment_no_proxy()` and its tests, and revisit the `should_be_proxied()` implementation to use urllib's helper. :-)

florimondmanca:
Also actually, we may want to have the contrary, i.e. `should_not_be_proxied()`, because even if `NO_PROXY` isn't set we may not actually end up proxy the URL (e.g. because we're not using proxies!). If it's set though, and the URL matches `NO_PROXY`, then we definitely should *not* proxy it. :)

mariaS210:
well, i tried to use `proxy_bypass` or `proxy_bypass_environment`, but the mypy check did not recognize it:
```
httpx$ mypy httpx
httpx/utils.py:13: error: Module 'urllib.request' has no attribute 'proxy_bypass'
Found 1 error in 1 file (checked 31 source files)
```
Maybe because it was not declared in `_all` [here](    https://github.com/python/cpython/blob/f5ed41c1ae9a575e965d55c6a5e86fb59181eee8/Lib/urllib/request.py#L130)?

tests/test_utils.py
mariaS210: added a test that shows that the https:// prefix is invalid
no_proxy is not retrieved via urllibs getproxies() function

httpx/utils.py
mariaS210: changed to using urllibs implementation, with a regular expression

httpx/utils.py
mariaS210: added strip and a test case with space between hosts

tests/test_utils.py
mariaS210: the list is in the `mit.edu,edu.info` examples

mariaS210:
@florimondmanca @yeraydiazdiaz @cansarigol thank you for your reviews

can you please take another look and let me know if you'd like something changed?

httpx/utils.py
```diff
@@ -212,6 +213,28 @@ def kv_format(**kwargs: typing.Any) -> str:
...
+    for name in no_proxy_list:
+        if name == "*":
+            return True
+        if name:
+            name = name.lstrip(".")  # ignore leading dots
+            name = re.escape(name)
+            pattern = r"(.+\.)?%s$" % name
+            if re.match(pattern, url.host, re.I) or re.match(
+                pattern, url.authority, re.I
+            ):
```
florimondmanca: Just a note that we have uncompiled regexes here, which will harm performance as soon as a user users `NO_PROXY` (we'll keep compiling regex over and over on each request).

I can't think of an obvious way right now to turn this into a single compiled regex because the regex is composed from the `name` coming from `NO_PROXY`, so I won't turn this into a merge blocker. :)

tests/test_utils.py
```diff
@@ -201,3 +202,27 @@ def test_obfuscate_sensitive_headers(headers, output):
...
+@pytest.mark.parametrize(
+    ["url", "no_proxy", "expected"],
+    [
+        ("http://127.0.0.1", {"NO_PROXY": ""}, False),
+        ("http://127.0.0.1", {"NO_PROXY": "127.0.0.1"}, True),
+        ("http://127.0.0.1", {"NO_PROXY": "https://127.0.0.1"}, False),
+        ("http://127.0.0.1", {"NO_PROXY": "1.1.1.1"}, False),
+        ("http://courses.mit.edu", {"NO_PROXY": "mit.edu"}, True),
+        ("https://mit.edu.info", {"NO_PROXY": "mit.edu"}, False),
+        ("https://mit.edu.info", {"NO_PROXY": "mit.edu,edu.info"}, True),
```
florimondmanca: Can we add a test case just above for `("https://mit.edu.info", {"NO_PROXY": "edu.info"}, True)`,to make extra clear that the `edu.info` being contained in `mit.edu.info` is what causes us to decide not to proxy this URL?

httpx/utils.py
sethmlarson: @florimondmanca FYI: `re.compile()` caches compiled patterns based on pattern+flags. :)

florimondmanca: Yes, but `re.match()` doesn't, or does it? 🤔

mariaS210: [this](https://stackoverflow.com/a/452142) stackoverflow answer says the difference is negligible
[match](https://github.com/python/cpython/blob/7773d391f8261a97af502ecfa74a2a588ee05f36/Lib/re.py#L189) -> [_compile](https://github.com/python/cpython/blob/7773d391f8261a97af502ecfa74a2a588ee05f36/Lib/re.py#L287) code suggests patterns are cached
notice that due to the no_proxy host (name) being in the regex the same pattern is used twice, not very many times

httpx/utils.py
```diff
@@ -212,6 +213,28 @@ def kv_format(**kwargs: typing.Any) -> str:
...
+    if not no_proxy:
+        return False
+    no_proxy_list = [host.strip() for host in no_proxy.split(",")]
+    for name in no_proxy_list:
+        if name == "*":
+            return True
+        if name:
+            name = name.lstrip(".")  # ignore leading dots
+            name = re.escape(name)
+            pattern = r"(.+\.)?%s$" % name
```
yeraydiazdiaz: Nit: I might be missing something, but do we absolutely need to use a regex here? This pattern seems like it doing `endswith`?

mariaS210: it does nearly just `endswith`, if it contains data before name (the `.+`) then it has to have a `.` before name, so whole domains match (see test with example2.com matching against ample2.com, it think)

tests/test_utils.py
mariaS210: sure, though test at line `214` with `http://courses.mit.edu` and `mit.edu` meets that branch
i could also add comments for what each test does

florimondmanca: Oh absolutely, I didn’t see the test on L214.

sethmlarson:
I'm gonna merge this in an hour or so pending no new comments :)

PR #477, opened 2019-10-16
Change python 3.8-dev to python 3.8 in travis

jtmiclat:
Change python 3.8-dev to python 3.8 in Travis as python 3.8 is officially released

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
       env: NOX_SESSION=test-3.6
     - python: 3.7
       env: NOX_SESSION=test-3.7
-    - python: 3.8-dev
+    - python: 3.8
       env: NOX_SESSION=test-3.8
       dist: bionic # Required to get OpenSSL 1.1.1+
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,6 @@ matrix:
     # FIX
     # Some tests not yet resolved for Windows. (In progress)
     - os: windows
-    # Allow builds to fail when using '-dev' Python versions.
-    # We should still investigate failures periodically for new APIs.
-    - python: 3.8
 
 install:
   - pip install --upgrade nox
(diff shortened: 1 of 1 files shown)


sethmlarson:
Took 3.8 out of the allow failures list since we need to now support 3.8.

florimondmanca:
@sethmlarson We still have the `start_tls()` test fail on 3.8, like it did back in #423 - see my https://github.com/encode/httpx/pull/423#discussion_r330871311.

I spent 30mins trying to debug what it is that we're doing wrong, but I can't seem to find anything. Where did you find the info/resources about how to implement the transport switching (see #263)? Could it be there's something we're missing there?

florimondmanca:
Locally I'm receiving an asyncio error/warning:

```console
ERROR    asyncio:base_events.py:1692 An open stream object is being garbage collected; call "stream.close()" explicitly.
```

which is emitted by [this piece of code](https://github.com/python/cpython/blob/54b74fe9df89f0e5646736f1f60376b4e37c422c/Lib/asyncio/streams.py#L238-L252). This is a new warning in 3.8 that was contributed here: https://github.com/python/cpython/pull/9201.

The warning is even emitted using:

```python
import asyncio

async def main(host, port):
        asyncio.open_connection(host, port), 5
    )
    stream_writer.close()
    await asyncio.sleep(0)
    print(stream_reader._transport)
    del stream_writer

asyncio.run(main("localhost", 8000))
```

By adding `await asyncio.wait(0)` just after `stream_writer.close()` I am able to remove the error; so I added that to our `Stream.close()` implementation to see what was left.

Well, the error is still there when running the `start_tls()` test — which makes me think we might not be cleaning up either the old `stream_reader` or `stream_writer` properly?

(I also tried the `await asyncio.sleep(0)` trick *inside* `.start_tls()`, i.e. just after we've called `start_tls()` and obtained a new transport. But the effect is that a `ConnectionResetError` is raised when we write the request during the test. So… 🤷‍♂)

florimondmanca:
Hi @jtmiclat, now that #484 I believe you can rebase from `master` and the build with 3.8 should pass. :)

sethmlarson:
Kicked another build off, let's see if we can merge now without a rebase. :)

sethmlarson:
@jtmiclat Thank you for this change :) :tada:

Issue #481, opened 2019-10-17
HTTP Post with files works with Requests but not HTTPX


phishstang65:
Hi. The following works with Requests, but not HTTPX. Can you please help me understand how to make it work:

```
import requests
import uuid
import subprocess
import json
import os
import httpx

newName = uuid.uuid4().hex
event = uuid.uuid4().hex

subprocess.call("cp index.jpeg /tmp/%s.jpeg"%(newName),shell=True)

url = 'http://myserver.us-west-2.compute.amazonaws.com'

event = "ASDGASFDASDF"
payload = {'event': "%s"%(event) , 'confidence': 95}

thermalName = "/tmp/%s.jpeg"%(newName)
newName = thermalName

files = {

        'json': (None, json.dumps(payload), 'application/json'),
        'image': (os.path.basename(newName), open(newName, 'rb'), 'application/octet-stream'),
        'image1': (os.path.basename(newName), open(newName, 'rb'), 'application/octet-stream'),
        'image2': (os.path.basename(newName), open(newName, 'rb'), 'application/octet-stream')

        }

# Do the HTTP POST
#r = requests.post(url, files=files)
r = httpx.post(url, files=files)
print(r.text)
```

The error I get is:

```
Traceback (most recent call last):
  File "test-post.py", line 32, in <module>
    r = httpx.post(url, files=files)
  File "/usr/local/lib/python3.6/dist-packages/httpx/api.py", line 183, in post
    trust_env=trust_env,
  File "/usr/local/lib/python3.6/dist-packages/httpx/api.py", line 53, in request
    trust_env=trust_env,
  File "/usr/local/lib/python3.6/dist-packages/httpx/client.py", line 718, in request
    cookies=cookies,
  File "/usr/local/lib/python3.6/dist-packages/httpx/client.py", line 336, in build_request
    cookies=cookies,
  File "/usr/local/lib/python3.6/dist-packages/httpx/models.py", line 696, in __init__
    content, content_type = self.encode_data(data, files, json)
  File "/usr/local/lib/python3.6/dist-packages/httpx/models.py", line 619, in encode_data
    content, content_type = multipart_encode(data or {}, files)
  File "/usr/local/lib/python3.6/dist-packages/httpx/multipart.py", line 103, in multipart_encode
    body.write(field.render_headers())
  File "/usr/local/lib/python3.6/dist-packages/httpx/multipart.py", line 66, in render_headers
    filename = _format_param("filename", self.filename)
[...]

PR #482, opened 2019-10-18
Multipart files tweaks

yeraydiazdiaz:
Fixes https://github.com/encode/httpx/issues/481

The tuple values of the `files` dictionary passed to `multipart_encode` now support:
- `None` filenames, the header will not include the `filename` section
- `str` objects as the contents, as opposed to file-like objects

I wasn't sure about adding docs about it as [requests seems to not include it](https://requests.readthedocs.io/en/master/user/advanced/#post-multiple-multipart-encoded-files) and it seemed more of an implementation detail that doesn't quite fit in the quickstart guide. But I'm happy to include it if we think it's worth it.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -205,17 +205,22 @@ with httpx.Client(proxies=proxy) as client:
     information at `Client` initialization.
 
 ## Timeout fine-tuning
-HTTPX offers various request timeout management options. Three types of timeouts are available: **connect** timeouts, 
-**write** timeouts and **read** timeouts.
 
-* The **connect timeout** specifies the maximum amount of time to wait until a connection to the requested host is established.   
-If HTTPX is unable to connect within this time frame, a `ConnectTimeout` exception is raised.
-* The **write timeout** specifies the maximum duration to wait for a chunk of data to be sent (for example, a chunk of the request body).   
-If HTTPX is unable to send data within this time frame, a `WriteTimeout` exception is raised.
-* The **read timeout** specifies the maximum duration to wait for a chunk of data to be received (for example, a chunk of the response body).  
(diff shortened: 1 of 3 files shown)


florimondmanca:
Nice work @yeraydiazdiaz! I might be in favor of a files section in the advanced docs. :)

docs/advanced.md
```diff
@@ -269,9 +276,53 @@ with httpx.Client(timeout=None) as client:
...
+  ...
+  "files": {
+    "upload-file": "<... binary content ...>"
+  },
+  ...
+}
+```
+
+More specifically, this tuple must have at least two elements and maximum of three:
+- The first one an optional file name which can be set to `None`.
```
florimondmanca:
```suggestion
- The first one is an optional file name which can be set to `None`.
```

yeraydiazdiaz:
Thanks all! 💚

docs/advanced.md
```diff
@@ -269,9 +276,53 @@ with httpx.Client(timeout=None) as client:
...
+```python
+>>> files = {'upload-file': (None, 'text content', 'text/plain')}
+>>> r = httpx.post("https://httpbin.org/post", files=files)
+>>> print(r.text)
+{
+  ...
+  "files": {},
+  "form": {
+    "upload-file": "text-content"
+  },
```
tomchristie: I'm a bit confused by this example - wouldn't we expect the item to be in `"files"` here?

yeraydiazdiaz: I was confused about that too.

That's the actual output of httpbin, the headers we send are correct as shown in the tests and I verified requests gets the same output from httpbin so I left it as is.

It's a bit confusing for the docs, maybe we should show the request headers instead of the response from httpbin?

PR #484, opened 2019-10-19
Proposal: Make start_tls a method on streams & return a new stream

JayH5:
Fixes Python 3.8.0 TLS issue (see second commit & #477)

I like this approach better than mutating the insides of a `TCPStream` when switching to TLS. Instead, a new `TCPStream` instance is returned. This pattern better matches the way Trio abstracts streams which also seems to be the way asyncio is headed for Python 3._next_.

Fixing the Python 3.8 issue does not require this refactor, but this makes the fix a lot cleaner.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -51,6 +51,8 @@ class TCPStream(BaseTCPStream):
         self.stream_writer = stream_writer
         self.timeout = timeout
 
+        self._inner: typing.Optional[TCPStream] = None
+
     async def start_tls(
         self, hostname: str, ssl_context: ssl.SSLContext, timeout: TimeoutConfig
     ) -> BaseTCPStream:
@@ -80,7 +82,12 @@ class TCPStream(BaseTCPStream):
             transport=transport, protocol=protocol, reader=stream_reader, loop=loop
         )
 
-        return TCPStream(stream_reader, stream_writer, self.timeout)
+        ssl_stream = TCPStream(stream_reader, stream_writer, self.timeout)
+        # When we return a new TCPStream with new StreamReader/StreamWriter instances,
+        # we need to keep references to the old StreamReader/StreamWriter so that they
+        # are not garbage collected and closed while we're still using them.
+        ssl_stream._inner = self
+        return ssl_stream
 
     def get_http_version(self) -> str:
         ssl_object = self.stream_writer.get_extra_info("ssl_object")
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -51,6 +51,37 @@ class TCPStream(BaseTCPStream):
         self.stream_writer = stream_writer
         self.timeout = timeout
 
+    async def start_tls(
+        self, hostname: str, ssl_context: ssl.SSLContext, timeout: TimeoutConfig
+    ) -> BaseTCPStream:
+        loop = asyncio.get_event_loop()
+        if not hasattr(loop, "start_tls"):  # pragma: no cover
+            raise NotImplementedError(
+                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
+            )
+
+        stream_reader = asyncio.StreamReader()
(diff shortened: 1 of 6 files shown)


httpx/concurrency/asyncio.py
```diff
@@ -51,6 +51,44 @@ def __init__(
...
+            ),
+            timeout=timeout.connect_timeout,
+        )
+
+        stream_reader.set_transport(transport)
+        stream_writer = asyncio.StreamWriter(
+            transport=transport, protocol=protocol, reader=stream_reader, loop=loop
+        )
+
+        ssl_stream = TCPStream(stream_reader, stream_writer, self.timeout)
```
JayH5: A question now is whether the timeout on the new stream should be the `timeout` passed to this function or the existing timeout on the stream (`self.timeout`)... or should the method not take a `timeout` parameter at all and just use the existing stream's timeout?

florimondmanca: Given the current state of timeout management on `.start_tls()`*, I think it's OK to not accept a `timeout` here and propagate the stream's `timeout` to this new SSL stream.

*It's a bit tricky, because there seems to be cases when the `timeout` passed to `.start_tls()` and `stream.timeout` may not be the same. This is because `.start_tls()` is called in `HTTPProxy.tunnel_start_tls()`, which doesn't itself accept a `timeout`, so the `timeout` it gives to `.start_tls()` is the one on the `HTTPConnection`. *But* `stream.timeout` refers to *either* the `connection.timeout`, *or* the custom `timeout` that was passed when making the request. So, yeah. I think we've got a bigger problem with managing timeouts anyway, so we can probably just use `self.timeout` and not introduce another level complexity.

JayH5: So are we saying...
```python
async def start_tls(self, hostname: str, ssl_context: ssl.SSLContext) -> BaseTCPStream:
    ...
```
...and just change `tunnel_start_tls` to use the stream's connection timeout for the SSL timeout.

Or I guess one could do something like:
```python
async def start_tls(
    self, hostname: str, ssl_context: ssl.SSLContext, ssl_timeout: typing.Optional[float] = None
) -> BaseTCPStream:
    # Default to using the stream's `timeout.connect_timeout` for the SSL timeout
    ...
```
In asyncio there is actually a [`ssl_handshake_timeout` parameter for `start_tls`](https://docs.python.org/3/library/asyncio-eventloop.html#tls-upgrade) -- which defaults to 60 seconds.

florimondmanca: Well, looking at it again, we could take the path of what `.read()` and `.write()` do:

```python
async def start_tls(self, hostname: str, ssl_context: ssl.SSLContext, timeout: TimeoutConfig = None) -> BaseTCPStream:
    if timeout is None:
        timeout = self.timeout
    ...
```

But *explicitly* pass `timeout=stream.timeout` in `.tunnel_start_tls()` (which amounts to not passing `timeout`, but explicitly shows that we're not using `timeout=connection.timeout`). What do you think?

(I have the intuition there's something fundamentally wrong in the way we're handling timeout/certs/verify configuration though, and we need to review it, so maybe let's not think too hard about this.)

> In asyncio there is actually a ssl_handshake_timeout parameter for start_tls -- which defaults to 60 seconds.

The current approach allows us to say "we want the entire process of switching to TLS to not take longer than X", which I think is a nice async library-agnostic way of putting things. (And `connect_timeout` is a good candidate or X, in my opinion.)

JayH5: Timeouts are hard :cold_sweat:. I don't have strong feelings about how things should work here. I'll think about it a bit more, but happy to defer to others. I think this currently is the smallest change to existing behaviour (it shouldn't really change anything) so I'll leave it as is.

> The current approach allows us to say "we want the entire process of switching to TLS to not take longer than X"

👍

florimondmanca: Sounds good, I also think the current state of things is satisfying enough.

sethmlarson:
:tada: Thank you so much @JayH5!

PR #490, opened 2019-10-20
Disable timeouts using timeout=False

### LINKED ISSUE #433: Passing 'timeout=None' should always result in no timeout

florimondmanca:
Fixes #433, alternative for #463 (cc @jcugat).

Essentially, this PR proposes the following behaviors:

- `timeout=None` (or not passing a `timeout`) uses the default timeout, *always*. (This is a breaking change compared to the current situation in which e.g. `Client(timeout=None)` disables timeouts.)
- `timeout=False` disables timeouts, *always*.

This leads us to have a consistent behavior for timeout values across entrypoints and clients. This also allows us to keep the footprint of the change minimal (as opposed to #463).

I updated the docs to be very explicit and illustrative on all these cases.

I do believe that `timeout=None` is not the most intuitive API to expose to our users, but my logic here is that using `timeout=False` (as in "Don't timeout!") is merely a matter of getting used to it.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -294,29 +294,57 @@ resp = httpx.get('http://example.com/api/v1/example', timeout=timeout)
 
 ### Default timeouts
 
-By default all types of timeouts are set to 5 second.
+By default all types of timeouts are set to 5 seconds.
 
-### Disabling timeouts
+Default timeouts are applied whenever you don't specify a timeout, or you pass `None` as a timeout value.
 
-To disable timeouts, you can pass `None` as a timeout parameter.
-Note that currently this is not supported by the top-level API.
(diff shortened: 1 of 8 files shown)


florimondmanca:
It *might* be possible to split this PR in two: 1) make `timeout=None` always apply timeouts (temporarily resulting in HTTPX not supporting disabled timeouts), 2) make `timeout=False` disable timeouts. Let me know if that'd be useful.

florimondmanca:
Closed in favor of #493.

PR #491, opened 2019-10-20
Reorganize timeout config tests

florimondmanca:
Just a bit of moving timeout config tests around so they're all close to one another, and add one or two tests for the default cases. Would help reduce the scope of #490.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -158,17 +158,6 @@ def test_empty_http_version():
         httpx.HTTPVersionConfig([])
 
 
-def test_timeout_repr():
-    timeout = httpx.TimeoutConfig(timeout=5.0)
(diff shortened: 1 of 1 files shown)


tomchristie:
Happy enough with this, sure.
Not obvs that eg. `test_timeout_from_nothing` is ultimately the behavior that we want or not, but if it expresses the existing behaviour and is then adapted then that's fine.

florimondmanca:
> Not obvs that eg. `test_timeout_from_nothing` is ultimately the behavior that we want or not, but if it expresses the existing behaviour and is then adapted then that's fine.

Exactly — this PR explicits the current behavior we have so that we can potentially adapt it when fixing #433. :)

PR #492, opened 2019-10-20
Always treat timeout=None as default timeout

florimondmanca:
Spun up from #490. Refs #433.

**Note**: this is based on #491, so #491 *must* be merged first. Will change base branch once it gets merged.

As a result of this PR:

- `timeout=None` **always** means "use the defaults". This applies to our users, but also components in our code (clients and dispatchers in particular).
- The logic for "how to apply default timeouts" is contained in `config.py` only. (And `DEFAULT_TIMEOUT_CONFIG` is only used there, and in some of the tests.)
- Support for disabling timeouts is (temporarily) removed, but it will be added again by #490.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -294,30 +294,37 @@ resp = httpx.get('http://example.com/api/v1/example', timeout=timeout)
 
 ### Default timeouts
 
-By default all types of timeouts are set to 5 second.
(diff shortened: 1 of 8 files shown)


florimondmanca:
Closed in favor of exploring https://github.com/encode/httpx/issues/433#issuecomment-544260647

PR #493, opened 2019-10-20
Disable timeouts using timeout=False

### LINKED ISSUE #433: Passing 'timeout=None' should always result in no timeout

florimondmanca:
Fixes #433. Spun up from #490. Based on #492.

**Note**: #492 must get merged first. I'll change the base branch once it gets merged.

This PR adds support for consistently treating `timeout=False` as "disable timeouts".

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -324,7 +324,28 @@ httpx.get(url, timeout=httpx.TimeoutConfig(connect_timeout=3))
 
 ### Disabling timeouts
 
-Support for disabling timeouts is planned, but not currently supported.
+To disable timeouts, you must pass `False` as a timeout parameter.
+
+For example, none of these calls will time out:
+
+```python
+url = "https://httpbin.org/delay/10"
+
+# Timeouts are disabled for this request.
+httpx.get(url, timeout=False)
+
+# Timeouts are disabled for all requests made with this client.
+with httpx.Client(timeout=False) as client:
+    client.get(url)
+
+with httpx.Client() as client:
+    # Timeouts are disabled for this request only.
+    client.get(url, timeout=False)
+
+# Only read timeout is disabled for this request.
+timeout = httpx.TimeoutConfig(read_timeout=False)
+httpx.get(url, timeout=timeout)
+```
 
 ## Multipart file encoding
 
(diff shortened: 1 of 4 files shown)


sethmlarson:
I'm still -1 on using `False` as "no timeout". "No timeout" has always been `None` in the Python world and the `timeout` parameter isn't a data structure like `headers`, `proxies`, `params`, `files`, etc so I don't see how adding this "unset" behavior is beneficial over using `timeout=httpx.DEFAULT_TIMEOUT` if you really want to revert for an unspecified default per-request.

florimondmanca:
@sethmlarson Hmm, so as I understand, you'd be prefer setting the default timeout across the code base (vs. passing `None` as we do currently)?

```python
def get(self, ..., timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG):
    ...
```

thebigmunch:
I sorta suggested the same as Seth in the issue. I suggested specifically setting them to 5 rather than using a DEFAULT object or variable. This is a bit more explicit/doesn't obscure the default value from users, especially in the API reference. We now have docs for fine-tuning timeout behavior: I think most of the rest of the docs and code should  be using plain values. It can be converted into a TimoutConfig object in the code where needed.

sethmlarson:
@florimondmanca Setting a proper default throughout our code base is much better in my opinion. :) I think @thebigmunch's suggestion of using `5` instead of an actual object is not bad either.

thebigmunch:
In the case of setting the defaults to something other than a sentinel value, the one place that would need some special handling is Client methods using the timeout set on Client if no timeout is set in the method call.

tomchristie:
> I think @thebigmunch's suggestion of using 5 instead of an actual object is not bad either.

That seems pretty sensible yeah.

The "empty" sentinal had a bigger impact on the API through all the different layers so I'm not so keen on that, also it doesn't document very easily "Hey this thing is empty, but you can set it to None, and no those aren't the same thing...". vs. "The default timeout is 5 seconds".

florimondmanca:
Closed in favor of exploring https://github.com/encode/httpx/issues/433#issuecomment-544260647

PR #494, opened 2019-10-20
asyncio: Wait for the stream to close when closing

JayH5:
We discussed calling `wait_closed()` on the `StreamWriter` in #369, but never implemented it due to a transient issue with SSL connections ([comment](https://github.com/encode/httpx/pull/369#discussion_r327187795)).

I now better understand the issue and have written a [lengthy explanation](https://github.com/aio-libs/aiohttp/issues/3535#issuecomment-544279431) on the related aiohttp issue (since that seems to have the most visibility).

The error makes sense if you think about it, and it's clear why it was showing up in our test--because we weren't making sure we consumed the entire response before closing the connection.

I'm not sure if we need to handle this error--I think in most cases the entire response should be consumed before we close a stream. My thinking was we try this and see if it shows up as an issue for users. But I'm happy to change that if people would like.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import ssl
+import sys
 import typing
 from types import TracebackType
 
@@ -170,6 +171,8 @@ class TCPStream(BaseTCPStream):
 
     async def close(self) -> None:
         self.stream_writer.close()
+        if sys.version_info >= (3, 7):
+            await self.stream_writer.wait_closed()
 
 
 class PoolSemaphore(BasePoolSemaphore):
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -50,7 +50,21 @@ async def test_start_tls_on_socket_stream(https_server, backend, get_cipher):
         assert get_cipher(stream) is not None
 
         await stream.write(b"GET / HTTP/1.1\r\n\r\n")
-        assert (await stream.read(8192, timeout)).startswith(b"HTTP/1.1 200 OK\r\n")
+
+        # stream.read() only gives us *up to* as much data as we ask for. In order to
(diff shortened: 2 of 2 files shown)


JayH5:
Also relevant: the `https_compatible` parameter for [`SSLStream`s in Trio](https://trio.readthedocs.io/en/stable/reference-io.html#trio.SSLStream). Maybe we should be using this?

florimondmanca:
Thank you so much for digging into this @JayH5. As for `https_compatible`, from reading it as well as your explaination on the `aiohttp` issue, that would seem like a very reasonable thing to do, indeed.

httpx/concurrency/asyncio.py
```diff
@@ -170,6 +171,8 @@ def is_connection_dropped(self) -> bool:
 
     async def close(self) -> None:
         self.stream_writer.close()
+        if sys.version_info >= (3, 7):
+            await self.stream_writer.wait_closed()
```
florimondmanca: So, is waiting for the write half of the connection to close good enough? Or could we still get the `K5S_INIT` error code somehow, as you mentioned in https://github.com/encode/httpx/pull/369#discussion_r327187795?

florimondmanca: Or is this part of the "try and see" note at the end of the PR description? :)

JayH5: Try and see 😛 but I'm really in two minds about it 😕. We could still get the error here if we somehow close the stream before receiving all the data from the other side.

I _think_ it should be fine to catch the error here. Once the error happens, the error will persist in the connection and whenever we try to read/write to the stream it will be raised again. But this seems correct since we shouldn't try to use a closed stream.

florimondmanca: OK. I'm good with merging in this state, given that I think we have enough context to come back here and try something else out if users start encountering issues.

PR #323, opened 2019-09-06
Pin uvloop<0.13 to fix 3.8-dev build

florimondmanca:
Refs https://github.com/MagicStack/uvloop/issues/266, and prompted by https://github.com/encode/httpx/pull/153 and https://github.com/encode/httpx/pull/276#issuecomment-528998323.

Opening this mostly to see if this fixes the build. We don't know how long the issue may stay though, so perhaps we could consider keeping this until things are sorted out?

(Also, this is probably affecting uvicorn as well.)

PR #498, opened 2019-11-02
Remove pin on uvloop in test-requirements

florimondmanca:
uvloop has shipped a 0.14 release candidate which uvicorn now uses, so the temporary fix from #323 should not be necessary anymore.


florimondmanca:
Taking @tomchristie's thumbs up as approval :-)

tomchristie:
Please do, yup

PR #501, opened 2019-11-02
Automatically request review via CODEOWNERS

florimondmanca:
Something I mentioned in our Keybase chat, is experimenting with adding a `CODEOWNERS` file ([docs](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners)) so that a review from maintainers is automatically requested.

The goal is to reduce a bit of the PR triaging burden as well as showing the intent to contributors that we will be looking into their PR soon.

If all goes well, `@encode/httpx-maintainers` should be requested for review on this one. 💯

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# For docs on code owners, see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+* @encode/httpx-maintainers


florimondmanca:
> If all goes well, `@encode/httpx-maintainers` should be requested for review on this one. 💯

Well, actually not, because I suppose the code owners check is applied based on the target branch (here `master`, which doesn't have a `CODEOWNERS` file yet).

Curious to hear @encode/httpx-maintainers's thoughts on whether this is a good idea at all?

tomchristie:
I’m not sure. I’d guess this is particularly suited for large codebases with subsets owned by particular devs/teams. Eg I could see how it might be useful for splitting out teams for something like Django, where there might be some folks best suited to reviewing orm PRs, others for admin PRs etc. I’m not sure if a project wide catch all is useful here or noise. (Already get notifications about all PRs). Not necessarily against it, just unsure ATM. Also, I do need to do better at making sure we’re adopting a consistent flow/style across all encode projects.

florimondmanca:
I hear you. We had a similar use case somewhere I worked at where migration files were "owned" by a specific team, and that's when code owners came in handy. Your point about already having email notifications is fair. I'm not too sold on the idea either, just poking around the available tooling. I guess let's keep that aside for now. :-)

PR #502, opened 2019-11-02
Add debug logs

### LINKED ISSUE #495: Logging levels
I've been spending some time looking at logging levels with `uvicorn` lately, see eg. https://github.com/encode/uvicorn/pull/474 and would like to take a consistent approach across other projects in `encode`.

We've currently got some very detailed logging, that's available, tho it's really at a "dig into the internals" level, rather than general purpose log outputs.

Something that would tie in really nicely with the approach I'm taking elsewhere would be to have DEBUG-level outputs that issue a single line per-request...

`[DEBUG] HTTP Request: "GET https://api.github.com/user HTTP/1.1" 200 OK`

And then have a `TRACE` level outputs reserved for the really low-level stuff.

florimondmanca:
Fixes #495 

This PR is a follow-up for #500 which:

- Adds debug logs on each HTTP request. (@tomchristie I chose to issue logs from clients, instead of e.g. `ConnectionPool`, so that they get issued regardless of the dispatcher in use (e.g. `ConnectionPool` vs `ASGIDispatcher`).)
- Tweaks the general log format, which was e.g. missing the date and the log level name.

@tomchristie I noticed the date and logger name are missing in Uvicorn logs. Perhaps a good idea to add those there as well? In particular, not having the logger name makes the origin of logs unclear: do they come from Uvicorn, HTTPX, Databases…? See example here:

Screenshot 2019-11-02 at 23 27 56

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -16,9 +16,9 @@ and what function they serve:
 
 Valid values: `debug`, `trace` (case-insensitive)
 
-If set to `trace`, then low-level details about the execution of HTTP requests will be logged to `stderr`. This can help you debug issues and see what's exactly being sent over the wire and to which location.
+If set to `debug`, then HTTP requests will be logged to `stderr`. This is useful for general purpose reporting of network activity.
 
-The `debug` log level is currently ignored, but is planned to issue high-level logs of HTTP requests.
+If set to `trace`, then low-level details about the execution of HTTP requests will be logged to `stderr`, in addition to debug log lines. This can help you debug issues and see what's exactly being sent over the wire and to which location.
 
 Example:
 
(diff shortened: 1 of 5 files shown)


tomchristie:
Great stuff! Welcome to open an issue on unicorn wrt. logger names etc.
Thought: Ideally we’d probably want log all requests in any redirect chain, right?

florimondmanca:
> Thought: Ideally we’d probably want log all requests in any redirect chain, right?

Think so, and I think this PR already supports this. I’ll need to check, but my thinking is that the redirect functionality calls get_response() on repeat for each redirect step, so it should work. :)

docs/environment_variables.md
```diff
@@ -30,23 +30,36 @@ with httpx.Client() as client:
     r = client.get("https://google.com")
 ```
 
+Debug output:
+
+```console
+$ HTTPX_LOG_LEVEL=debug python test_script.py
+DEBUG [2019-11-02 23:39:09] httpx.client - HTTP Request: www.google.com - "GET / HTTP/2" 200 OK
```
tomchristie: I'm wondering how/if we should indicate http/https here?

Options:

* `HTTP Request: https://www.google.com - "GET / HTTP/2" 200 OK` # Include it in the origin.
* `HTTPS Request: www.google.com - "GET / HTTP/2" 200 OK` - Include it in the message itself.
* `HTTP Request: "GET https://www.google.com/ HTTP/2" 200 OK` - Include it in the request line.

florimondmanca:
@tomchristie I added some tests for the logging behavior on redirect. I reckon this PR now does a bunch of different things, so I opened https://github.com/encode/httpx/pull/506 which should get merged first, and then we can get this in. :+1:

docs/environment_variables.md
florimondmanca: That might be useful, yeah. In general I don't think logs do include HTTPS, though (e.g. the current format is inspired by the [Common Log Format](https://httpd.apache.org/docs/1.3/logs.html#common), and that doesn't show the HTTP scheme). So I'm more inclined towards including it in the message.

tomchristie: "the current format is inspired by the Common Log Format, and that doesn't show the HTTP scheme". Yeah I'm not sure here. Logging outgoing HTTP requests isn't neccessarily the same as logging requests to a webserver. (Eg. we want to see the host up front and center, wheras in the server case you typically don't need to see that.)

florimondmanca: Yeah sure, it's not like these debug logs are designed to be ingested by production log systems anyway. :) I prefer first option then — include it in the origin.

florimondmanca: Or actually, option 3 looks sensible too. Especially if we want to make it *not* look like a proper log line.

tomchristie: I'm keen on 3, personally. 👍

sethmlarson: Same for me, 3 looks good.

florimondmanca: Cool, changed to use 3 then!

httpx/client.py
```diff
@@ -241,6 +243,12 @@ def merge_queryparams(
                 finally:
                     await response.close()
 
+            request_line = (
```
StephenBrown2: Looks like is short enough to be on a single line according to black:

```suggestion
            request_line = f"{request.method} {request.url} {response.http_version}"
```

tomchristie:
A thought on formats:

Right now we've got...

* `HTTP Request:  "GET https://www.google.com HTTP/2" 200 OK`

We're indicating something *like* the request line in the quotes, except that it's not quite the actual HTTP request line, since it has the complete URL, and the HTTP version comes from the response. (I *guess* it'll always be the same at the moment, but if we supported `Upgrade` then the request line would indicate HTTP/1.1, and the response line would indicate HTTP/2)

Something we *could* do instead would be to log the literal *response line* instead, so...

* `HTTP Request:  GET https://www.google.com "HTTP/2 200 OK"`

This then has a kind of nice symmetry, in that server logs output a literal request line, and client logs output a literal response line.

Maybe?

florimondmanca:
@tomchristie I like that, yep!

florimondmanca:
@tomchristie Should be ready for re-review. :-)

florimondmanca:
Thanks all!

PR #510, opened 2019-11-06
Add docs on SSL certificates

### LINKED ISSUE #507: Hints on self-signed certificates
In https://github.com/encode/httpx/issues/503, @gvbgduh mentioned the need to make HTTP/2 requests to a local server. Because of our current implementation (require TLS for HTTP/2, as browsers do), this typically requires to setup a local self-signed certificate for use in tests.

I'm probably not the one most aware of how this can be done — @sethmlarson may have some pointers? But we could at least gather some resources on how to set things up, and then see if/how we can include those to our documentation. This might take the form of a note added [here](https://www.encode.io/httpx/advanced/#specify-the-version-of-the-http-protocol) with a link to a relevant how-to guide.

florimondmanca:
Fixes #507

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -365,9 +365,6 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
 
 ## SSL certificates
 
-!!! warning
-    This is an advanced section. You typically won't need this, unless you have special needs regarding your SSL configuration.
-
 When making a request over HTTPS, HTTPX needs to verify the identity of the requested host. To do this, it uses a bundle of SSL certificates (a.k.a. CA bundle) delivered by a trusted certificate authority (CA).
 
 ### Default CA bundle
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -383,11 +383,9 @@ r = httpx.get("https://example.org", verify="path/to/client.pem")
 
 ### Making HTTPS requests to a local server
 
-In some situations, you may have to make HTTPS requests to a server running on `localhost` (a.k.a. `127.0.0.1`, the loopback address). This is typically the case, for example, if you are making requests to a development server.
+When making requests to local servers, such as a development server running on `localhost`, you will typically be using unencrypted HTTP connections.
 
-Because `localhost` is a local and non-uniquely owned domain, you'll need to create your own certificates for HTTPS to work.
-
-Here's one way to create such certificates, and then use them in HTTPX:
+If you do need to make HTTPS connections to a local server, for example to test an HTTPS-only service, you will need to create and use your own certificates. Here's one way to do it:
 
 1. Use [trustme-cli](https://github.com/sethmlarson/trustme-cli/) to generate a pair of server key/cert files, and a client cert file.
 1. Pass the server key/cert files when starting your local server. (This depends on the particular web server you're using. For example, [Uvicorn](https://www.uvicorn.org) provides the `--ssl-keyfile` and `--ssl-certfile` options.)
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
   ...
 }
 ```
+
+## SSL certificates
+
+!!! warning
+    This is an advanced section. You most likely won't need to use the features documented here, unless you know what you're doing.
+
+When making a request over HTTPS, HTTPX needs to verify the identity of the requested host. To do this, it uses a bundle of SSL certificates delivered by a trusted certificate authority (CA).
+
+By default, HTTPX uses the CA bundle provided by [Certifi](https://pypi.org/project/certifi/). This works fine in most cases, however in some advanced situations you may want to tweak this.
+
+### Using a custom CA bundle
+
+If you'd like to use a custom CA bundle, you can use the `verify` parameter that is available on the high-level API as well as clients. For example:
+
+```python
+import httpx
+
+r = httpx.get("https://example.org", verify="path/to/client.pem")
+```
+
+### Making HTTPS requests to a local server
+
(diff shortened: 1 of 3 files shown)


docs/advanced.md
```diff
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
...
+
+Here's one way to do this, and then use them in HTTPX:
+
+1. Use [trustme-cli](https://github.com/sethmlarson/trustme-cli/) to generate a pair of server key/cert files, and a client cert file.
+1. Pass the server key/cert files when starting your local server. (This depends on the particular web server you're using. For example, [Uvicorn](https://www.uvicorn.org) provides the `--ssl-keyfile` and `--ssl-certfile` options.)
+1. Tell HTTPX to use the certificates stored in `client.pem`:
+
+```python
+>>> import httpx
+>>> r = httpx.get("https://localhost:8000", verify="/tmp/client.pem")
```
yeraydiazdiaz: This was surprising to me, I thought based on [our docs](https://www.encode.io/httpx/api/) that `verify` was a boolean to perform SSL checks defaulting to `True`.

Should we update those as well?

docs/advanced.md
```diff
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
   ...
 }
 ```
+
+## SSL certificates
+
+!!! warning
+    This is an advanced section. You most likely won't need to use the features documented here, unless you know what you're doing.
```
yeraydiazdiaz: Nit: I'd drop the "unless you know what you're doing" given you've said "most likely won't need".

docs/advanced.md
```diff
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
...
+
+When making a request over HTTPS, HTTPX needs to verify the identity of the requested host. To do this, it uses a bundle of SSL certificates delivered by a trusted certificate authority (CA).
+
+By default, HTTPX uses the CA bundle provided by [Certifi](https://pypi.org/project/certifi/). This works fine in most cases, however in some advanced situations you may want to tweak this.
```
yeraydiazdiaz: Nit: I'd phrase it as "This is what you want in most cases, however..." as opposed to "works fine" which makes it sound like it's "just OK" :)

docs/advanced.md
```diff
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
...
+
+### Using a custom CA bundle
```
yeraydiazdiaz: Is it CA bundle that we're passing or a certificate?

I'm not an expert by any means but I thought `pem` files are certificates?

docs/advanced.md
florimondmanca: Very confusing indeed. Turns out there's more logic done on `verify`:

https://github.com/encode/httpx/blob/2fcf23bbfe41c5526cb35a65d32dd98dde786988/httpx/config.py#L126-L145

Basically, `verify=True` uses the default CA bundle (currently Certifi's), but it also accepts a path to a CA bundle (`.pem` file).

docs/advanced.md
florimondmanca: Well, only keeping "You most likely won't need to use the features documented here" feels a bit like "we document these features, but they aren't useful!". 😕 

I guess the core problem here is that "unless you know what you're doing" might be a bit patronizing?

Anyway, would you be OK with…?

> This is an advanced section. You typically won't need this, unless you have special needs regarding your SSL configuration.

docs/advanced.md
florimondmanca: Yeah, I'm not sure on the wording myself. According to [this](https://serverfault.com/questions/200881/whats-a-ssl-certificate-bundle) a CA bundle is a group of root and intermediate certificates that work together… and I think `.pem` files can contain any number of certificates? Not sure, really. Ping @sethmlarson? 😄

sethmlarson: CA bundle is a file with one or more CA certificates concatenated together typically.  PEM can contain multiple certificates. (fun-fact: `certifi.where()` returns a .pem file)

florimondmanca: Thanks @sethmlarson. I pushed a commit based on this info and that in https://github.com/encode/httpx/issues/507#issuecomment-550371125 and elsewhere to more precisely explain in the developer interface what `verify` and `cert` are doing. :-)

docs/advanced.md
florimondmanca: See https://github.com/encode/httpx/pull/510#discussion_r343339286 - I just pushed a commit to update these docs. :)

florimondmanca:
@yeraydiazdiaz I pushed some commits addressing your comments, feel free to give this another look. 😄

docs/advanced.md
```diff
@@ -362,3 +362,41 @@ is set to `None` or it cannot be inferred from it, HTTPX will default to
...
+If you'd like to use a custom CA bundle, you can use the `verify` parameter that is available on the high-level API as well as clients. For example:
+
+```python
+import httpx
+
+r = httpx.get("https://example.org", verify="path/to/client.pem")
+```
+
+### Making HTTPS requests to a local server
+
```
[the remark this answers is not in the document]
florimondmanca: Thanks, I pushed a commit rephrasing this section towards that. :)

florimondmanca:
@tomchristie I figured we could just drop the warning — after all, the whole page is labelled "Advanced Usage", right? 😄

florimondmanca:
Mental note that it seems @sethmlarson might have trustme-cli brought into trustme directly (see https://github.com/sethmlarson/trustme-cli/issues/1), so we’ll want to update the link here when that happens. :)

Issue #133, opened 2019-07-21
Implement http+unix dispatcher


sethmlarson:
Lots of users for Docker, Sentry, etc use the `http+unix://...` URLs. Might necessitate additional backend interface if we intend this to be first-class support.

tomchristie:
Any good reference describing this scheme?

I'd expect that this is more of a good case for us ensuring that the dispatcher interface is sufficiently capable, and sufficiently well documented. If we can allow third party packages to support themore complex use cases, without building them into the core package itself, then I think that'd be a great thing.

tomchristie:
Let's treat this as out-of-scope at this point in time.

Issue #509, opened 2019-11-06
dns-over-https DoH


graingert:
Normally a DNS resolver would be something you plug in at a different level, however for DoH you need an http client already and for some optimizations (eg using cloudflare DoH to only access the Cloudflare web should only require one connection) you need to be using the same machinery for your DoH and regular web requests

As such httpx should implement DoH directly

sethmlarson:
DoH can be implemented as a middleware via capturing the request URL domain name and rewrite it to an IP address that's been resolved via DoH. I agree we should implement it once and distribute it via a package.

tomchristie:
> DoH can be implemented as a middleware via capturing the request URL domain name and rewrite it to an IP address that's been resolved via DoH.

That's a great call!

graingert:
It would also be good to support both the Chromium style lookup current provider in table approach, and the Firefox everything via Cloudflare approach, and allow a custom DoH provider

sethmlarson:
Do you have a link to Chromium's approach? I haven't heard of that.

graingert:
@sethmlarson https://blog.chromium.org/2019/09/experimenting-with-same-provider-dns.html

graingert:
The table is enumerated here: https://www.chromium.org/developers/dns-over-https

graingert:
Whatever resolver system is used to resolve the initial DoH server should also be pluggable, eg aiodns etc

tomchristie:
Also: I’d really like to see this sort of thing tackled as third party middleware, rather than including it in the core package.

If we have third party packages that later become sufficiently proven and well used, then we can take a call on including directly in the package itself.

toppk:
> DoH can be implemented as a middleware via capturing the request URL domain name and rewrite it to an IP address that's been resolved via DoH. I agree we should implement it once and distribute it via a package.

It's a little bit tricky if your are doing TLS, as the http hostname often must stay as the fqdn for certificate validation reasons.

I do think that plugable name resolution could be useful in several use cases:
1. DoH, DoT
2. DNS SRV records
3. Service discovery (eureka, consul, etcd, zookeeper based implementations)
4. custom DNS RR loadbalancing/high availability handling
5. minimize dns traffic generated by agents.

[...]

PR #511, opened 2019-11-06
Add support for unix domain sockets

lundberg:
This PR adds support for connecting through unix domain sockets.

New client argument `uds` is added to define unix domain socket path to use.

**Usage:**
```py
async with httpx.AsyncClient(uds="/var/run/docker.sock") as client:
    # This request will connect through the socket file
    r = await client.get("http://docker/version")
```

Solves previously discussed topic in #133

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -74,6 +74,7 @@ class BaseClient:
         app: typing.Callable = None,
         backend: ConcurrencyBackend = None,
         trust_env: bool = True,
+        uds: str = None,
     ):
         if backend is None:
             backend = AsyncioBackend()
@@ -99,6 +100,7 @@ class BaseClient:
                 pool_limits=pool_limits,
                 backend=backend,
                 trust_env=self.trust_env,
+                uds=uds,
             )
         elif isinstance(dispatch, Dispatcher):
             async_dispatch = ThreadedDispatcher(dispatch, backend)
@@ -721,6 +723,7 @@ class Client(BaseClient):
     async requests.
     * **trust_env** - *(optional)* Enables or disables usage of environment
     variables for configuration.
+    * **uds** - *(optional)* A path to a Unix domain socket to connect through
     """
 
     def check_concurrency_backend(self, backend: ConcurrencyBackend) -> None:
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -723,7 +723,7 @@ class Client(BaseClient):
     async requests.
     * **trust_env** - *(optional)* Enables or disables usage of environment
     variables for configuration.
-    * **uds** - *(optional)* A path to a Unix domain socket to connect through
+    * **uds** - *(optional)* A path to a Unix domain socket to connect through.
     """
 
     def check_concurrency_backend(self, backend: ConcurrencyBackend) -> None:
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -263,10 +263,35 @@ class AsyncioBackend(ConcurrencyBackend):
         ssl_context: typing.Optional[ssl.SSLContext],
         timeout: TimeoutConfig,
     ) -> SocketStream:
+        return await self._open_stream(
+            asyncio.open_connection(hostname, port, ssl=ssl_context), timeout
+        )
+
+    async def open_uds_stream(
+        self,
+        path: str,
+        hostname: typing.Optional[str],
+        ssl_context: typing.Optional[ssl.SSLContext],
+        timeout: TimeoutConfig,
+    ) -> SocketStream:
+        server_hostname = hostname if ssl_context else None
+        return await self._open_stream(
+            asyncio.open_unix_connection(
+                path, ssl=ssl_context, server_hostname=server_hostname
+            ),
+            timeout,
+        )
+
+    async def _open_stream(
+        self,
+        socket_stream: typing.Awaitable[
+            typing.Tuple[asyncio.StreamReader, asyncio.StreamWriter]
+        ],
+        timeout: TimeoutConfig,
+    ) -> SocketStream:
         try:
             stream_reader, stream_writer = await asyncio.wait_for(  # type: ignore
-                asyncio.open_connection(hostname, port, ssl=ssl_context),
-                timeout.connect_timeout,
+                socket_stream, timeout.connect_timeout,
             )
         except asyncio.TimeoutError:
             raise ConnectTimeout()
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -263,8 +263,16 @@ class AsyncioBackend(ConcurrencyBackend):
         ssl_context: typing.Optional[ssl.SSLContext],
         timeout: TimeoutConfig,
     ) -> SocketStream:
-        return await self._open_stream(
-            asyncio.open_connection(hostname, port, ssl=ssl_context), timeout
+        try:
+            stream_reader, stream_writer = await asyncio.wait_for(  # type: ignore
+                asyncio.open_connection(hostname, port, ssl=ssl_context),
+                timeout.connect_timeout,
+            )
+        except asyncio.TimeoutError:
+            raise ConnectTimeout()
+
+        return SocketStream(
+            stream_reader=stream_reader, stream_writer=stream_writer, timeout=timeout
         )
 
     async def open_uds_stream(
@@ -275,23 +283,13 @@ class AsyncioBackend(ConcurrencyBackend):
         timeout: TimeoutConfig,
     ) -> SocketStream:
         server_hostname = hostname if ssl_context else None
-        return await self._open_stream(
-            asyncio.open_unix_connection(
-                path, ssl=ssl_context, server_hostname=server_hostname
-            ),
-            timeout,
-        )
 
-    async def _open_stream(
-        self,
-        socket_stream: typing.Awaitable[
-            typing.Tuple[asyncio.StreamReader, asyncio.StreamWriter]
-        ],
-        timeout: TimeoutConfig,
-    ) -> SocketStream:
         try:
             stream_reader, stream_writer = await asyncio.wait_for(  # type: ignore
-                socket_stream, timeout.connect_timeout,
+                asyncio.open_unix_connection(
+                    path, ssl=ssl_context, server_hostname=server_hostname
+                ),
+                timeout.connect_timeout,
             )
         except asyncio.TimeoutError:
             raise ConnectTimeout()
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -156,7 +156,4 @@ async def test_uds(uds_server, backend):
         response = await client.get(url)
     assert response.status_code == 200
     assert response.text == "Hello, world!"
-    assert response.http_version == "HTTP/1.1"
-    assert response.headers
-    assert repr(response) == "<Response [200 OK]>"
-    assert response.elapsed > timedelta(seconds=0)
+    assert response.encoding == "iso-8859-1"
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -146,3 +146,17 @@ async def test_100_continue(server, backend):
 
     assert response.status_code == 200
     assert response.content == data
+
+
+async def test_uds(uds_server, backend):
+    url = uds_server.url
+    uds = uds_server.config.uds
+    assert uds is not None
+    async with httpx.AsyncClient(backend=backend, uds=uds) as client:
+        response = await client.get(url)
+    assert response.status_code == 200
+    assert response.text == "Hello, world!"
+    assert response.http_version == "HTTP/1.1"
+    assert response.headers
+    assert repr(response) == "<Response [200 OK]>"
+    assert response.elapsed > timedelta(seconds=0)
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -138,6 +138,25 @@ def test_base_url(server):
     assert response.url == base_url
 
 
+def test_uds(uds_server):
+    url = uds_server.url
+    uds = uds_server.config.uds
+    assert uds is not None
+    with httpx.Client(uds=uds) as http:
+        response = http.get(url)
+    assert response.status_code == 200
+    assert response.url == url
+    assert response.content == b"Hello, world!"
+    assert response.text == "Hello, world!"
+    assert response.http_version == "HTTP/1.1"
+    assert response.encoding == "iso-8859-1"
+    assert response.request.url == url
+    assert response.headers
+    assert response.is_redirect is False
+    assert repr(response) == "<Response [200 OK]>"
+    assert response.elapsed > timedelta(0)
+
+
 def test_merge_url():
     client = httpx.Client(base_url="https://www.paypal.com/")
     url = client.merge_url("http://www.paypal.com")
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -145,16 +145,8 @@ def test_uds(uds_server):
     with httpx.Client(uds=uds) as http:
         response = http.get(url)
     assert response.status_code == 200
-    assert response.url == url
-    assert response.content == b"Hello, world!"
     assert response.text == "Hello, world!"
-    assert response.http_version == "HTTP/1.1"
     assert response.encoding == "iso-8859-1"
-    assert response.request.url == url
-    assert response.headers
-    assert response.is_redirect is False
-    assert repr(response) == "<Response [200 OK]>"
-    assert response.elapsed > timedelta(0)
 
 
 def test_merge_url():
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -5,26 +5,30 @@ from httpx import AsyncioBackend, HTTPVersionConfig, SSLConfig, TimeoutConfig
 from httpx.concurrency.trio import TrioBackend
 
 
+def get_asyncio_cipher(stream):
+    return stream.stream_writer.get_extra_info("cipher", default=None)
+
+
+def get_trio_cipher(stream):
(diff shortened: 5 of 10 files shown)


httpx/concurrency/asyncio.py
```diff
@@ -232,7 +237,7 @@ def loop(self) -> asyncio.AbstractEventLoop:
     ) -> BaseTCPStream:
         try:
             stream_reader, stream_writer = await asyncio.wait_for(  # type: ignore
-                asyncio.open_connection(hostname, port, ssl=ssl_context),
+                self._open_connection(hostname, port, ssl_context),
```
florimondmanca: Is there a way we could generalize the backend interface so that it exposes a way to open a unix socket, rather than create a separate backend?

This way, users won't need to use a specific backend, and we can later look at adding this piece of functionality to the trio backend, too.

I'm thinking, maybe a new `.open_unix_socket()` method similar to the current `.open_tcp_socket()`, and then make the switch between the two in `HTTPConnection.connect()` (if I remember correctly that that's where we create a stream instance).

tomchristie:
Similarly to https://github.com/encode/httpx/issues/509#issuecomment-550463187 I’d probably like to see any support here via a third party package first, rather than including it in the core package.

I was also slightly surprised to see that the implementation was a ComcurrencyBackend class, rather than a Dispatcher implementation, which is what I would have expected to see, but I’ve not dug into that.

httpx/concurrency/asyncio.py
lundberg: Great feedback! I was thinking the same but wasn't sure if stream dispatching would fit that high in the stack.

If we dispatch between tcp and unix connection in `HttpConnection`, we need to determine what makes it select one or the other.

One example could be the `origin.scheme` being `unix` or `http+unix` to favour the unix socket route. But that affects other bits too much maybe, and also the path to the socket file needs to be passed some way, e.g. as `origin.host` which in turn hides any wanted real hostname to be passed through the unix socket.

For now, it might be better to add the unix socket path as an optional argument to the base class `ConcurrencyBackend` and implement the dispaching there in new `.open_stream()` method, which in turn calls your suggested `.open_tcp_stream()` / `.open_unix_stream()`.

That way both `AsyncioBackend`, `TrioBackend` etc gets support for the socket file feature, but one needs to instantiate the backend with the socket file arg as my current implementation in the separate backend does.

florimondmanca:
> I was also slightly surprised to see that the implementation was a ComcurrencyBackend class, rather than a Dispatcher implementation

It turns out that opening a Unix socket is very much an async library-level piece of functionality (here you see that we must call `asyncio.open_unix_connection()` instead of `asyncio.open_connection()`), so it kind of makes sense that adding support for Unix sockets requires us to play with concurrency backends.

That said, I think #133 initially discussed using a special `http+unix` scheme, which doesn't seem to be the way we're heading at here.

lundberg:
> I was also slightly surprised to see that the implementation was a ComcurrencyBackend class, rather than a Dispatcher implementation, which is what I would have expected to see, but I’ve not dug into that.

It's true that this could be implemented as a *Dispatcher*, but that would prevent the use of other existing dispatchers through a socket file, like if you're running some `wsgi` server bound to a unix socket and want to use `WSGIDispatch`.

httpx/concurrency/asyncio.py
florimondmanca: Hmm. Well, if we put aside the whole concurrency backend aspect (which is only us trying to abstract I/O functionality so that we aren't particularly tied to any async library), *maybe* there's a way to, as @tomchristie suggested, to implement this as a separate, very narrow-focused dispatcher.

It's clearer to me why it should be doable if we look at [requests-unixsocket](https://pypi.org/project/requests-unixsocket/), which provides unix sockets in the form of a `Session` subclass that uses its own adapter (a concept somewhat similar to dispatchers in HTTPX). The implementation is based on HTTP connection and connection pool subclasses. In theory we should be able to take the same path in HTTPX.

And then in terms of API, maybe aim at something like…

```python
import httpx
from httpx_unixsocket import UnixSocketAsyncDispatcher

dispatch = UnixSocketAsyncDispatcher("/var/run/docker.sock")
async with httpx.AsyncClient(dispatch=dispatch) as client:
    r = await client.get("http+unix://docker/version")
```

florimondmanca:
> like if you're running some `wsgi` server bound to a unix socket and want to use `WSGIDispatch`.

Except `WSGIDispatch` doesn't actually make network calls, does it? (It merely passes the correct `environ`/`start_response` components to the WSGI app.)

I think it's also fine that clients equiped with a Unix socket dispatcher can't make regular `http://` requests. We aren't preventing users from using multiple clients, one for each use case (e.g. HTTP/TCP, HTTP/Unix requests, HTTP/WSGI, etc).

lundberg:
> I think it's also fine that clients equiped with a Unix socket dispatcher can't make regular `http://` requests.

Maybe I'm misunderstanding, but tcp or unix socket is just the connections. A client with an available unix socket, probably needs the surrounding logic being triggered as is, like connection pooling, proxies etc, only that the actual communication in the end is over the socket file rather than over tcp?

florimondmanca:
Yeah I think you're mostly right there @lundberg. :-) Unix vs TCP is merely a matter of transport, which ideally should be transparent to the end user in terms of dispatchers.

But I also kind of agree with @tomchristie that for HTTPX's own sustainability, pushing for building third-party packages is definitely worthwhile. I reckon that currently HTTPX doesn't provide enough hooks to allow adding Unix socket support as a third party package. But it never will expose such hooks that might be useful for plenty of other use cases if we don't actually challenge the status quo and do the work of uncovering where those hooks or entrypoints should be. At least that's how I'm seeing it. :-)

So, just as a thought experiment, this is the place where I think the switch between TCP or Unix would make sense:

https://github.com/encode/httpx/blob/2fcf23bbfe41c5526cb35a65d32dd98dde786988/httpx/dispatch/connection.py#L88

If this line was refactored into a `.open_stream()` method on `HTTPConnection`, we'd be able to build a subclass of it, say `UnixHTTPConnection` that opens a Unix socket stream instead of a TCP stream.

Then, we could also refactor this part:

https://github.com/encode/httpx/blob/2fcf23bbfe41c5526cb35a65d32dd98dde786988/httpx/dispatch/connection_pool.py#L136-L145

into a `.create_connection(origin)` method, that could be overridden in a `UnixConnectionPool` subclass that creates an `UnixHTTPConnection` and also stores the socket path.

At that point, basic needs would be covered as we'd be able to use `dispatch=UnixConnectionPool()` to make connections over a Unix socket, even if we'd have a split between TCP and Unix. (*Except* that that functionality would be maintained by the community. HTTPX maintainers do have limited resources as well. :wink:)

Then, if we decide to bring the functionality in core, we'll be able to use the newly-created hooks to simplify the implementation, and we'd have pre-existing knowledge about how things work, should be tested, etc.

So, what do you all think about the idea of introducing these hooks? (`HTTPConnection.open_stream()`, `ConnectionPool.create_connection()`)

(My core intuition is that there might be a fundamental abstraction we're missing on. I think @tomchristie mentioned before that there may be key insights to learn from ASGI in that regard. I agree, but the to-be-discovered abstraction isn't here yet.)

lundberg:
> So, what do you all think about the idea of introducing these hooks? (`HTTPConnection.open_stream()`, `ConnectionPool.create_connection()`)

These are great ideas and would make extending connections a lot easier @florimondmanca!

One thought only;  
Let's say your suggested hooks exists, and in a future core/3rd `UnixHTTPConnection` class  we can override the suggested `open_stream()` method to open a unix socket connection. We still would want to use the client's given/selected backend, like `AsyncioBackend` or `TrioBackend`, leading us to use `self.backend.open???` in the implementation of `???Connection.open_stream()`, right?

So even with the hooks in place, the `ConcurrencyBackend` still, *in my opinion*, needs both tcp and unix interface methods and implementations in related backends, since it's where the actual transport connections currently are being made.

florimondmanca:
> We still would want to use the client's given/selected backend, like `AsyncioBackend` or `TrioBackend`

Well, I think that's how we'd to things if we were to implement this in core, but that's not an immediate requirement for a 3rd party implementation.

The 3rd party package could just have its own `open_unix_socket()` helper, and use it there instead of calling into the concurrency backend.

It doesn't even *have* to support `trio` right away — asyncio-only can be defined as a constraint, which might be resolved by someone contributing trio support later. :-)

florimondmanca:
The only bit that might be off is that right now HTTP connections manipulate `TCPStream` instances. There's nothing particularly "TCP" about this interface — what's important is that it's a *socket* of some sort, that exposes read/write operations. So *maybe* we should rename that to `SocketStream` or something? (This is the terminology that trio uses.)

---


> It doesn't even _have_ to support `trio` right away — asyncio-only can be defined as a constraint, which might be resolved by someone contributing trio support later. :-)

Just to solidify this claim, trio also has [`open_unix_socket()`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.open_unix_socket) so extending an initial asyncio-only implementation should be easy enough. :)

lundberg:
> The 3rd party package could just have its own `open_unix_socket()` helper, and use it there instead of calling into the concurrency backend.

This is true @florimondmanca , but i really think that an http client should have the option of connecting through a unix socket in its core, since it's part of pythons core to support both tcp and unix socket connections, which in turn is transparent of what protocol being used, HTTP in httpx's case.

>  I’d probably like to see any support here via a third party package first, rather than including it in the core package.

Let's take `curl` as an example, which has the option `--unix-socket` to set transport over a specific unix socket file instead of TCP. That said, I think the URL shouldn't be used to hold information about any socket file or mess with the scheme, since it's still http/https but over a unix socket.

Another example is `uvicorn`, which has the option `--uds` to bind to a unix domain socket, why wouldn't we want `httpx` to have core support to request uvicorn that way @tomchristie ?  
*By the way, this is how I test unix socket in this PR, i.e. by starting uvicorn bound to an uds file.*

tomchristie:
I'm certainly okay with us *considering* this functionality for core (tho we may want to think about the top-level API if that's what we're doing.) I'd expect us to have an `open_uds_connection` on the concurrency backend, and then expose `usd=...` on the client, and have the connection pool switch between `open_tcp_connection` or `open_uds_connection` depending on if `uds` is set.

lundberg:
> I'd expect us to have an `open_uds_connection` on the concurrency backend, and then expose `usd=...` on the client, and have the connection pool switch between `open_tcp_connection` or `open_uds_connection` depending on if `uds` is set.

Spot on @tomchristie , this was my initial idea but started easy with the previous backend. I have now refactored accordingly! Any suggestions of more tests that should be done?

> The only bit that might be off is that right now HTTP connections manipulate `TCPStream` instances. There's nothing particularly "TCP" about this interface — what seems to be important is that it's a _socket_ of some sort, that exposes read/write operations. So _maybe_ we should rename that to `SocketStream` or something? (This is the terminology that trio uses.)

Since backends now return both tcp and uds streams, your renaming suggestion `TCPStream` -> `SocketStream` makes sense @florimondmanca . Should I do this aswell?

florimondmanca:
> Should I do this aswell?

Thanks for suggesting! If you’re up for it let’s try to get this in via a separate PR maybe? :)

I’ll try to have a new look at this one soon!

lundberg:
> If you’re up for it let’s try to get this in via a separate PR maybe? :)

Well, a new PR would be ok i guess, but since the only code instantiating and returning `TCPStream` are these `open_???_steam` methods I'd say it belongs to this PR, OK @florimondmanca ?

Question, do we need to do this *rename* as a non-breaking change, like keeping the old class and deprecate it but make it extend the new new named one? Don't know if any 3rd party package directly would import `BaseTCPStream`. Maybe a simple `from .concurrency.base import BaseSocketStream, BaseSocketStream as BaseTCPStream` is enough during deprecation window, if needed at all?

florimondmanca:
> I'd say it belongs to this PR, OK @florimondmanca ?

It would be easier for all to propose this change separately.

We tend to encourage focused and isolated PRs as much as possible, even if that means having to deal with changes/conflicts with another pending PRs, as it facilitates reviewing and auditing. Two isolated PRs tend to be reviewed more efficiently, and merged more quickly, than a single PRs that tries to add/change two things at once. :-)

FYI, a rename has already happened in the past, if it can be of inspiration: #339

> do we need to do this rename as a non-breaking change

Probably not — the only public aspect about concurrency backends is that class name and the import path; methods and internal classes like `TCPStream` aren't used directly by our users. (Anyway we're in 0.x, so I think backwards compatibility / strict semver isn't a concern yet. :-) e.g. #482.)

lundberg:
> It would be easier for all to propose this change separately.

Note taken, I'll do a separate PR then.

> FYI, a rename has already happened in the past, if it can be of inspiration: #339

`Stream` -> `TCPStream` -> `SocketStream`  ;-)

> > do we need to do this rename as a non-breaking change
> 
> Probably not

Ok, great. Only raised the question since it is exposed in `__init__`, i.e. `httpx.BaseTCPStream`

lundberg:
Rebased and in line with #517

florimondmanca:
@lundberg Sorry to have let this drag a bit. I'm just a bit wary on the attempt to DRY out the "open_x_stream" methods — that's probably one level of indirection too much. Having to pass awaitables (being paranoid — who knows if they'll actually be awaited?) also feels a bit strange. Can we keep them separate, for the sake of explicitness and clarity?

lundberg:
> I'm just a bit wary on the attempt to DRY out the "open_x_stream" methods

Just to clarify, I moved the timeout and ssl logic into the generic "private" `_open_stream` method just to not duplicate that code, since it's the same for both the `uds` and `tcp` coroutines, in both backends. But intentionally did not add it to the "interface" `ConcurrencyBackend`, to not force it to be implemented like that.

But, I have no problem removing the `_open_stream` methods. Should I @florimondmanca?

florimondmanca:
> Should I?

I think so, if that sounds OK to you. :)

lundberg:
Fixed and rebased @florimondmanca

tests/client/test_client.py
```diff
@@ -138,6 +138,25 @@ def test_base_url(server):
...
+    assert response.url == url
+    assert response.content == b"Hello, world!"
+    assert response.text == "Hello, world!"
+    assert response.http_version == "HTTP/1.1"
+    assert response.encoding == "iso-8859-1"
+    assert response.request.url == url
+    assert response.headers
+    assert response.is_redirect is False
+    assert repr(response) == "<Response [200 OK]>"
+    assert response.elapsed > timedelta(0)
```
florimondmanca: Do we *really* need all these asserts, or would just ones on `status_code`, `.text`, `.encoding` (if that's uds-specific?) be enough? The rationale for me is that we're *only* testing that we're able to make a request via a Unix domain socket, not the entire response API. What do you think?

tests/client/test_async_client.py
```diff
@@ -146,3 +146,17 @@
...
+    uds = uds_server.config.uds
+    assert uds is not None
+    async with httpx.AsyncClient(backend=backend, uds=uds) as client:
+        response = await client.get(url)
+    assert response.status_code == 200
+    assert response.text == "Hello, world!"
+    assert response.http_version == "HTTP/1.1"
+    assert response.headers
+    assert repr(response) == "<Response [200 OK]>"
+    assert response.elapsed > timedelta(seconds=0)
```
florimondmanca: Same comment than in `test_client.py` — are all these extra asserts necessary w.r.t. strictly testing uds usage?

httpx/client.py
```diff
@@ -721,6 +723,7 @@ class Client(BaseClient):
     async requests.
     * **trust_env** - *(optional)* Enables or disables usage of environment
     variables for configuration.
+    * **uds** - *(optional)* A path to a Unix domain socket to connect through
```
florimondmanca: Copy nit :)

```suggestion
    * **uds** - *(optional)* A path to a Unix domain socket to connect through.
```

tests/test_concurrency.py
```diff
@@ -5,36 +5,46 @@
...
+    return stream.stream_writer.get_extra_info("cipher", default=None)
+
+
+def get_trio_cipher(stream):
+    return stream.stream.cipher() if isinstance(stream.stream, trio.SSLStream) else None
+
+
 @pytest.mark.parametrize(
-    "backend, get_cipher",
+    "backend, test_uds, get_cipher",
```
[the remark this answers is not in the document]
lundberg: Thats a much better use of parametrize!

tests/test_concurrency.py
```diff
@@ -5,36 +5,46 @@
...
     [
         pytest.param(
-            AsyncioBackend(),
-            lambda stream: stream.stream_writer.get_extra_info("cipher", default=None),
-            marks=pytest.mark.asyncio,
+            AsyncioBackend(), False, get_asyncio_cipher, marks=pytest.mark.asyncio
```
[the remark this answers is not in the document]
lundberg: Agree, only changed those lambdas to not duplicate them for both `test_uds` *True* and *False*. Will revert to the lambdas.

tests/client/test_client.py
lundberg: No, I guess not. Just copy-pasted the tcp-asserts for safety, will leave out the redundant assertions.

lundberg:
> I left several final comments on the tests, but nothing critical.

Legit comments, thanks. Refactored and pushed @florimondmanca .

florimondmanca:
We may want to add some prose to the "Advanced" section of the docs, to quickly demonstrate why and how to use Unix domain sockets. A short paragraph and a code snippet should do the trick. Would you like to tackle that as well, @lundberg? Up to you whether to add it to this PR, or we can merge and add the docs via a separate PR. Let me know!

lundberg:
> We may want to add some prose to the "Advanced" section of the docs....Up to you whether to add it to this PR, or we can merge and add the docs via a separate PR. Let me know!

Let's merge this. I'll happily open a separate PR for the docs section, will start working on it now.

Issue #349, opened 2019-09-16
[Feature Request]: raise httpx exception instead of buildin ConnectionResetError and ConnectionError


Trim21:
This is a feature requests instead of bug report.

I'm using `requests` in my web server to request upstream api. So I can handler all `requests.RequestException` in my middleware and return a `502 upstream` error.

But after I migrated to `httpx`, httpx doesn't have such exception when issuing a http(s) request, it will raise `httpx.ConnectTimeout`, buildin `ConnectionError` and `ConnectResetError`. And last 2 exception are not only raised by `httpx`.

Maybe `httpx` should also wrap `ConnectionError` and other buildin connection exception?

florimondmanca:
Hi @Trim21,

I think we're going to need a bit more context to address this issue. :-)

- What were you trying to do?
- Can you share a minimal reproducible example?
- A traceback of the error?

Thanks!

Trim21:
@florimondmanca Sorry, I submit this issue by accident and I'm still writting details.

florimondmanca:
Thanks for updating the description. :) I'm not entirely sure on this myself, I'll ping @tomchristie on this one.

sethmlarson:
Probably because we don't capture any `socket.error` or `socket.gaierror`.

florimondmanca:
I remember that Trio already wraps errors such as ConnectionResetError into its own exceptions, eg BrokenResourceError. That means whatever wrapping we do needs to occur at the concurrency backend level. :)

Trim21:
@florimondmanca I'm sorry, I'm not familiar with `trio`. What's the relation between `trio` and `httpx`?

florimondmanca:
@Trim21 HTTPX supports the async/await syntax, which under the hood requires us to work with a library able to perform asynchronous operations. Right now we only support asyncio (which is part of the standard library), but we’ve got plans for supporting alternatives such as trio - see #120. The « concurrency backend » I mentioned in my comment is an internal interface we use to abstract away from the individual async libraries.

I guess the point in my comment was that which exceptions we need to wrap may depend on the async library in use, so we should do the wrapping at the concurrency backend level.

Trim21:
@florimondmanca thanks for your explanation

Trim21:
will this be a feature of 1.0 ?

sethmlarson:
It'll probably be in before 1.0, just requires a contributor to do the work! :)

tomchristie:
[...]

PR #512, opened 2019-11-06
Improve robustness of live HTTP/2 test

florimondmanca:
Right now the `master` build is [failing](https://travis-ci.org/encode/httpx/builds/608437975) because the live HTTP/2 request made to nghttp2.org is timing out, probably because the website is suffering some issues right now.

This PR makes the test xfail on some predictable failure cases.

Note that the `socket.gaierror` block is not ideal — when resolving #349 we should switch to whichever HTTPX-native exception is introduced.

### DIFF
--- a/tests/dispatch/test_http2.py
+++ b/tests/dispatch/test_http2.py
@@ -1,10 +1,12 @@
 import json
+import socket
 
 import h2.connection
 import h2.events
+import pytest
 from h2.settings import SettingCodes
 
-from httpx import AsyncClient, Client, Response
+from httpx import AsyncClient, Client, Response, Timeout
 
 from .utils import MockHTTP2Backend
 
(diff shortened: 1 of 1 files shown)


florimondmanca:
I triggered a new build on master, which is now fixed, and the build here passed without xfailing. However I was bitten by the issue in #510, so I'd say this change still has value.

florimondmanca:
@yeraydiazdiaz Yes, though the main concern for #412 was possible overhead associated to using a non-Encode project for one single test. Uvicorn HTTP/2 development might be starting again soon ( https://github.com/encode/uvicorn/issues/47), though there are no guarantees when that will land.

yeraydiazdiaz:
Right, I just feel I'd rather have a dependency on a non-Encode project for a while, potentially until HTTP/2 support lands in Uvicorn, than having a dependency on an external service. 🙂

PR #513, opened 2019-11-07
Allow using Client in an async environment

### LINKED ISSUE #508: Incompatible with jupyter
```
import httpx
httpx.get('https://httpbin.org/get')
```

When run from Jupyter this snippet produces "Cannot run the event loop while another loop is running" error. Works fine in REPL. Apparently the use of asyncio conflicts with Jupters's own event loop.

httpx version: 0.7.6
os: Fedora 29
python: Python 3.6.9
ipykernel: 5.1.2
ipython: 7.8.0
jupyter: 5.4.1

florimondmanca:
Fixes #508 

This is an attempt at handling the situation in which users try to use `httpx.get()` or `httpx.Client()` in an async environment that has a running loop (e.g. in a Jupyter notebook).

I documented several potential caveats in the form of code comments.

I still need to verify this actually solves the Jupyter use case, though the tests give me hope that it does. :-)

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -1,12 +1,16 @@
 import asyncio
+import concurrent.futures
 import functools
 import ssl
 import sys
+import threading
+import time
 import typing
 from types import TracebackType
 
 from ..config import PoolLimits, TimeoutConfig
 from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
+from ..utils import get_logger
 from .base import (
     BaseBackgroundManager,
     BaseEvent,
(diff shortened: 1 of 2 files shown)


florimondmanca:
@encode/httpx-maintainers OK, I think I've got something that passes the tests. Basically we're checking if the event loop is running and when relevant run the coroutine in a worker thread instead. This thing is inspired by `asgiref`'s `AsyncToSync` utility. Absurdly complex, but at least it  seems to work.

florimondmanca:
Closing as I don't think this will make it, probably in favor of a `SyncBackend` (unless #522 is accepted) - see https://github.com/encode/httpx/issues/508#issuecomment-552079957.

PR #516, opened 2019-11-07
changed timeout, cert, verify, trust_env args as client instance.

### LINKED ISSUE #505: trust_env is partially ignored for httpx.get calls
I have a socks5 proxy used in the system and it's not supported by httpx. `all_proxy` environmental variable points to socks5 proxy, therefore `httpx.get('https://httpbin.org/get')` call results in an error (Unknown proxy for URL). 

To bypass that I specify trust_env=False like this `httpx.get('https://httpbin.org/get', trust_env=False)`, but still get the same error.

I looked up the request method implementation in the api.py and apparently when the Client instance is created none of the request method parameters are being passed to the constructor. 

https://github.com/encode/httpx/blob/master/httpx/api.py#L80

Therefore inside the BaseClient's __init__ method trust_env has the default value (True). And accordingly to this value tries to deduce the proxy settings from the environment variable, which causes the error.

It would be great if you passed the trust_env value to __init__ when creating Client's instance.

cansarigol:
Closes #505

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -265,17 +265,13 @@ You can set timeouts on two levels:
 # Using top-level API
 httpx.get('http://example.com/api/v1/example', timeout=5)
 
-# Or, with a client:
-with httpx.Client() as client:
-    client.get("http://example.com/api/v1/example", timeout=5)
+# Or, with a client instance:
+with httpx.Client(timeout=5) as client:
+    client.get("http://example.com/api/v1/example")
 ```
 
 - On a client instance, which results in the given `timeout` being used as a default for requests made with this client:
 
-```python
-with httpx.Client(timeout=5) as client:
-    client.get('http://example.com/api/v1/example')
-```
 
 Besides, you can pass timeouts in two forms:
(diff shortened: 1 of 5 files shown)


tomchristie:
This PR is a much bigger footprint that I was expecting for #505 - I don't think it matches what I'd intended there. The only changes ought to be in the `request` function of `api.py`.

In that function, `timeout` and other arguments should be passed to the client instantiation, rather than to the call.

(Although If I've missed or misunderstood something that's needed here then happy to talk that through.)

cansarigol:
Ahh sorry, I thought that you want to change to pass these arguments as Client instance. 
> The only changes ought to be in the request function of api.py.

api.request calls client.request with these arguments. this PR is to change this per-request approach. I will check the issue again.

florimondmanca:
Hi here, I think we can close this one in favor of #505. Thanks for initially tackling this @cansarigol. :)

cansarigol:
I couldn't find any time to pursue this. Thanks

PR #517, opened 2019-11-08
Rename TCPStream to SocketStream

lundberg:
This PR renames `BaseTCPStream`, backend subclasses `asyncio.TCPStream` and `trio.TCPStream`, to `BaseSocketStream` / `SocketStream`, to not limit name/usage to TCP specific streams.

Related to #511

### DIFF
--- a/tests/dispatch/utils.py
+++ b/tests/dispatch/utils.py
@@ -5,7 +5,7 @@ import h2.config
 import h2.connection
 import h2.events
 
-from httpx import AsyncioBackend, BaseTCPStream, Request, TimeoutConfig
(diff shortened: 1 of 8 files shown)


tests/dispatch/utils.py
```diff
@@ -189,13 +189,13 @@ def __getattr__(self, name: str) -> typing.Any:
         return getattr(self.backend, name)
 
 
-class MockRawSocketStream(BaseTCPStream):
+class MockRawSocketStream(BaseSocketStream):
```
florimondmanca: Us already using the term “socket stream” in this mock seems like a good indication that renaming is the right thing to do. :)

PR #520, opened 2019-11-08
Files without a filename should not set a Content-Type in multipart data.

### LINKED ISSUE #519: No way to replicate specific multipart requests POST?
With the requests library, I could do something like this:
```
sub = {'field1': (None, 'test')}
requests.post('https://enygrrtwf27.x.pipedream.net', files=sub)
```
And get this result:
```
--13489f816b24eea1f1ae00720a15d8da
Content-Disposition: form-data; name="field1"

test
--13489f816b24eea1f1ae00720a15d8da--
```
<details>
<summary>
Reading through the docs, it's unclear how I would replicate this. Here are some of the methods I've tried with `httpx.post(url, files=sub)` and their results:</summary>

#### Input
```
sub = {'field1': (None, 'test', None)}
```
#### Ouput
```
AttributeError: 'NoneType' object has no attribute 'encode'
```

#### Input
```
sub = {'field1': 'test'}
```
#### Output
```
--d0eeaee30c2025a463cc2eebc6455a22
Content-Disposition: form-data; name="field1"; filename="upload"
Content-Type: application/octet-stream

test
--d0eeaee30c2025a463cc2eebc6455a22--
```

#### Input
```
sub = {'field1': (None, 'test')}
```
#### Output
```
--7ef3cb0ed2b1789c79c2ec46db9aa4bb
Content-Disposition: form-data; name="field1"
Content-Type: application/octet-stream

test
--7ef3cb0ed2b1789c79c2ec46db9aa4bb--
```

#### Input
```
sub = {'field1': (None, 'test', 'text/plain')}
```
#### Output
```
--7ef3cb0ed2b1789c79c2ec46db9aa4bb
Content-Disposition: form-data; name="field1"
Content-Type: application/octet-stream

test
--7ef3cb0ed2b1789c79c2ec46db9aa4bb--
```
</details>

I think the best way to rectify this would be to allow `None` as a `Content-Type` and then just not set any `Content-Type`

Mattwmaster58:
Fixes #519

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -84,17 +84,17 @@ RequestData = typing.Union[dict, str, bytes, typing.Iterator[bytes]]
 RequestFiles = typing.Dict[
     str,
     typing.Union[
-        typing.IO[typing.AnyStr],  # file
+        # file
+        typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+        # (filename, file (or str))
+        typing.Tuple[
+            typing.Optional[str], typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+        ],
+        # (filename, file (or str), content_type)
         typing.Tuple[
             typing.Optional[str],
-            typing.Union[
-                typing.IO[typing.AnyStr], typing.AnyStr
-            ],  # (filename, file (or str))
-            typing.Tuple[
-                typing.Optional[str],
-                typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
-                typing.Optional[str],
-            ],  # (filename, file (or str), content_type)
+            typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+            typing.Optional[str],
         ],
     ],
 ]
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -84,7 +84,7 @@ RequestData = typing.Union[dict, str, bytes, typing.Iterator[bytes]]
 RequestFiles = typing.Dict[
     str,
     typing.Union[
-        # file
+        # file (or str)
         typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
         # (filename, file (or str))
         typing.Tuple[
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -85,10 +85,17 @@ RequestFiles = typing.Dict[
     str,
     typing.Union[
         typing.IO[typing.AnyStr],  # file
-        typing.Tuple[str, typing.IO[typing.AnyStr]],  # (filename, file)
         typing.Tuple[
-            str, typing.IO[typing.AnyStr], str
-        ],  # (filename, file, content_type)
+            typing.Optional[str],
+            typing.Union[
+                typing.IO[typing.AnyStr], typing.AnyStr
+            ],  # (filename, file (or str))
+            typing.Tuple[
+                typing.Optional[str],
+                typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+                typing.Optional[str]
+            ],  # (filename, file (or str), content_type)
+        ],
     ],
 ]
 
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -93,7 +93,7 @@ RequestFiles = typing.Dict[
             typing.Tuple[
                 typing.Optional[str],
                 typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
-                typing.Optional[str]
+                typing.Optional[str],
             ],  # (filename, file (or str), content_type)
         ],
     ],
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -136,9 +136,8 @@ def test_multipart_encode_files_allows_filenames_as_none():
 
         assert content_type == f"multipart/form-data; boundary={boundary}"
         assert body == (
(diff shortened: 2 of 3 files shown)


Mattwmaster58:
I hadn't even considered type annotations, I will add that ASAP. Setting `guess_content_type` to return `None` when filename is `None` would make this feature 1:1 with requests, and it seems to have worked out just fine for them over the years.

Mattwmaster58:
Looks like I auto-formatted with my IDE, will fix this later.

tests/test_multipart.py
```diff
@@ -136,9 +136,8 @@ def test_multipart_encode_files_allows_filenames_as_none():
 
         assert content_type == f"multipart/form-data; boundary={boundary}"
         assert body == (
-            '--{0}\r\nContent-Disposition: form-data; name="file"\r\n'
-            "Content-Type: application/octet-stream\r\n\r\n<file content>\r\n"
-            "--{0}--\r\n"
+            '--{0}\r\nContent-Disposition: form-data; name="file"\r\n\r\n'
```
sethmlarson: Partially unrelated to this change, but would you be filling to add a test case that exercises the mimetypes guess? (Maybe `example.json` -> `Content-Type: application/json` and `test.log` -> `Content-Type: application/octet-stream`?)

httpx/models.py
```diff
@@ -85,10 +85,17 @@
...
+            typing.Optional[str],
+            typing.Union[
+                typing.IO[typing.AnyStr], typing.AnyStr
+            ],  # (filename, file (or str))
+            typing.Tuple[
+                typing.Optional[str],
+                typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+                typing.Optional[str],
+            ],  # (filename, file (or str), content_type)
+        ],
```
florimondmanca: I don't think these annotations are correct, are they? If we want to allow a file, a (optional filename, file/str), or an (optional filename, file/str, content_type), this should probably be:

```python
typing.Union[
    # file
    typing.IO[typing.AnyStr],
    # (filename, file (or str))
    typing.Tuple[
        typing.Optional[str],
        typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
    ],
    # (filename, file (or str), content_type)
    typing.Tuple[
        typing.Optional[str],
        typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
        typing.Optional[str],
    ],
]
```

tests/test_multipart.py
```diff
@@ -136,13 +136,34 @@ def test_multipart_encode_files_allows_filenames_as_none():
...
+        boundary = binascii.hexlify(os.urandom(16)).decode("ascii")
+
+        body, content_type = multipart.multipart_encode(data={}, files=files)
+
+        assert content_type == f"multipart/form-data; boundary={boundary}"
+        assert body == (
+            '--{0}\r\nContent-Disposition: form-data; name="file"; filename="{2}"\r\n'
+            "Content-Type: {1}\r\n\r\n<file content>\r\n"
+            "--{0}--\r\n"
+            "".format(boundary, expected_content_type, file_name).encode("ascii")
```
florimondmanca: Would 3.6+ f-strings be a tad more readable here?

```python
assert body == (
    f'--{boundary}\r\nContent-Disposition: form-data; name="file"; filename="{filename}"\r\n'
    f"Content-Type: {expected_content_type}\r\n\r\n<file content>\r\n"
    f"--{boundary}--\r\n"
).encode("ascii")
```

Mattwmaster58: Definitely would. I was trying to mirror the original style of code by copying from a similar test, but this is far better.

httpx/models.py
Mattwmaster58: When I diffed your annotation with the current ones they seemed to be identical, just different comment positions? Either way, the comment before the specific type annotation is more readable IMO so we'll go with that.

I just noticed this now: shouldn't 
```
    # file
    typing.IO[typing.AnyStr],
```
be
```
   # file
   typing.Union[typing.IO[typing.AnyStr], typing.AnyStr]
```
Since we do allow strings instead of IO objects?

tomchristie:
So this looks good... I've got one question.

It looks like there's two seperate changes here:

* No filename -> No Content-Type header, instead of `application/octet-stream`.
* Changing the type annotation to indicate that the file can either be passed as a plain str argument, or as  a file object.

I'm totally on board with the first one of those, since that's exactly what this ticket describes.

Wrt. the second one - is that because we've got the existing annotation incorrect there at the moment?

httpx/models.py
```diff
@@ -84,11 +84,18 @@
 RequestFiles = typing.Dict[
     str,
     typing.Union[
-        typing.IO[typing.AnyStr],  # file
-        typing.Tuple[str, typing.IO[typing.AnyStr]],  # (filename, file)
+        # file
```
tomchristie: Let's mirror the other comments precisely and use "file (or str)" here.

Mattwmaster58:
Sorry, I should've clearly stated what I was doing with the type annotations. First modified them to allow for Content-Type to be set to `None`  and to allow for `filename` to be `None`. 

I also noticed that we allow `typing.AnyStr` instead of `typing.IO[typing.AnyStr]` (there's even a test written for it: `test_multipart_encode_files_allows_str_content`), so I updated the type annotations to match that.

tomchristie:
Crackin' bit of work this.
Thank you! 🍾

PR #521, opened 2019-11-11
Backport start_tls() support for Python 3.6

### LINKED ISSUE #515: HTTPX claims Python 3.6+ support when asyncio backend TCPStream.start_tls explicitly does not support Python 3.6
As stated, HTTPX explicitly does not support Python 3.6 [here](https://github.com/encode/httpx/blob/master/httpx/concurrency/asyncio.py#L61) but declares Python 3.6 support [here](https://github.com/encode/httpx/blob/master/setup.py#L40). This has caused problems with users of a package of mine that is supposed to support Python 3.6.

sethmlarson:
This adds support for proxy tunnels for Python 3.6 + asyncio.
Based almost entirely on this comment: https://github.com/urllib3/urllib3/issues/1323#issuecomment-362494839 Massive thanks to @florimondmanca for finding this!

Closes #515.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -58,16 +58,49 @@ class SocketStream(BaseSocketStream):
         self, hostname: str, ssl_context: ssl.SSLContext, timeout: TimeoutConfig
     ) -> "SocketStream":
         loop = asyncio.get_event_loop()
-        if not hasattr(loop, "start_tls"):  # pragma: no cover
-            raise NotImplementedError(
-                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
-            )
 
         stream_reader = asyncio.StreamReader()
         protocol = asyncio.StreamReaderProtocol(stream_reader)
         transport = self.stream_writer.transport
 
-        loop_start_tls = loop.start_tls  # type: ignore
+        if hasattr(loop, "start_tls"):
+            loop_start_tls = loop.start_tls  # type: ignore
+        else:
+
+            async def loop_start_tls(
+                transport: asyncio.BaseTransport,
+                protocol: asyncio.BaseProtocol,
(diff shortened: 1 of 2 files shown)


tomchristie:
Fantastic stuff!

I guess this PR should also update the [`test_start_tls_on_socket_stream`](https://github.com/encode/httpx/blob/1a32cf036a825f6eb35395af5388a3b23180a82e/tests/test_concurrency.py#L35) test case, right?

tomchristie:
At a later date I may look at moving this out to a coverage-excluded `compat.py` module, so that we can tweak our tests to expect 100% coverage on any single run, without requiring collection across multiple Python versions. (Buit I'm very happy to treat that seperately)

httpx/concurrency/asyncio.py
```diff
@@ -58,16 +58,49 @@ def __init__(
...
-                "asyncio.AbstractEventLoop.start_tls() is only available in Python 3.7+"
-            )
 
         stream_reader = asyncio.StreamReader()
         protocol = asyncio.StreamReaderProtocol(stream_reader)
         transport = self.stream_writer.transport
 
-        loop_start_tls = loop.start_tls  # type: ignore
+        if hasattr(loop, "start_tls"):
+            loop_start_tls = loop.start_tls  # type: ignore
```
JayH5: Removing the `type: ignore` after changing the if statement to be `if sys.version_info >= (3, 7)` should pass CI with `mypy>=0.730`.

sethmlarson: Unfortunately we can't assume that loops on Python 3.6 don't have start_tls(). Uvloop supports 3.6 and supports start_tls(). That's interesting that mypy knows about features though! Would an assert work here?

tomchristie: getattr would work around any typing issues too.

PR #528, opened 2019-11-14
pass on cert, verify, timeout, trust_env to Client

AtomsForPeace:
Think this is what's needed for #505

I'd love to add a test for it but couldn't think of any sensible way of doing. Tried messing around with mocking the Client object and seeing the attributes that were set in the request function call but didn't pan out as hoped.

Any suggestions for tests?

I'd be happy to add one.

Let me know if I should remove timeout, cert, verify and leave only trust_env.

### DIFF
--- a/httpx/api.py
+++ b/httpx/api.py
@@ -80,7 +80,13 @@ def request(
     <Response [200 OK]>
     ```
     """
-    with Client(http_versions=["HTTP/1.1"]) as client:
+    with Client(
+        http_versions=["HTTP/1.1"],
+        cert=cert,
+        verify=verify,
+        timeout=timeout,
+        trust_env=trust_env,
+    ) as client:
         return client.request(
             method=method,
             url=url,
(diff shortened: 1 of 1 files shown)


AtomsForPeace:
Hmmm, I see that mypy is failing.

Would it possibly be better to have the defaults for timeout and trust_env the same in request as they are in Client?

Client has trust_env: bool = True but request has trust_env: bool = None.

That would solve the type issue.

tomchristie:
> Hmmm, I see that mypy is failing.

Right - this PR should also change the signatures in `api.py` so that the defaults match the defaults we're using on `Client`. That ought to resolve the issue.

Eg. we want `trust_env: bool = True` in our `api.py` signatures.

tomchristie:
Oh right, I'd skimmed this. Yes your observation here is spot on...

> Would it possibly be better to have the defaults for timeout and trust_env the same in request as they are in Client?

😃 Absolutely, yes. Go for it!

AtomsForPeace:
Ready to merge?

tomchristie:
Thanks for having put in the work here. Incorperated into https://github.com/encode/httpx/pull/544

PR #529, opened 2019-11-15
Fix redirect cookie behavior

tomchristie:
Cookies are not being properly handled by redirect responses.
(Specifically, expired cookies in a response, are still included in any redirect request)

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -6,7 +6,6 @@ brotlipy==0.7.*
 cryptography
 flake8
 flake8-bugbear
-flake8-comprehensions
 flake8-pie
 isort
 mypy
--- a/tests/client/test_redirects.py
+++ b/tests/client/test_redirects.py
@@ -301,7 +301,6 @@ class MockCookieDispatch(AsyncDispatcher):
         timeout: TimeoutTypes = None,
     ) -> AsyncResponse:
         if request.url.path == "/":
-            print(request.headers)
             if "cookie" in request.headers:
                 content = b"Logged in"
             else:
--- a/tests/client/test_redirects.py
+++ b/tests/client/test_redirects.py
@@ -312,7 +312,10 @@ class MockCookieDispatch(AsyncDispatcher):
             status_code = codes.SEE_OTHER
             headers = {
                 "location": "/",
-                "set-cookie": "session=eyJ1c2VybmFtZSI6ICJ0b21; path=/; Max-Age=1209600; httponly; samesite=lax",
+                "set-cookie": (
+                    "session=eyJ1c2VybmFtZSI6ICJ0b21; path=/; Max-Age=1209600; "
+                    "httponly; samesite=lax"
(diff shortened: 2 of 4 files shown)


tomchristie:
Needs a test case really.
Linter failures are unrelated. https://travis-ci.org/encode/httpx/jobs/612451968?utm_medium=notification&utm_source=github_status (grumble grumble 🤣)

florimondmanca:
The desired behavior here is that any cookies on a request should be dropped in case of redirect, right?

test-requirements.txt
```diff
@@ -6,7 +6,6 @@ brotlipy==0.7.*
 cryptography
 flake8
 flake8-bugbear
-flake8-comprehensions
```
StephenBrown2: Why drop this plugin?

tomchristie:
> The desired behavior here is that any cookies on a request should be dropped in case of redirect, right?

Not quite. It's that whatever the *client* has in it's cookie store is what we should use.

The case where this issue occurs is going to typically be in a logout flow...

1. A logged-in httpx client issues a POST request to the logout URL, including it's current session cookie.
2. Server clears the session cookie (sets an expired cookie on the response) and returns a redirect (eg. back to the homepage/login page).
3. The httpx client makes the homepage redirect request, but because it has cloned the headersof the original request, it erronously includes a cookie on the request, even tho the client cookie store cleared properly when it got the response at (2)

test-requirements.txt
tomchristie: It's unneccessarily causing breakages...

Eg. expressions such as...

`self.content = b"".join([part async for part in self.stream()])`

Are failing with "C416 Unnecessary list comprehension - rewrite using list()."

tomchristie:
Righty - I'm happy for this to be reviewed now.
The test cases fail without the two-line fixes, and pass after them.

tests/client/test_redirects.py
```diff
@@ -290,3 +290,72 @@ class MockDispatch(AsyncDispatcher):
...
+class MockCookieDispatch(AsyncDispatcher):
+    async def send(
+        self,
+        request: AsyncRequest,
+        verify: VerifyTypes = None,
+        cert: CertTypes = None,
+        timeout: TimeoutTypes = None,
+    ) -> AsyncResponse:
+        if request.url.path == "/":
+            print(request.headers)
```
florimondmanca: Gotcha :)

```suggestion
```

test-requirements.txt
StephenBrown2: Can be resolved with `# noqa: C416`, as we want to keep the comprehension _there_, but perhaps not other places (non-async comprehensions).

florimondmanca: Related: https://github.com/adamchainz/flake8-comprehensions/issues/196

adamchainz: Fix released if you want to add it back :)

PR #531, opened 2019-11-15
Drop flake8 and friends

tomchristie:
Some discussion on this over at https://gitter.im/encode/community
In short. I don't think flake8 et. al. are working in our favour, and are giving us more busy work than benefit.

No immediate decisions neccesary here. I will aim to provide some proper context/motivation but just wanted to start by issuing the PR.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -10,7 +10,7 @@ source_files = ("httpx", "tools", "tests", "setup.py", "noxfile.py")
 @nox.session
 def lint(session):
     session.install(
-        "--upgrade", "autoflake", "black", "flake8", "isort", "seed-isort-config"
+        "--upgrade", "autoflake", "black", "isort", "seed-isort-config"
     )
 
     session.run("autoflake", "--in-place", "--recursive", *source_files)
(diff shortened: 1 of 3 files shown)


florimondmanca:
I think we'd need to judge whether only having `mypy` as a linter is enough for day-to-day development and ensuring overall quality.

Some counter-arguments I can think of in favor of not dropping `flake8`:

- There are *some* code quality cases that `flake8` catches while `mypy` doesn't (e.g. unused variables).
- If we keep autoflake in the lint pipeline, what guarantees us that `autoflake` is actually run periodically (e.g. before pushing to the remote)?

yeraydiazdiaz:
I had to look up what those `flake8-*` packages do:

- https://pypi.org/project/flake8-bugbear/
- https://pypi.org/project/flake8-pie/
- https://pypi.org/project/flake8-comprehensions/

I think the first two are probably good to have, `flake8-comprehensions` might be a bit much.

I wouldn't mind dropping all of them but I would definitely miss the standard `flake8`. That being said I routinely ignore flake8 issues that interfere with black. I'd prefer tweaking the flake8 configuration rather than dropping it completely.

StephenBrown2:
> I'd prefer tweaking the flake8 configuration rather than dropping it completely.

Ditto.

florimondmanca:
Seems `flake8-comprehensions` is gone (probably part of #544 or related PRs?). As I wrote previously, personally I'd like to keep flake8 (and it seems others do too), so is this a close? 😄

florimondmanca:
@tomchristie I'll close but if you have pending arguments against flake8 let's rediscuss!

tomchristie:
Sure. There might be some remaining tidy up we can do here, but I don't think there's anything pressing.

PR #534, opened 2019-11-17
Document when to use AsyncClient

### LINKED ISSUE #179: Use the existing runloop for sync calls in async handler
Hi, this library looks awesome and I really want to use it; but I seem to have hit a blocker and can't seem to find a solution in the docs.

TL;DR: Is there any way to provide our own runloop when using the `sync` calls as looking through the source the sync calls are actually using the async request call.

Basically I'm writing a client api that provides 2 interfaces; an async and sync interface.

The issue that i am having is that if I try to use the sync methods with an existing runloop (provided by Sanic in this scenario), it complains with the exception: `Cannot run the event loop while another loop is running`.

Essentially I'm just calling:
```
import httpx
...
class SomeClass(object):
    def get_some_url(self):
        r = httpx.get('some url')
```

note that nothing here is async; apart from the fact that the `.get_some_url` call is made from within an async function like so:

```
import SomeClass

async def handler(request):
    o = SomeClass()
    o.get_some_url()
```

---

This works fine when using the async calls like this:

```
import httpx
...
class SomeClass(object):
    async def get_some_url(self):
        client = httpx.AsyncClient()
        r = await client.get('some url')
```

```
import SomeClass

async def handler(request):
    o = SomeClass()
    await o.get_some_url()
```

So I was just wondering if its possible to provide our own runloop when using the sync calls? Sanic uses uvloop internally, would be nice to be able to just pass that straight in somehow when constructing the `httpx` object, or setting it as a property after import.

Thanks!

florimondmanca:
Fixes #179, refs (or fixes?) #508.

Screenshot 2019-11-17 at 11 47 51

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -139,44 +139,4 @@ async def main():
 asyncio.run(main())
 ```
 
-Note that being in an async environment may not be immediately obvious. For example, you may be in a regular function that is called from an asynchronous function. So, this won't work either:
-
-```python
-import asyncio
-import httpx
-
-class Service:
-    def fetch_data(self):
-        r = httpx.get("https://example.org")
-        return r.text
-
-async def main():
-    service = Service()
-    data = service.fetch_data()
-
-asyncio.run(main())
-```
-
-Here, you should make `.fetch_data()` asynchronous. To use a shared client and benefit from connection pooling, you can also inject the `AsyncClient` as a parameter. This leads to the following code:
-
-```python
-import asyncio
-import httpx
-
-class Service:
-    async def fetch_data(self, client):
-        r = await client.get("https://example.org")
-        return r.text
-
-async def main():
-    service = Service()
-    async with httpx.AsyncClient() as client:
-        data = await service.fetch_data(client)
-        more_data = await service.fetch_data(client)
-
-asyncio.run(main())
-```
-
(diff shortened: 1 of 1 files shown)


docs/async.md
```diff
@@ -105,3 +105,78 @@ trio.run(main)
...
+async def main():
+    service = Service()
+    async with httpx.AsyncClient() as client:
+        data = await service.fetch_data(client)
+        more_data = await service.fetch_data(client)
+
+asyncio.run(main())
+```
+
+If you do not have control over making `.fetch_data()` asynchronous (for example, if it is a synchronous hook provided by a third-party framework), then the only solution would be to make the HTTP request in a separate thread. At this point, you should probably consider using a regular threaded HTTP client such as [Requests](https://github.com/psf/requests).
```
yeraydiazdiaz: Nit: I think this bit might going a bit too much into specific use cases. Maybe a general advise to consider using Requests +  threads if you get into a pickle?

I'd like to hear @tomchristie's and @sethmlarson's opinions though.

tomchristie: Agreed. I'd probably suggest scratching everything from "may not be immediately obviously" through to "If you do not have control". I think it dilutes the important stuff too much.

Otherwise great, tho.

florimondmanca:
Thanks @tomchristie and @yeraydiazdiaz, sounds legit! Last commit simplifies the write up.

florimondmanca:
Merging now, thanks! :)

PR #544, opened 2019-11-22
Drop sync

tomchristie:
Trawling through and dropping sync support everywhere.

There's a bunch of stuff that I think we *really* need to slim down on in order to focus on a really tight 1.0 release. I also don't think we've got a good approach onto sync support as things stand.

Subclassing in order to provide sync+async, leaading to BaseRequest, BaseResponse, AsyncRequest, AsyncResponse, Request, Response, Client, AsyncClient, Dispatcher, AsyncDispatcher just *doesn't* feel legible when getting to grips with any particular part of the codebase.

I think we need to get the asyncio+trio support absolutely squared away, and have any sync support come in a 2.0 version, that also keeps any bridging more clearly isolated.


httpx/dispatch/base.py
```diff
@@ -96,16 +45,16 @@ def send(
...
 
-    def __exit__(
+    async def __aexit__(
         self,
         exc_type: typing.Type[BaseException] = None,
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        self.close()
+        await self.close()
```
florimondmanca: Interesting diff on this class, right? Related to the codegen approach folks from urllib3-async pointed out :)

florimondmanca:
Also, refs #522.

tests/client/test_auth.py
```diff
@@ -89,80 +89,86 @@ def challenge_send(self, request: AsyncRequest) -> AsyncResponse:
...
-def test_basic_auth():
+@pytest.mark.asyncio
+async def test_basic_auth():
     url = "https://example.org/"
     auth = ("tomchristie", "password123")
 
-    with Client(dispatch=MockDispatch()) as client:
-        response = client.get(url, auth=auth)
+    client = Client(dispatch=MockDispatch())
+    response = await client.get(url, auth=auth)
```
florimondmanca: Thoughts on keeping `async with` usage here and elsewhere, out of consistency?

tests/test_api.py
```diff
@@ -1,63 +1,63 @@
...
+#
+# def test_delete(server):
+#     response = httpx.delete(server.url)
+#     assert response.status_code == 200
+#     assert response.reason_phrase == "OK"
+#
+#
+# def test_get_invalid_url(server):
+#     with pytest.raises(httpx.InvalidURL):
+#         httpx.get("invalid://example.org")
```
florimondmanca: Commented out until we flesh out the async high-level API, right? :-)

tests/test_concurrency.py
```diff
@@ -1,91 +1,91 @@
...
+#     stream = await backend.open_tcp_stream(
+#         server.url.host, server.url.port, ssl_context=None, timeout=TimeoutConfig(5)
+#     )
+#     try:
+#         await stream.write(b"GET / HTTP/1.1\r\n\r\n")
+#         await run_concurrently(
+#             backend, lambda: stream.read(10), lambda: stream.read(10)
+#         )
+#     finally:
+#         await stream.close()
```
florimondmanca: Any reason this has been commented out? Doesn't seem related to sync to me.

tests/test_decoders.py
```diff
@@ -128,18 +129,21 @@ def test_decoding_errors(header_value):
...
+async def test_text_decoder(data, encoding):
+    async def iterator():
         nonlocal data
         for chunk in data:
             yield chunk
 
     response = httpx.Response(200, content=iterator())
-    assert "".join(response.stream_text()) == (b"".join(data)).decode(encoding)
+    await response.read()
+    assert "".join(response.text) == (b"".join(data)).decode(encoding)
```
florimondmanca: While this should *work*, we can simplify:

```suggestion
    assert response.text == (b"".join(data)).decode(encoding)
```

tests/client/test_auth.py
tomchristie: Really not needed. We’re not closing a connection pool, so I think we should keep the style simple.

httpx/client.py
```diff
@@ -675,537 +653,6 @@ class AsyncClient(BaseClient):
...
-    * **dispatch** - *(optional)* A dispatch class to use for sending requests
-    over the network.
-    * **app** - *(optional)* A WSGI or ASGI application to send requests to,
-    rather than sending actual network requests.
-    * **backend** - *(optional)* A concurrency backend to use when issuing
-    async requests.
-    * **trust_env** - *(optional)* Enables or disables usage of environment
-    variables for configuration.
-    * **uds** - *(optional)* A path to a Unix domain socket to connect through.
-    """
```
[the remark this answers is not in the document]
tomchristie: Yup, defo.

tests/test_concurrency.py
tomchristie: Sorted.

tests/test_api.py
tomchristie: Included top-level async API. We'll probably want to switch the top-level stuff to a global client/connection pool rather than a one-shot, but it's still valuable in the meantime.

httpx/client.py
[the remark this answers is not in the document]
tomchristie: Done.

tomchristie:
I think this should be ~up to scratch now.

Big jump to make, and likely to be unpopular in places, but I think we've got more than enough on our plate to fully square away before coming back to a decision on best tack for bridging-to-sync.

tomchristie:
Screenshot 2019-11-26 at 16 05 10

httpx/api.py
```diff
@@ -25,18 +25,16 @@ def request(
...
     allow_redirects: bool = True,
     verify: VerifyTypes = True,
     cert: CertTypes = None,
     stream: bool = False,
-    trust_env: bool = None,
+    trust_env: bool = True,
 ) -> Response:
     """
     Sends an HTTP request.
-
```
StephenBrown2: These spaces are critical, otherwise it gets mashed up into a single paragraph:
Screen Shot 2019-11-26 at 10 25 44 AM
vs.
Screen Shot 2019-11-26 at 10 29 39 AM
...
Screen Shot 2019-11-26 at 10 29 56 AM

tomchristie: Thanks! Not sure how those got mashed up.

florimondmanca:
Boom 💥

Issue #546, opened 2019-11-22
Merging query params eats list type paramaters


kcen:
When merging two `QueryParam` objects list/repeated param options in the merge-from object stomp on themselves. This is particularly obnoxious because we call `self.merge_queryparams` from `httpx.client.Client().build_request`

**Simplified example**
```
$ python
Python 3.7.4 (default, Sep 16 2019, 17:52:53)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import httpx
>>> from httpx.models import QueryParams
>>> first = QueryParams(); first
QueryParams('')
>>> second = QueryParams(r=['abc', 'def']); second
QueryParams('r=abc&r=def')
>>> first.update(second); first
QueryParams('r=def')
>>> httpx.__version__
'0.7.8'
```

**User impact example**
```
>>> with httpx.Client(base_url='https://postman-echo.com/') as c:
...   c.get('/get', params={'myarg':['abc','def']}).json()
...
{'args': {'myarg': 'def'}, 'headers': {'x-forwarded-proto': 'https', 'host': 'postman-echo.com', 'accept': '*/*', 'accept-encoding': 'gzip, deflate', 'user-agent': 'python-httpx/0.7.8', 'x-forwarded-port': '443'}, 'url': 'https://postman-echo.com/get?myarg=def'}
```

tomchristie:
Closed via #547

PR #547, opened 2019-11-23
Preserve list type query paramaters when merging QueryParams objects

kcen:
Addresses: https://github.com/encode/httpx/issues/546

Attempting to address the issue, whilst keeping the element of surprise low for existing use cases.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -351,7 +351,12 @@ class QueryParams(typing.Mapping[str, str]):
 
         params = QueryParams(params)
         for param in params:
-            self[param] = params[param]
+            item, *extras = params.getlist(param)
+            self[param] = item
+            if extras:
+                self._list.extend((param, e) for e in extras)
+                # ensure getter matches merged QueryParams getter
+                self._dict[param] = params[param]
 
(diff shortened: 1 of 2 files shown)


httpx/models.py
```diff
@@ -351,7 +351,12 @@ def update(self, params: QueryParamTypes = None) -> None:  # type: ignore
 
         params = QueryParams(params)
         for param in params:
-            self[param] = params[param]
+            item, *extras = params.getlist(param)
```
tomchristie: wondering if there's an scope for refactoring for simplificity here.
Would the following also work?...

```python
items = params.getlist(param)
self._list.extend((param, item) for item in items)
self._dict[param] = items[0]
```

kcen: I don't think that will work if the params getting replaced are of a list/repeated value.

We need to call `self[param]=` at some point to trigger the search delete for other values in `QueryParams._list` via `QueryParams.__setitem__`

tomchristie: Good point!

tomchristie:
Thanks!

PR #552, opened 2019-11-26
Client handles redirect + auth

tomchristie:
Builds on #544

Drops redirect and auth from being middleware, and just handles them on the `Client` instance.


tomchristie:
Screenshot 2019-11-26 at 16 05 27

florimondmanca:
@tomchristie The only thing itching me on this is that we’re back with having lots of redirect utility helper methods on the Client.

They used to be neatly encapsulated in the middleware class.

Any chance we can keep redirect-as-middleware *and* the benefits from the refactoring here?

tomchristie:
I'm not really fussed about it - they're neatly demarked. (In the same way as the `merge_<...>` methods.)

I'd *far* rather we have a mostly linear chain of control to follow through, like this.

PR #553, opened 2019-11-26
Add mkautodoc to nox "serve" session

StephenBrown2:
`nox -s serve` doesn't work without the `mkautodoc` extension installed, so adding it to the session to fix that.


StephenBrown2:
Closing as https://github.com/encode/httpx/pull/566 makes this obsolete.

PR #566, opened 2019-11-29
Drop nox in favor of vanilla scripts

### LINKED ISSUE #239: scripts as a single top-level point of project automation.

florimondmanca:
Fixes #239. Also based on some discussion @tomchristie and I had offline.


florimondmanca:
Hmm… `pytest` is showing a "Could not load plugin" error:

```
pluggy.manager.PluginValidationError: Plugin 'trio' could not be loaded: (attrs 18.2.0 (/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages), Requirement.parse('attrs>=19.2.0'), {'trio'})!
```

This happened to me in `asgi-lifespan` earlier today. Probably due to a silly dependency resolution issue… 😕 Not sure why it's only happening now; there hasn't been a release in `trio`, `attrs`, `pytest` or `pluggy` recently that seems to change the behavior.

Anyway, forcing `attrs>=19.2` should fix it on our end, but it's definitely a hack…

florimondmanca:
It seems the Windows build is stalling, and has been since the addition of UDS (#511 - see build [here](https://travis-ci.org/encode/httpx/builds/614401432?utm_medium=notification&utm_source=github_status)).

We didn't notice because the job was running last, and was allowed to fail, so the build was green even though the Windows job hadn't finished yet (and would finish minutes later).

I pushed a commit here to revert to the old situation, though I might be inclined towards removing the Windows job for now since we don't actively seek Windows support (although we probably should).

docs/contributing.md
```diff
@@ -37,85 +37,54 @@ your GitHub username:
...
-
 ## Documenting
 
 Documentation pages are located under the `docs/` folder.
 
 To run the documentation site locally (useful for previewing changes), use:
 
 ```shell
-$ nox -s serve
+$ scripts/serve
```
tomchristie: I’m guessing we want `scripts/docs serve` here?

.travis.yml
```diff
@@ -7,42 +7,44 @@ branches:
...
-      dist: bionic # Required to get OpenSSL 1.1.1+
-
-    - os: windows
+    -
+      stage: check
+      python: 3.7
+      script: scripts/check
+    - stage: docs
+      python: 3.7
+      script: scripts/docs
```
tomchristie: Should we use `scripts/docs build` and have the script accept $@ command line arguments?

docs/contributing.md
florimondmanca: What would we think about `scripts/docs-serve` and `scripts/docs-build`? (We'd even get autocomplete without having to deal with CLI arguments in plain bash.)

florimondmanca: Also the `serve` script is actually missing right now. 😆

florimondmanca:
@tomchristie Are we good running with `scripts/docs-{build,serve}`?

tomchristie:
@florimondmanca - *Personally* I'd probably prefer `scripts/docs build` and `scripts/docs serve`, but either way let's pull this in for now and we can figure out if there's anything else we want to iterate on later. 👍

Issue #565, opened 2019-11-29
Drop background manager (in favour of a two-function fork primitive).


tomchristie:
I think we're running against the grain with our current `read`, `write`, `background_manager` API on the backends.

Switching to the same style as the `hip` project is using for its concurrency backends is likely a smart thing for us to do here... https://github.com/python-trio/hip/tree/master/src/urllib3/_backends

What would it take for us to switch over to a `send_and_receive_for_a_while` API?

@florimondmanca - I seem to remember you did some work on scoping this out at some point?

florimondmanca:
@tomchristie I believe I had a working implementation of this in #526.

florimondmanca:
But I remember we said somewhere that the parallel send request/receive response flow may not actually be necessary - what’s the latest status on that?

tomchristie:
Yeah I dug back into that... and no, it's not neccessary for HTTP/1.1.
I was forgetting tho that it is absolutely neccessary for HTTP/2.

(I tried dropping background manager out everywhere to help figure out if we had any obvious failures once we did that, and an HTTP/2 "send a large request" test case started failing, which retrospectively I should have realised would occur.)

tomchristie:
Part of the motivation here is that if we fall into line with how `hip` have tackled it, we can draw from their work on twisted, sync, and anyio backends.

(Now that we've dropped out all the sync-bridging code, there's a much clearer path onto how we'd *ought* to be supporting the sync case.)

tomchristie:
As it *happens* I'm probably fairly agnostic on the benefits of `read`+`write`+`background manager` vs. `read_and_write_for_a_while`, but I think we're *probably* better of following their lead here, than not, simply for the benefits of all singing from the same song-sheet, right?

florimondmanca:
Yeah, I’m good either way. #526 taught be that this won’t necessarily result in simpler code, though — bit of the opposite, since we have to provide some extra abstraction to deal with HTTP/1.1 vs HTTP/2. There may be extra simplifications made possible since we dropped sync. But yeah I’d be fine with taking the same approach than hip here. :)

tomchristie:
Right.

Also the *big* advantage with this is that a SyncBackend with `read_and_write_for_a_while` makes sense, but a SyncBackend with `read`+`write`+`background manager` isn't really possible. (At least without introducing multi-threading, which we clearly wouldn't want.)

pquentin:
> simply for the benefits of all singing from the same song-sheet

[...]

PR #569, opened 2019-11-30
Don't use background task in HTTP/1.1 dispatcher

florimondmanca:
As mentioned in https://github.com/encode/httpx/issues/565#issuecomment-559798817, sending the request and receiving the response concurrently is not necessary for HTTP/1.1

### DIFF
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -46,10 +46,8 @@ class HTTP11Connection:
         timeout = None if timeout is None else TimeoutConfig(timeout)
 
         await self._send_request(request, timeout)
-
-        task, args = self._send_request_data, [request.stream(), timeout]
-        async with self.backend.background_manager(task, *args):
-            http_version, status_code, headers = await self._receive_response(timeout)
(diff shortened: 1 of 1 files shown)


tomchristie:
Great, so...

This gives us marginally simpler code & execution flow, at the cost of poorer performance in the case of early responses, where we won't start getting the response until we've finalised sending all the data.

I think it's very feasible that we might move back to a concurrent read/write in the future, by addressing #565, but at this point in time I think we should opt for the simplest possible case. In any case, this is the same style that urllib3/requests uses, so we're not regressing here

A *hugely* important thing that this PR would allow us to do is to make it possible for us to write an HTTP/1.1 supporting SyncBackend, without needing to address #565.

It think doing that is *the single most important technical task* for us to cover, since we can validate if we're missing anything 

The SyncBackend can look to https://github.com/python-trio/hip/blob/master/src/urllib3/_backends/sync_backend.py for implementation, although at least for our first pass we'd just want `.read()` and `.write()` implementations, rather than `send_and_receive_for_a_while()`.

PR #570, opened 2019-11-30
Decouple request bytes generation/sending in HTTP/1.1 dispatcher

florimondmanca:
Builds upon #569.

Considering #565, it might be a good idea in general to decouple generating the bytes that should be sent to the network (i.e. interacting with `h11`/`h2`) from actually sending those bytes to the network (i.e. interacting with the `stream`).

(A general pattern is to build a pair of strictly distinct methods, i.e. `._get_XYZ_bytes()` / `._send_XYZ()` (and have the latter call the former to know what to send). Generally it will lead to a bit more code, but the result is IMO easier to read and extend, esp. if we want to go towards implementing #565 — although that issue does not actually apply for HTTP/1.1.)

So, this PR does that for sending the request and its body.

### DIFF
--- a/httpx/dispatch/http11.py
+++ b/httpx/dispatch/http11.py
@@ -69,11 +69,12 @@ class HTTP11Connection:
             pass
         await self.stream.close()
 
-    async def _send_request(
+    def _get_request_bytes(
         self, request: Request, timeout: TimeoutConfig = None
-    ) -> None:
+    ) -> bytes:
         """
-        Send the request method, URL, and headers to the network.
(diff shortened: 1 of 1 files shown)


florimondmanca:
I already have the pattern implemented for reading the response and its body as well, but it's a bit more involved so I wanted to get some feedback on the simple request case first. :-)

tomchristie:
I think we should probably hold off on this.

florimondmanca:
Okay, let's close this off for now then. :-) We can still reference this later if need be.

PR #574, opened 2019-11-30
Add `Response.is_error`

### LINKED ISSUE #543: Add equivalent to requests.response `ok` property
Hey there, I'm wondering do we have some equivalent of `requests.response:ok` property in `httpx` library? If no, Is it planning to be added sometime in the future? thx!

tomchristie:
Adds a `response.is_error` property, which closes #543

Our section in the requests compat docs details httpx's usage of `response.is_error` vs. requests' usage of `response.is_ok`...

> We don't support `response.is_ok` since the naming is ambiguous there, and might incorrectly imply an equivalence to `response.status_code == codes.OK`. Instead we provide an `.is_error` property.

What do we think?


tomchristie:
We could have a `.is_success` property, but folks *might* expect that to be equivelent to `100 <= response.status_code <= 399`... Or am I being overly cautious there?

florimondmanca:
Nice about `.is_error`.

What do we think about `.is_200_ok`?

tomchristie:
Nah, that'd just be cleaner as `response.status_code == 200`

Having `response.is_error` is sensible, because it's the same check as `response.raise_for_status(...)`, but without raising an exception.

PR #583, opened 2019-12-02
Minor test fixes

tomchristie:


tomchristie:
The previous multipart test included an "example.log" -> "application/octet-stream" was evidentally ambiguous since it was mapping to `"text/plain"` for me locally.

I've updated the test to include less ambiguous cases.

Issue #496, opened 2019-10-31
Require `pip install httpx[http2]` for HTTP/2 support


tomchristie:
I wonder if we should consider having HTTP/2 be an optional requirement, rather than always-on.

I've been working towards minimal installs in other projects eg. Starlette (no hard dependencies) and Uvicorn - we're planning to move towards `pip install uvicorn` for a basse install of just `click` and `h11`, and `pip install uvicorn[default]` to pull in websockets plus platform specfic stuff like uvloop and httptools. And generally with the other projects in the encode async space too.

We *could* consider having minimal and default install options with httpx too... eg:

* `pip install httpx` - Just http/1.1 support.
* `pip install httpx[defaults]` - Also http/2. Potentially http/3 and/or websocket support at a future date. Possible that hsts isn't strictly a hard requirement either.

Any thoughts?

florimondmanca:
I think having a consistent policy wrt installs across Encode projects is a neat idea. Minimal installs + `<project>[default]` for regular features sounds good to me!

florimondmanca:
With #584, we could consider *multiple* extras, i.e. one per extra feature:

```
pip install httpx[http2]
```

And in the future, possibly…

```
pip install httpx[http2,websocket,http3,...]
```

tomchristie:
Retitling this to specifically reflect what we'd like to do for HTTP/2.

I had a look at it this morning, and it's actually not as easy as I'd expected to only lazily import the HTTP2Connection.

* Really we want a base `EstablishedConnection` type, that provides base interface the both `HTTP2Connection` and `H11Connection` implement.
* That way the `HTTPConnection` class can hold a connections of the type `Optional[ActiveConnection]`, rather than specifically having the `H2Connection` and `H11Connection` types.

However, right now we're blocked by the proxy implementation reaching directly into the H11 implementation.

We'd need to switch things around so that the connection classes are responsible for implementing a `start_tls` method, which the proxy implementation can then call into.

tomchristie:
I'm going to suggest we drop this.

Our HTTP/2 implemenentation has greatly improved, and even tho it's currently not enabled by default I could see us switching that once we judge it to be sufficiently mature. I'm no longer convinced that "minimal installs" are worth the additional hoop for users to jump through and codebase complexity.

florimondmanca:
[...]

PR #584, opened 2019-12-02
Better HTTP/2 defaults.

### LINKED ISSUE #582: HTTP/2 should be off by default.
So...

I'm super excited that we've got an HTTP/2 implementation in here, but we also need to consider *what default behavior we should be aiming for*, and what we should prioritise.

Our HTTP/2 implementation is neccessarily going to be more complex, and less mature than our HTTP/1.1 implementation. Even if that *wasn't* the case, then HTTP/2 infrastructure that we're connecting too is *more* likely to have implementation issue and quirks than HTTP/1.1 versions.

I think we ought to clearly be *prioritizing for reliablity*, while still supporting users that want the more cutting-edge functionality.

I'd suggest that:

* We ought to have a documentation page on HTTP/2, with a brief overview of the primary differences, and motivation as to *why* user might or might not want to switch HTTP/2 support on.
* We ought to tidy up our interface for switching between HTTP versions.

We currently have this...

```python
client = Client(http_versions=["HTTP/1.1", "HTTP/2"])  # Really clunky
client = Client(http_versions=["HTTP/1.1"]) 
client = Client(http_versions=["HTTP/2"])  # Doesn't really make sense - you *might* still end up with an HTTP/1.1 connection.
```

I think we should just switch to...

```python
client = Client(http_2=True)
client = Client(http_2=False)
```

Furthermore, I think the default ought to be false. There's no point in exposing our users to this class of issues https://github.com/encode/httpx/issues?q=is%3Aissue+is%3Aopen+label%3Ahttp%2F2 *as our default behaviour*, but we can still make it easy for our users to upgrade to the HTTP/2 implementation.

tomchristie:
* HTTP/2 is off by default.
* `Client(http2=True)` to enable
* Added HTTP/2 docs.

Closes #582

Still TODO:

- [x] Interlink to the HTTP/2 section.
- [x] Link to https://http2-explained.haxx.se/content/en/ as a comprehensive HTTP/2 guide.

API changes for `http_version=...` to `http_2=[True|False]`, so we should leave this to a 0.9 release.

### DIFF
--- a/docs/http2.md
+++ b/docs/http2.md
@@ -1,0 +1,42 @@
+# HTTP/2
+
+HTTP/2 is a major new iteration of the HTTP protocol, that provides a far more
+efficient transport, with potential performance benefits. HTTP/2 does not change
+the core semantics of the request or response, but alters the way that data is
+sent to and from the server.
+
+Rather that the text format that HTTP/1.1 uses, HTTP/2 is a binary format.
+The binary format provides full request and response multiplexing, and efficient
+compression of HTTP headers. The stream multiplexing means that where HTTP/1.1
+requires one TCP stream for each concurrent request, HTTP/2 allows a single TCP
+stream to handle multiple concurrent requests.
+
+HTTP/2 also provides support for functionality such as response prioritization,
+and server push.
+
+## Enabling HTTP/2
+
+The HTTPX client provides provisional HTTP/2 support.
+
+The HTTP/2 support is not enabled by default, because HTTP/1.1 a mature,
+battle-hardened transport layer. With HTTP/2 being newer and significantly more
+complex, it should be considered a less robust option at this point in time.
+
+However, if you're issuing highly concurrent requests you might want to consider
+trying out our HTTP/2 support. You can do so by instantiating a client with
+HTTP/2 support enabled.
+
+Simply instantiating a client...
+
+```python
+client = httpx.Client(http_2=True)
+...
+```
+
+Instantiating a client as a context manager, to ensure that all connections
+are nicely closed at the end of it's usage...
+
+```python
+async with httpx.Client(http_2=True) as client:
+    ...
+```
--- a/docs/http2.md
+++ b/docs/http2.md
@@ -14,29 +14,47 @@ stream to handle multiple concurrent requests.
 HTTP/2 also provides support for functionality such as response prioritization,
 and server push.
 
+For a comprehensive guide to HTTP/2 you may want to check out "[HTTP2 Explained](https://http2-explained.haxx.se/content/en/)".
+
 ## Enabling HTTP/2
 
 The HTTPX client provides provisional HTTP/2 support.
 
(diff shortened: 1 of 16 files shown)


docs/http2.md
```diff
@@ -0,0 +1,42 @@
...
+stream to handle multiple concurrent requests.
+
+HTTP/2 also provides support for functionality such as response prioritization,
+and server push.
+
+## Enabling HTTP/2
+
+The HTTPX client provides provisional HTTP/2 support.
+
+The HTTP/2 support is not enabled by default, because HTTP/1.1 a mature,
```
florimondmanca: Typo: *is* a mature

docs/http2.md
```diff
@@ -0,0 +1,42 @@
...
+battle-hardened transport layer. With HTTP/2 being newer and significantly more
+complex, it should be considered a less robust option at this point in time.
+
+However, if you're issuing highly concurrent requests you might want to consider
+trying out our HTTP/2 support. You can do so by instantiating a client with
+HTTP/2 support enabled.
+
+Simply instantiating a client...
```
florimondmanca: This line is probably superfluous — how about adding a colon at the end of the sentence above?

Also let’s not assume things are « simple »? 😄 (I remember someone did tremendous work to remove these expressions from the Django docs last month to make them more inclusive, but I can’t find the PR/tweet anymore).

docs/http2.md
```diff
@@ -0,0 +1,42 @@
...
+
+```python
+client = httpx.Client(http_2=True)
+...
+```
+
+Instantiating a client as a context manager, to ensure that all connections
+are nicely closed at the end of it's usage...
```
florimondmanca: Typo: its?

docs/http2.md
tomchristie: Thanks!

docs/http2.md
tomchristie: Yes, and yes!

docs/http2.md
tomchristie: Going to change that around in any case, yeah. Thanks again!

tomchristie:
> I suppose this moves us forward relative to minimal dependencies, eg we’ll now be able to not require h2 unless users want to use HTTP/2?

Ah yeah I'd forgotten about that. We *can also* make that switch here, yup.
I guess we would drop `h2` out of our default install, and into `extra_requires`, so...

`pip install httpx[h2]`

And a check on the client for "if http_2 is True and h2 is not installed then error".

florimondmanca:
> We can also make that switch here, yup.

I’d rather we deal with it later, this PR is quite a big chunk already. :-)

tomchristie:
Good call.

tomchristie:
Should we use `httpx.Client(http_2=True)` or `httpx.Client(http2=True)`?

I think the later is probably neater, right?

florimondmanca:
Just remembered we had a "Minimal installs" issue here — https://github.com/encode/httpx/issues/496.

PR #586, opened 2019-12-02
Rename http2 switch

tomchristie:
Follow up to #584

I *guess* we want to prefer the style `Client(http2=True)` over the style `Client(http_2=True)`, right?


florimondmanca:
Clear yes for me. Doesn’t seem like this branch is up to date yet, is it?

PR #592, opened 2019-12-04
Differentiate between `timeout=None` and `timeout=UNSET`.

### LINKED ISSUE #433: Passing 'timeout=None' should always result in no timeout

tomchristie:
Builds on #591. Closes #433.


tomchristie:
Behaviour now is...

```python

# Default behavior, timeouts after 5 seconds.
await httpx.get(url)

# Default behavior, timeouts after 5 seconds.
client = httpx.Client()
await client.get(url)

# Override timeout on client, does not timeout, returns response after 10s
await client.get(url)

# Override timeout on method, does not timeout, returns response after 10s
client = httpx.Client()
await client.get(url, timeout=None)

# Override timeout on method, does not timeout, returns response after 10s
client = httpx.Client(timeout=5)
await client.get(url, timeout=None)
```

PR #593, opened 2019-12-04
Allow default+override timeout style

### LINKED ISSUE #435: Fine-tuning timeouts from the default timeout config
Currently, if we want to change only one of the three timeouts (connect, read, write) we need to build a full-blown `TimeoutConfig`. But there is no way to use the default timeout values.

```python
>>> import httpx
>>> httpx.config.DEFAULT_TIMEOUT_CONFIG
TimeoutConfig(timeout=5.0)
>>> httpx.TimeoutConfig(read_timeout=1)
# We'd expect connect_timeout and write_timeout to be 5 (the default)
TimeoutConfig(connect_timeout=None, read_timeout=1, write_timeout=None)
```

Changing `TimeoutConfig` to automatically use the default values in case some values aren't given is probably not the good way to go, because `TimeoutConfig` shouldn't contain any smart behavior.

Instead, we should probably allow explicitly building a `TimeoutConfig` instance from `DEFAULT_TIMEOUT_CONFIG`:

I'd see it materialize as factory method on `TimeoutConfig`:

```python
>>> httpx.TimeoutConfig.from_defaults(read_timeout=1)
TimeoutConfig(connect_timeout=5, read_timeout=1, write_timeout=5)
```

Does this sound reasonable @tomchristie?

### LINKED ISSUE #578: TimeoutConfig improvements.
1. `TimeoutConfig(5.0)` should really just be `Timeout(5.0)`. We don't want to break the API prior to 1.0 so I guess let's switch it to `Timeout` but have a `TimeoutConfig = Timeout` synonym.

2. Right now we either support `Timeout(5.0)` or `Timeout(read_timeout=5.0)`, however we should actually support a style that provides the default case, and the exceptional cases. Eg...

```python
# Connect, Read, Write have a 5 second timeout. Pool acquiry never times out.
timeout = httpx.Timeout(5.0, pool_timeout=None)
client = Client(timeout=timeout)
```

tomchristie:
Allow eg. `Timeout(5.0, pool_timeout=None)`

Closes #435, Closes #578


docs/advanced.md
```diff
@@ -250,83 +250,76 @@ async with httpx.Client(proxies=proxy) as client:
...
-)
-
-resp = await httpx.get('http://example.com/api/v1/example', timeout=timeout)
+client = httpx.Client()              # Use a default 5s timeout everywhere.
+client = httpx.Client(timeout=10.0)  # Use a default 10s timeout everywhere.
+client = httpx.Client(timeout=None)  # Disable all timeouts by default.
 ```
 
-### Default timeouts
+#### Fine tuning the configuration.
```
florimondmanca: Nit: extra full stop isn’t aligned with the rest of the headers :)

docs/advanced.md
```diff
@@ -250,83 +250,76 @@ async with httpx.Client(proxies=proxy) as client:
...
-* The **connect timeout** specifies the maximum amount of time to wait until
-a connection to the requested host is established. If HTTPX is unable to connect
-within this time frame, a `ConnectTimeout` exception is raised.
-* The **write timeout** specifies the maximum duration to wait for a chunk of
-data to be sent (for example, a chunk of the request body). If HTTPX is unable
-to send data within this time frame, a `WriteTimeout` exception is raised.
-* The **read timeout** specifies the maximum duration to wait for a chunk of
-data to be received (for example, a chunk of the response body). If HTTPX is
-unable to receive data within this time frame, a `ReadTimeout` exception is raised.
+The default behavior is to raise a `RequestException` after 5 seconds of
```
florimondmanca: Isn’t it a `RequestTimeout`?

docs/advanced.md
florimondmanca: Also should this be an h3-level section?

docs/advanced.md
```diff
@@ -250,83 +250,76 @@ async with httpx.Client(proxies=proxy) as client:
...
-)
-
-resp = await httpx.get('http://example.com/api/v1/example', timeout=timeout)
+client = httpx.Client()              # Use a default 5s timeout everywhere.
+client = httpx.Client(timeout=10.0)  # Use a default 10s timeout everywhere.
+client = httpx.Client(timeout=None)  # Disable all timeouts by default.
 ```
 
-### Default timeouts
+#### Fine tuning the configuration
```
florimondmanca: Same here - h3 level? The rationale is that otherwise we’ll have a single h3 (setting/disabling timeouts).

docs/advanced.md
tomchristie: Thanks!

What do we *actually* want the base timeout exception to be called? `TimeoutError` is a python built-in, so I guess we've got...

* `TimeoutException`
* `BaseTimeout`
* `RequestTimeout`

?...

florimondmanca: I like `TimeoutException` best: explicit, boring and unsurprising. (`RequestTimeout` has an odd vibe to it that I can't really explain, and `BaseTimeout` sounds like a base class.)

tomchristie:
Thanks for the review!

PR #596, opened 2019-12-05
Update changelog to include 'master'

tomchristie:
I *think* that this is everything we need to call out from a user-facing perspective.
There's a few other refactorings that I've omitted, since they're strictly behind-the-scenes.

https://github.com/encode/httpx/commits/master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,24 +8,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added concurrency backend autodetection.
-- Added `Client(backend='trio')` and `Client(backend='asyncio')` API.
-- Added `response.stream_lines()` API.
-- Added `response.is_error` API.
-- Added support for `timeout=Timeout(5.0, connect_timeout=60.0)` styles.
+- Added concurrency backend autodetection. (Pull #585)
+- Added `Client(backend='trio')` and `Client(backend='asyncio')` API. (Pull #585)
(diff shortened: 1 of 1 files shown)


florimondmanca:
FYI, what I usually do is go to the `/compare` view (here https://github.com/encode/httpx/compare/0.8.0...HEAD) and select things that aren't docs/tests/tooling-related, i.e. things that have an impact on users.

We also usually include the pull request ref, for users that might be interested in it without having to dig through PRs manually. Could you add those in here too?

CHANGELOG.md
```diff
@@ -4,11 +4,34 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Master
+
+### Added
+
+- Added concurrency backend autodetection.
```
florimondmanca: So, e.g…

```suggestion
- Added concurrency backend autodetection. (Pull #585)
```

(That's why we tend to edit the title of the PR squash commit — personally, I just go to `/compare` and copy/paste the entries here, making small copy edits if needed.)

tomchristie: Yes great call.
Will start to make sure that titles get reviewed at the point of merge.
I've erronously got two cases where I've had to reference a commit rather than a PR.

We'll want to get really tight about *only ever* allowing changes-as-PRs, partly because it's *so much nicer* for history/interlinking, and partly to make sure *everything* is getting fully reviewed.

tomchristie:
Right then, I'm going to merge this as it currently stands, and update the last bits once they make it in.

Issue #554, opened 2019-11-27
Drop per-request `cert`/`verify`/`trust_env`


tomchristie:
I think we should drop per-request `cert`/`verify` on the client methods, and only allow them on the client itself.

The reasoning for this is that you *don't know* when you make a request on the client if you'll actually acquire (and verify) a connection, or if you'll end up re-acquiring a connection (without any verification).

Really the configuration ought to always just be per-connection-pool, rather than per-request.

Dropping these two would also mean removing them from the dispatcher API, which feels cleaner as a result.

We can leave them on the top-level standalone functions, since they apply to the client in that case anyways.

florimondmanca:
I'd be interested in seeing this happen! It always felt strange to me that *these particular two* parameters were part of the dispatcher API. To reinforce your point, IMO the use case of verifying/not verifying or using different certs for different requests is narrow enough, and is already supported by the ability to instantiate multiple clients.

PR #597, opened 2019-12-05
Warn if cert / verify / trust_env are passed to client.request()

tomchristie:
Refs #554.

Adds a warning not to use `cert=...`, `verify=...`, or , `truts_env=...` on client methods, but instead pass them to the client init (and use a different client instance if you need different cert/verify/trust_env configuration.)

I *think* that addressing #554 is the right thing to do, particularly wrt. honing in on a 1.0. (It's much easier for us to add them back in if we decide we really want to, than it is to take them out.)

Benefits here are that:

* Connection pooling will be seperate for differing `cert`/`verify` cases.
* The change will positively impact the dispatcher API, and improve implementation clarity all the way through.

**TODO**:

- [x] Documentation in the advanced section.
- [x] Note in the "requests compat" docs.
- [x] Add `trust_env` here, too.

As I say, we *could* come back to including these at a future date if we really want them, but I think it's better that we start from the more constrained, but still just-as-functionally-capable variant.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -174,6 +174,12 @@ password example-password
 ...
 ```
 
+When using `Client` instances, `trust_env` should be set on the client itself, rather that on the request methods:
+
+```python
+client = httpx.Client(trust_env=False)
+```
+
 ## Unix Domain Sockets
 
 You can configure an `httpx` client to connect through a unix domain socket via the `uds` parameter. This is useful when making requests to a server that is bound to a socket file rather than an IP address.
@@ -373,19 +379,35 @@ MIME header field.
 
 When making a request over HTTPS, HTTPX needs to verify the identity of the requested host. To do this, it uses a bundle of SSL certificates (a.k.a. CA bundle) delivered by a trusted certificate authority (CA).
 
-### Default CA bundle
+### Changing the verification defaults
 
 By default, HTTPX uses the CA bundle provided by [Certifi](https://pypi.org/project/certifi/). This is what you want in most cases, even though some advanced situations may require you to use a different set of certificates.
 
-### Using a custom CA bundle
-
-If you'd like to use a custom CA bundle, you can use the `verify` parameter that is available on the high-level API as well as clients. For example:
+If you'd like to use a custom CA bundle, you can use the `verify` parameter.
 
 ```python
 import httpx
 
 r = await httpx.get("https://example.org", verify="path/to/client.pem")
 ```
+
+You can also disable the SSL verification...
+
+```python
+import httpx
+
+r = await httpx.get("https://example.org", verify=False)
+```
+
+### SSL configuration on client instances
+
+If you're using a `Client()` instance, then you should pass any SSL settings when instantiating the client.
+
+```python
+client = httpx.Client(verify=False)
+```
+
+The `client.get(...)` method and other request methods *do not* changing the SSL settings on a per-request basis. If you need different SSL settings in different cases you should use more that one client instance, with different settings on each. Each client will then be using an isolated connection pool with a specific fixed SSL configuration on all connections within that pool.
 
 ### Making HTTPS requests to a local server
 
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -1,5 +1,6 @@
 import functools
 import typing
+import warnings
 from types import TracebackType
 
 import hstspreload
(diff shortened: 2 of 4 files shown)


httpx/client.py
```diff
@@ -222,6 +223,19 @@ def params(self, params: QueryParamTypes) -> None:
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
         trust_env: bool = None,
     ) -> Response:
+        if cert is not None:
+            warnings.warn(
+                "Passing a 'cert' argument when making a request on a client "
+                "is due to be deprecated. Instantiate a new client instead, "
+                "passing any 'cert' or 'verify' arugments to the client itself."
```
florimondmanca: Typo:

```suggestion
                "passing any 'cert' or 'verify' arguments to the client itself."
```

httpx/client.py
```diff
@@ -222,6 +223,19 @@ def params(self, params: QueryParamTypes) -> None:
...
+            )
+        if verify is not None:
+            warnings.warn(
+                "Passing a 'verify' argument when making a request on a client "
+                "is due to be deprecated. Instantiate a new client instead, "
+                "passing any 'cert' or 'verify' arugments to the client itself."
```
florimondmanca: Typo:

```suggestion
                "passing any 'cert' or 'verify' arguments to the client itself."
```

tomchristie: I really need to find a good process for making sure those aren't creeping in all the time!

florimondmanca: Time for `mkspellcheck`? 🤷‍♂

tomchristie:
> I think we're ready to release this and the host of other stuff we've got on master in a 0.9 release at this point?

Pretty much, yes. The *only* other thing I'd like to do is:

* Add a `client.stream(...)` method.
* Raise warnings on `request.get(..., stream=True)`.

That won't include any fancy `Response` vs `StreamingResponse` stuff, but it *will* get our streaming API fully up to scratch.

httpx/client.py
```diff
@@ -222,6 +223,25 @@ def params(self, params: QueryParamTypes) -> None:
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
         trust_env: bool = None,
     ) -> Response:
+        if cert is not None:
+            warnings.warn(
+                "Passing a 'cert' argument when making a request on a client "
+                "is due to be deprecated. Instantiate a new client instead, "
+                "passing any 'cert' arguments to the client itself."
+            )
```
StephenBrown2: Don't forget to specify the category of warning with the second parameter:
`warn(message, category, ...)`
```suggestion
            , PendingDeprecationWarning)
```

httpx/client.py
```diff
@@ -222,6 +223,25 @@ def params(self, params: QueryParamTypes) -> None:
...
+        if verify is not None:
+            warnings.warn(
+                "Passing a 'verify' argument when making a request on a client "
+                "is due to be deprecated. Instantiate a new client instead, "
+                "passing any 'verify' arguments to the client itself."
+            )
```
StephenBrown2: Same category suggestion, pending formatting because no multi-line suggestions yet.
```suggestion
            , PendingDeprecationWarning)
```

httpx/client.py
```diff
@@ -222,6 +223,25 @@ def params(self, params: QueryParamTypes) -> None:
...
+        if trust_env is not None:
+            warnings.warn(
+                "Passing a 'trust_env' argument when making a request on a client "
+                "is due to be deprecated. Instantiate a new client instead, "
+                "passing any 'trust_env' argument to the client itself."
+            )
```
StephenBrown2: Same category suggestion.
```suggestion
            , PendingDeprecationWarning)
```

tomchristie:
@StephenBrown2 I'm happy with adding PendingDeprecationWarning if we know that'll default to visible?

(For this 0.x case the most important thing is just to ensure it's a visible warning, and I'm otherwise not too fussed about warning categories *at this point*. We'll warn about this for 0.9.x, and it'll be removed in the next major release.)

tomchristie:
Okay, I've updated the requests compat docs to take account of "per-client SSL config" rather than "per-request SSL config". I've also given them a referesh.

docs/advanced.md
```diff
@@ -373,20 +379,36 @@ MIME header field.
...
+
+### SSL configuration on client instances
+
+If you're using a `Client()` instance, then you should pass any SSL settings when instantiating the client.
+
+```python
+client = httpx.Client(verify=False)
+```
+
+The `client.get(...)` method and other request methods *do not* changing the SSL settings on a per-request basis. If you need different SSL settings in different cases you should use more that one client instance, with different settings on each. Each client will then be using an isolated connection pool with a specific fixed SSL configuration on all connections within that pool.
```
florimondmanca: Erf…

```suggestion
The `client.get(...)` method and other request methods *do not* support changing the SSL settings on a per-request basis. If you need different SSL settings in different cases you should use more that one client instance, with different settings on each. Each client will then be using an isolated connection pool with a specific fixed SSL configuration on all connections within that pool.
```

tomchristie: Thanks

PR #598, opened 2019-12-05
Fix exception name in test case.

tomchristie:
* `Timeout` -> `TimeoutException` in an HTTP/2 test case.


tomchristie:
Merging this as a trivially correct case.

Issue #588, opened 2019-12-03
Reviewing the streaming API


tomchristie:
Right now we issue streaming requests using...

```python
response = await client.get(url, stream=True)
```

And have the following stream-specific methods...

* `.read()`
* `.raw()`
* `.stream()`
* `.stream_text()`
* `.stream_lines()`
* `.close()`

For the 1.0 release I think we should consider? droping `stream=True`, as it makes it *exceptionally* easy to leave unclosed connections hanging around.

I'm interested in if we might want to use the following instead...

```python
async with client.stream('GET', url) as response:  # Can only be used as a context manager.
    ...
```

That way users never have to think about "did I call close on the response", or leave open connections hanging around accidentally.

This is related to #393, but it's a stricter API, that doesn't rely on "users *might* use a response as a context manger". And also makes it far clearer *where and when* the streaming methods should be used. Ie. they only make sense inside a `.stream()` block.

I also think we could rationalize the streaming methods, so...

* `.read()`
* `.stream_raw()`
* `.stream_bytes()`
* `.stream_text()`
* `.stream_lines()`
* `.close()`

florimondmanca:
Thanks for bringing this up! I remember we mentioned it briefly elsewhere but it's good it has its own issue now.

I agree with the idea of a dedicated method, because passing `stream=True` induces an important change in behavior that should be better reflected in the API.

I also agree on enforcing context manager usage.

Thinking about alternatives, the fact that we have "one method per HTTP verb" for the non-streaming case and "one method accepting the HTTP verb" for the streaming case might look a bit odd.

(Strictly speaking, users *can* use `.request("GET", ...)` for the non-streaming case, but we don't really advertise that possibility e.g. in the docs.)

The only constructive idea I might have regarding this would be…

```python
async with client.stream.get(url) as response: ...
```

But a) flat is better tha nested and b) that would require us to implement *yet another host of methods* for each HTTP verb.

So, I think `.stream("<method", url, **kwargs)` is fine; at least *much better* than `.get(url, stream=True`).

> I also think we could rationalize the streaming methods

:100:, yes.

tomchristie:
> Thinking about alternatives, the fact that we have "one method per HTTP verb" for the non-streaming case and "one method accepting the HTTP verb" for the streaming case might look a bit odd.

Indeed yeah.

Options I can see here are...

Provide nested `.get()` and pals...

```python
async with client.stream.get(url) as response:
    ...
```

Explicit request methods, for the streaming case...

[...]

PR #599, opened 2019-12-05
Consistent suite of `stream_*` methods

tomchristie:
Renames `.raw()` to `.stream_raw()`, and renames `.stream()` to `.stream_bytes()`, so that we have a nice consistent suite of streaming operations...

* `response.stream_raw()`
* `response.stream_bytes)`
* `response.stream_text()`
* `response.stream_lines()`

Plus also...

* `response.read()`, `response.close()`

The existing `.raw` and `.stream` methods remain for compat, but issue warnings.

This is the first part of #588

We don't yet have any streaming docs. We might want to wait until we also address the second part of this work, and add the documentation as part of *that* pull request rather than *this* one, since we've got some new API that we want to add there.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -4,6 +4,7 @@ import email.message
 import json as jsonlib
 import typing
 import urllib.request
+import warnings
 from collections.abc import MutableMapping
 from http.cookiejar import Cookie, CookieJar
 from urllib.parse import parse_qsl, urlencode
@@ -701,7 +702,7 @@ class Request:
 
     async def read(self) -> bytes:
         """
-        Read and return the response content.
+        Read and return the request content.
         """
         if not hasattr(self, "content"):
             self.content = b"".join([part async for part in self.stream()])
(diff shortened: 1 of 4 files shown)


httpx/models.py
```diff
@@ -918,18 +919,34 @@ def __repr__(self) -> str:
...
+            self._content = b"".join([part async for part in self.stream_bytes()])
         return self._content
 
-    async def stream(self) -> typing.AsyncIterator[bytes]:
+    @property
+    def stream(self):  # type: ignore
+        warnings.warn(
+            "Response.stream() is due to be deprecated. "
+            "Use Response.stream_bytes() instead."
+        )
```
StephenBrown2: PendingDeprecationWarning?

```suggestion
        , PendingDeprecationWarning)
```

httpx/models.py
```diff
@@ -918,18 +919,34 @@ def __repr__(self) -> str:
...
+        return self.stream_bytes
+
+    @property
+    def raw(self):  # type: ignore
+        warnings.warn(
+            "Response.raw() is due to be deprecated. "
+            "Use Response.stream_raw() instead."
+        )
```
StephenBrown2: PendingDeprecationWarning?

```suggestion
        , PendingDeprecationWarning)
```

Issue #393, opened 2019-09-26
Allow Responses to be used as Context Managers


sethmlarson:
Requests API allows Responses to be used and closed as a context manager.

We should allow the following use-cases:

```python
async def run_async():
    client = httpx.AsyncClient()
    async with client.get("https://example.com") as resp:
        resp.raise_for_status()

def run_sync():
    client = httpx.Client()
    with client.get("https://example.com") as resp:
        resp.raise_for_status()
```

See: PendragonLore/anido#1

florimondmanca:
I think this is a good thing beyond Requests compatibility — ensures that all resources are properly closed in case something goes wrong. I'm kind of growing the intuition that ideally we should be following trio's lessons and enforce (async) context manager usage *everywhere* — on clients, on responses, …

tomchristie:
Okay, *so*...

This is relevant *only* for the streaming case.
In other cases the response has been closed by the point it's returned from the function.

There's a question here to be had in that `client.get(..., stream=True)` is actually fundamentally different to `client.get(...)` because you *really* want to be using a context manager for the first case.

Requests has exactly the same issue, and actually makes it really easy to *not* close streaming responses. One option that we could do here that's a bit radical would be to diverge from the `requests` API here, even through `stream=True` is a fairly widely used case.

Eg. `client.stream()` could return a context managed response. (It might have a default `method='get'` that could be overidden.)

One other bugbear to be wary of here is that "stream" in this context specifically means "stream the response", not "stream the request" (which yes, we support, but that's different), but the naming isn't very clear about the expectation.

gdhameeja:
I'd like to work on this.

tomchristie:
Main question here is where we want to go with the API...

* Add a `.stream(request='GET')` method which is *always* what we use for streaming cases, so that we can ensure that they're always context-manged.
* Stay in line with `requests` and just use `stream=True` on the existing methods. Keeps the API the same, but really easy to fail to close a response, which isn't *awful* but will leave the connection hanging.

sethmlarson:
Maybe we can put a warning on the Responses garbage collector that it wasn't closed properly, similar to files?

jaKass3:
Idk what to do I’m still learning 
    On Friday, October 4, 2019, 07:09:29 AM CDT, Seth Michael Larson <notifications@github.com> wrote:  
 
 
Maybe we can put a warning on the Responses garbage collector that it wasn't closed properly, similar to files?

tomchristie:
[...]

PR #600, opened 2019-12-05
Add Client.stream() method.

### LINKED ISSUE #555: Garbage collected responses should warn if still open.
See https://github.com/encode/httpx/issues/393#issuecomment-538370554

### LINKED ISSUE #588: Reviewing the streaming API

tomchristie:
Streaming a request looks like this...

```python
async with httpx.stream("GET", "https://www.example.com") as response:
    async for line in response.stream_lines():
        print(line)
```

This ensures that our streaming I/O is always properly context scoped.

Essentially closes #555
Closes #588

**TODO**:

- [x] Documentation.
- [x] Notes in Requests compat section.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -10,11 +10,10 @@ Pretty much any API mentioned in the `requests` QuickStart should be identical
 to the API in our own documentation. The following exceptions apply:
 
 * `Response.url` - Returns a `URL` instance, rather than a string. Use `str(response.url)` if you need a string instance.
-* `httpx.codes` - In our documentation we prefer the uppercased versions, such as `codes.NOT_FOUND`,
-but also provide lower-cased versions for API compatibility with `requests`.
-* `stream=True`. - Streaming responses provide the `.stream()` and `.raw()` byte iterator interfaces, rather than the `.iter_content()` method and the `.raw` socket interface.
+* `httpx.codes` - In our documentation we prefer the uppercased versions, such as `codes.NOT_FOUND`, but also provide lower-cased versions for API compatibility with `requests`.
+* `stream()`. - HTTPX provides a `.stream()` interface rather than using `stream=True`. This ensures that streaming responses are always properly closed outside of the stream block, and makes it visually clearer at which points streaming I/O APIs may be used with a response. Streaming request data is made avialable with `.stream_bytes()`, `.stream_text()`, `.stream_lines()`, and `.stream_raw()`.
 * `.get`, `.delete`, `.head`, `.options` -  These methods do not support `files`, `data`, or `json` arguments. Use `.request` if you need to need to send data using these http methods.
-* We don't support `response.is_ok` since the naming is ambiguous there, and might incorrectly imply an equivalence to `response.status_code == codes.OK`. Instead we provide an `.is_error` property.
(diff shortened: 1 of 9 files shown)


tomchristie:
Thanks, yes! (Amended now)

tomchristie:
@florimondmanca Thanks for having looked this over! Got the docs in now, too.

This one makes me particularly happy - just seems *might* nicer having any streaming I/O tightly scoped like this. 🤩

docs/compatibility.md
```diff
@@ -10,11 +10,10 @@ Pretty much any API mentioned in the `requests` QuickStart should be identical
 to the API in our own documentation. The following exceptions apply:
 
 * `Response.url` - Returns a `URL` instance, rather than a string. Use `str(response.url)` if you need a string instance.
-* `httpx.codes` - In our documentation we prefer the uppercased versions, such as `codes.NOT_FOUND`,
-but also provide lower-cased versions for API compatibility with `requests`.
-* `stream=True`. - Streaming responses provide the `.stream()` and `.raw()` byte iterator interfaces, rather than the `.iter_content()` method and the `.raw` socket interface.
+* `httpx.codes` - In our documentation we prefer the uppercased versions, such as `codes.NOT_FOUND`, but also provide lower-cased versions for API compatibility with `requests`.
+* `stream()`. - HTTPX provides a `.stream()` interface rather than using `stream=True`. This ensures that streaming responses are always properly closed outside of the stream block, and makes it visually clearer at which points streaming I/O APIs may be used with a response. Streaming request data is made avialable with `.stream_bytes()`, `.stream_text()`, `.stream_lines()`, and `.stream_raw()`.
```
florimondmanca:
```suggestion
* `stream()` - HTTPX provides a `.stream()` interface rather than using `stream=True`. This ensures that streaming responses are always properly closed outside of the stream block, and makes it visually clearer at which points streaming I/O APIs may be used with a response. Streaming request data is made available with `.stream_bytes()`, `.stream_text()`, `.stream_lines()`, and `.stream_raw()`.
```

vjpr:
This was marked as making #393 obsolete, however it now departs from the aiohttp api which allows:

```
  async with aiohttp.ClientSession() as client:
            async with client.request() as resp:
                ...
```

PR #602, opened 2019-12-05
Refactor Response.next() to perform no I/O

florimondmanca:
Refs https://github.com/encode/httpx/issues/588#issuecomment-562057229

As we want to progress towards separating responses that may perform I/O from those that don't, this PR refactors `.next()` so that it returns an opaque object that can be passed as-is to `client.send_handling_redirects()`.

(We can't just request a `Request` to pass to `client.send()` because we need to persist the history and some other parameters. But this means the implementation is not as elegant as I'd like it to be, though.)

This makes `.send_handling_redirects()` part of the public API, *which I think is fine*, but we'll probably need to document this and `response.next()` into the developer interface at some point?

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -44,6 +44,7 @@ from .models import (
     Request,
     RequestData,
     RequestFiles,
+    RequestForRedirect,
     Response,
     URLTypes,
 )
@@ -432,13 +433,18 @@ class Client:
 
     async def send_handling_redirects(
         self,
-        request: Request,
-        timeout: Timeout,
+        request: typing.Union[Request, RequestForRedirect],
+        timeout: Timeout = None,
         verify: VerifyTypes = None,
         cert: CertTypes = None,
         allow_redirects: bool = True,
         history: typing.List[Response] = None,
     ) -> Response:
+        if isinstance(request, RequestForRedirect):
+            return await self.send_handling_redirects(request.request, **request.kwargs)
+
+        assert timeout is not None
+
         if history is None:
             history = []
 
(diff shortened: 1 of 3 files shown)


tomchristie:
I’m not sure we should bother preserving the parameters or history or anything.

We don’t have to go completely down the `requests` style here, since we’ve got some different constraints in place.

I’d consider going lighter here, and just setting `.redirect_request = ...` and leaving it up to the user to call `.send` again.

(send_handling_redirects probably shouldn’t be public API)

I guess if we really wanted we could have a `request.redirect_history` in order to preserve the redirect response history in the same way as we currently do. (Incidentally I wonder if response.redirect_history might not also be a clearer name than .history?)

tomchristie:
Hrm. Actually not entirely sure about what’ll be best here. We’ll probably want to sketch out examples of how it’d look. Realised that it’s odd on the top level API, since there’s no .send there anyways.

Maybe we’re going down the wrong route here altogether.

tomchristie:
If we just annotate next as a callable with no extra info, and just underspecify “this will end up being an async callable if allow_redirects is used in the async case, or a sync callable in the sync case” then there’s no reason to drop response.next.

Type checkers wouldn’t know that it returns a response, but that’s okay.

*Or* we could have a next/anext pair, if we *did* want tighter type checking.

tomchristie:
I think we *probably* just don’t need this after all. (But I could be wrong, will be clearer once we think about splitting out an AsyncResponse/SyncResponse again.

Sorry for all the noise. 😑

florimondmanca:
Yeah, I'm not sure either. Mostly to open this as a way to get the thinking going. Your point about the top-level API is interesting, and means it's probably OK to let the response retain the callback to get the next response in the chain (synchronously or asynchronously). Closing for now!

PR #604, opened 2019-12-06
Add `OpenedConnection` base class.

tomchristie:
Establish a common interface that both `HTTP11Connection` and `HTTP2Connection` implement.

This would make it possible for `HTTPConnection` to hold a single `connection = typing.Optional[OpenedConnection]` rather than having specific instances for each of the H11 and H2 cases.

That'd also then allow us to switch to conditional imports of HTTP2Connection, only if `h2` is installed.

The blocker here right now is that `http_proxy` currently reaches into the H11 implementation.

### DIFF
--- a/httpx/dispatch/base.py
+++ b/httpx/dispatch/base.py
@@ -1,6 +1,7 @@
 import typing
 from types import TracebackType
 
+from ..concurrency.base import BaseSocketStream, ConcurrencyBackend
 from ..config import CertTypes, Timeout, VerifyTypes
 from ..models import (
     HeaderTypes,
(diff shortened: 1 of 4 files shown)


tomchristie:
I'm going to punt on this for now...

* It'll be slightly easier once `cert`/`verify` are no longer per-request / part of the dispatcher API, since that'll drop out a bit of cruft that we don't want.
* We ought to instead start on this the other way around. The initial part of approaching this incrementally would be...

Move the `start_tls` implementation stuff out of the proxy class and onto the `HTTPConnection` class. (No other changes, no base OpenedHTTPConnection class, *just* the "the HTTPConnection class is closer to the h11 state then the proxy class, so we should *start* to move the implementation in that direction.)

For that first pass we can be okay with HTTPConnection.start_tls *still* reaching into the h11 implementation. We'll then be in a nicer position of tackling the next step and moving the implementation of `start_tls` properly onto the H11Connection class.

PR #606, opened 2019-12-06
Version 0.9

tomchristie:
I think we're about ready to go here.

Final bits to decide:

- [x] Should we use `.aiter_content()`, `.aiter_text()`, `.aiter_lines()`, `.aiter_raw()` instead of `.stream_content()` etc?... #610
- [x] Should we drop the type annotations out of the top-level exports? #608


tomchristie:
⚡️✨🦋✨⚡️

PR #608, opened 2019-12-06
Tighten up top-level API to only expose public API

tomchristie:
Drops various private API from the top-level API.

We should recommend that our users *only* import the top-level `httpx` package, rather than importing sub-modules, in order to help guide them towards public API only.

Eg.

✅ Approved usage:

```python
import httpx
httpx.URL(...)
```

⚠️ Uh-oh:
```python
from httpx.models import URL
URL(...)
```

Refs #606

### DIFF
--- a/httpx/dispatch/proxy_http.py
+++ b/httpx/dispatch/proxy_http.py
@@ -1,4 +1,3 @@
-import enum
 import typing
 from base64 import b64encode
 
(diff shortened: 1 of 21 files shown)


httpx/dispatch/proxy_http.py
```diff
@@ -24,10 +23,9 @@
 logger = get_logger(__name__)
 
 
-class HTTPProxyMode(enum.Enum):
-    DEFAULT = "DEFAULT"
-    FORWARD_ONLY = "FORWARD_ONLY"
-    TUNNEL_ONLY = "TUNNEL_ONLY"
+DEFAULT_MODE = "DEFAULT"
+FORWARD_ONLY = "FORWARD_ONLY"
+TUNNEL_ONLY = "TUNNEL_ONLY"
```
florimondmanca: Dropping this enum is a breaking change — I believe `HTTPProxyMode` is documented under the proxy docs. Should we refrain from getting this change here, and see what we do with it later?

tomchristie: Thanks! Switched to a retain-for-compat style, with warnings raised.

I think it's a reasonable change, right? Better to have a minimal API surface area, than expose configuration enums.

tomchristie: Can roll back the change completely if you're unsure on it, mind.

PR #609, opened 2019-12-06
Move tunnel_start_tls() to HTTPConnection

florimondmanca:
Prompted by https://github.com/encode/httpx/issues/496#issuecomment-562505159

Removes pending HTTP version-specific usage from `HTTPProxy`, in order to make it easier to disable HTTP/2 by default in the future (by lazily importing `h2`).

I added a few extra comments in there to further clarify how and why `.tunnel_start_tls()` works.

### DIFF
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -2,9 +2,11 @@ import functools
 import ssl
 import typing
 
+import h11
+
 from ..concurrency.base import ConcurrencyBackend, lookup_backend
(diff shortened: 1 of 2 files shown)


tomchristie:
Yes! Lovely incremental refactor.

tomchristie:
I guess we may also have a second half of this where we move the implementation *all* the way onto `HTTP11Connection`, so that `HTTPConnection` is not reaching into its implementation details.

Issue #236, opened 2019-08-19
Cannot connect to yahoo.com, receiving a ProtocolError on Stream 0 after connection


sethmlarson:
Couple of issues here: we're not raising an exception after receiving a `ProtocolError` from HTTP/2 when connecting to `yahoo.com`. We should be bubbling that exception up to users.

The second issue is is we don't know why yahoo is giving us this error, we have to figure that out as I'm sure it'll impact more than just them.

I found this by running interop tests against Alexa 1000 websites.

tomchristie:
> I found this by running interop tests against Alexa 1000 websites.

Good call - were you just making a standard GET request to every site ranked in there, or something more complex than that?

sethmlarson:
I found plenty of problems with just getting the index. I've got to sort through them all cuz the Yahoo one kept hanging.

tomchristie:
> I found plenty of problems with just getting the index.

Okeydokes. At least we've got something to work from eh. 😃

tomchristie:
Something that'd be really interesting would be adding our proposed http version control switch, and runnning compat tests *only* against HTTP/1.1.

yeraydiazdiaz:
I've been looking into this issue with the help of #240 and been having trouble understanding our h2 flow.

Here's the logging for a request to `https://yahoo.com`

```
[DEBUG 2019-08-19 16:52:46,220] (httpx.dispatch.http2) H2 Connection initiating
[DEBUG 2019-08-19 16:52:46,221] (httpx.dispatch.http2) Stream [1] - Sending headers: b'\x00\x00/\x01\x04\x00\x00\x00\x01\x82A\x87\xf49\xceu\xc8z\x7f\x87\x84z\x8d\xaf\xd2g=KN\x94\xd7\xe5\x80.\xea\xe1S\x83\xf9c\xe7P\x8d\x9b\xd9\xab\xfaRB\
xcb@\xd2_\xa5#\xb3'
[DEBUG 2019-08-19 16:52:46,222] (httpx.dispatch.http2) Stream [1] - Ending request stream with data: b'\x00\x00\x00\x00\x01\x00\x00\x00\x01'
[...]

PR #612, opened 2019-12-06
HTTP/2 refactoring

tomchristie:
This is my first draft onto refactoring the HTTP/2 implementation into something a bit more comprehensible and less brittle.

It's *possible* that the current implementation might be slow - I think we need to be listening in the background for WindowUpdated events while we're sending the data, rather than sending data until the flow control window hits zero, and then listening for a WindowUpdated event. I've not thought that all through enough.

Probably worth trying it out on a couple of sites to see if I've got that right or not.

It's a really nice, clear, clean-up tho. We *actually* have a proper isolated HTTP2Stream rather than just mangling all the state on a single class.

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -104,39 +104,59 @@ class HTTP2Connection:
     async def close(self) -> None:
         await self.socket.close()
 
-    async def receive_event(self, stream_id: int, timeout: Timeout) -> h2.events.Event:
+    async def wait_for_outgoing_flow(self, stream_id: int, timeout: Timeout) -> int:
         """
-        Streams call into `connection.receive_event()` in order to read events
-        from the network. The connection manages holding onto any events that
-        have not yet been pulled by a stream.
+        Returns the maximum allowable outgoing flow for a given stream.
+
+        If the allowable flow is zero, then waits on the network until
+        WindowUpdated frames have increased the flow rate.
+
+        https://tools.ietf.org/html/rfc7540#section-6.9
+        """
+        local_flow = self.state.local_flow_control_window(stream_id)
+        connection_flow = self.state.max_outbound_frame_size
+        flow = min(local_flow, connection_flow)
+        while flow == 0:
+            await self.receive_events(timeout)
+            local_flow = self.state.local_flow_control_window(stream_id)
+            connection_flow = self.state.max_outbound_frame_size
+            flow = min(local_flow, connection_flow)
+        return flow
+
+    async def wait_for_event(self, stream_id: int, timeout: Timeout) -> h2.events.Event:
+        """
+        Returns the next event for a given stream.
+
+        If no events are available yet, then waits on the network until
+        an event is available.
         """
         while not self.events[stream_id]:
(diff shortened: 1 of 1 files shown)


tomchristie:
I've given this a run around and it's looking really good to me so far.

Hanging on www.yahoo.com, which I can't figure out - I don't recall if we'd resolved that fully or not. Haven't tested with 0.8 so see if it's a regression or not.

pquentin:
> Hanging on [www.yahoo.com](http://www.yahoo.com), which I can't figure out - I don't recall if we'd resolved that fully or not. Haven't tested with 0.8 so see if it's a regression or not.

I think you're looking for https://github.com/encode/httpx/issues/236. You memory served you well, it's not a regression as the issue was only partially resolved.

tomchristie:
Btw. I'd recommend reviewing by looking [at the new implementation itself](https://github.com/encode/httpx/blob/e00a2babe3f6ac27ff0348bb83b5a4b431e4ff0f/httpx/dispatch/http2.py) rather than reviewing the diff.

tomchristie:
Note that this change will also let use drop TimeoutFlags and fork, given that we’ve got a nice simple “send the request, read the response” flow. As it turns out we just don’t need to perform concurrent read/writes. As a result it’ll also give use sync support for HTTP/2.

httpx/dispatch/http2.py
```diff
@@ -116,16 +88,113 @@ def init_complete(self) -> BaseEvent:
...
+        """
+        local_flow = self.state.local_flow_control_window(stream_id)
+        connection_flow = self.state.max_outbound_frame_size
+        flow = min(local_flow, connection_flow)
+        while flow == 0:
+            await self.receive_events(timeout)
+            local_flow = self.state.local_flow_control_window(stream_id)
+            connection_flow = self.state.max_outbound_frame_size
+            flow = min(local_flow, connection_flow)
+        return flow
```
florimondmanca: Some duplicated lines here - possible refactor?

```python
        while True:
            local_flow = self.state.local_flow_control_window(stream_id)
            connection_flow = self.state.max_outbound_frame_size
            flow = min(local_flow, connection_flow)
            if flow == 0:
                await self.receive_events(timeout)
                continue
            return flow
```

I can see that the current implementation really fits the docstring though, so not particularly urging about this.

httpx/dispatch/http2.py
```diff
@@ -135,67 +204,38 @@ def init_complete(self) -> BaseEvent:
...
         """
         Read the response status and headers from the network.
         """
         while True:
-            event = await self.receive_event(stream_id, timeout)
-            # As soon as we start seeing response events, we should enable
-            # read timeouts, if we haven't already.
-            self.timeout_flags[stream_id].set_read_timeouts()
+            event = await self.connection.wait_for_event(self.stream_id, timeout)
+
```
florimondmanca: Nit - superfluous blank line

```suggestion
```

httpx/dispatch/http2.py
tomchristie: Yup, may well be worthwhile. I’ll get this in as-is for now, and can follow up.

PR #614, opened 2019-12-07
Add 'fork' to auto backend

tomchristie:
Seems that we don't have coverage for this yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.9.3 (December 7th, 2019)
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


tomchristie:
No sweat! 😃

tomchristie:
Incidentally, #612 actually means we *won't* end up needing fork or a background manager *at all*.

florimondmanca:
Is this actually causing errors? Curious about why we didn't catch this… Probably worth refactoring our test setup to test all possible backends - right now I believe we force `backend=AsyncioBackend()` and `backend=TrioBackend()`, so don't cover the `auto` part nor passing a string.

tomchristie:
Yes - causing a `NotImplementedError` when using HTTP/2.

We'll probably want to test using `backend='auto'`, `backend='trio'`, `backend='asyncio'`, but there may be a bit of rejigging to do wrt. running the test server.

florimondmanca:
@tomchristie I'm taking a look at it. :-)

PR #615, opened 2019-12-07
Refactor tests in the light of backend auto-detection

florimondmanca:
Following up comments from #614, this PR updates the tests so that:

- They prefer passing the string-form backend (`"auto"`, `"asyncio"`, `"trio"`) to all pieces of public API.
- The `"auto"` option is put under test as well.

Fixes a bunch of `NotImplementedError`s in the `concurrency` test utils as well.

Tests that use the `backend` are now twice as many (because the `backend` fixture now has two more params - auto-in-asyncio and auto-in-trio). I figured the test suite was still fast enough that running all these tests for each situation was acceptable, rather than e.g. unit-testing that passing `"asyncio"` looks up `AsyncioBackend`. Now that I wrote about it I'm less sure this wouldn't be preferable.

### DIFF
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -160,3 +160,17 @@ async def test_uds(uds_server, backend):
     assert response.status_code == 200
     assert response.text == "Hello, world!"
     assert response.encoding == "iso-8859-1"
+
+
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pytest.param("asyncio", marks=pytest.mark.asyncio),
+        pytest.param("trio", marks=pytest.mark.trio),
+    ],
+)
+async def test_explicit_backend(server, backend):
+    async with httpx.Client(backend=backend) as client:
+        response = await client.get(server.url)
+    assert response.status_code == 200
+    assert response.text == "Hello, world!"
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -7,7 +7,7 @@ import httpx
 
 async def test_get(server, backend):
     url = server.url
-    async with httpx.Client(backend=backend) as client:
+    async with httpx.Client() as client:
         response = await client.get(url)
     assert response.status_code == 200
     assert response.text == "Hello, world!"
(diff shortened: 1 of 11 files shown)


florimondmanca:
Okay, I've moved things around a bit more so that:

- We only have two params for `backend`: auto-in-asyncio, auto-in-trio. (In particular this allows running tests for only one of the async libraries using `pytest -m [asyncio|trio]`, which we'd lose with the previous pass.)
- We rely on auto-detection, *as most users should*, i.e. don't pass `backend=backend` to clients, connection pools or connections.
- We test passing the backend explicitly in a dedicated test.

tests/client/test_async_client.py
```diff
@@ -7,7 +7,7 @@
 
 async def test_get(server, backend):
```
StephenBrown2: Can also remove the `backend` fixture param from all these test functions since it's not being referenced anywhere.

florimondmanca: Actually we can’t; even though we don’t use the backend fixture in the test it’s responsible for setting up the asyncio/trio environment. I agree the usage as a function parameter is misleading though. We can probably switch to pytest.mark.usefixtures.

florimondmanca: > We can probably switch to pytest.mark.usefixtures.

... and rename the fixture to « async_environment » or something? It’s true that the fixture doesn’t even return a backend class — it’s always « auto ». Hmm.

StephenBrown2: Yeah, I only suggested it since it is supposed to be auto-detected now, and they could be parameterized like https://github.com/encode/httpx/pull/615/files#diff-21bb0534f2645b17f8572db206999d00R165-R171
Which I now see is done here: https://github.com/encode/httpx/pull/615/files#diff-484462fced51d1a06b1d93b4a44dd535R52-R64 so, well done! Sorry I missed that before.
It is a little confusing as a function param, so I would agree with renaming it "async_backend" and then using it as `@pytest.mark.usefixtures("async_backend")` since all it's doing is adding the mark for asyncio/trio to do their thing.

PR #616, opened 2019-12-07
Add OpenConnection base class

florimondmanca:
To progress towards resolving #496, we need to isolate the manipulation of `HTTP2Connection` in `HTTPConnection`.

This PR (inspired by #604) introduces a base `OpenConnection` interface that `HTTP11Connection` and `HTTP2Connection` implement, which allows `HTTPConnection` to hold a single `open_connection` object instead of one of `h11_connection` or `h2_connection`.

### DIFF
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -4,11 +4,11 @@ import typing
 
 import h11
 
-from ..concurrency.base import ConcurrencyBackend, lookup_backend
+from ..concurrency.base import BaseSocketStream, ConcurrencyBackend, lookup_backend
 from ..config import CertTypes, SSLConfig, Timeout, VerifyTypes
 from ..models import URL, Origin, Request, Response
 from ..utils import get_logger
-from .base import Dispatcher
+from .base import Dispatcher, OpenConnection
 from .http2 import HTTP2Connection
 from .http11 import HTTP11Connection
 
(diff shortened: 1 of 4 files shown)


httpx/dispatch/connection.py
```diff
@@ -150,13 +141,21 @@ def __init__(
...
+
+    def set_open_connection(
+        self,
+        http_version: str,
+        socket: BaseSocketStream,
+        on_release: typing.Optional[typing.Callable],
+    ) -> None:
         if http_version == "HTTP/2":
-            self.h2_connection = HTTP2Connection(
+            self.open_connection = HTTP2Connection(
```
florimondmanca: This is the only place where with reference `HTTP2Connection` directly, so we'll be able to lazily-import it here, and fail if `h2` isn't installed. :+1:

httpx/dispatch/connection.py
```diff
@@ -92,13 +88,7 @@ def __init__(
...
         logger.trace(f"connected http_version={http_version!r}")
 
-        if http_version == "HTTP/2":
-            self.h2_connection = HTTP2Connection(
-                stream, backend=self.backend, on_release=on_release
-            )
-        else:
-            assert http_version == "HTTP/1.1"
-            self.h11_connection = HTTP11Connection(stream, on_release=on_release)
+        self.set_open_connection(http_version, socket=stream, on_release=on_release)
```
florimondmanca: Note that we've got an outstanding `stream` vs `socket` instance here (following up #601). Wondering if we'd want to push the renaming further with `.open_tcp_socket()` and `.open_uds_socket()` on the `ConcurrencyBackend` interface.

tomchristie: Seems reasonable, yup.

florimondmanca:
Thanks @tomchristie - I see what you mean about the remaining h11 internals being accessed in HTTPConnection. Seems separate from the scope of this PR so I’ll merge and we can address it later!

PR #617, opened 2019-12-07
Drop per-request `cert`, `verify`, and `trust_env`

### LINKED ISSUE #554: Drop per-request `cert`/`verify`/`trust_env`

florimondmanca:
Fixes #554.

This PR restricts configuration to a per-client, per-connection-pool and per-connection basis, instead of allowing to override it per-request:

- Remove the `cert`, `verify` and `trust_env` parameters from the request methods on `Client`.
- Drop `cert` and `verify` from the dispatcher API (and concrete dispatchers).

It probably shouldn't be merged before 1.0 lands, though. I just felt like going through this to see how it would end up looking like. :+1:


tomchristie:
So keen to get this in! 😃

florimondmanca:
@tomchristie Is that a PR approval in disguise…? :-)

tomchristie:
We shouldn’t merge yet, no, since it’d block us making 0.9.x point releases.

We *could* potentially have a `version-1.0` branch tho?

florimondmanca:
Ah, of course.

I'm good with leaving this as-is for now. Managing v1 branches is a bit of a pain, especially since we don't know exactly when we're going to get there.

tomchristie:
Might be worth bringing this up to date now and starting a 1.0 branch?

florimondmanca:
Yup. What workflow would you recommend for managing the `1.0` branch? Merge `master` into `1.0` after every PR?

tomchristie:
> Yup. What workflow would you recommend for managing the 1.0 branch? Merge master into 1.0 after every PR?

I guess so, yea.

We probably want to:

* Get #636 sorted out and in.
* Issue a 0.9.5 release.
* Get #617, #639, and #644 into a 1.0 branch.

That'll get us into a position where we're ready to start adding in a SyncClient again.

tomchristie:
So, to get this in for a 0.10.0 we may want leave `cert`/`verify`/`timeout` on the public API but have them raise errors now, and don't include them in the docstrings. We can drop them throughout the internals and dispatch API.

Seem like a reasonable approach?

florimondmanca:
Sounds good - will update the removal of warnings to raise errors instead. 👍

florimondmanca:
@tomchristie Should be ready for re-review :tada:

tomchristie:
Looks great! How do we want to do this... should we gradually merge everything on #667 into a version-0.10.0 branch, or are we confident enough that it's a small set of stuff now, and just work on getting these bits into master and then releasing?

florimondmanca:
I'd be in favor of merging bits from #667 into master. :-) We've got the 0.9.5 tag available to us, so if there's any hot fix we need to release as a 0.9.6 patch we can still cherry-pick it there from master.

PR #623, opened 2019-12-09
Dropped unreliable test cases

tomchristie:
We've at least already got coverage here, and these two tests are proving too unreliable, so I think the pragmatic option here is just to drop them.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -173,22 +173,3 @@ def test_merge_url():
 
     assert url.scheme == "https"
     assert url.is_ssl
-
-
-@pytest.mark.asyncio
-async def test_elapsed_delay(server):
-    url = server.url.copy_with(path="/slow_response/100")
-    async with httpx.Client() as client:
-        response = await client.get(url)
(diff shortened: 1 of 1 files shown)


tomchristie:
Eg see test run on #622

tests/client/test_client.py
```diff
@@ -173,22 +173,3 @@ def test_merge_url():
...
     assert url.scheme == "https"
     assert url.is_ssl
-
-
-@pytest.mark.asyncio
-async def test_elapsed_delay(server):
-    url = server.url.copy_with(path="/slow_response/100")
-    async with httpx.Client() as client:
-        response = await client.get(url)
-    assert response.elapsed.total_seconds() == pytest.approx(0.1, rel=0.2)
```
florimondmanca: Can we maybe keep this one and assert that the number of seconds is between 0 and 0.5, manually? I don’t think we actually access response.elapsed anywhere else in the test suite.

tomchristie:
Ah darn. Pushed https://github.com/encode/httpx/commit/02d16bfb34e2af183b1b014b3109e1e322611758 to master by accident.

Have now turned on "Require 1 review" and "Include administrators" in the protected branches settings.

Issue #527, opened 2019-11-14
Python 3.7: RuntimeError: read() called while another coroutine is already waiting for incoming data 


PrimozGodec:
As we already discussed with @florimondmanca in https://github.com/encode/httpx/issues/382, I am reporting this issue here. 

I get the following error: `RuntimeError: read() called while another coroutine is already waiting for incoming data` when I am sending images to the server. As you can see from the code it is implemented that only max 100 requests are sent to the server at the same time. I know there could be lazy loading used but the limitation is made custom because before we send images to the server we check whether they are in the local cache or not. The limit of 100 is set that not all images are loaded in the memory at the same time (there can be also 10000 images).

Here is the trace:
```
Traceback (most recent call last):
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 63, in <module>
    asyncio.get_event_loop().run_until_complete(cl.embedd_batch())
  File "/Users/primoz/miniconda3/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 18, in embedd_batch
    embeddings = await asyncio.gather(*requests)
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 41, in _send_to_server
    emb = await self._send_request(client, im, url)
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 56, in _send_request
    data=image
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 484, in post
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 626, in request
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 650, in send
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 265, in _get_response
    return await get_response(request)
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/middleware/redirect.py", line 31, in __call__
    response = await get_response(request)
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 226, in get_response
    request, verify=verify, cert=cert, timeout=timeout
[...]

PR #628, opened 2019-12-11
Ensure H2 state is only accessed by the connection, not per-stream.

tomchristie:
This refactor moves all the H2 state management onto the Connection class, rather than being per-stream.

I think we'll want to eventually move our stream-locks out of the stream implementation and onto  HTTP2Connection instead, so eg...

```python
async with self.write_lock:  # Ensure that outgoing data cannot get sent out-of-sequence.
    data_to_send = self.state.data_to_send()
    await self.socket.write(data_to_send, timeout)
```

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -64,7 +64,7 @@ class HTTP2Connection(OpenConnection):
             await self.init_complete.wait()
             stream_id = self.state.get_next_available_stream_id()
 
-        stream = HTTP2Stream(stream_id=stream_id, connection=self, state=self.state)
+        stream = HTTP2Stream(stream_id=stream_id, connection=self)
         self.streams[stream_id] = stream
         self.events[stream_id] = []
         return await stream.send(request, timeout)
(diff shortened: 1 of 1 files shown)


florimondmanca:
> I think we'll want to eventually move our stream-locks out of the stream implementation and onto HTTP2Connection instead

It's probably a good idea, provided those locks were indeed added to solve issues for HTTP/2. (It was the case for the `read_lock` - see #527. For `write_lock`, I *think* we only have it on trio for now - it was added as part of #276, and I *think* that was also to fix an HTTP/2 concurrency issue, see https://github.com/encode/httpx/pull/276#issuecomment-525469745.)

tomchristie:
Thanks, those are great pointers.

httpx/dispatch/http2.py
```diff
@@ -172,15 +194,9 @@ def is_connection_dropped(self) -> bool:
 
 
 class HTTP2Stream:
-    def __init__(
-        self,
-        stream_id: int,
-        connection: HTTP2Connection,
-        state: h2.connection.H2Connection,
-    ) -> None:
+    def __init__(self, stream_id: int, connection: HTTP2Connection,) -> None:
```
florimondmanca: Formatting nit (I've already had these, not sure why Black doesn't reformat them):

```suggestion
    def __init__(self, stream_id: int, connection: HTTP2Connection) -> None:
```

tomchristie: Thanks!

StephenBrown2: https://github.com/psf/black#trailing-commas was changed recently, I believe https://github.com/psf/black/pull/826 was the ~culprit~ responsible PR. There appears to be a follow-up PR https://github.com/psf/black/pull/1164 to remedy it.

Issue #581, opened 2019-12-02
Random KeyErrors when performing GET requests (possible race condition)


KyleKreutzer:
When performing a get request using httpx.client.get, I am (with every 1/10 requests or so) getting random key errors. The only argument being handed to this call is a 5 second timeout, a google "NID" cookie, and the URL. I've attached a text file with the error.
[KeyError.txt](https://github.com/encode/httpx/files/3909364/KeyError.txt)

tomchristie:
Thanks for reporting this.

What's the simplest possible example that reproduces this reliably?
Is this only occuring against a specific URL, or can you determine various cases where its failing?

KyleKreutzer:
I've only tried it against the Google Trends API (https://trends.google.com/trends/api/*). The simplest way to reproduce the error would be to pull [](https://github.com/KyleKreutzer/pytrends-async) and run the unit tests. The problem is occurring intermittently so it may take more than 1 run to have the issue appear, but any of those tests seem to reliably reproduce the problem.

KyleKreutzer:
`        for i in range(100):
            async with httpx.Client() as c:
                resp = await c.get('https://trends.google.com/?geo={geo}'.format(geo='US'), timeout=5)
                time.sleep(1)`

This will produce the bug

millefalcon:
Is this still reproducible ?
I couldn't reproduce it with `python3.6.9`,

millefalcon:
Is this still reproducible ?
Couldn't reproduce it with:
```
python3.6.9
trio==0.13.0
git+https://github.com/encode/httpx@75a439d7e2120edcddd8213df27dfeea28a2dc8c#egg=httpx
```

KyleKreutzer:
I was able to reproduce it again with above code snippet on both mac and windows using 3.7.0 and 3.7.3. I am not using trio and am using version httpx==0.8.0 from pypi.

tomchristie:
Closed in master via #612

PR #629, opened 2019-12-12
Bump up flow control defaults

### LINKED ISSUE #262: HTTP/2 download speeds / flow control settings

tomchristie:
Have confirmed that this closes #262

😊😊😊

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -96,6 +96,7 @@ class HTTP2Connection(OpenConnection):
         del self.state.local_settings[h2.settings.SettingCodes.ENABLE_CONNECT_PROTOCOL]
 
         self.state.initiate_connection()
+        self.state.increment_flow_control_window(2**24)
         data_to_send = self.state.data_to_send()
         await self.socket.write(data_to_send, timeout)
 
@@ -165,6 +166,7 @@ class HTTP2Connection(OpenConnection):
         timeout: Timeout,
     ) -> None:
         self.state.send_headers(stream_id, headers)
+        self.state.increment_flow_control_window(2**24, stream_id=stream_id)
         data_to_send = self.state.data_to_send()
         await self.socket.write(data_to_send, timeout)
 
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -96,7 +96,7 @@ class HTTP2Connection(OpenConnection):
         del self.state.local_settings[h2.settings.SettingCodes.ENABLE_CONNECT_PROTOCOL]
 
         self.state.initiate_connection()
-        self.state.increment_flow_control_window(2**24)
+        self.state.increment_flow_control_window(2 ** 24)
         data_to_send = self.state.data_to_send()
         await self.socket.write(data_to_send, timeout)
 
@@ -166,7 +166,7 @@ class HTTP2Connection(OpenConnection):
         timeout: Timeout,
     ) -> None:
         self.state.send_headers(stream_id, headers)
-        self.state.increment_flow_control_window(2**24, stream_id=stream_id)
+        self.state.increment_flow_control_window(2 ** 24, stream_id=stream_id)
         data_to_send = self.state.data_to_send()
         await self.socket.write(data_to_send, timeout)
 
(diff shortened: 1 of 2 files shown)


httpx/dispatch/http2.py
```diff
@@ -96,6 +96,7 @@ def init_complete(self) -> BaseEvent:
         del self.state.local_settings[h2.settings.SettingCodes.ENABLE_CONNECT_PROTOCOL]
 
         self.state.initiate_connection()
+        self.state.increment_flow_control_window(2 ** 24)
```
florimondmanca: Our of curiosity, what’s the default value for the window?

florimondmanca:
Also we’re getting strange KeyErrors on CI. Not sure if this is related to #581?

httpx/dispatch/http2.py
tomchristie: 65535

tomchristie:
> Could you verify the results from #262 (comment) got improved?

Yup that was what I was testing against.

tomchristie:
@florimondmanca No I don't think it's related to that. We've just got a bit of a flaky mock HTTP/2 stream implementation in our tests cases, that fails once we make this change. I've cut out some stuff that we don't appear to need now that we have a neater send/receive flow.

florimondmanca:
I tested it out locally too, and can confirm the performance improvement!

Code:

```python
import asyncio
import contextlib
import httpx
import time
import requests

url = "https://quic.aiortc.org/50000000"


@contextlib.contextmanager
def timer(flavor: str):
    start = time.time()
    try:
        yield
    finally:
        elapsed = time.time() - start
        print(f"{flavor} {elapsed:.1f}s")


async def main() -> None:
    with timer("requests"):
        requests.get(url)

    async with httpx.Client() as h1_client, httpx.Client(http2=True) as h2_client:
        with timer("httpx HTTP/1.1"):
            r = await h1_client.get(url)
            assert r.http_version == "HTTP/1.1"

        with timer("httpx HTTP/2"):
            r = await h2_client.get(url)
            assert r.http_version == "HTTP/2"


asyncio.run(main())

```

Results:

```console
requests 3.3s
httpx HTTP/1.1 3.3s
httpx HTTP/2 4.3s
```

On `master` this gave me:

```
requests 3.0s
httpx HTTP/1.1 3.3s
httpx HTTP/2 30.5s
```

There's a lot of variance between runs, but yeah we're now on the same order of magnitude between HTTP/1.1 and HTTP/2! :tada:

httpx/dispatch/http2.py
florimondmanca: Okay. Should this line require a comment of any sort? It's the kind of line I'd bump into and wonder "Why?".

tomchristie:
> There's a lot of variance between runs

Indeed yes. I've sometime seen it a touch faster than HTTP/1.1, sometimes a touch slower. I think it comes out a shade slower overall, but that's not surprising, and it's clearly in the same range now.

tomchristie:
A bunch of run throughs...

```shell
$ venv/bin/python ./example.py 
requests 5.7s
httpx HTTP/1.1 7.0s
httpx HTTP/2 6.1s

$ venv/bin/python ./example.py 
requests 6.3s
httpx HTTP/1.1 6.2s
httpx HTTP/2 6.1s

$ venv/bin/python ./example.py 
requests 6.3s
httpx HTTP/1.1 6.1s
httpx HTTP/2 9.2s

$ venv/bin/python ./example.py 
requests 5.3s
httpx HTTP/1.1 5.3s
httpx HTTP/2 5.2s

$ venv/bin/python ./example.py 
requests 6.6s
httpx HTTP/1.1 13.2s
httpx HTTP/2 6.6s

$ venv/bin/python ./example.py 
requests 11.9s
httpx HTTP/1.1 7.9s
httpx HTTP/2 7.4s
```

lundberg:
> A bunch of run throughs...

[hyperfine](https://github.com/sharkdp/hyperfine) would be useful here 😉

PR #635, opened 2019-12-13
Using charset_normalizer instead of chardet in models.py and decoder.py

Ousret:
Hi,

With this PR I am proposing to replace chardet by charset_normalizer.
This new package is MIT licensed and support more encoding, also young. But it is slower for now. (not by much)

Complete details are available at : https://github.com/Ousret/charset_normalizer

charset_normalizer is still an ongoing project, so there is still maintenance as of now. (by me, mostly)

If anyone is willing to give it a try and send feedbacks, I would be thrilled to listen.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,5 +22,8 @@ trio
 trustme
 uvicorn
 seed-isort-config
+rfc3986
+certifi
+h2
 
 attrs>=19.2  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,7 @@ trio
 trustme
 uvicorn
 seed-isort-config
+charset_normalizer>=1.3,<2
 rfc3986
 certifi
 h2
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,8 +23,5 @@ trustme
 uvicorn
 seed-isort-config
 charset_normalizer>=1.3,<2
-rfc3986
-certifi
-h2
 
 attrs>=19.2  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The httpx project relies on these excellent libraries:
 * `h2` - HTTP/2 support.
 * `h11` - HTTP/1.1 support.
 * `certifi` - SSL certificates.
-* `chardet` - Fallback auto-detection for response encoding.
+* `charset_normalizer` - Fallback auto-detection for response encoding.
 * `hstspreload` - determines whether IDNA-encoded host should be only accessed via HTTPS.
 * `idna` - Internationalized domain name support.
 * `rfc3986` - URL parsing & normalization.
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,7 +111,7 @@ The HTTPX project relies on these excellent libraries:
 * `h2` - HTTP/2 support.
 * `h11` - HTTP/1.1 support.
 * `certifi` - SSL certificates.
-* `chardet` - Fallback auto-detection for response encoding.
+* `charset_normalizer` - Fallback auto-detection for response encoding.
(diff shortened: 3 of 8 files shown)


tomchristie:
That's really interesting.

There's some failing test cases ATM. There's also an issue with the iterative decoding, which doesn't "feed some data, then see where we are now", but instead re-runs the decoding on the buffer each time, which *could* potentially be problematic?

Promising, tho...

Shivelight:
Why not giving users the option to choose which detector they want to use? Maybe one wishes to use cChardet for performance or whatever reason.

I would love to have this as the default detector tho :+1:

Ousret:
Tests concerned are 

```py
@pytest.mark.parametrize(
    ["data", "encoding"],
    [
        ((b"Hello,", b" world!"), "ascii"),
        ((b"\xe3\x83", b"\x88\xe3\x83\xa9", b"\xe3", b"\x83\x99\xe3\x83\xab"), "utf-8"),
        ((b"\x83g\x83\x89\x83x\x83\x8b",) * 64, "shift-jis"),
        ((b"\x83g\x83\x89\x83x\x83\x8b",) * 600, "shift-jis"),
        (
            (b"\xcb\xee\xf0\xe5\xec \xe8\xef\xf1\xf3\xec \xe4\xee\xeb\xee\xf0",) * 64,
            "MacCyrillic",
        ),
        (
            (b"\xa5\xa6\xa5\xa7\xa5\xd6\xa4\xce\xb9\xf1\xba\xdd\xb2\xbd",) * 512,
            "euc-jp",
        ),
    ],
)
@pytest.mark.asyncio
async def test_text_decoder(data, encoding):
    async def iterator():
        nonlocal data
        for chunk in data:
            yield chunk

    response = httpx.Response(200, content=iterator())
    await response.read()
    assert response.text == (b"".join(data)).decode(encoding)

```

tomchristie:
> Why not giving users the option to choose which detector they want to use?

Where possible we should generally prefer less configuration rather than more.

Shivelight:
> Where possible we should generally prefer less configuration rather than more.

I was thinking of handling it like aiohttp, use cChardet if it's installed (but it's not better than charset_normalizer except for performance).

I agree with less configuration and the case might rather specific, people can monkey patch this easily anyway.

requirements.txt
```diff
@@ -22,5 +22,9 @@ trio
 trustme
 uvicorn
 seed-isort-config
+charset_normalizer>=1.3,<2
+rfc3986
+certifi
+h2
```
florimondmanca: Were these dependencies added because we miss them for our base development install?

Ousret: Hi, When cloning the base project I was unable to use it in my py3.8 venv. theses packages were missing. running `python setup.py install` was not enough. Maybe I missed something. `model.py` require `rfc3986`. The base package fail and also the test suite.

Ousret:
> > Where possible we should generally prefer less configuration rather than more.
> 
> I was thinking of handling it like aiohttp, use cChardet if it's installed (but it's not better than charset_normalizer except for performance).
> 
> I agree with less configuration and the case might rather specific, people can monkey patch this easily anyway.

cChardet is not python native, so without whl it would reduce multi platform support.

Shivelight:
> cChardet is not python native, so without whl it would reduce multi platform support. and conditional import is not a very good idea I think.

Yea I agree. I didn't think much when writing that comment. Please ignore

requirements.txt
Shivelight: Did you run `scripts/install` first?

florimondmanca: @Ourset Yup, we have an install script that should take care of installing all package and development dependencies.

Ousret: My bad. I will revert this change.

Ousret: Done

tomchristie:
I'd like to close this off for now. It is *feasible* that we could consider this a little further down the line, tho.
At the moment switching across to a much newer project would introduce a bit of a risk factor, that we could do without.

PR #636, opened 2019-12-13
Added RequestContent

tomchristie:
## The RequestContent interface

This pull requests adds `RequestContent` as the internal interface for all request body construction. The RequestContent interface provides:

* `.get_headers()` - Return the headers for the encoding.
* `.is_rewindable()` - True if the content can be replayed, False otherwise.
* `async def __aiter__() -> bytes` - Returns the byte stream for the encoded data.

There are a few concrete implementations.

* `RequestContent` - An empty request content.
* `BytesRequestContent` - Handles plain bytes and str content.
* `StreamingRequestContent` - Handles byte async iterables. Not rewindable.
* `JSONRequestContent` - Handles JSON data.
* `URLEncodedRequestContent` - Handles URL encoded form data.
* `MultipartEncodedRequestContent` - Handles Multipart encoded form data.

There is also a top level function for handling returning a `RequestContent` instance, given `data, files, json` arguments.

* `encode(data, files, json) -> RequestContent`

## Motivation

Short: Generally a much nicer seperation of logic & will allow us to support both sync+async cases on a single `Request` class.

* All the request encoding logic is nicely seperated for the `Request` class itself, and testable in isolation.
* We'll be able to have both async and sync variants by implementing `__iter__` on the the interface alongside the existing `__aiter__` - most encodings will support both. The StreamingRequestContent will end up as two distinct cases sync+async, which will only support one or the other case, and will error if used incorrectly. (Eg. passing an async iterable as a `data=` argument will raise an error when used with a sync client.)
* This gets us much closer to being able to implement a streaming, rewindable multipart implementation.

### DIFF
--- a/httpx/content.py
+++ b/httpx/content.py
@@ -1,0 +1,149 @@
+import typing
+from json import dumps as json_dumps
+from urllib.parse import urlencode
+
+from .multipart import multipart_encode
+
+RequestData = typing.Union[dict, str, bytes, typing.AsyncIterator[bytes]]
+
+RequestFiles = typing.Dict[
+    str,
+    typing.Union[
+        # file (or str)
+        typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+        # (filename, file (or str))
+        typing.Tuple[
+            typing.Optional[str], typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+        ],
+        # (filename, file (or str), content_type)
+        typing.Tuple[
+            typing.Optional[str],
+            typing.Union[typing.IO[typing.AnyStr], typing.AnyStr],
+            typing.Optional[str],
+        ],
+    ],
+]
+
+
+class RequestContent:
+    def headers(self) -> typing.Dict[str, str]:
+        return {}
+
+    def can_rewind(self) -> bool:
+        return True
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        yield b""  # pragma: nocover
+
+    async def aread(self) -> bytes:
+        return b"".join([part async for part in self])
+
+
+class BytesRequestContent(RequestContent):
+    """
+    Request content encoded as plain bytes.
+    """
+
+    def __init__(self, body: typing.Union[str, bytes]) -> None:
+        self.body = body.encode("utf-8") if isinstance(body, str) else body
+
+    def headers(self) -> typing.Dict[str, str]:
+        content_length = str(len(self.body))
+        return {"Content-Length": content_length}
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        yield self.body
+
+
+class StreamingRequestContent(RequestContent):
+    """
+    Request content encoded as plain bytes, using an async byte iterator.
+    """
+
+    def __init__(self, aiterator: typing.AsyncIterator[bytes]) -> None:
+        self.aiterator = aiterator
+
+    def can_rewind(self) -> bool:
+        return False
+
+    def headers(self) -> typing.Dict[str, str]:
+        return {"Transfer-Encoding": "chunked"}
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        async for part in self.aiterator:
+            yield part
+
+
+class JSONRequestContent(RequestContent):
+    """
+    Request content encoded as JSON.
+    """
+
+    def __init__(self, json: typing.Any) -> None:
+        self.body = json_dumps(json).encode("utf-8")
+
+    def headers(self) -> typing.Dict[str, str]:
+        content_length = str(len(self.body))
+        content_type = "application/json"
+        return {"Content-Length": content_length, "Content-Type": content_type}
+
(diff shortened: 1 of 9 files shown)


tomchristie:
NB. The `get_headers()` interface could potentially be plainer (eg. list of byte tuples) - we can consider any further tweaks like that if needed easily enough, since this is currently a private interface.

We'll want to think a little about exactly about if/how we want to expose any of this on a `Request` instance. On this pass I've wired the existing `request.read()` and `request.stream()` methods onto the new implementation. It's feasible that we'll want to drop that an instead expose `request.content` as a public interface. (Not that most folks will be working at the level, mind.)

tomchristie:
Much design credit here is due to @sethmlarson.
Also ties in with how request/response content on any lower-level Dispatcher API would look.

httpx/content.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    if data is None:
+        if json is not None:
+            return JSONRequestContent(json)
+        elif files:
+            return MultipartRequestContent({}, files, boundary=boundary)
+        else:
+            return RequestContent()
+    elif isinstance(data, dict):
+        if files:
```
florimondmanca: For consistency/explicitness:

```suggestion
        if files is not None:
```

httpx/content.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    Base class for request content.
+    Defaults to a "no request body" implementation.
+    """
+
+    def get_headers(self) -> typing.Dict[str, str]:
+        return {}
+
+    def can_rewind(self) -> bool:
+        return True
```
florimondmanca: What is this property supposed to enable in practice? Does that mean we can `async for chunk in content` again once we've done so at least once? I can't think of better names (`.is_replayable()`? `.can_replay()`? `.can_read_multiple_times`? `.is_isomorphic()`? 🤣), but maybe a docstring on this base implementation could help.

httpx/content.py
```diff
@@ -0,0 +1,162 @@
...
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        yield b""
+
+    async def aread(self) -> bytes:
```
florimondmanca: Also, do we really want `.aread()` as a name here?

Thinking about `unasync`, this means that the sync version of this would be `def aread(self)` by default (unless `unasync` also supports removing any "a" prefix, but I don't think it does/should).

So maybe this should be `async def read(self) -> bytes` which would map to `def read(self) -> bytes` on the `SyncResponseContent` variant?

Or is that not the `unasync` use case you're thinking of here?

florimondmanca: Thinking about it we may not need `unasync` at all to support sync here, right? Kind of the point of this whole thing I suppose. 😄 We'll just implement `def read()` and `__iter__()`, right?

httpx/content.py
tomchristie: > Does that mean we can async for chunk in content again

Exactly, yes. It is used when handling redrect responses, that need to generate another request with a body. It happens to be true in all cases except if you pass `data=<async iterator>` in which case we're not able to replay/rewind the request.

And yes, a decent docstring there would make sense.

httpx/content.py
tomchristie: I've commited the suggestion, yup. It does *somewhat* depend on exactly what behavior we want to expect here... ie should `files={}` result in a multipart encoded request, or a urlencoded request?

httpx/content.py
tomchristie: > Thinking about it we may not need unasync at all to support sync here, right? Kind of the point of this whole thing I suppose. 😄 We'll just implement def read() and __iter__(), right?

Exactly, yes. `Request` and `Response` *won't* have two different class variants.

httpx/content.py
florimondmanca: Well, I'd say if `files` is given then as a user I'd expect HTTPX to use multipart. The previous behavior (which we still have in the `data is None` branch?) would use urlencoded for `data=..., files={}`, which I'd find surprising.

(But I'm also wondering whether using `files={}` has any practical sense in the first place?)

httpx/content.py
tomchristie: I've updated this to `can_replay` and added a proper docstring.

tomchristie:
Other thoughts...

* I'm not sure about the module naming. `encoders.py` might be nicer than `content.py`. It doesn't really mirror the `decoders` (It's data structure encoding, rather than content-encoding) but it does encapsulate the functionality reasonably.
* It'd probably be nice to bring the multipart implementation into the module, pehaps all encapsulated on the `MultipartRequestContent` class.

None of that is public API tho, so we could treat it incrementally.

florimondmanca:
>  `encoders.py` might be nicer than `content.py`

What about `request_content.py`? Simple and consistent.

> It'd probably be nice to bring the multipart implementation into the module

I haven't looked at our multipart implementation, so I can't say yet, but definitely in favor of treating that as a follow-up. :-)

tomchristie:
> This looks good to me as-is, feel free to resolve the question on the content.py module name.

Gotcha. Not 100% sure yet, but I think that's one that is ok to tackle incrementally. 👍

Issue #587, opened 2019-12-03
Deprecation Policy.


tomchristie:
At this point in time we ought to have a *clear* warning on the docs to pin to a median version. eg. 0.8.x

Using this issue to describe what I'd like our deprecation policy to be once we hit 1.0.

* Upgrading from X to X+1 (1.0 to 2.0) never breaks any public API, but may raise warnings.
* Upgrading from X to X+2 (2.0 to 3.0) warnings esclate into errors.
* Upgrading from X to X+3 (3.0 to 4.0) code is deprecated.

The deprecation policy would apply to all publically documented code.

We'd recommend that users pin to whichever major version they're using.
(Actually pinning to X or X+1 would also be okay.)

We *might* initially want to mark a couple of things as provisional (eg. I very much doubt we'd have the Backend API covered by the policy, or publicly documented. Our exception heirachy might not be fully nailed down.)

florimondmanca:
Interesting!

So I assume « escalate into errors » still means that we track the code path and raise an appropriate exception/error message, right?

What are the pros of having a « pad major » between breaking changes? Have you seen it help adoption/upgrades compared to strict semver?

StephenBrown2:
I was thinking it would follow marking things with [PendingDeprecationWarning](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) and then [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning), but those aren't _errors_.

Could also borrow from the [deprecation](https://pypi.org/project/deprecation/) [library](https://github.com/briancurtin/deprecation/blob/master/deprecation.py) for an easy decorator.

plizonczyk:
Not sure if it's the right place, but 0.8 has brought fairly breaking change with removal of AsyncClient, without any prior warning. CHANGELOG.md hasn't been updated neither. The reason I'm bringing it up here is that maybe it's worth to enforce deprecation warnings before v1.0 (unless it's the next version)?

florimondmanca:
Thanks for raising this @plizonczyk :-)

> CHANGELOG.md hasn't been updated neither.

Ah, correct. @tomchristie Seems only the package was pushed to PyPI, without an extra `Release 0.8` commit + tag?

[...]

PR #639, opened 2019-12-14
Make HTTP/2 optional via 'pip install httpx[http2]'

### LINKED ISSUE #496: Require `pip install httpx[http2]` for HTTP/2 support

florimondmanca:
Fixes #496 

This PR makes HTTP/2 support entirely optional, and requires installing with `pip install httpx[http2]` to use HTTP/2, in particular to pass `http2=True` to a `Client`.

Note that our development setup always installs HTTP/2 dependencies. If we go and `import h2` again in the future in places where it shouldn't be installed, we won't catch the issue locally nor on CI. I think a good option might be to add a "Test with a base install" job to CI, but we'll need to shuffle things around in `requirements.txt` and our install scripts to make that happen. Happy to treat it as a follow-up. :-)


florimondmanca:
This is *technically* a breaking change (although users may not catch it immediately because they'll already have `h2` installed locally, so they'll only see `MissingDependencies` on CI or a fresh install), and I believe there's no way of being backwards-compatible here. The error message is clear enough that I might be OK with releasing this as part of 0.10, but if we stick to discussions in #587 then this will need to wait for 1.0?

httpx/__init__.py
```diff
@@ -58,6 +59,7 @@
     "DecodingError",
     "HTTPError",
     "InvalidURL",
+    "MissingDependencies",
```
tomchristie: I don't think we should introduce exception classes for this sort of case. A `RuntimeError` or something similar would be suitable here.

florimondmanca: Fair enough :-) Dropped this exception class in the latest commit.

florimondmanca:
The build's broken due to a crazy `ImportError` caused by `regex`: https://travis-ci.org/encode/httpx/jobs/625866049. Not able to reproduce locally. 😕 

```console
$ scripts/check
+black --check --diff --target-version=py36 httpx tests
  File "/home/travis/virtualenv/python3.7.1/bin/black", line 6, in <module>
    from black import patched_main
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/black.py", line 15, in <module>
    import regex as re
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/regex.py", line 402, in <module>
    import _regex_core
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/_regex_core.py", line 21, in <module>
    import regex._regex as _regex
ModuleNotFoundError: No module named 'regex._regex'; 'regex' is not a package
```

The `regex` folks are experimenting with wheels for Linux (see https://bitbucket.org/mrabarnett/mrab-regex/issues/343/wheel-for-linux) so let me try tweaking our `install` script for now…

httpx/extras.py
```diff
@@ -0,0 +1,33 @@
+"""
```
tomchristie: I'm wondering about `optionals.py`?

florimondmanca: Sure, I like that name too. :-)

florimondmanca: Sure, I can roll with that. :-)

tomchristie:
I'm holding back on this slightly, because I feel like we ought to weigh up if we're being overly heavyhanded here. "We've got this ace new HTTP/2 feature. Btw, it's only much use on async. And it's not installed by default. Any you need to turn it on explicitly even once it is installed."

Not totally sure either way, but feeling like *maybe* we don't want to jump on it just yet, for 0.10. ?

florimondmanca:
> We've got this ace new HTTP/2 feature. Btw, it's only much use on async. And it's not installed by default. Any you need to turn it on explicitly even once it is installed.

Put like that, it sure starts to sound like we don't even *want* users to use HTTP/2? 🤣

Maybe the `http2` flag is a strong enough barrier-to-usage already.

An alternative experience we might want to offer is to automatically switch `http2` on if we detect HTTP/2 dependencies are there — i.e. have `http2="auto"` be the default? That way users only have to `pip install httpx[http2]`, and they'll have HTTP/2 enabled when possible, right away.

tomchristie:
> An alternative experience we might want to offer is to automatically switch http2 on if we detect HTTP/2 dependencies are there

That's probably a no. It's too painful that HTTP behavior might switch depending on *other* packages being installed. (Eg. A user installs hypercorn or uvicorn and suddenly their `httpx` behavior changes.)

tomchristie:
Closing as per #496

PR #640, opened 2019-12-14
asyncio: Don't wait for the stream to close

### LINKED ISSUE #634: Client.close() can hang. (asyncio only).
I've been doing some async requests with HTTPX for a while now without any issue.

## Summary

However, I've stumbled upon today a case I cannot explain. While doing a request to Microsoft Graph OAuth2 access token endpoint, I correctly receive a response but HTTPX client seems to hang and never return. I didn't find any other API with the same behaviour.

## Reproduction

```py
import asyncio

import httpx


async def main():
        response = await client.post(
            "https://login.microsoftonline.com/common/oauth2/v2.0/token",
            # You can keep those parameters, the issue happens also for 4XX responses
            data={
                "grant_type": "authorization_code",
                "code": "code",
                "redirect_uri": "http://localhost:8000/redirect",
                "client_id": "CLIENT_ID",
                "client_secret": "CLIENT_SECRET",
            },
        )

        print(response.http_version)  # HTTP/1.1
        print(len(response.content))  # 485 (same as in Content-Length header)
        print(response.headers)
    # Never happens
    print("Done")
    return response


asyncio.run(main())
```

<details>
<summary>Response headers</summary>

```
Headers([('cache-control', 'no-cache, no-store'), ('pragma', 'no-cache'), ('content-type', 'application/json; charset=utf-8'), ('expires', '-1'), ('strict-transport-security', 'max-age=31536000; includeSubDomains'), ('x-content-type-options', 'nosniff'), ('x-ms-request-id', 'dd57f328-14dd-4a99-8ed4-c81566f10800'), ('x-ms-ests-server', '2.1.9707.19 - AMS2 ProdSlices'), ('p3p', 'CP="DSP CUR OTPi IND OTRi ONL FIN"'), ('set-cookie', 'fpc=AhPq7wxRjzlCr42-XJpjplo; expires=Sun, 12-Jan-2020 06:43:04 GMT; path=/; secure; HttpOnly; SameSite=None'), ('set-cookie', 'x-ms-gateway-slice=prod; path=/; SameSite=None; secure; HttpOnly'), ('set-cookie', 'stsservicecookie=ests; path=/; SameSite=None; secure; HttpOnly'), ('date', 'Fri, 13 Dec 2019 06:43:04 GMT'), ('content-length', '485')])
```

</details>

<details>
<summary>After waiting a while, a `ConnectionResetError` is raised. Here is the stacktrace</summary>

```
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/client.py", line 884, in __aexit__
    await self.close()
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/client.py", line 873, in close
    await self.dispatch.close()
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/dispatch/connection_pool.py", line 212, in close
    await connection.close()
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/dispatch/connection.py", line 171, in close
    await self.open_connection.close()
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/dispatch/http11.py", line 72, in close
    await self.socket.close()
  File "/Users/fvoron/.local/share/virtualenvs/httpx-oauth-5xgbpDkq/src/httpx/httpx/concurrency/asyncio.py", line 167, in close
    await self.stream_writer.wait_closed()
  File "/Users/fvoron/.pyenv/versions/3.7.5/lib/python3.7/asyncio/streams.py", line 323, in wait_closed
    await self._protocol._closed
  File "/Users/fvoron/.pyenv/versions/3.7.5/lib/python3.7/asyncio/selector_events.py", line 804, in _read_ready__data_received
    data = self._sock.recv(self.max_size)
ConnectionResetError: [Errno 54] Connection reset by peer
```

</details>

## Versions

* Python 3.7
* HTTPX 0.9.3

florimondmanca:
Fixes #634 

Basically reverts #494.

Opening this as a POC - not sure this is the best way to fix things, or that this won't break other stuff.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -1,7 +1,6 @@
 import asyncio
 import functools
 import ssl
-import sys
 import typing
 
 from ..config import Timeout
@@ -163,8 +162,6 @@ class SocketStream(BaseSocketStream):
 
     async def close(self) -> None:
         self.stream_writer.close()
-        if sys.version_info >= (3, 7):
-            await self.stream_writer.wait_closed()
 
 
 class PoolSemaphore(BasePoolSemaphore):
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -161,6 +161,13 @@ class SocketStream(BaseSocketStream):
         return self.stream_reader.at_eof()
 
     async def close(self) -> None:
+        # NOTE: StreamWriter instances expose a '.wait_closed()' coroutine function,
+        # but using it has caused compatibility issues with certain sites in
+        # the past (see https://github.com/encode/httpx/issues/634), which is
+        # why we don't call it here.
+        # This is fine, though, because '.close()' schedules the actual closing of the
+        # stream, meaning that at best it will happen during the next event loop
+        # iteration, and at worst asyncio will take care of it on program exit.
         self.stream_writer.close()
 
 


florimondmanca:
Note that #494 was originally added when we had still had #369 in. But in the meantime, the `asyncio.Stream` interface which #369 aimed at using to prevent `DeprecationWarning`s was removed from Python 3.8.

So in effect most discussions in #494 about awaiting `stream_writer.close()` or not are outdated.

This is why I *think* the best way to go for us is to do as we did before #369 et. al.: just `stream_writer.close()`, without waiting on the stream to actually close.

(What this does in schedule the closing of the stream in the background, i.e. via a future, and I guess that future will be closed by asyncio on program exit anyway, so we should be safe.)

gdhameeja:
Just a question.
Should we be removing call to `stream_closed()`? The docs suggest using `stream_closed` along with `close()`
https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.close

PR #641, opened 2019-12-16
Drop attrs version pin hot fix

florimondmanca:
Originally added via #566, but the issue now seems to be solved on the `trio` side - need to confirm via CI.


florimondmanca:
Ah, well, seems like the issue isn't fixed yet — I was not seeing it as it works on 3.8 (which I use), but still fails on 3.6 and 3.7. Closing!

PR #644, opened 2019-12-17
No I/O auth

tomchristie:
Switching our auth API around so that implementations do not perform I/O themselves. (Eg. ensure that DigestAuth can be used either with a sync client or an async client.)

Starts by adding a 'send_handling_auth' method, and an empty Auth class.


tomchristie:
Since #645 is in, I think this is ready for re-review.

florimondmanca:
I don't think this includes breaking changes, does it? The [docs](https://www.encode.io/httpx/quickstart/#authentication) only advertise `(username, password)` and `httpx.DigestAuth`, so we're probably safe releasing this with 0.10.

tomchristie:
> I don't think this includes breaking changes, does it? 

It doesn't nope. We're good for this to go in now. 👍

PR #647, opened 2019-12-18
Fix-453: Create new connections when stream ids expire

gdhameeja:
Fixes #453 
Creating new connections when stream ids expire by catching `NoAvailableStreamIDError` in `send` method of `HTTP2Connection`.

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -3,6 +3,7 @@ import typing
 import h2.connection
 import h2.events
 from h2.config import H2Configuration
+from h2.exceptions import NoAvailableStreamIDError
 from h2.settings import SettingCodes, Settings
 
 from ..concurrency.base import (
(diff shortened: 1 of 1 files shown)


httpx/dispatch/http2.py
```diff
@@ -55,16 +56,23 @@ def init_complete(self) -> BaseEvent:
...
+                self.init_complete.set()
+            else:
+                # All other streams need to wait until the
+                # connection is established.
+                await self.init_complete.wait()
+                stream_id = self.state.get_next_available_stream_id()
+        except NoAvailableStreamIDError:
+            self.state = h2.connection.H2Connection(config=self.CONFIG)
+            self.init_stated = False
+            self.send(request, timeout)
```
aviramha: It's recursively calling send. Maybe limit it/change behaviour in case NoAvailableStreamIDError is reoccurring?

aviramha: Also, shouldn't you await ?

gdhameeja: > It's recursively calling send. Maybe limit it/change behaviour in case NoAvailableStreamIDError is reoccurring?

`NoAvailableStreamIDError` shouldn't be raised again since we're opening a new connection. I suppose the exception should only be raised when the identifier for stream ids expire, which is 2^31 on a particular connection I guess. If it does get raised, that would be a bug in the hyper lib?



> Also, shouldn't you await ?

yes, I missed that.

aviramha: I'm not very experienced with HTTP2, but I can assume that the exception can also come from the serverside? If so, a 'bad' server could lead to such scenario.

Also I'd add tests to cover the new scenario.

gdhameeja: > I'm not very experienced with HTTP2, but I can assume that the exception can also come from the serverside? If so, a 'bad' server could lead to such scenario.
> 
> Also I'd add tests to cover the new scenario.

Yes, so I believe according to the HTTP2 spec, even numbers are reserved for the server to identify it's streams with and odd numbers are for the client to label it's streams with, so it is quite possible that the stream id's expire on the server side, but I'm not sure if `NoAvailableStreamIDError` would be raised in that case. Also, I was hoping maybe we could discuss on how to test this in the review, because I guess we'd need to spin up a `MockHTTP2Server` and configure it to accept only one stream id, so it gets exhausted. But I couldn't find any other tests doing this thing so I was hoping I could get some advice on how to test it here.

aviramha: The test you suggested is blackbox, why not just patch H2 Connection to raise the exception? 
I don't see the a lot of value in doing blackbox, especially when the exception is coming from another library (h2).

httpx/dispatch/http2.py
```diff
@@ -55,16 +56,23 @@ def init_complete(self) -> BaseEvent:
...
+                await self.send_connection_init(timeout)
+                stream_id = self.state.get_next_available_stream_id()
+                self.init_complete.set()
+            else:
+                # All other streams need to wait until the
+                # connection is established.
+                await self.init_complete.wait()
+                stream_id = self.state.get_next_available_stream_id()
+        except NoAvailableStreamIDError:
+            self.state = h2.connection.H2Connection(config=self.CONFIG)
```
tomchristie: We can't do this, since it'll break existing streams.

Getting this right is a little bit complex, and I think we should probably pause on it, pending some more comprehensive work around connection state management along the lines in https://github.com/encode/httpx/pull/624

tomchristie:
@gdhameeja Really appreciate your time on having taken a crack at this.

Given https://github.com/encode/httpx/pull/647#discussion_r359262937 I'm going to close this PR off.

PR #648, opened 2019-12-18
Use Streams API for both requests and responses.

tomchristie:
* `content.py` -> `content_streams.py`
* Fold `multipart.py` into `streams.py`
* Request instances have a `request.stream` property for accessing the body.
* Responses accept a stream argument, rather than a pair of `content: async interator`, `on_close: async callable`.


florimondmanca:
Going to let this settle on my mind while I forget that « stream » is a term we often use to refer to asyncio streams. 😛 I *think* we still even have it in your socket abstraction layer, don’t we?

tomchristie:
True enough.

Also I should be clearer about the motivation behind *why* I've tackled this...

* Request & Response streams get us much closer to a vanilla dispatch interface.
* The sync+async client work means we need sync+async capable response content, in the same way that our request content is now sync+async capable.

httpx/client.py
```diff
@@ -499,10 +499,10 @@ def build_redirect_request(self, request: Request, response: Response) -> Reques
         method = self.redirect_method(request, response)
         url = self.redirect_url(request, response)
         headers = self.redirect_headers(request, url, method)
-        content = self.redirect_content(request, method)
+        stream = self.redirect_content(request, method)
```
florimondmanca: Strange mismatch here between stream/content, isn’t it?

httpx/dispatch/asgi.py
```diff
@@ -77,13 +77,20 @@ def __init__(
         status_code = None
         headers = None
         body_parts = []
-        request_stream = request.stream()
```
florimondmanca: How about just changing this into…?

```python
request_body_chunks = request.stream.__aiter__()
```

httpx/dispatch/proxy_http.py
```diff
@@ -167,8 +167,9 @@ def build_auth_header(self, username: str, password: str) -> str:
                 response=proxy_response,
             )
         else:
-            proxy_response.on_close = None
-            await proxy_response.read()
+            # Hack to ingest the response, without closing it.
+            async for chunk in proxy_response._raw_stream:
+                pass
```
florimondmanca: It's a hack, yes 😕 But it seems there's no public interface on `Response` that allows us to do this. Why is it important to not close the response in the first place? We don't do anything with it afterwards. Would it close the connection?

httpx/dispatch/http2.py
```diff
@@ -209,13 +210,13 @@ def __init__(self, stream_id: int, connection: HTTP2Connection) -> None:
...
-        content = self.body_iter(timeout)
+        stream = ResponseStream(iterator=self.body_iter(timeout), close=self.close)
+
         return Response(
             status_code=status_code,
             http_version="HTTP/2",
             headers=headers,
-            content=content,
-            on_close=self.close,
+            content=stream,
```
florimondmanca: Wouldn't we want this to be `Response(stream=...)`?

httpx/models.py
```diff
@@ -646,11 +644,7 @@ def __repr__(self) -> str:
         """
         Read and return the request content.
         """
-        return await self.content.aread()
-
-    async def stream(self) -> typing.AsyncIterator[bytes]:
-        async for part in self.content:
-            yield part
+        return await self.stream.aread()
```
florimondmanca: Not a direct consequence of this PR, but should this method on the request really be called `.aread()`? As in: will we need a sync `def read(self):` for sync support?

httpx/models.py
```diff
@@ -660,8 +654,7 @@ def __init__(
         *,
         http_version: str = None,
         headers: HeaderTypes = None,
-        content: ResponseContent = None,
-        on_close: typing.Callable = None,
+        content: typing.Union[bytes, Stream] = None,
```
florimondmanca: `HTTP11Connection` and `HTTP2Connection` pass a `Stream` instance, but do we every pass raw `bytes` anywhere else in the code base? I'm thinking maybe this can be scoped down to `stream: Stream = None`.

florimondmanca:
Marking this as `breaking` because of `request.content` -> `request.stream`. We may want to provide a compatibility property as for `Response.stream()`?

httpx/models.py
florimondmanca: See also your own comment in the PR description: "Responses accept a stream argument".

tomchristie: So, right now we're only ever passing `content=<bytes>` in the test cases.

We may want to have seperate `content: bytes = None` and `stream: Stream = None` arguments. Instantiating `Response` instances isn't something that users will typically do, but we probably do want it to be doable, and we probably *don't* want to force users to pass a stream instance for the simple case.

tomchristie: Once we have a sync interface on the `Response` we can have a more graceful implementation here, too, along these lines...

```python
if stream is None:
    self.stream = ByteStream(content)
    self.read()
else:
    self.stream = stream
```

And drop the `_raw_content` special casing that we currently have in a few places.

httpx/models.py
tomchristie: It should yes. It's also not *actually* clear if we should expose this method at all.

httpx/dispatch/proxy_http.py
tomchristie: Actually this point in the code surprised me too. Yes I think we're making sure that we maintain the connection. However I *actually* think that this is highlighting an issue here. I *think* that we probably actually want an explicit "proxy being established state" on the HTTP Connection, that will play into the stuff I was starting to think about on https://github.com/encode/httpx/pull/624. It wouldn't at all surprise me if we have some buggy behavior aroung this, such as later maintaining proxy connections that ought to have become closed.

httpx/dispatch/asgi.py
tomchristie: Thanks!

httpx/models.py
tomchristie: Actually we happen to be passing `content=...` in the ASGI case at the moment, but only for convience. We could easily change that around to passing a concrete stream instance.

tomchristie:
`because of request.content -> request.stream` Right, although the `.content` interface isn't *really* public API at this point (we don't have it doc'ed or any documentation on what type it returns) so I think we don't need to be *too* precious here. (We *might*  choose not to include it in a minor bump I guess, but I think we *probably* could if we thought it neccessary.)

httpx/models.py
florimondmanca: Well, I'm not sure we actually. Do we use it anywhere internally (in the package or in the tests) already?

florimondmanca:
> we don't have it doc'ed or any documentation on what type it returns

Oh okay, yeah, let's have it in 0.10 then! :+1:

httpx/streams.py
```diff
@@ -0,0 +1,297 @@
...
+    """
+
+    def __init__(self, aiterator: typing.AsyncIterator[bytes]) -> None:
+        self.aiterator = aiterator
+
+    def can_replay(self) -> bool:
+        return False
+
+    def get_headers(self) -> typing.Dict[str, str]:
+        return {"Transfer-Encoding": "chunked"}
```
florimondmanca: Is `Transfer-Encoding` actually a valid/actionable request header? The [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding) present it as a response-only header. 🤔 

Even then, we're not making sure that the parts in the `aiterator` are separated with `\r\n`, are we?

(Not directly related to this PR, we probably were doing this before.)

tests/models/test_responses.py
```diff
@@ -190,7 +190,7 @@ def test_response_force_encoding():
 
 @pytest.mark.asyncio
 async def test_streaming_response():
-    response = httpx.Response(200, content=async_streaming_body())
+    response = httpx.Response(200, stream=async_streaming_body())
```
florimondmanca: This *works*, but it isn't actually valid from a type hint point of view, correct? Async generators are missing the `.aread()` method.

(Albeit tedious, I tend to be in favor of type-hinting/type-checking tests. Helps catch this kind of minor inconsistency as well as see how users will actually interact with our code *including type hints*.)

httpx/streams.py
```diff
@@ -0,0 +1,297 @@
...
+
+class Stream:
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        yield b""
+
+    async def aclose(self) -> None:
+        pass
+
+    async def aread(self) -> bytes:
+        return b"".join([part async for part in self])
```
florimondmanca: We never actually use this method on responses — only on requests.

So should this be moved to `RequestStream`?

This would mean `Stream` ends up being equivalent to `AsyncIterator[bytes]` — which sounds good to me — and resolve a few type hint issues (we don't currently catch) in our tests.

tomchristie: Dropped this entirely.

tests/models/test_responses.py
tomchristie: > I tend to be in favor of type-hinting/type-checking tests.

Yeah, I've come to the view that it'd *probably* be a good plan for us, since there's *loads* of stuff that it would help catch.

(I'd need to see how things look in order to take a more informed view on it, but it *feels* like it'd be a good thing to be doing.)

httpx/streams.py
tomchristie: Not sure why they list it as response only, since it's required for streaming unknown-length request bodies too. Wrt. the framing, `h11` has us covered... https://h11.readthedocs.io/en/latest/api.html#message-body-framing-content-length-and-all-that

httpx/content_streams.py
```diff
@@ -0,0 +1,283 @@
...
+def encode(
+    data: RequestData = None,
+    files: RequestFiles = None,
+    json: typing.Any = None,
+    boundary: bytes = None,
+) -> ContentStream:
+    """
+    Handles encoding the given `data`, `files`, and `json`, returning
+    a `ContentStream` implementation which provides a byte iterator onto
+    the content, as well as `.is_rewindable()` and `.get_headers()` interfaces.
```
florimondmanca: Small inconsistency on `.is_rewindable()` here. I think slimming this down to the following is sufficient:

```
    Handles encoding the given `data`, `files`, and `json`, returning
    a `ContentStream` implementation.
```

tests/test_decoders.py
```diff
@@ -137,7 +137,7 @@ def test_decoding_errors(header_value):
         for chunk in data:
             yield chunk
 
-    response = httpx.Response(200, content=iterator())
+    response = httpx.Response(200, stream=iterator())
```
florimondmanca: Strictly speaking these would need to be `stream=AsyncIteratorStream(iterator())`, correct?

I'd be in favor of making that change, since we're otherwise setting false expectations that `stream=` accepts async generators when it actually doesn't (it just so happens to work because responses don't call into `.get_headers()` or `.can_replay()`).

tomchristie: 100% yes! I’d missed these - thanks for the catch!

tomchristie:
Thanks so much for the review! Great catches, there. 👍

PR #651, opened 2019-12-19
Reintroduce sync

tomchristie:
At this point this PR:

* Response classes support both sync+async interfaces.
* ContentStream is unified.

Breaking changes:

* `async def read()` -> `async def aread()`
* `async def close()` -> `async def aclose()`
* Dropped Request.read. Not used in tests, not required by users. `.stream` interface is available, although shouldn't typically be accessed by end users.


florimondmanca:
I assume this includes changes from #648?

Not sure if that’s the planned methodology, but if possible I’d rather we find a way to reintroduce sync incrementally. Reviewing the drop of sync was fine because it was all removals, but this is already a bigger chunk to review 😅. Eg the ContentStream unification can go in earlier than the rest. I can setup the v1 branch soon to get this workflow started, if need be.

tomchristie:
>  if possible I’d rather we find a way to reintroduce sync incrementally.

Absolutely agreed, yup.

I'd suggest that...

* We get #649 in, and issue a 0.9.5 release.
* I've updated #648 based on this work - it's *reasonably* large, but I think it's a sensibly reviewable piece of work. I reckon we could merge that with a view to a 0.10.0 release.

I'd expect the 0.10.0 release to include the following API changes...

* Drop `request.read()` (Really it's internal, only used for tests, and `request.stream` is a sufficient alternative.)
* Change streaming reads on the response to `await response.aread()`. (Only effects users performing conditional reads inside a stream block)
* Change `close()` on the response and client to `.aclose()`. (Generally users don't see/use these anyway, since the `.stream()` interface handles response closes automatically, and manually closing a client is a rare usage.)
* Upgrade the usage of `cert`+`verify` to raise an error (and drop them from the dispatch API.)

Those set of changes would enable us to *then* start adding back in sync support incrementally.

Issue #551, opened 2019-11-25
Write error on HTTP/2: "AssertionError: assert waiter is None or waiter.cancelled()"


PrimozGodec:
Since #527 is now closed I am opening a new issue. We already discussed this error in #527 but we could not find a simple test that would reproduce the issue. 

**Issue** When I run a code appended below I get the following error.

```
Traceback (most recent call last):
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 57, in <module>
    asyncio.get_event_loop().run_until_complete(cl.embedd_batch())
  File "/Users/primoz/miniconda3/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 19, in embedd_batch
    embeddings = await asyncio.gather(*requests)
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 43, in _send_to_server
    emb = await self._send_request(client, im, url)
  File "/Users/primoz/PycharmProjects/orange3-imageanalytics/example.py", line 50, in _send_request
    response = await client.post(url, headers=headers, data=image)
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 492, in post
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 634, in request
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 658, in send
    trust_env=trust_env,
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 273, in _get_response
    return await get_response(request)
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/middleware/redirect.py", line 31, in __call__
    response = await get_response(request)
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/client.py", line 228, in get_response
    request, verify=verify, cert=cert, timeout=timeout
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/dispatch/connection_pool.py", line 126, in send
    raise exc
  File "/Users/primoz/venv/orange/lib/python3.7/site-packages/httpx/dispatch/connection_pool.py", line 121, in send
    request, verify=verify, cert=cert, timeout=timeout
[...]

PR #660, opened 2019-12-20
Rationalize backend Semaphore interface slightly

tomchristie:
Some internal refactoring, to make the usage of backend 'semaphore' and 'event' classes more consistent.

* Switch to `create_semaphore()` for consistency with `create_event()`.
* Stop using `BasePoolSemaphore`, and instead just use a generic `BaseSemaphore`. Pass the `exc_class=PoolTimeout`, so that backend semaphores are *not* specific to any particular use-case. (We don't strictly need to do this, but I think it feels neater?)
* Move implementations into more consistent locations, so that event and semaphore implementations are kept close.

### DIFF
--- a/httpx/concurrency/asyncio.py
+++ b/httpx/concurrency/asyncio.py
@@ -4,8 +4,8 @@ import ssl
 import typing
 
 from ..config import Timeout
-from ..exceptions import ConnectTimeout, PoolTimeout, ReadTimeout, WriteTimeout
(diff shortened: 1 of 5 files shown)


tomchristie:
After this I'm planning on:

* Switching from using an Event in h2 to using a Lock instead. (So backends expose a lock implementation, rather than an event implementation.)
* Move read/write locks *outside* the stream, and onto the HTTP classes instead, with a view to resolving #551.

PR #661, opened 2019-12-20
Add VScode stuff to .gitignore

BigChief45:


tomchristie:
Heya, and thanks. We’ll pass on this one. Ignores that are *specific to your local development environment* ought to live in your own global gitignore config.

I think you can create one like so:

`git config --global core.excludesfile $HOME/.gitignore`

Hope that helps!

PR #664, opened 2019-12-20
Clean up 'backend' fixture

florimondmanca:
https://github.com/encode/httpx/pull/615 removed passing `backend=...` to most of our tests, but we still had a `backend` fixture. It was declared as a test function parameter but not actually used.

This PR replaces this fixture with another one, `async_environment`. It is invoked using `@pytest.mark.usefixtures()` and should be applied when we want to run a test against all async environments.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,7 +38,21 @@ ENVIRONMENT_VARIABLES = {
         pytest.param("trio", marks=pytest.mark.trio),
     ]
 )
-def async_environment(request) -> str:
+def async_environment(request: typing.Any) -> str:
+    """
+    Mark a test function to be run on both asyncio and trio.
+
+    Equivalent to having a pair of tests, each respectively marked with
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -32,6 +32,16 @@
 }
 
 
+@pytest.fixture(
+    params=[
+        pytest.param("asyncio", marks=pytest.mark.asyncio),
+        pytest.param("trio", marks=pytest.mark.trio),
+    ]
+)
+def async_environment(request) -> str:
```
tomchristie: Should we have a small docstring here, noting usage, and that this is equivelent to having a pair of tests, wrapped wth `@pytest.mark.asyncio` and `@pytest.mark.trio`. Or perhaps not neccessary?

florimondmanca: Sounds good, yup

Issue #650, opened 2019-12-19
Type-checking our tests


florimondmanca:
I know this is not a standard thing to do across Encode projects, but I've been wondering if it would be worth starting to type-hint our tests.

I've seen at least two instances of this recently:

- In HTTPX: https://github.com/encode/httpx/pull/648#discussion_r359862603
- In Starlette: https://github.com/encode/starlette/issues/722

My rationale is based on two aspects:

- It improves our upfront knowledge about how users will actually use HTTPX — currently their usage of type hints in the wild is not reflected anywhere.
- It helps us catch type hint inconsistencies we wouldn't see in the core package.

The main counter-argument, I suppose, is that type hinting tests is tedious. I think that's fair, but I believe the two pro's above make it compelling.

Thoughts?

florimondmanca:
Alternatively, we could tweak the `mypy` config:

```ini
[mypy]
files = httpx, tests
ignore_missing_imports = True
disallow_untyped_defs = False
check_untyped_defs = True

[mypy-httpx]
disallow_untyped_defs = True
check_untyped_defs = False
```

Running this locally raises 141 errors, but mostly on internal APIs, so, hmm.

BigChief45:
I think this is a great idea, would like to work on this. Since I am pretty new to httpx, adding these type hints to the tests will give me a better picture of how httpx works under the hood.

florimondmanca:
I think this is still a good idea, though it's a bit of busy work to upgrade our existing test code.

Let's perhaps just remind ourselves to add types whenever we add/touch existing tests?

Pradhvan:
@florimondmanca I see a lot of merged PR. I guess you should close this ? 😄

florimondmanca:
I think there’s one small last step which would be to add tests/ to the mypy arguments in the check script. :-) Once this is enforced then yup, we can close this off.

Pradhvan:
I am currently working on #871 and when that is done. I would be happy to pick up what's remaining in this PR so we can close this one too. 😄

PR #665, opened 2019-12-20
Type-check test_auth.py

florimondmanca:
Refs #650 

First pass at type-checking our tests, starting (arbitrarily) with the `test_auth.py` module.


tests/client/test_auth.py
```diff
@@ -179,34 +180,37 @@ def auth(request):
...
-async def test_auth_hidden_header():
+async def test_auth_hidden_header() -> None:
     url = "https://example.org/"
     auth = ("example-username", "example-password")
 
     client = Client(dispatch=MockDispatch())
     response = await client.get(url, auth=auth)
 
-    assert "'authorization': '[secure]'" in str(response.request.headers)
+    request = typing.cast(Request, response.request)
```
florimondmanca: This might prove to be bad UX for our users: `response.request` is *almost always* set, *except* for some tests where *we* create a `Response` without a request. May look into those situations and then turn on `request` as non-optional on `Response`.

tomchristie: Good call. Let's defer this until #666 is in, and we can drop the cast.

PR #673, opened 2019-12-21
Fix out-of-date streaming methods on Response API docs

florimondmanca:
The `Response` API reference still showcased the `stream_*` methods instead of `aiter_*` - refs #610.


florimondmanca:
Will merge this as trivially acceptable. :-)

Issue #667, opened 2019-12-21
Version 0.10.0


tomchristie:
Let's get squared away what we need for Version 0.10.0.

The key here is we need to make sure we've sorted out any API that'd block us from reintroducing the sync API, so...

* [x] Let's drop `.read()` from `Request` - It's not documented anywhere, not *actually* required by users, and the content is available on `request.stream` if absolutely needed. #679
* [x] We need `Response.read()` and `Response.close()` to be `Response.aread()` and `Response.aclose()`. The only point where this is relevant is users performing conditional reads inside a `with httpx.stream(method, url) as response` block. #674
* [x] We ought to have `Client.close()` become `Client.aclose()` for consistency. #675
* [x] Good point to get #617 in.
* [x] We'll need to change `response.next()`. We could *either* underspecify the return type, and allow it to be either sync or async depending on the context, *or* we use `response.anext()` and `response.next()`. #676
* [ ] ~Good point to address~ #656.
* [x] *Potentially* we could introduce an `httpx.AsyncClient` synonm for `httpx.Client`, and advise our users to switch towards that usage, so that there's no breaking changes for them once we fully reintroduce a sync API all the way through. #680

florimondmanca:
> *or* we use `response.anext()` and `response.next()`

Keen on this. :+1:

florimondmanca:
Regarding `.aclose()` — I think we probably want to switch to that naming for all our other `.close()` operations too, if only for consistency.

In particular, this would be consistent with the overall convention in the async ecosystem — e.g. [`AsyncExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack.aclose), `async_generator`'s [`aclosing`](https://github.com/python-trio/async_generator/blob/9f0e08c762dc2b9e18a38e8dd714790a319ed836/async_generator/_util.py#L6) (async equivalent of [`contextlib.closing`](https://docs.python.org/3/library/contextlib.html#contextlib.closing)), etc.

tomchristie:
Pragmatically I think we probably just want to make the response/request changes and document them. If we don’t do that we’ll be blocked from reintroducing the sync API. They only affect anyone using conditional reads in streaming responses, and we’ve got clear messaging about pinning versions.

florimondmanca:
> If we don’t do that we’ll be blocked from reintroducing the sync API.

Yes, that's a good point. Same for `Client`, then, yes?

tomchristie:
Probably yes.

[...]

PR #676, opened 2019-12-21
Rename 'next' to 'anext' on Response

florimondmanca:
Refs #667

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -476,12 +476,7 @@ class Client:
 
             if not allow_redirects:
                 response.call_next = functools.partial(
-                    # NOTE: not using 'self.send_handling_redirects' here, because
-                    # 'call_next' must reference a function (instead
-                    # of a method). This ensures that 'inspect.iscoroutinefunction()'
-                    # checks behave properly.
-                    Client.send_handling_redirects,
-                    self,
+                    self.send_handling_redirects,
                     request=request,
                     auth=auth,
                     verify=verify,
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -476,7 +476,12 @@ class Client:
 
             if not allow_redirects:
                 response.call_next = functools.partial(
-                    self.send_handling_redirects,
+                    # NOTE: not using 'self.send_handling_redirects' here, because
+                    # 'call_next' must reference a function (instead
+                    # of a method). This ensures that 'inspect.iscoroutinefunction()'
+                    # checks behave properly.
+                    Client.send_handling_redirects,
+                    self,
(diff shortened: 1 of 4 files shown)


httpx/client.py
```diff
@@ -476,7 +476,12 @@ def setup_auth(
 
             if not allow_redirects:
                 response.call_next = functools.partial(
-                    self.send_handling_redirects,
+                    # NOTE: not using 'self.send_handling_redirects' here, because
+                    # 'call_next' must reference a function (instead
+                    # of a method). This ensures that 'inspect.iscoroutinefunction()'
+                    # checks behave properly.
+                    Client.send_handling_redirects,
+                    self,
```
florimondmanca: Yes, that's surprising, and essentially boils down to…

```python
import inspect

class Foo:
    async def bar(self):
        pass

foo = Foo()
print(inspect.iscoroutinefunction(foo.bar))  # False
```

`functools.partial` is not in fault here, since `inspect.is*function` checks unwrap any `partial` down to the underlying function/method.

The "issue" is that `inspect.is*function` has a check on `inspect.isfunction()` (which, er, seems logicial because they both have "function" in there). But in this case `foo.bar` refers to a *method*, [which is different from a function](https://stackoverflow.com/questions/17019949/why-is-there-a-difference-between-inspect-ismethod-and-inspect-isfunction-from-p) in Python 3.

To get a function, we must reach to the unbound method, so:

```python
print(inspect.iscoroutinefunction(Foo.bar))  # True
```

httpx/models.py
```diff
@@ -922,13 +923,14 @@ def raw(self):  # type: ignore
...
 
-    async def next(self) -> "Response":
+    async def anext(self) -> "Response":
         """
         Get the next response from a redirect response.
         """
         if not self.is_redirect:
             raise NotRedirectResponse()
         assert self.call_next is not None
+        assert inspect.iscoroutinefunction(self.call_next)
```
tomchristie: I wouldn’t bother making this functional change, or the associated `partial` change. Let’s just do a straight name change and leave it at that. We can look at any extra type smart when we introduce `.next()`

httpx/client.py
florimondmanca: Strangely, this workaround doesn't work on 3.6 and 3.7 — CI is failing for those versions. 😕

httpx/models.py
florimondmanca: Yeah, sure, that'd fix it *for now* — but we might bump into this again when looking at `.next()`. Anyway, will remove this check for now so we can move on.

PR #677, opened 2019-12-21
Type-check test_cookies.py

florimondmanca:
More tests type-checking, refs #650.

### DIFF
--- a/tests/client/test_cookies.py
+++ b/tests/client/test_cookies.py
@@ -22,10 +22,12 @@ class MockDispatch(Dispatcher):
         elif request.url.path.startswith("/set_cookie"):
             headers = {"set-cookie": "example-name=example-value"}
             return Response(200, headers=headers, request=request)
+        else:
+            raise NotImplementedError  # pragma: no cover
 
 
 @pytest.mark.asyncio
-async def test_set_cookie():
+async def test_set_cookie() -> None:
     """
     Send a request including a cookie.
     """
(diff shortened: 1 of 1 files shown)


tests/client/test_cookies.py
```diff
@@ -99,21 +101,21 @@ class MockDispatch(Dispatcher):
...
         comment_url=None,
-        rest={"HttpOnly": None},
+        rest={"HttpOnly": ""},
         rfc2109=False,
     )
     cookies.set_cookie(cookie)
 
     client = Client(dispatch=MockDispatch())
-    client.cookies = cookies
+    client.cookies = cookies  # type: ignore
```
florimondmanca: mypy doesn't yet support setters that accept arguments of a different type than what is returned by the getter. The official position is to `# type: ignore` for now. See: https://github.com/python/mypy/issues/3004

tomchristie: Hrm. Well, it *is* a strange usage. Got me thinking that maybe we just shouldn’t be doing that in the first place...

Instead, we could enforce this style:

`client.cookies = httpx.Cookies(...)`

That’s possibly less surprising than setting the cookies to a dict, but having it return a Cookies instance.

We would probably want to hard check the type.

Obvs same would apply to the other setters too.

Thoughts?

florimondmanca: Yup, sure, that would sound good to me. :-)

florimondmanca: -> https://github.com/encode/httpx/pull/678

PR #678, opened 2019-12-21
Make client.cookies setter expect a Cookies instance

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/677#discussion_r360668301

@tomchristie IMO the only thing that *does* look odd (at least from a typing perspective) is the setter magic, so this PR focuses on that. In particular, not sure we really want to get rid of `CookieTypes` and change the `cookies` parameters to prevent using the convenience `dict`/`CookieJar` types.

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -189,8 +189,8 @@ class Client:
         return self._cookies
 
     @cookies.setter
-    def cookies(self, cookies: CookieTypes) -> None:
-        self._cookies = Cookies(cookies)
+    def cookies(self, cookies: Cookies) -> None:
+        self._cookies = cookies
 
     @property
     def params(self) -> QueryParams:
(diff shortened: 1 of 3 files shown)


httpx/client.py
```diff
@@ -189,8 +189,8 @@ def cookies(self) -> Cookies:
         return self._cookies
 
     @cookies.setter
-    def cookies(self, cookies: CookieTypes) -> None:
-        self._cookies = Cookies(cookies)
+    def cookies(self, cookies: Cookies) -> None:
+        self._cookies = cookies
```
florimondmanca: Would we want to add an `assert` or any kind of runtime type check here? I'm not sure it's really needed. Also this change makes the `cookies` property equivalent to a simple variable, so we might as well switch it to just that — a plain and public `.cookies` variable?

tomchristie:
I think we should probably *only* do this if we're also doing the same for `Headers` and `Params`.
It's very possibly a good thing to do in all cases, but perhaps let's defer that decision for now, and allow the `type: ignore` on https://github.com/encode/httpx/pull/677/files#diff-2aeaac4c23c1f6bf49dfbd99885a3aa7L108 for the moment.

florimondmanca:
Closing this one for now to be addressed later with a pass on `Headers` and `Params`, then.

PR #680, opened 2019-12-23
Rename Client to AsyncClient (with compat synonym)

florimondmanca:
Refs #667 

- Rename `Client` to `AsyncClient`
- Update references to `Client` in core and tests.
- Change docs to encourage using `AsyncClient` from 0.10.0 and on.
- Add `Client` synonym for backwards compat with `0.9.*`.

Regex search & replace was a good mate on this one. 💯

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -10,7 +10,7 @@ all outgoing requests.
     A Client instance is equivalent to a Session instance in `requests`.
 
 !!! note
-    Starting from `httpx==0.10.0`, the default and recommended Client class is `AsyncClient`.
+    Starting from `httpx==0.10.0`, the default and recommended Client class is `AsyncClient`. This should help prevent breaking changes once sync support is reintroduced.
 
     A `Client` synonym remains for compatibility with `httpx==0.9.*` releases, but you are encouraged to migrate to `AsyncClient` as soon as possible.
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -9,12 +9,17 @@ all outgoing requests.
 !!! hint
     A Client instance is equivalent to a Session instance in `requests`.
 
+!!! note
+    Starting from `httpx==0.10.0`, the default and recommended Client class is `AsyncClient`.
+
+    A `Client` synonym remains for compatibility with `httpx==0.9.*` releases, but you are encouraged to migrate to `AsyncClient` as soon as possible.
+
 ### Usage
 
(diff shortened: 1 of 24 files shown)


florimondmanca:
Originally, `Client` was referenced in the docs because sync was the default, and `Client` was really `SyncClient`.

IMO we should end up with `AsyncClient`/`SyncClient` (instead of `AsyncClient`/`Client` as we had in 0.7.* and before), so demonstrating usage with one explicit type of client is a good idea.

docs/advanced.md
```diff
@@ -9,12 +9,17 @@ all outgoing requests.
 !!! hint
     A Client instance is equivalent to a Session instance in `requests`.
 
+!!! note
+    Starting from `httpx==0.10.0`, the default and recommended Client class is `AsyncClient`.
```
tomchristie: Should we note that it’s because we’re hoping to reintroduce sync+async support shortly, or something along those lines?

docs/advanced.md
```diff
@@ -9,12 +9,17 @@ all outgoing requests.
 !!! hint
     A Client instance is equivalent to a Session instance in `requests`.
```
StephenBrown2: Is this also to be `AsyncClient`?

docs/compatibility.md
```diff
@@ -52,7 +52,7 @@ We don't support `response.is_ok` since the naming is ambiguous there, and might
 
 ## Client instances
 
-The HTTPX equivalent of `requests.Session` is `httpx.Client`.
+The HTTPX equivalent of `requests.Session` is `httpx.AsyncClient`.
```
StephenBrown2: Since requests is sync-only... will this be changed _back_ once sync is reintroduced?

tomchristie: Yes, pending some final decisions on exact naming for the sync vs async variants.

docs/advanced.md
florimondmanca: Sure, added a note about that. :-)

docs/advanced.md
florimondmanca: Well, here (and elsewhere on this page) "Client" refers to the general idea of a class for making requests — sync or async.

It's a bit blurry at the moment since we we don't have anything else than `AsyncClient` yet, but I'd say for now it doesn't look *too* unnatural?

florimondmanca:
@tomchristie Updated the PR!

PR #684, opened 2019-12-25
`DigestAuth` will raise an exception if non replayable request is passed #670

### LINKED ISSUE #670: DigestAuth should raise a clear error if the request cannot replay.
Our DigestAuth implementation cannot work with non-replayable requests.
We ought to raise a nice clear error if `request.stream.is_replayable` is not True.

kousikmitra:
Fixes #670

1. `DigestAuth` will now raise an `RedirectBodyUnavailable` exception if a non-replayable request is passed to its `__call__` method.

Feedback will be appreciated. :)

### DIFF
--- a/httpx/auth.py
+++ b/httpx/auth.py
@@ -6,7 +6,7 @@ import typing
 from base64 import b64encode
 from urllib.request import parse_http_list
 
-from .exceptions import ProtocolError
+from .exceptions import ProtocolError, RedirectBodyUnavailable
 from .models import Request, Response
 from .utils import to_bytes, to_str, unquote
 
@@ -104,6 +104,8 @@ class DigestAuth(Auth):
         self.password = to_bytes(password)
 
     def __call__(self, request: Request) -> AuthFlow:
+        if not request.stream.can_replay():
+            raise RedirectBodyUnavailable("Non-replayable request is passed")
         response = yield request
 
         if response.status_code != 401 or "www-authenticate" not in response.headers:


httpx/auth.py
```diff
@@ -104,6 +104,8 @@ def __init__(
         self.password = to_bytes(password)
 
     def __call__(self, request: Request) -> AuthFlow:
```
gdhameeja: Wasn't the issue specific to `DigestAuth` ?

kousikmitra: I guess it is. Can you be more specific what the issue?

kousikmitra:
@florimondmanca I will rename it and add the tests. Thanks for the feedback.

httpx/auth.py
gdhameeja: Yeah, my bad, I thought the changes were in Base class

kousikmitra:
> 
> 
> Thanks for this @kousikmitra! Edited the description to reference the PR.
> 
> Should we look at renaming the RedirectBodyUnavailable exception into something more generic, eg RequestBodyUnavailable? This could apply to both the redirect and digest auth cases.
> 
> Also probably worth adding a test for this as well? :-)

Should I just rename the exception or add new exception class `RequestBodyUnavailable` and derive that class from RedirectBodyUnavailable` for compatibility?

kousikmitra:
I deleted my fork by mistake so creating a new pull request and closing this one.

PR #685, opened 2019-12-26
DigestAuth will raise exception if non-replayble request is passed #670

### LINKED ISSUE #670: DigestAuth should raise a clear error if the request cannot replay.

kousikmitra:
Reference to PR https://github.com/encode/httpx/pull/684

Fixes #670

`DigestAuth` will now raise `RequestBodyUnavailable` exception if a non-replayable request is passed to its `__call__` method.

Feedback will be appreciated. :)


httpx/exceptions.py
```diff
@@ -93,6 +93,12 @@ class RedirectBodyUnavailable(RedirectError):
     """
 
 
+class RequestBodyUnavailable(RedirectBodyUnavailable):
+    """
+    Got a request body, but the request body is no longer available.
+    """
```
florimondmanca: Hmm… So maybe we can simply re-purpose the `RedirectBodyUnavailable` exception above, by moving it to the `StreamError` tree… Something like:

```python
# Stream exceptions...

# ...
class RequestBodyUnavailable(StreamError):
    """
    Had to send the request again but the request body was streaming, and is
    no longer available.
    """
```

We can (should?) address this change separately, and come back to this afterwards. I'll issue a quick PR right away to see what this would look like. :-)

florimondmanca:
@kousikmitra Issued https://github.com/encode/httpx/pull/690. Feel free to have a look!

httpx/exceptions.py
kousikmitra: Yeah it looks  better. once the PR gets merged I will make the changes here.

florimondmanca: Cool! Should merge soon.

On a separate note — ideally we'd have some testing for this new behavior. Could we look at adding one along with the other `DigestAuth` tests?

florimondmanca: @kousikmitra #690 has landed in master — you can update this PR when ready. 😄

kousikmitra: @florimondmanca Merged The PR and added a test function to test the behavior. 😃

PR #686, opened 2019-12-26
Save history if authentication flow makes more than one request

### LINKED ISSUE #669: Authentication flows should preserve history.
If the authentication flow makes more that one request, then any prior requests should be visible in `response.history`.

gdhameeja:
Fixes #669 
Saves the response to `history` if authentication flow makes more than one request. This history is assigned to `response.history`

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -451,7 +451,7 @@ class AsyncClient:
                 raise RedirectLoop()
 
             response = await self.send_handling_auth(
-                request, auth=auth, timeout=timeout,
+                request, history, auth=auth, timeout=timeout, verify=verify, cert=cert
             )
             response.history = list(history)
 
@@ -566,12 +566,21 @@ class AsyncClient:
         return request.stream
 
     async def send_handling_auth(
-        self, request: Request, auth: Auth, timeout: Timeout,
+        self,
+        request: Request,
+        history: typing.List[Response],
+        auth: Auth,
+        timeout: Timeout,
+        verify: VerifyTypes = None,
+        cert: CertTypes = None,
     ) -> Response:
         auth_flow = auth(request)
         request = next(auth_flow)
+        requested_more_than_once = False
         while True:
             response = await self.send_single_request(request, timeout)
+            if requested_more_than_once:
+                history.append(response)
             try:
                 next_request = auth_flow.send(response)
             except StopIteration:
@@ -582,6 +591,7 @@ class AsyncClient:
             else:
                 request = next_request
                 await response.aclose()
+            requested_more_than_once = True
 
     async def send_single_request(
         self, request: Request, timeout: Timeout,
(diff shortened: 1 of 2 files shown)


gdhameeja:
There is an ugly counter variable that we use to keep track of the number of requests made in `send_handling_auth`. Maybe this can be done in a better way, that I could not think of currently. Please suggest.

httpx/client.py
```diff
@@ -582,15 +582,20 @@ def redirect_stream(
...
         request: Request,
+        history: typing.List[Response],
         auth: Auth,
         timeout: Timeout,
         verify: VerifyTypes = None,
         cert: CertTypes = None,
     ) -> Response:
         auth_flow = auth(request)
         request = next(auth_flow)
+        counter = -1
```
florimondmanca: I understand the intent, but I'd agree it's not immediately obvious. One way to make things clearer could be to use a verbosely-named boolean flag:

```python
made_more_than_one_request = False

while True:
    response = ...

    if made_more_than_one_request:
        history.append(response)

    try:
        # ...
    else:
        request = next_request
        made_more_than_one_request = True
```

gdhameeja: Fixed. Yeah, this is better. Thanks.

httpx/client.py
```diff
@@ -601,6 +605,7 @@ def redirect_stream(
             else:
                 request = next_request
                 await response.close()
+            requested_more_than_once = True
```
florimondmanca: Nit: should we keep this in the `else` block, since it represents what happens when we're asked to make a new request?

```suggestion
                requested_more_than_once = True
```

gdhameeja:
> Great, thanks! Just a final nit.

Fixed.

j178:
From codes in these lines: https://github.com/encode/httpx/blob/master/httpx/client.py#L456-L463, I think the history desigin is that every response in the history also got a history at its own perspective, but those responses created during handing auth don't have its own history, I think this is a little inconsistent.

florimondmanca:
@j178 As documented in [Redirection and History](https://www.encode.io/httpx/quickstart/#redirection-and-history) the `history` property on responses is only aimed at inspecting the redirection history, so we shouldn't include requests made in other situations (such as when authenticating, in this case).

gdhameeja:
@florimondmanca I added a test. Can you please review this?
Thanks.

florimondmanca:
@gdhameeja Thanks! After thinking quite deeply about it, I agree with @j178 that there's a slight inconsistency compared to redirects in that the `response.history` for intermediate responses isn't populated.

Since I ended up with a complete implementation and tests, and before I saw your last comment :-), I pushed the result to #718. Would you like to give it a look? Then if that sounds good to you we can close this PR and roll with #718.

gdhameeja:
Closing this as there is another PR for this.
https://github.com/encode/httpx/pull/718

PR #687, opened 2019-12-27
Changed behaviour of elapsed on response

### LINKED ISSUE #655: Make `.elapsed` reflect the entire response time.
I think we should probably switch the behavior of `.elapsed` so that rather than reflecting the "time for the response to start", it is instead only available once `.close()` has been called, and is calculated from the start of the request to the point that the response is closed.

That way it'll actually properly reflect the time taken for the entire request/response cycle.

gdhameeja:
Fixes #655 
Makes elapsed time only available after response is closed

Took the django-rest-framework approach here (as making .data available only when `is_valid` is called)
Used `ElapsedTimer` on the request to start the timer and assigned the `elapsed` from request timer to response.

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -34,6 +34,7 @@ from .exceptions import (
 )
 from .status_codes import StatusCode
 from .utils import (
+    ElapsedTimer,
     flatten_queryparams,
     guess_json_utf,
     is_known_encoding,
@@ -607,6 +608,7 @@ class Request:
         else:
             self.stream = encode(data, files, json)
 
+        self.timer = ElapsedTimer()
         self.prepare()
 
     def prepare(self) -> None:
@@ -667,14 +669,12 @@ class Response:
         stream: ContentStream = None,
         content: bytes = None,
         history: typing.List["Response"] = None,
-        elapsed: datetime.timedelta = None,
     ):
         self.status_code = status_code
         self.http_version = http_version
         self.headers = Headers(headers)
 
         self.request = request
-        self.elapsed = datetime.timedelta(0) if elapsed is None else elapsed
         self.call_next: typing.Optional[typing.Callable] = None
 
         self.history = [] if history is None else list(history)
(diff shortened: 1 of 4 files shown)


httpx/models.py
```diff
@@ -688,6 +688,20 @@ def __init__(
...
+        """
+        Returns the time taken for the complete request/response
+        cycle to complete.
+        """
+        if not hasattr(self, "_elapsed"):
+            msg = (
+                "`elapsed` may only be called after calling `close()`"
+                " on the response"
+            )
+            raise AssertionError(msg)
```
florimondmanca: Perhaps this would be better served by a `RuntimeError`? Also some stylistic and copy suggestions:

```python
raise RuntimeError(
    "'.elapsed' may only be accessed after the response "
    "has been read or closed."
)
```

gdhameeja: Fixed.

gdhameeja:
> Just realised there's a description of this property in the Response Developer Interface docs. Should we update it here as well?

I can update it. Should I?

florimondmanca:
> Should I?

Would probably be best to have it with this PR, yes. :-)

httpx/models.py
```diff
@@ -940,6 +953,7 @@ def raw(self):  # type: ignore
             self.is_closed = True
             if hasattr(self, "_raw_stream"):
                 await self._raw_stream.aclose()
+        self._elapsed = self.request.timer.elapsed
```
tomchristie: Let's drop this inside the `if not self.is_closed` block.

httpx/models.py
```diff
@@ -667,14 +669,12 @@ def __init__(
...
     ):
         self.status_code = status_code
         self.http_version = http_version
         self.headers = Headers(headers)
 
         self.request = request
-        self.elapsed = datetime.timedelta(0) if elapsed is None else elapsed
         self.call_next: typing.Optional[typing.Callable] = None
 
         self.history = [] if history is None else list(history)
```
tomchristie: We probably also want to add `self._elapsed = datetime.timedelta(0)` inside the `if stream is None` block.

tomchristie:
Great work on this! Pulled this into #692, with a few additional tweaks.

PR #688, opened 2019-12-27
Drop `Origin` from public API

### LINKED ISSUE #656: Drop `Origin` from public API.
We don't really need to expose `Origin` publically. For the sake of keeping the public API space slim we should...

* Drop the `url.origin` property.
* Drop `Origin` from the top-level export.
* Move `Origin` into `utils.py`.
* Use `origin = Origin(url)` in our internal usage, rather than `url.origin`.

Minor breaking change here, although I'm not sure we *document* origin anywhere or not.

BigChief45:
- [x] Drop the `url.origin` property.
- [x] Drop `Origin` from the top-level export.
- [ ] ~Move `Origin` into `utils.py`.~
- [x] Use `origin = Origin(url)` in our internal usage, rather than `url.origin`.

Closes #656

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -26,7 +26,7 @@ from .exceptions import (
     TooManyRedirects,
     WriteTimeout,
 )
-from .models import URL, Cookies, Headers, Origin, QueryParams, Request, Response
+from .models import URL, Cookies, Headers, QueryParams, Request, Response
 from .status_codes import StatusCode, codes
 
 __all__ = [
@@ -75,7 +75,6 @@ __all__ = [
     "StatusCode",
     "Cookies",
     "Headers",
-    "Origin",
     "QueryParams",
     "Request",
     "TimeoutException",
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -37,6 +37,7 @@ from .models import (
     CookieTypes,
     Headers,
     HeaderTypes,
+    Origin,
     ProxiesTypes,
     QueryParams,
     QueryParamTypes,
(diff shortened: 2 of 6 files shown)


tomchristie:
The awkwardness here is the cyclical reference between `utils.py` and `models.py`.
I bumped into the same thing having a look at this, and I'm not sure what the most graceful option for us is. We *might* decide that actually we just want to drop this issue, and that `Origin` is actually perfectly sensible in the public API anyways?

florimondmanca:
Another option is to leave `Origin` in `models.py`, and only exclude it from the exports in `__init__.py`?

tomchristie:
Indeed yup, makes sense too. Possibly feels a bit funny that it's the only model in there that's not public API. 🤷‍♂️

BigChief45:
Updated to keep `Origin` in `models.py` but excluding it from top-level exports.

BigChief45:
@florimondmanca Oops, forgot to update that origin reference when fixing conflicts. Should be good now :heavy_check_mark:

florimondmanca:
About whether we should drop this in the first place… Currently `Origin` _is_ a documented piece of public API, i.e. in the Developer Interface we document `.origin` on [`URL`](https://www.encode.io/httpx/api/#url) and [`Origin`](https://www.encode.io/httpx/api/#origin) itself. I don't think we mention it in the usage guides, though.

I don't think users *should* be accessing `request.url.origin`, less so build `Origin` instances themselves, so I think I'm still in favor of dropping, and this PR makes sense to me.

tomchristie:
Yup, I think this is great too, and yes I think those are all good arguments in favour of going ahead with this. I'd *guess* we want to wait until the next proper version bump before we pull this in?

PR #692, opened 2019-12-29
response.elapsed now reflects entire request/response time.

### LINKED ISSUE #655: Make `.elapsed` reflect the entire response time.

tomchristie:
Couple of further tweaks on top of #687, with thanks to @gdhameeja ✨

* `Response(content=...)` is a closed response and should have `response.elapsed == datetime.timedelta(0)`.
* `._elapsed` should only be set when the response is closed. Any further calls to `.close` should have no effect.

Closes #655

### DIFF
--- a/httpx/models.py
+++ b/httpx/models.py
@@ -34,6 +34,7 @@ from .exceptions import (
 )
 from .status_codes import StatusCode
 from .utils import (
+    ElapsedTimer,
     flatten_queryparams,
     guess_json_utf,
     is_known_encoding,
@@ -607,6 +608,7 @@ class Request:
         else:
             self.stream = encode(data, files, json)
 
+        self.timer = ElapsedTimer()
         self.prepare()
 
     def prepare(self) -> None:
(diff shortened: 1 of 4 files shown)


httpx/models.py
```diff
@@ -683,11 +683,25 @@ def __init__(
             self.is_closed = True
             self.is_stream_consumed = True
             self._raw_content = content or b""
+            self._elapsed = datetime.timedelta(0)
```
florimondmanca: Shouldn't this be `request.timer.elapsed`? Otherwise we don't account for the time spent receiving the response headers in the raw-bytes case.

tomchristie: Ah that'd be better yup. As it *happens* we'll always use `stream=...` for actual network responses, it's really only the "mock response instance" case that we use `content=...` for, but yes it makes sense to account for the time properly in that case too.

Issue #514, opened 2019-11-07
Issuing requests in close concurrency can unneccessarily result in multiple HTTP/2 connections.


PrimozGodec:
When I do several requests with the AsyncClient it seems that it opens multiple connections to the server. My code looks like this:

```
async def _send_request(self, client, image, url):
    headers = {
        'Content-Type': 'image/jpeg',
        'Content-Length': str(len(image))
    }
    return await client.post(
        url,
        headers=headers,
        data=image
    )

async with httpx.AsyncClient(timeout=timeout, base_url=self.server_url) as client:
    requests = []
    for image in images:
        requests.append(self._send_request(client, image, url))
    responses = await asyncio.gather(*requests)
```

As far as I understand it should use a connection pooling, store the connection and reuse it. In my case I get the following debug message for every `image`- post request I send with a loop:

`DEBUG:httpx.dispatch.connection:start_connect host='api.garaza.io' port=443 timeout=TimeoutConfig(connect_timeout=None, read_timeout=60, write_timeout=5)`

Am I doing anything wrong or is it a bug?

PrimozGodec:
The issue seems to appear here:

https://github.com/encode/httpx/blob/1a32cf036a825f6eb35395af5388a3b23180a82e/httpx/dispatch/connection_pool.py#L132-L150

Since first `self.pop_connection(origin)` is called then there is `await` call in the line 137 and at the end there is `self.active_connections.add(connection)`. Since there is await in between them first pops for all requests are called (there is no connection in the pool at that time) and the connection is added to the pool at the end when all pop already happens. It would be required to get rid of await in between but I do not know how since I do not understand what exactly `await self.max_connections.acquire()` does.

yeraydiazdiaz:
I noticed this happening on a similar situation as well and it also confused me. I fully expected connections to only be created once per origin, but it seems the current behaviour is by design according to this test?

https://github.com/encode/httpx/blob/master/tests/dispatch/test_connection_pools.py#L69-L80

@PrimozGodec, `max_connections` is a [`BoundedSemaphore`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.BoundedSemaphore) which HTTPX uses to limit the number of connections of the pool. The default is 100.

PrimozGodec:
@yeraydiazdiaz thank you for the response. So it looks OK then. Can establishing more connections instead of one at the same time decrease the speed of getting the responses?

tomchristie:
[...]

PR #693, opened 2019-12-29
Switch from an Event primitive to a Lock primitive

tomchristie:
As noted in https://github.com/encode/httpx/pull/660#issuecomment-567906879

This will allow us to then:

* Move read/write locks *outside* the stream, and onto the HTTP classes instead, with a view to resolving #551.
* Revisit https://github.com/encode/httpx/pull/624, which needs locking behavior to get fully sorted. (With a view to resolving #514)

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -6,7 +6,7 @@ from h2.config import H2Configuration
 from h2.settings import SettingCodes, Settings
 
 from ..backends.base import (
-    BaseEvent,
+    BaseLock,
     BaseSocketStream,
     ConcurrencyBackend,
     lookup_backend,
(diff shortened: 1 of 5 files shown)


httpx/dispatch/http2.py
```diff
@@ -39,32 +39,28 @@ def __init__(
...
-            self.init_complete.set()
-        else:
-            # All other streams need to wait until the connection is established.
-            await self.init_complete.wait()
+        async with self.init_lock:
+            if not self.sent_connection_init:
+                # The very first stream is responsible for initiating the connection.
+                await self.send_connection_init(timeout)
+                self.sent_connection_init = True
             stream_id = self.state.get_next_available_stream_id()
```
florimondmanca: Took me some time to realise this works as expected, but it does seem to!

PR #694, opened 2019-12-29
Lock at HTTP level

tomchristie:
Moves the read/write locking out of the socket level, and onto the HTTP implementation level.

I've been more coarse grained about what logic the read/write locks go around in anticipation of having to also support threaded concurrency from the same codebase, which could switch between threads at any point. As a result I'm making sure we're locking around any operations that mutate `self.state`.

Not 100% about this yet, but opening as a draft to look over.

### DIFF
--- a/httpx/dispatch/http2.py
+++ b/httpx/dispatch/http2.py
@@ -6,7 +6,7 @@ from h2.config import H2Configuration
 from h2.settings import SettingCodes, Settings
 
 from ..backends.base import (
-    BaseEvent,
+    BaseLock,
     BaseSocketStream,
     ConcurrencyBackend,
     lookup_backend,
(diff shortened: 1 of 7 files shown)


httpx/dispatch/http2.py
```diff
@@ -107,7 +119,9 @@ def is_connection_dropped(self) -> bool:
         return self.socket.is_connection_dropped()
 
     async def close(self) -> None:
-        await self.socket.close()
+        async with self.write_lock:
+            async with self.read_lock:
```
florimondmanca: Alternatively…

```python
async with self.write_lock, self.read_lock:
    await self.socket.close()
```

tomchristie: TIL

florimondmanca: Works with a regular `with:` and supports `as`, too!

```python
with open("file1.txt") as f1, open("file2.txt") as f2:
    ...
```

The more you know…

tomchristie:
Related to this...

Seems like the *minimal* fix for https://github.com/encode/httpx/issues/551 is that the asyncio `SocketStream` backend ought to have a write lock, which it acquires on write and close operations. It might be that we want to go with that *simpler* fix first, and keep this work on the backburner.

Issue #572, opened 2019-11-30
Supporting Sync. Done right.


tomchristie:
So, with the 0.8 release we dropped our sync interface completely.

The approach of bridging a sync interface onto an underlying async one, had some issues, in particular:

* Have Sync and Async variant classes both subclassing Base classes was leading to poorer, less comprehensible code.
* There were niggly issues, such as being unable to run in some environments where top-level async support was provided. (Eg. failing under Jupyter)

The good news is that I think the `unasync` approach that the `hip` team are working with is *actually* a much better tack onto this, and is something that I think `httpx` can adopt.

We'll need to decide:

* How we name/scope our sync vs. async variants.
* If we provide the top-level API for sync+async, or just sync.

(Essentially the same discussion as https://github.com/python-trio/hip/issues/168)

One option here could be...

* `httpx.get()`, `httpx.post()` and pals. *Sync only - they're just a nice convenience. Good for the repl and casual scripting, but no good reason why async code shouldn't use a proper client.*
* `httpx.sync.Client`, `httpx.async.Client` and pals. *Keep the names the same in each sync vs async case, but just scope them into two different submodules. Our `__repr__`'s for these clases could make sure to be distinct, eg. `<async.Response [200 OK]>`*

So eg...

**Sync**:

```python
client = httpx.sync.Client()
client.get(...)
```

**Async**:

```python
client = httpx.async.Client()
await client.get(...)
```

---

**Update**

Here's how the naming/scoping could look...

Top level:

```python
response = httpx.get(...)
response = httpx.post(...)
response = httpx.request(...)
with httpx.stream(...) as response:
    ...
```

```python
response = await httpx.aget(...)
response = await httpx.apost(...)
response = await httpx.arequest(...)
async with httpx.astream(...) as response:
    ...
```

Client usage:

```python
client = httpx.SyncClient()
```

```python
client = httpx.AsyncClient()
```

(Unlike our previous approach there's no `Client` case there, and no subclassing. Just a pure async implementation, and a pure sync implementation.)

We can just have a single `Request` class, that'd accept whatever types on __init__ and expose both streaming and async streaming interfaces to be used by the dispatcher, that raises errors in invalid cases eg. the sync streaming case is called, but it was passed an async iterable for body data.

Most operations would return a plain ol' `Response` class, with no I/O available on it. See https://github.com/encode/httpx/issues/588#issuecomment-562057229
When using `.stream`, either an `AsyncResponse` or a `SyncResponse` would be returned.

[...]

PR #695, opened 2019-12-30
Sync streaming interface on responses

tomchristie:
Refs #572 

Add sync interfaces for streaming on responses.

Draft PR at the moment just to demonstrate how it'll look... The tests will need filling out to.

Related to this will be:

* Ability to support both sync and async request bodies... We need an `IteratorStream` in `content_streams.py`.
* `.next()`

### DIFF
--- a/httpx/content_streams.py
+++ b/httpx/content_streams.py
@@ -108,10 +108,8 @@ class IteratorStream(ContentStream):
         for part in self.iterator:
             yield part
 
-    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+    def __aiter__(self) -> typing.AsyncIterator[bytes]:
         raise RuntimeError("Attempted to call a async iterator on an sync stream.")
-        async for part in ():  # pragma: nocover
-            yield b""
 
     def close(self) -> None:
         if self.close_func is not None:
--- a/httpx/content_streams.py
+++ b/httpx/content_streams.py
@@ -9,7 +9,9 @@ from urllib.parse import urlencode
 
 from .utils import format_form_param
 
-RequestData = typing.Union[dict, str, bytes, typing.AsyncIterator[bytes]]
+RequestData = typing.Union[
+    dict, str, bytes, typing.Iterator[bytes], typing.AsyncIterator[bytes]
+]
 
 RequestFiles = typing.Dict[
     str,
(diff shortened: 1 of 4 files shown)


httpx/models.py
```diff
@@ -917,18 +979,15 @@ def raw(self):  # type: ignore
         """
         A byte-iterator over the raw response content.
         """
-        if hasattr(self, "_raw_content"):
```
tomchristie: We can drop `_raw_content` here since we're not special casing that anymore.

httpx/models.py
```diff
@@ -673,15 +679,13 @@ def __init__(
...
-        else:
-            self.is_closed = False
-            self.is_stream_consumed = False
+        self.is_closed = False
+        self.is_stream_consumed = False
+        if stream is not None:
             self._raw_stream = stream
+        else:
+            self._raw_stream = ByteStream(body=content or b"")
+            self.read()
```
tomchristie: This is a little nice now that we can use a sync `.read()` inside `__init__` rather than having to special-case `self._raw_content`.

httpx/content_streams.py
```diff
@@ -68,10 +76,48 @@ def get_headers(self) -> typing.Dict[str, str]:
...
+        if self.is_stream_consumed:
+            raise StreamConsumed()
+        self.is_stream_consumed = True
+        for part in self.iterator:
+            yield part
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        raise RuntimeError("Attempted to call a async iterator on an sync stream.")
+        async for part in ():  # pragma: nocover
+            yield b""
```
florimondmanca: Does this have any actual runtime consequence, or is this to please linters?

Turns out that we can make this work without tricks by dropping the `async` annotation…

```python
def __aiter__(self) -> typing.AsyncIterator[bytes]:
    raise RuntimeError("Attempted to call a async iterator on an sync stream.")
```

tests/test_content_streams.py
```diff
@@ -130,3 +187,9 @@
             b"--+++--\r\n",
         ]
     )
+
+
+@pytest.mark.asyncio
+async def test_invalid_argument():
```
florimondmanca: We don't *need* the `asyncio` marker here, do we? Possible simplification…

```python
def test_invalid_argument():
    with pytest.raises(TypeError):
        encode(123)
```

httpx/content_streams.py
tomchristie: Nice one, yup!

tomchristie:
Thanks for the great review!
I think we may as well get this one in now.

PR #698, opened 2019-12-31
Set `open_connection` before adding the connection to `active_connections` (see https://github.com/encode/httpx/issues/689)

### LINKED ISSUE #689: Unexpected AssertionError (httpx 0.9.5)
I'm getting an unexpected `AssertionError` while trying to do a relatively simple `await client.get()` call.  Since the last time this worked I've done two big things, upgraded my project to use Python3.8 (not really important, but the assignment operator is useful for my project) as well as collapsing several httpx clients into a single, package-wide client, that is now being used to make several concurrent, but unrelated requests to the same host (FastAPI/Uvicorn, so no HTTP/2).

I'm not super familiar with the codebase, but at a quick glance it looks like this might be a problem with issuing the concurrent requests?  I can try to throw a lock around the requests to start, though I think that shouldn't really be necessary?

Not sure if it's useful, but the calling code is using uvloop and running on Linux (specifically the python:3.8 Docker container).

```
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 679, in get
    return await self.request(
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 259, in request
    response = await self.send(
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 403, in send
    response = await self.send_handling_redirects(
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 465, in send_handling_redirects
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 591, in send_handling_auth
    response = await self.send_single_request(request, timeout, verify, cert)
  File "/opt/venv/lib/python3.8/site-packages/httpx/client.py", line 618, in send_single_request
    response = await dispatcher.send(
  File "/opt/venv/lib/python3.8/site-packages/httpx/dispatch/connection_pool.py", line 147, in send
    connection = await self.acquire_connection(
  File "/opt/venv/lib/python3.8/site-packages/httpx/dispatch/connection_pool.py", line 165, in acquire_connection
    connection = self.pop_connection(origin)
  File "/opt/venv/lib/python3.8/site-packages/httpx/dispatch/connection_pool.py", line 215, in pop_connection
    connection = self.active_connections.pop_by_origin(origin, http2_only=True)
  File "/opt/venv/lib/python3.8/site-packages/httpx/dispatch/connection_pool.py", line 46, in pop_by_origin
    if http2_only and not connection.is_http2:
  File "/opt/venv/lib/python3.8/site-packages/httpx/dispatch/connection.py", line 175, in is_http2
    assert self.open_connection is not None
AssertionError
```

I don't have a minimal example yet, I can try to put one together though if that'd be helpful.

kshramt:
Fixes #689

### DIFF
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -51,7 +51,9 @@ class HTTPConnection(Dispatcher):
 
         return response
 
-    async def connect(self, timeout: Timeout) -> None:
+    async def connect(self, timeout: Timeout = None) -> None:
+        timeout = Timeout() if timeout is None else timeout
+
         host = self.origin.host
         port = self.origin.port
         ssl_context = await self.get_ssl_context(self.ssl)
--- a/httpx/dispatch/connection_pool.py
+++ b/httpx/dispatch/connection_pool.py
@@ -174,6 +174,7 @@ class ConnectionPool(Dispatcher):
                 trust_env=self.trust_env,
                 uds=self.uds,
             )
+            await connection.connect(timeout=timeout)
             logger.trace(f"new_connection connection={connection!r}")
         else:
             logger.trace(f"reuse_connection connection={connection!r}")
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1,3 +1,4 @@
+import asyncio
 from datetime import timedelta
 
 import pytest
@@ -180,3 +181,18 @@ async def test_elapsed_delay(server):
     async with httpx.AsyncClient() as client:
         response = await client.get(url)
     assert response.elapsed.total_seconds() > 0.0
+
+
+@pytest.mark.asyncio
+async def test_open_connection_invariance(server):
+    """
+    Check that `assert self.open_connection is not None` statement is not violated.
+    See https://github.com/encode/httpx/issues/689 .
+    """
+
+    async def run_many(client, n):
+        for _ in range(n):
+            await client.get(server.url)
+
+    async with httpx.AsyncClient() as client:
+        await asyncio.gather(*[run_many(client, 10) for _ in range(11)])


kshramt:
Test code:

```
import asyncio
import httpx


def main():
    asyncio.run(run())


async def run():
        await asyncio.gather(*[runn(client, 20) for _ in range(20)])


async def runn(client, n):
    for _ in range(n):
        await run1(client)


def run1(client):
    return client.get(<URL>)


main()
```

florimondmanca:
Edited the issue description to reference the original issue.

tomchristie:
An alternate fix is to have `is_http2` return `False` if no connection has been acquired yet.

I'd *consider* leaning towards that instead. It's a lower footprint, and a less fundamental change than calling directly into `.connect` from the connection pool.

We've got some work remaining to do on dealing with stricter locking around the HTTPConnection state changes, and handling "this is a pending connection, that might (HTTP/2) or might not (HTTP/1.1) be able to support concurrent requests". (Draft PR #624) - I'm not sure that work would fit in so well with the "acquire the connection and wait until it's connected", because in the HTTP/2 case we'd like to be able to acquire a connection even if it's still in a PENDING state. (#514)

florimondmanca:
Hmm, yes. I can confirm #700 fixes the original issue that this addresses with a much smaller footprint, so it seems we should consider it instead.

Still seems we don't have any test for the behavior of `hard_limit` with concurrent requests (esp. when using asyncio's `.gather()`), so I think we should add one eventually.

tomchristie:
@kshramt Thanks for having put your time into this. Closing in favour of #700.

PR #699, opened 2019-12-31
Lock around `stream.write` and `stream.close` operations

tomchristie:
Closes #551

PR #700, opened 2019-12-31
Drop `OpenConnection`

### LINKED ISSUE #689: Unexpected AssertionError (httpx 0.9.5)

tomchristie:
Since we're not 100% on if we want #496 or not, I think we should drop the `OpenConection` abstraction for now - really we should only introduce that if we need it.

I've also made some other refactoring tweaks:

* `.open_connection` -> `.connection`.
* Drop `.set_open_connection` in favor of just instantiating the connection where used.
* More robust behavior for `.is_http2`, `.is_closed`, `.is_connection_dropped` properties.

Also closes #689

### DIFF
--- a/httpx/dispatch/base.py
+++ b/httpx/dispatch/base.py
@@ -50,26 +50,3 @@ class Dispatcher:
         traceback: TracebackType = None,
     ) -> None:
         await self.close()
-
-
-class OpenConnection:
-    """
-    Base class for connection classes that interact with a host via HTTP.
-    """
-
-    @property
-    def is_http2(self) -> bool:
-        raise NotImplementedError()  # pragma: no cover
-
-    async def send(self, request: Request, timeout: Timeout = None,) -> Response:
-        raise NotImplementedError()  # pragma: no cover
-
-    @property
-    def is_closed(self) -> bool:
-        raise NotImplementedError()  # pragma: no cover
-
-    def is_connection_dropped(self) -> bool:
-        raise NotImplementedError()  # pragma: no cover
-
-    async def close(self) -> None:
-        raise NotImplementedError()  # pragma: no cover
--- a/httpx/dispatch/connection.py
+++ b/httpx/dispatch/connection.py
@@ -4,11 +4,11 @@ import typing
 
 import h11
 
-from ..backends.base import BaseSocketStream, ConcurrencyBackend, lookup_backend
+from ..backends.base import ConcurrencyBackend, lookup_backend
 from ..config import CertTypes, SSLConfig, Timeout, VerifyTypes
 from ..models import URL, Origin, Request, Response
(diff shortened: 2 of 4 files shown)


florimondmanca:
Adding a ref to the original PR for `OpenConnection` here: #616.

tomchristie:
Thanks for the review @florimondmanca!
I'd suggest we aim for a 0.10.1 release with this and #699 both in.

PR #701, opened 2019-12-31
Version 0.10.1

tomchristie:
### Fixed

- Fix issue with concurrent connection acquiry. (Pull #700)
- Fix write error on closing HTTP/2 connections. (Pull #699)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.10.1 (December 31st, 2019)
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


florimondmanca:
Can we please wait a bit for a confirmation from @PrimozGodec on the fixing of https://github.com/encode/httpx/issues/551 with this release? It's not clear to me that we validated that the fix *actually* works (although in theory it *should*), did we?

To install from this branch:

```
pip install git+https://github.com/encode/httpx@version-0.10.1
```

tomchristie:
I think #699 is a clear enough valid improvement in it's own right, so not too fussed either way, but happy to do so, yup! 😃

PrimozGodec:
I see that version 0.10.1 is already made, but I am confirming that it is fixed now, anyway. Thank you @tomchristie

PR #704, opened 2019-12-31
Make timeout parameter None-able & cleanup __all__.

j178:
[The doc](https://www.encode.io/httpx/advanced/#setting-and-disabling-timeouts) says we can disable timeout by setting it to None, but currently `TimeoutTypes`  doesn't allow a None, so I added it.
Plus, cleans up two removed entries from `__all__`.


tomchristie:
Great, thank you!

PR #706, opened 2020-01-01
SSLConfig refactor

tomchristie:
The HTTPConnection class now takes any ssl configuration as a single `ssl: SSLConfig` argument.
As a result, the ConnectionPool class holds a single `SSLConfig` argument, that only ever ends up being loaded once.

I've dropped the `run_in_threadpool` behavior since we're only making a single load-from-disk per Client instance here.

Should provide slightly better performance since we're not reloading the SSLContext once-per-connection anymore.

I'd like to iterate on this further so that the context loading just happens directly when instantiating `SSLConfig`. However, I'm going to leave that to a follow-up PR, in order to keep this managably reviewable.

### DIFF
--- a/tests/client/test_proxies.py
+++ b/tests/client/test_proxies.py
@@ -29,27 +29,6 @@ def test_proxies_parameter(proxies, expected_proxies):
     assert len(expected_proxies) == len(client.proxies)
 
 
-def test_proxies_has_same_properties_as_dispatch():
-    client = httpx.AsyncClient(
-        proxies="http://127.0.0.1",
-        verify="/path/to/verify",
-        cert="/path/to/cert",
-        trust_env=False,
-        timeout=30,
-    )
-    pool = client.dispatch
-    proxy = client.proxies["all"]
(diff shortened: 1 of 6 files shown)


tests/client/test_proxies.py
```diff
@@ -29,27 +29,6 @@ def test_proxies_parameter(proxies, expected_proxies):
...
-    proxy = client.proxies["all"]
-
-    assert isinstance(proxy, httpx.HTTPProxy)
-
-    for prop in [
-        "verify",
-        "cert",
-        "pool_limits",
-    ]:
-        assert getattr(pool, prop) == getattr(proxy, prop)
```
tomchristie: `verify` and `cert` properties no longer exist.
I don't think this test case is sufficiently meaningful at this point, so dropping it. 😃

tomchristie:
> I hadn't realized we had been loading SSL configs for each new connection.

Same, actually. Really it's something that's fallen out of #617

PR #690, opened 2019-12-27
Repurpose `RedirectBodyUnavailable` exception as `RequestBodyUnavailable`

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/685#discussion_r361761039

This PR generalizes the semantics of `RedirectBodyUnavailable` so that it fits situations when we can't replay a request stream that aren't related to redirection (e.g. Digest Auth in #685).

PR #707, opened 2020-01-01
Wrap network errors as HTTPX-specific exceptions

### LINKED ISSUE #349: [Feature Request]: raise httpx exception instead of buildin ConnectionResetError and ConnectionError

florimondmanca:
Fixes #349 

- Wrap `OSError` exceptions raised in backends `.read()`, `.write()`, `.close()`, as well as `.open_xxx_stream()` methods.
- For trio, handle `BrokenResourceError` exceptions as well when we can expect trio to raise them instead of `OSError`-derived exceptions.
- Specialize `ConnectionClosed` to be a subclass of `NetworkError`.

Note that we're not doing special handling for `ConnectionError` because it's a subclass of `OSError`.

Not married to the naming/API specifics of the `as_network_error()` construct — could also be e.g. `raise_as(NetworkError, OSError)`. But I think it's a good idea to have a dedicated utility for actually performing the wrapping, instead of bare `try/except`. Eg. this centralizes which behavior we'd like to expose w.r.t. exposing or hiding the `__cause__` of the `NetworkError`.

### DIFF
--- a/httpx/backends/asyncio.py
+++ b/httpx/backends/asyncio.py
@@ -5,6 +5,7 @@ import typing
 
 from ..config import Timeout
 from ..exceptions import ConnectTimeout, ReadTimeout, WriteTimeout
+from ..utils import as_network_error
 from .base import BaseLock, BaseSemaphore, BaseSocketStream, ConcurrencyBackend
 
 SSL_MONKEY_PATCH_APPLIED = False
@@ -126,9 +127,10 @@ class SocketStream(BaseSocketStream):
     async def read(self, n: int, timeout: Timeout) -> bytes:
         try:
             async with self.read_lock:
-                return await asyncio.wait_for(
-                    self.stream_reader.read(n), timeout.read_timeout
-                )
+                with as_network_error(OSError):
+                    return await asyncio.wait_for(
+                        self.stream_reader.read(n), timeout.read_timeout
+                    )
         except asyncio.TimeoutError:
             raise ReadTimeout() from None
 
@@ -138,10 +140,11 @@ class SocketStream(BaseSocketStream):
 
         try:
             async with self.write_lock:
-                self.stream_writer.write(data)
-                return await asyncio.wait_for(
-                    self.stream_writer.drain(), timeout.write_timeout
-                )
+                with as_network_error(OSError):
+                    self.stream_writer.write(data)
+                    return await asyncio.wait_for(
+                        self.stream_writer.drain(), timeout.write_timeout
+                    )
         except asyncio.TimeoutError:
             raise WriteTimeout() from None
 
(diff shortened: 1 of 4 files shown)


tomchristie:
This is a really nice approach!

* Would it make sense to be more specific than `OSError`?
* Should we differentiate between `CouldNotConnect` and `BrokenConnection` (Or whatever other naming we want here?)

florimondmanca:
> Would it make sense to be more specific than OSError?

I don’t think so:

- For trio, we use OSError in connection methods because that’s what is explicitly documented as being possible to get from those methods.
- For asyncio, it might be possible to switch to ConnectionError in read/write methods. The asyncio docs are very sparse regarding the exception hierarchy, so that would need checking. At least with OSError we’d be catching _everything_.

> Should we differentiate between CouldNotConnect and BrokenConnection (Or whatever other naming we want here?)

Probably... CouldNotConnect for connection methods, and BrokenConnection for read/write?

florimondmanca:
> Do we want to merge this in now, or should we defer to a major release?

You meant _minor_? :) I think it’s good to have this release as soon as possible, but it’s half a fix/half a behavior change, so I’m not too inclined on releasing as 0.10.1 either...

I think we’d have way enough material to cut out 0.11 with...

- This PR
- #688
- #685
- #686
- #690

I.e. no need to wait for the urllib3 sync client, and that could probably be for 0.12? 😄

florimondmanca:
Oh wait, the sync client is just around the corner: #716 🎉

tomchristie:
@florimondmanca I think we’re ready to start getting it in, yup. But we might (?) choose to do that incrementally as smaller individual PRs.

Also: We could either create a 0.11 branch, or just plan on merging the PRs to master since we’re expecting this to be our next release.

We *might* (?) also choose to start by issuing a 0.11 prerelease, so that we can get a bit of test usage in before making any big “httpx has sync&async support” proclamations.

dalf:
Using httpx (from 0.7.5 to 0.9.5), I've seen these errors:
* asyncio.TimeoutError(concurrent.futures._base.Error)
* ssl.CertificateError(ValueError)
* h11._util.RemoteProtocolError(Exception)

florimondmanca:
@dalf None of these would be handled differently via this PR. There was a recent fix for the last one, but the TimeoutError is unexpected since we should already be handling those by raising the proper HTTPX timeout exception. If that still replicates the latest version, feel free to open an issue with a replication example to track it. :)

PR #710, opened 2020-01-02
Drop `run` and `run_in_threadpool`

tomchristie:
We're no longer using the backend methods `run` and `run_in_threadpool` within the `httpx` codebase, so this PR removes them.

We *are* however using them within the test cases in order to perform the server restarts.

I've changed things around so that `.restart()` is a standard blocking function (It may as well be, and it runs from a different thread from the server, so it won't block the server itself from shutting down.)

However, the tests that use `server.restart()` are currently failing in the `asyncio` case. (With h11 errors that the server close event is occuring at a point that it should not.) Really not clear to me why we'd see that with asyncio but not with trio. I *think* that are resolution here might be that restart should perform a full shutdown, and reopen the server on an entirely new thread, to keep things cleaner?

Any thoughts on this @florimondmanca?


florimondmanca:
Yes, I remember server restarts have already been the cause of some headaches in the past… https://github.com/encode/httpx/pull/276#issuecomment-525621680

As far as understanding the error message, it's already been discussed in https://github.com/encode/httpx/issues/96 — basically the server abruptly closed the connection (which it did), and I think we're getting issues on asyncio because of the blocking call to `time.sleep()` (trio maybe handles socket closure more cleanly in that case).

Actually, if we turn `restart()` into a coroutine again, with calls to `asyncio.sleep(0.2)` instead of `time.sleep()`, the test runs fine on asyncio — but it fails on trio, obviously, because of the call to `asyncio.sleep()`.

*Luckily*, we have an agnostic `sleep()` implementation in `tests/concurrency.py`, so since there are no other asyncio-specific async calls in `restart()`, we can *probably* call `await sleep()` directly, and things will work on trio too.

Lemme push a commit to demonstrate the idea…

florimondmanca:
Note that these connection pool tests run faster now, because we don't have to setup/teardown a thread on each restart. 😄

tomchristie:
@florimondmanca Thanks, great stuff!

I've now also dropped `AsyncioBackend.loop`, since it's not used anywhere.

Also cleaned up the `tests/concurrency.py` util functions to use a more direct style than `functools.single_dispatch` - we don't really need to be that clever.

PR #713, opened 2020-01-02
Prep for introducing SyncClient

tomchristie:
* Reorganize method ordering in client, with sync methods appearing first. Keeping all the I/O methods seperate will ensure that 
* Add `init_dispatch`, so that we can have seperate instantiation for the sync and async cases.
* Use `httpx.Proxy(...)` for any fine-grained proxy configuration, rather than instantiating a dispatcher class directly. This ensures that proxy configuration looks the same in the API regardless of sync vs. async.

Eg...

```python
proxies = httpx.Proxy("https://12.34.56.78:910", mode="CONNECT_ONLY")
```

I've ensured that the existing proxy API is currently still supported here, so that we'd be able to bring this change in without breaking API.


tomchristie:
Can potentially split this into a smaller set of PRs if helpful.

PR #716, opened 2020-01-03
Reintroduce sync client

tomchristie:
Builds on #713 and reintroduces a sync Client.

```python
>>> import httpx
<Response [200 OK]>
```

🎉✨ **TA-DA!** ✨🎉

We're using `urllib3` here as the backing service for our sync calls.

Potentially we could later switch away to using our own `unasync`'ed dispatcher. (Although it's actually feasible that we might just want to stick with urllib3 for the sync case.)

If we wanted to tackle this as a set of incremental PRs we would want:

* Switch `Dispatcher` to `AsyncDispatcher` throughout.
* Introduce a `SyncDispatcher` class.
* Introduce a `Proxy` config class.
* Split the client into the two sync and async cases, with a base class of `BaseClient`.
* Update `StreamContextManager` to handle both sync and async cases.
* Switch the top-level API back to sync.

Also still TODO here:

* Tie in timeout and pool limits config properly.
* Introduce a `WSGIDispatcher` class.
* Introduce a `URLLib3ProxyDispatcher` class.
* Plenty of docs updating.
* Plenty of test updating.
* Wrap `urllib3` exceptions (Need to cover `pool.urlopen` and `conn.stream` methods)


tomchristie:
Now closing in favour of #735

PR #718, opened 2020-01-03
Record history of requests made during authentication

### LINKED ISSUE #669: Authentication flows should preserve history.

florimondmanca:
Fixes #669 

Based on @gdhameeja's work in #686 🌟🙇

The updated tests contain:

- Asserts on `response.history` in the digest auth case.
- An isolated test case using an imaginary auth scheme, for in-depth verification of what gets pushed to the history.

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -451,7 +451,7 @@ class AsyncClient:
                 raise RedirectLoop()
 
             response = await self.send_handling_auth(
-                request, auth=auth, timeout=timeout,
+                request, history, auth=auth, timeout=timeout,
             )
             response.history = list(history)
 
@@ -566,7 +566,11 @@ class AsyncClient:
         return request.stream
 
     async def send_handling_auth(
-        self, request: Request, auth: Auth, timeout: Timeout,
+        self,
+        request: Request,
+        history: typing.List[Response],
+        auth: Auth,
+        timeout: Timeout,
     ) -> Response:
         auth_flow = auth(request)
         request = next(auth_flow)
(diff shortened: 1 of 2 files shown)


httpx/client.py
```diff
@@ -580,8 +584,10 @@ def redirect_stream(
                 await response.aclose()
                 raise exc from None
             else:
+                response.history = list(history)
+                await response.aread()
                 request = next_request
-                await response.aclose()
+                history.append(response)
```
florimondmanca: As suggested by @j178, this part closely follows the behavior we have when handling redirection: copy the current history onto the latest response, then add the response to the history.

The only difference is that here we mutate `history` in-place instead of creating a new list (`history = history + [response]`) so that the change is propagated to `.send_handling_redirects()`.

It's possible to change this so that we build and maintain an `auth_history` list instead, and merge it with the initial `history` both here and `.send_handling_redirects()`, but I'm not sure the gained immutability would be worth the extra complexity.

Separate note: I don't think we test the case of redirect + auth yet, but that would be a good idea. Definitely out of scope for this PR though?

florimondmanca:
As a follow-up, I think we'll also want to update the wording in [Redirection and History](https://www.encode.io/httpx/quickstart/#redirection-and-history), as it currently describes the history as "a list of all any redirect responses that were followed". *Technically* that's true, but now the history would *also* list intermediate auth responses.

httpx/client.py
```diff
@@ -580,8 +584,10 @@ def redirect_stream(
                 await response.aclose()
                 raise exc from None
             else:
+                response.history = list(history)
+                await response.aread()
```
tomchristie: Yup, good call.

It's possible that at a later date we *might* want to ensure that streaming responses *don't* load an intermediate response bodies (in redirect or auth), but that's only a *"might"*, and would be a seperate issue if we *did* choose to tackle it.

florimondmanca:
Merging, thanks for taking the time to look into this @gdhameeja! 💯

Issue #715, opened 2020-01-02
Getting to 1.0


tomchristie:
An informal issue here to talk about one route of getting to 1.0, that I think makes a bunch of sense...

* Remove `httpx.HTTPProxy` from the public API, in favor of `httpx.Proxy`, which is an object that just holds the proxy config, rather than a `Dispatcher` instance.  #713
* Add a `SyncClient` implementation, initially *requiring* `dispatch=...` be passed, at this point we can introduce tests for the sync client.
* Add `WSGIDispatcher`, and support `SyncClient(app=...)`.
* Add `URLLib3ConnectionPool`, and `URLLib3ProxyManager` implementations, and use those by default for the sync case. This gives us `urllib3` backing for the sync case, and our own dispatch for the async case.
* Switch the top-level API to the sync case.
* Drop any remaining bits of API that we currently have warnings for.
* Cut a 1.11 release.

At this point we're in great shape - we've got all our public API how we think we want it to look for 1.0, and we've got sync+async support.

(We *won't* have sync HTTP/2 support at that point, because we're backed by `urllib3` there.)

We can then:

* Use an `unasync` approach to provide a sync dispatch implementation.
* Work on resolving any criticial issues that we think would block our users from wanting to switch away from the `urllib3` dispatcher implementation.
* Switch the sync dispatcher to us our built-in version, with a `URLLib3Client` provided as an alternate case.
* Issue a 1.0 release.

What I think is compelling about this route is that it gets us to an "API complete" state as quickly as possible, while still allowing us to eventually switch to using a single unified dispatch implementation for the 1.0 release.

tomchristie:
Alrighty, I wanted to check that there's no obvious blockers to 1.11, so pressed on with a draft PR to demonstate... #716

Anyways, *yay*, I reckon we're just about there!

We may want to think about our review process, and address the changes bit by bit to make sure they get properly reviewed, but that PR demonstates how it'll look.

florimondmanca:
That’s cool! But let’s hold off onto #716 for a bit, maybe?

The approach taken here is to duplicate a bunch of code, and reintroduce the BaseClient construct we had before 0.8.

Is that the final situation we’d like to end up in? I’m not sure…

I’d really see unasync help us out a lot here. If we go that path, while it’s _okay_ for us to duplicate code _for now_, I’m very much in favor of unasyncing the client for 1.0, and not after (if that was the strategy), for the sake of derisking and taking our time to _actually achieve stability_ (beyond API stability).

[...]

PR #722, opened 2020-01-05
Unasync'ed HTTPX

florimondmanca:
I wanted to get an actual feeling for what switching to `unasync` would look like and require, so…

This PR provides a working^1 implementation for an unasync'ed HTTPX, from concurrency backends to `SyncClient`… :tada:

Uses a sync backend adapted from #525, and a patched version of `unasync` waiting for https://github.com/python-trio/unasync/pull/53 to be released. Also adds a `test_sync_client.py` module.

~^1: the sync UDS implementation is still TODO, which is why CI fails.~

Below some notes about what I learnt. I'll start with…

## So what?

There are no *major* blockers to unasync'ing, however:

- Unasync'ing *does* move a bunch of code around, and that's a bit painful in terms of reviewing especially.
- Unasync'ing will require us to switch most of our generically named classes such as `Dispatcher`, `HTTPConnection`, etc. to the `Async`-prefixed variant. So: `AsyncDispatcher`, `AsyncHTTPConnection`, etc. (*Or* we could make `unasync` more clever than it needs to be, but we definitely don't want it to be too clever/configurable.)

In terms of doing this incrementally *after* adding an urllib3-backed 

## Achievements

- The public API hasn't changed at all. This PR is 100% backwards-compatible (except maybe some imports in the docs).
- Typing and editor support is as good as before, and linters are happy.
- Development process now involves a `scripts/compile` step whenever we want to use the sync code (e.g. for tests or local experimentation).

## Workflow

My main workflow for this PR was to move code to the `_async` folder, generate the `_sync` one, and run mypy to see what's wrong at least from a typing perspective. Eg. we don't want `SyncConnectionPool` to be importing the current `HTTPConnection`, because that's an async component. (And that hints us at unasync'ing `HTTPConnection` too. In general unasync'ing component `X` requires to unasync all its dependencies, which isn't surprising.)

## Typing support

I'm very pleased to say that once the `_sync` version is compiled, mypy is 100% happy with the output. Editor support is great too, e.g. when importing code from `._sync` we get autocompletions on modules and then on the imported components.

## Top-level API

The top-level API here is switched to sync (in line with #715). (Not using unasync here since we wouldn't be using the async top-level API.)

## Renaming stuff

Besides handling conversion for builtins, the only mechanism that unasync provides for renaming stuff is in the form of `AsyncXYZ -> SyncXYZ`.

(I think this is nice, because it brings a lot of consistency w.r.t. to "how do I unasync this component.")

But it doesn't handle (1) `aread() -> read()` / `aiter_xxx() -> iter_xxx()`, (2a) `AsyncIteratorStream -> IteratorStream`, (2b) `func(aiterator=...) -> func(iterator=...)`, certainly not (2) `ASGIDispatch -> WSGIDispatch`.

To get things working, I solved (1) by defining the mappings manually via the `AsyncFixes` and `SyncFixes` classes, that e.g. have `read_response()`, which is async on `AsyncFixes`, but sync on `SyncFixes`. Then the `_async` part of the code imports `AsyncFixes`, which gets unasync'ed to `SyncFixes`, and everything works great. Another approach would be to unasync the response, i.e. drop our recent approach of "let's put both sync and async methods on `Response`", and rename `.aiter_xxx()` to `.iter_xxx()` so that we expose `.iter_xxx()` in all cases. (That doesn't make it clear that you'd need to `async for` the result though, does it?)

For (2a), `(Async)IteratorStream` doesn't have to be defined in the `_(a)sync` part of the code (it's in `content_streams.py` right now), so I left it there but just renamed `IteratorStream` to the more standard `SyncIteratorStream`. This allows `unasync` to seamlessly rename references to `AsyncIteratorStream` to `SyncIteratorStream`.

(2b) is similar: internal API, so I went ahead and standardized the `aiterator` parameter name to `iterator`.

For (3), I renamed `ASGIDispatch` to `AsyncAppDispatch`, so that it gets unasync'ed to `SyncAppDispatch`. We can then easily re-export things to `ASGIDispatch` and `WSGIDispatch`.

## Proxy modules

There are several "proxy modules" left (i.e. modules that import stuff from `_async` and `_sync` to re-export them), mostly because I didn't want to break the import paths in the tests. We'd probably need to handle that eventually, but at least this is a technique for breaking things up a bit.

## To commit or not to commit the sync code?

This PR git-ignores the `_sync` part of the code, which means it won't show up in the source tree in version control, and `scripts/compile` is a prerequisite to even running the tests locally. But the developer experience feels cleaner that way. Eg. it makes it clearer that `_sync` should never be edited manually, and re-compiles don't trigger git highlighting in the explorer.


httpx/backends/sync.py
```diff
@@ -0,0 +1,243 @@
...
+            raise self.exc_class()
+
+    def release(self) -> None:
+        self.semaphore.release()
+
+
+class SyncBackend(SyncConcurrencyBackend):
+    """
+    Concurrency backend that performs synchronous I/O operations
+    while exposing async-annotated methods.
```
florimondmanca: Ah, this is outdated!

httpx/models.py
```diff
@@ -51,8 +51,9 @@
     str_query_param,
 )
 
-if typing.TYPE_CHECKING:  # pragma: no cover
-    from .dispatch.base import Dispatcher  # noqa: F401
+if typing.TYPE_CHECKING:
+    from ._async.dispatch.connection import AsyncHTTPConnection  # noqa: F401
+    from ._sync.dispatch.connection import SyncHTTPConnection  # noqa: F401
```
florimondmanca: Need to remove these.

tomchristie:
Amazing stuff! Ace bit of work. Crackin stuff on the SyncBackend too! 

* I’ve been meaning to raise an issue about pushing back on UDS support - I think it was a mistake to bring it in early on, since it adds a 1% use-case, at the expense of slowing us down in getting our 99% use cases to a feature complete 1.0 release.
* I’d still like to start by pulling in a raw SyncClient & AsyncClient based on #716 to get us to API completeness, so long as any PRs that get us there keep the test coverage all the way through. I probably think that’d actually be an okay approach longer term too, with unasync targeting the dispatch only. It’s possible that the dispatch implementation could even get split off into a really tightly scoped `httpcore` with a minimal no-models dispatch API, and we can just apply unasync to that package only, while `httpcore` has some small areas of duplication.

florimondmanca:
Yes, doing things the raw way for anything above dispatchers still sounds pretty valid. I think it's a viable path, though I'm still worried that some of the renaming hoops I had to go through to get this working result in issues down the line. Surely we'll need (and be able to?) to extend `unasync` to accomodate our needs.

Anyway, I pushed a last commit demonstrating how we could push the madness even further and have unasync *generate the tests* for the `SyncClient`. This is based off another little hack project from today, [`pytest-unasync`](https://github.com/florimondmanca/pytest-unasync). There are several edge cases related to handling imports that `pytest-unasync` really doesn't handle well, and the better approach would probably be to unasync the entire test modules instead. Well anyway, that was fun!

I don't see this going anywhere near `master` really, so I'm going to close it proactively. Just happy to see that `unasync`-ing everything is *possible*, and that we've now got proof of that if we want to refer to it in the future. Same than with `SyncBackend` and #525. :-)

PR #727, opened 2020-01-06
Proposed 0.11 docs

tomchristie:
There may well be a bunch of stuff still to think over here, but this is what I've got right now...

* Docs changes for sync API.
* Added new "async" section.
* HTTP/2 listed as "async only"
* Kept the UDS docs, listed as "async only" and moving it to the async section.
* Advanced section now covers mounting WSGI apps, async-specific section covers mounting ASGI apps.


README.md
```diff
@@ -16,40 +16,24 @@
...
-
-*In particular, the 0.8 release switched HTTPX into focusing exclusively on
-providing an async client, in order to move the project forward, and help
-us [change our approach to providing sync+async support][sync-support]. If
-you have been using the sync client, you may want to pin to `httpx==0.7.*`,
-and wait until our sync client is reintroduced.*
+**Note**: *HTTPX should be considered in beta. We believe we've got the public API to
+a stable point now, but would strongly recommend pinning your dependencies to the `0.11.*`
+release, so that you're able to properly review API changes between package updates.
+A 1.0 release is expected to be issued sometime on or before April 2020.
```
florimondmanca: Funny bug in GitHub's Markdown rendered: the first `*` for italics clashes with `0.11.*`. Consider switching to `_italics_`?

docs/advanced.md
```diff
@@ -68,79 +68,68 @@ Additionally, `Client` accepts some parameters that aren't available at the requ
...
 
 For example:
 
 ```python
-# Instantiate a client that makes ASGI requests with a client IP of "1.2.3.4",
-# on port 123.
-dispatch = httpx.dispatch.ASGIDispatch(app=app, client=("1.2.3.4", 123))
-async with httpx.AsyncClient(dispatch=dispatch) as client:
+# Instantiate a client that makes WSGI requests with a client IP of "1.2.3.4".
+dispatch = httpx.dispatch.WSGIDispatch(app=app, remote_addr="1.2.3.4")
```
florimondmanca: Would we want that to become `httpx.WSGIDispatch`?

docs/advanced.md
```diff
@@ -274,22 +247,22 @@ You can set timeouts for an individual request:
...
+with httpx.Client() as client:
+    client.get("http://example.com/api/v1/example", timeout=10.0)
 ```
 
 Or disable timeouts for an individual request:
 
 ```python
 # Using the top-level API:
-await httpx.get('http://example.com/api/v1/example', timeout=None)
+ httpx.get('http://example.com/api/v1/example', timeout=None)
```
florimondmanca: Nit:

```suggestion
httpx.get('http://example.com/api/v1/example', timeout=None)
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+
+```python
+>>> async with httpx.AsyncClient() as client:
+>>>     r = await client.get('https://www.example.com/')
+>>> r
+<Response [200 OK]>
+```
+
+!!! tip
+    Use [IPython](https://ipython.readthedocs.io/en/stable/) to try this code interactively, as it supports executing `async`/`await` expressions in the console.
```
florimondmanca: I think we mentioned 3.8+ elsewhere before, so maybe…
```suggestion
    Use [IPython](https://ipython.readthedocs.io/en/stable/) or Python 3.8+ with `python -m asyncio` to try this code interactively, as they support executing `async`/`await` expressions in the console.
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+### Opening and closing clients
+
+Use `async with httpx.AsyncClient()` if you want a context-managed client...
+
+```python
+async with httpx.AsyncClient() as client:
+    ...
+```
+
+Alternatively, use `await clien.aclose()` if you want to close a client explicitly.
```
florimondmanca:
```suggestion
Alternatively, use `await client.aclose()` if you want to close a client explicitly:
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+* `Response.aiter_text()` - For streaming the response content as text.
+* `Response.aiter_lines()` - For streaming the response content as lines of text.
+* `Response.aiter_raw()` - For streaming the raw response bytes, without applying content decoding.
+* `Response.aclose()` - For closing the response. You don't usually need this, since `.stream` block close the response automatically on exit.
+
+### Streaming requests
+
+When sending a request body with an `AsyncClient` instance, it is invalid to use a bytes generator, but it is valid to use an async bytes generator...
+
+```
```
florimondmanca:
```suggestion
```python
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+* `Response.aread()` - For conditionally reading a response inside a stream block.
+* `Response.aiter_content()` - For streaming the response content as bytes.
+* `Response.aiter_text()` - For streaming the response content as text.
+* `Response.aiter_lines()` - For streaming the response content as lines of text.
+* `Response.aiter_raw()` - For streaming the raw response bytes, without applying content decoding.
+* `Response.aclose()` - For closing the response. You don't usually need this, since `.stream` block close the response automatically on exit.
+
+### Streaming requests
+
+When sending a request body with an `AsyncClient` instance, it is invalid to use a bytes generator, but it is valid to use an async bytes generator...
```
florimondmanca:
```suggestion
When sending a streaming request body with an `AsyncClient` instance, you should use an async butes generator instead of a bytes generator:
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+
+```
+async def upload_bytes():
+    ...  # yield byte content
+
+client.post(url, data=upload_bytes())
```
florimondmanca:
```suggestion
await client.post(url, data=upload_bytes())
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+>>> r
+<Response [200 OK]>
+```
+
+!!! tip
+    Use [IPython](https://ipython.readthedocs.io/en/stable/) to try this code interactively, as it supports executing `async`/`await` expressions in the console.
+
+## API Differences
+
+If you're using streaming responses then there are a few bits of API that
```
florimondmanca:
```suggestion
If you're using an async client then there are a few bits of API that
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+for writing concurrent code with the async/await syntax.
+
+```python
+import asyncio
+import httpx
+
+async def main():
+    client = httpx.AsyncClient()
+    response = await client.get('https://www.example.com/')
+    print(response)
```
florimondmanca: Shouldn't we be encouraging the `async with` syntax here?

I'm not even sure why this runs fine without a "an open stream is being garbage collected" error, since the keepalive connection should still be open and have to be closed off. It does raise a warning:

```console
$ python -Wd debug/example.py
<Response [200 OK]>
sys:1: ResourceWarning: unclosed <socket.socket fd=9, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('2a01:cb04:49:9800:4128:7fd2:96fd:60d0', 50895, 0, 0), raddr=('2606:2800:220:1:248:1893:25c8:1946', 443, 0, 0)>
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/Users/florimond/.pyenv/versions/3.8.0/lib/python3.8/asyncio/sslproto.py:321: ResourceWarning: unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x103656160>
  _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/Users/florimond/.pyenv/versions/3.8.0/lib/python3.8/asyncio/selector_events.py:684: ResourceWarning: unclosed transport <_SelectorSocketTransport fd=9>
  _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

(*We* may know that there are *some* cases when it just happens to be fine to not close the client — in this case it's not — but I really think we should encourage users to always ensure the client gets closed at some point.)

docs/async.md
```diff
@@ -0,0 +1,210 @@
...
+dispatch = httpx.dispatch.ASGIDispatch(app=app, client=("1.2.3.4", 123))
+async with httpx.AsyncClient(dispatch=dispatch) as client:
+    ...
+```
+
+See [the ASGI documentation](https://asgi.readthedocs.io/en/latest/specs/www.html#connection-scope) for more details on the `client` and `root_path` keys.
+
+## Unix Domain Sockets
+
+The async client provides support for connect through a unix domain socket via the `uds` parameter. This is useful when making requests to a server that is bound to a socket file rather than an IP address.
```
florimondmanca:
```suggestion
The async client provides support for connecting through a unix domain socket via the `uds` parameter. This is useful when making requests to a server that is bound to a socket file rather than an IP address.
```

docs/async.md
```diff
@@ -0,0 +1,210 @@
+# Async Client
```
florimondmanca: I think I like the title proposed in the navigation better… Matches `HTTP/2 Support` as well:

```suggestion
# Async Support
```

docs/quickstart.md
```diff
@@ -329,17 +323,17 @@ HTTPX will use universal line endings, normalising all cases to `\n`.
...
->>> async with httpx.stream("GET", "https://www.example.com") as r:
-...     async for chunk in r.aiter_raw():
+>>> with httpx.stream("GET", "https://www.example.com") as r:
+...     for chunk in r.iter_raw():
 ...         print(chunk)
 ```
 
 If you're using streaming responses in any of these ways then the `response.content` and `response.text` attributes will not be available, and will raise errors if accessed. However you can also use the response streaming functionality to conditionally load the response body:
 
 ```
```
florimondmanca: These need a language for syntax highlighting — we can submit a separate PR though.

```suggestion
```python
```

docs/async.md
tomchristie: Good call, yup!

docs/advanced.md
tomchristie: Good catch, yes. Didn't notice the `.dispatch.` that crept in.

docs/async.md
tomchristie: Thanks, yes.

docs/async.md
tomchristie: Yeah good call.

docs/api.md
```diff
@@ -32,11 +32,11 @@
...
     :docstring:
 
-## `AsyncClient`
+## `Client`
 
-::: httpx.AsyncClient
+::: httpx.Client
     :docstring:
-    :members: headers cookies params request get head options post put patch delete build_request send aclose
+    :members: headers cookies params request get head options post put patch delete build_request send
```
florimondmanca: Don't we want to document `.close()` here?

florimondmanca:
@tomchristie Also there are three pending comments that GitHub shows under "3 more items hidden", have you seen those?

tomchristie:
Thanks @florimondmanca! Nope, I'd not noticed those other suggestions.

The `Client.close()` API docs reference was waiting for #735 which is now in.

docs/api.md
tomchristie: Thanks yup, now in.

tomchristie:
Fantastic, thanks! ✨

tomchristie:
Ah, I shoulda actually held off on merging this, since it updates the repo README.

PR #730, opened 2020-01-07
Drop 'cert' / 'verify' / 'trust_env' / 'stream' arguments.

tomchristie:
* Drop the per-request usage of `cert`, `verify`, `trust_env`.
* Drop the `stream=True` argument in favour of our streaming API.


florimondmanca:
A note that these arguments were already raising errors as of 0.10, so this doesn't necessarily register as "breaking".

Issue #731, opened 2020-01-07
Public API for auth flow.


tomchristie:
The `authlib` package provides support for oauth integration with `httpx` - they're currently reporting a breakage with the latest versions, which isn't particularly surprising since they were having to rely on private API (eg. Middleware)... https://github.com/lepture/authlib/issues/178

Really we need to make sure that we introduce sufficient public API for them to be able to implement against.

Our public Auth API currently accepts...

* A two-tuple of username/password.
* Either `httpx.BasicAuth` or `httpx.DigestAuth`.
* A callable taking a request, and returning a response.

In order to support their use-case we should also make the custom auth class into public API for the 0.11 release, so that we're also officially supporting:

* A subclass of `httpx.Auth`.

Also, our auth class is now sans-IO, so that it can support both the `Client` and the `AsyncClient` cases. However, some kinds of auth need to be able to sign the request body, so we need some API in order to support that.

I'd suggest the following API for custom auth classes:

* Subclass `httpx.Auth`.
* Implement `def auth_flow(request: Request) -> typing.Generator[Request, Response, None]`. (We're currently using `__call__`, but I think we may as well us a named method instead
* Optionally include `.requires_request_body = True`. (If specified then the client should read the request body, making it available as `request.content`. The client should error if the request body is not replayable.)

We'll also need to implement `.content` and `.read()`/`.aread()` on request instances.

PR #732, opened 2020-01-07
Public Auth API

tomchristie:
Refs #731

Starting this PR with the proposed documentation. Implementation will require:

- [x] Switching the auth method name to `.auth_flow`.
- [x] Implementing a `Request.aread()` method and `Request.content` property.
- [x] Implementing `.requires_request_body` functionality on auth flows.
- [x] Expose `httpx.Auth` as public API.
- [x] Drop `AuthFlow` in favor of explicit type, in order to help guide users away from importing bits of `httpx` that are outside public API.


florimondmanca:
Cc'ing @lepture relative to https://github.com/lepture/authlib/issues/178#issuecomment-571153749 — from the docs proposed here, would this fit the use case for `authlib`?

tests/models/test_requests.py
```diff
@@ -21,19 +21,38 @@ def test_content_length_header():
...
 
     assert request.headers["Content-Type"] == "application/json"
-    assert content == b'{"test": 123}'
+    assert request.content == b'{"test": 123}'
+
+
+@pytest.mark.asyncio
+async def test_read_and_stream_data():
+    # Ensure a request may still be streamed if it has been read.
+    #  Needed for cases such as authentication classes that read the request body.
```
florimondmanca: Nit:

```suggestion
    # Needed for cases such as authentication classes that read the request body.
```

Same on the comment in the other test case below.

tomchristie:
> I'm really happy with how the auth API turned out

I think you made a great call on the generator approach. 👍

It's also kinda nice that we've got a `callable[request] -> request` available for the simple cases, and a more powerful custom class + `.auth_flow` for multi-request flows.

lepture:
yes, that would work.

PR #734, opened 2020-01-07
Fixed redirect loop

yuriiohorodnik:
Redirect loop works wrong because it tries to get the same address with another HTTP method checking only for URL but not for request method that causes error while starting a new HTTPS connection. 

Redirection from:
POST /url
GET /url
Cause Redirect Loop error.
This code fixes this behavior.


tomchristie:
Great catch, yup!
You'll need to fix the linter errors with `./scripts/lint`.

yuriiohorodnik:
@tomchristie  Thank you for blazing-fast reaction and great software.
CI fixed.

tomchristie:
Fab, thanks.

yuriiohorodnik:
Thank you for such a great library and fabulous work

вт, 7 янв. 2020 г. в 18:15, Tom Christie <notifications@github.com>:

PR #735, opened 2020-01-07
Reintroduce sync API.

### LINKED ISSUE #572: Supporting Sync. Done right.

tomchristie:
Closes #572

- [x] Added `Client`.
- [x] Top-level API becomes sync.
- [x] Added `URLLib3Dispatcher`.
- [x] Wire up timeouts.
- [x] Wire up pool limits.
- [x] Catch and wrap urllib3 networking exceptions.
- [x] Add WSGIDispatcher.
- [x] Add urllib3 proxy dispatcher.

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -610,7 +610,8 @@ class Client(BaseClient):
         while True:
             if len(history) > self.max_redirects:
                 raise TooManyRedirects()
-            if request.url in (response.url for response in history):
+            urls = ((resp.request.method, resp.url) for resp in history)
+            if (request.method, request.url) in urls:
                 raise RedirectLoop()
 
             response = self.send_handling_auth(
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -884,8 +884,8 @@ class AsyncClient(BaseClient):
     Usage:
 
     ```python
-    >>> client = httpx.AsyncClient()
-    >>> response = client.get('https://example.org')
+    >>> async with httpx.AsyncClient() as client:
+    >>>     response = await client.get('https://example.org')
     ```
 
     **Parameters:**
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -22,9 +22,10 @@ from .config import (
 )
 from .content_streams import ContentStream
 from .dispatch.asgi import ASGIDispatch
-from .dispatch.base import AsyncDispatcher, Dispatcher
+from .dispatch.base import AsyncDispatcher, SyncDispatcher
 from .dispatch.connection_pool import ConnectionPool
 from .dispatch.proxy_http import HTTPProxy
+from .dispatch.urllib3 import URLLib3Dispatcher
 from .exceptions import (
     HTTPError,
     InvalidURL,
@@ -423,9 +424,6 @@ class Client(BaseClient):
     over the network.
     * **app** - *(optional)* An ASGI application to send requests to,
     rather than sending actual network requests.
-    * **backend** - *(optional)* A concurrency backend to use when issuing
-    async requests. Either 'auto', 'asyncio', 'trio', or a `ConcurrencyBackend`
-    instance. Defaults to 'auto', for autodetection.
     * **trust_env** - *(optional)* Enables or disables usage of environment
     variables for configuration.
     """
@@ -444,9 +442,8 @@ class Client(BaseClient):
         pool_limits: PoolLimits = DEFAULT_POOL_LIMITS,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
         base_url: URLTypes = None,
-        dispatch: AsyncDispatcher = None,
+        dispatch: SyncDispatcher = None,
         app: typing.Callable = None,
-        backend: typing.Union[str, ConcurrencyBackend] = "auto",
         trust_env: bool = True,
     ):
         super().__init__(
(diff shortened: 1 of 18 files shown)


tomchristie:
Figure we can probably mark this as ready for review here.

We probably still want to do a bit of work on wrapping exception classes, but that could be a follow-on PR.

* The pool limits don't quite match up to what `urllib3` uses, but it'll do.
* `write_timeout` isn't supported.
* I don't think we have any integration tests for proxy support, so I've not verified the proxy case properly yet.
* Not sure that `urllib3` supports the `CONNECT_ONLY`/`TUNNEL_ONLY` modes.

httpx/dispatch/wsgi.py
```diff
@@ -0,0 +1,156 @@
...
+        return Response(
+            status_code=int(seen_status.split()[0]),
+            http_version="HTTP/1.1",
+            headers=seen_response_headers,
+            stream=IteratorStream(chunk for chunk in result),
+            request=request,
+        )
+
+
+class BodyStream(io.RawIOBase):
```
florimondmanca: Can we call this `WSGIInput`, so that it doesn't clash with our own concept of "streams"?

I'm also a bit surprised by this class. Does WSGI necessarily require such an object to be passed? Can't an `io.BytesIO` do the trick?

httpx/client.py
```diff
@@ -437,88 +299,743 @@ def build_redirect_request(self, request: Request, response: Response) -> Reques
...
-        if url.is_relative_url:
-            url = request.url.join(url)
+    Usage:
 
-        # Attach previous fragment if needed (RFC 7231 7.1.2)
-        if request.url.fragment and not url.fragment:
-            url = url.copy_with(fragment=request.url.fragment)
+    ```python
+    >>> client = httpx.AsyncClient()
+    >>> response = client.get('https://example.org')
```
florimondmanca: Encourage block usage? Also the `await` keyword is missing:

```python
>>> async with httpx.AsyncClient():
...     response = await client.get("https://example.org")
```

httpx/dispatch/wsgi.py
tomchristie: If we read the request body rather than streaming it, then yes, `io.BytesIO` would work.

tomchristie: Righty, I've opted for a simpler implementation, with the constraint that the request body is no longer streamed. We can always update that at a later point if we choose, but let's keep it simple if possible.

httpx/client.py
tomchristie: Thanks!

tomchristie:
> Though I can't wait for us unasync'ing all this duplicated code in the client, maybe someday? 🤣

I think that one is a solid *maybe*, and could go either way. 😃

The only *non-trivial* chunk of stuff that unasync would deduplicate is https://github.com/encode/httpx/blob/sync/httpx/client.py#L1040-L1216 which is a fairly modest amount for us to maintain, so I could equally well see an `unasync` approach only ever applied to the dispatch layer. But open to either.

florimondmanca:
In any case I think we'll start with the dispatch layer, if/when we switch away from urllib3. :-)

PR #737, opened 2020-01-08
Version 0.11.0

### LINKED ISSUE #724: Version 0.11.0
Checklist for our 0.11.0 release.

This one is a big deal, since it reintroduces the sync client, and is essentially a 1.0 pre-release in terms of how the API looks.

```python
>>> import httpx
<Response [200 OK]>
```

🎉✨ **TA-DA!** ✨🎉

I'd expect that we'll likely end up waiting for a period of time after this release, and then end up releasing a 1.0 with either no API changes, or only very minimal API changes. (The only remaining area I can see us still wanting to refine/change, would be some review making sure we've got an exception heirarchy/naming that we're entirely happy to stick with for 1.0 onwards)

- [x] Reintroduce `Client` as a sync client. #735
- [x] Reintroduce `WSGIDispatch`. #735
- [x] Top-level API becomes sync, not async. #735
- [x] Drop `Origin` from public API. #688
- [x] Use `httpx.Proxy()` for proxy configuration, not the `httpx.HTTPProxy` dispatch class. #713
- [ ] ~Consider switching `client.params`, `client.headers`, `client.cookies` so that they don't have a setter/getter mismatch.~ Refs #678 #274
- [ ] ~Consider dropping UDS support.~ #723
- [x] Wrap IO Exceptions in httpx exceptions. #707
- [x] Docs #727
- [x] `httpx.Auth` becomes public API. #732 #731

tomchristie:
Closes #724

Also requires the following, before merge:

- [x] #727 
- [x] #735

I've currently listed the release date here as tomorrow (Thursday 9th January). I don't *think* we've got any blockers left at this point, @florimondmanca?

## 0.11.0 (January 9th, 2019)

The 0.11 release reintroduces our sync support, so that `httpx` now supports both a standard thread-concurrency API, and an async API.

Existing async `httpx` users that are upgrading to 0.11 should ensure that:

* Async codebases should always use a client instance to make requests.
* The async client is named as `httpx.AsyncClient()`.
* When instantiating proxy configurations use the `httpx.Proxy()` class, instead of the previous `httpx.HTTPProxy()`. This new configuration class works for configuring both sync and async clients.

We believe the API is now pretty much stable, and are aiming for a 1.0 release sometime on or before April 2020.

### Changed

- Top level API such as `httpx.get(url, ...)`, `httpx.post(url, ...)`, `httpx.request(method, url, ...)` becomes synchronous.
- Added `httpx.Client()` for synchronous clients, with `httpx.AsyncClient` being used for async clients.
- Switched to `proxies=httpx.Proxy(...)` for proxy configuration.
- Network connection errors are wrapped in `httpx.NetworkError`, rather than exposing lower-level exception types directly.

### Removed

- The `request.origin` property and `httpx.Origin` class are no longer available.
- The per-request `cert`, `verify`, and `trust_env` arguments are escalated from raising errors if used, to no longer being available. These arguments should be used on a per-client instance instead, or in the top-level API.
- The `stream` argument has escalated from raising an error when used, to no longer being available. Use the `client.stream(...)` or `httpx.stream()` streaming API instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,31 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.11.0 (January 9th, 2019)
+
+The 0.11 release reintroduces our sync support, so that `httpx` now supports both a standard thread-concurrency API, and an async API.
+
+Existing async `httpx` users that are upgrading to 0.11 should ensure that:
+
+* Async codebases should always use a client instance to make requests.
+* The async client is named as `httpx.AsyncClient()`.
+* When instantiating proxy configurations use the `httpx.Proxy()` class, instead of the previous `httpx.HTTPProxy()`. This new configuration class works for configuring both sync and async clients.
+
+We believe the API is now pretty much stable, and are aiming for a 1.0 release sometime on or before April 2020.
+
+### Changed
+
+- Top level API such as `httpx.get(url, ...)`, `httpx.post(url, ...)`, `httpx.request(method, url, ...)` becomes synchronous.
+- Added `httpx.Client()` for synchronous clients, with `httpx.AsyncClient` being used for async clients.
+- Switched to `proxies=httpx.Proxy(...)` for proxy configuration.
+- Network connection errors are wrapped in `httpx.NetworkError`, rather than exposing lower-level exception types directly.
+
+### Removed
+
+- The `request.origin` property and `httpx.Origin` class are no longer available.
+- The per-request `cert`, `verify`, and `trust_env` arguments are escalated from raising errors if used, to no longer being available. These arguments should be used on a per-client instance instead, or in the top-level API.
+- The `stream` argument has escalated from raising an error when used, to no longer being available. Use the `client.stream(...)` or `httpx.stream()` streaming API instead.
+
 ## 0.10.1 (December 31st, 2019)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.11.0 (January 9th, 2019)
+## 0.11.0 (January 8th, 2019)
 
 The 0.11 release reintroduces our sync support, so that `httpx` now supports both a standard thread-concurrency API, and an async API.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ We believe the API is now pretty much stable, and are aiming for a 1.0 release s
 
 ### Removed
 
-- The `request.origin` property and `httpx.Origin` class are no longer available.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,31 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.11.0 (January 9th, 2019)
+
+The 0.11 release reintroduces our sync support, so that `httpx` now supports both a standard thread-concurrency API, and an async API.
+
+Existing async `httpx` users that are upgrading to 0.11 should ensure that:
+
+* Async codebases should always use a client instance to make requests.
```
florimondmanca:
```suggestion
* Async codebases should always use a client instance to make requests, instead of the top-level API.
```

CHANGELOG.md
```diff
@@ -4,6 +4,31 @@ All notable changes to this project will be documented in this file.
...
+* The async client is named as `httpx.AsyncClient()`.
```
florimondmanca:
```suggestion
* The async client is named as `httpx.AsyncClient()`, instead of `httpx.Client()`.
```

CHANGELOG.md
```diff
@@ -4,6 +4,31 @@ All notable changes to this project will be documented in this file.
...
+### Changed
+
+- Top level API such as `httpx.get(url, ...)`, `httpx.post(url, ...)`, `httpx.request(method, url, ...)` becomes synchronous.
+- Added `httpx.Client()` for synchronous clients, with `httpx.AsyncClient` being used for async clients.
+- Switched to `proxies=httpx.Proxy(...)` for proxy configuration.
+- Network connection errors are wrapped in `httpx.NetworkError`, rather than exposing lower-level exception types directly.
+
+### Removed
+
+- The `request.origin` property and `httpx.Origin` class are no longer available.
```
florimondmanca: Wasn't `.origin` on the `URL`?

```suggestion
- The `request.url.origin` property and `httpx.Origin` class are no longer available.
```

tomchristie:
> Also, we should probably link to this changelog entry in the README intro and in the docs home page afterwards? :-)

Yup good call

tomchristie:
> Also, we should probably link to this changelog entry in the README intro and in the docs home page afterwards? :-)

Alternately, we move it into the docs.

tomchristie:
Updated the release date to today.
Reckon once this one get approved we're ready to merge & release. 😅

florimondmanca:
> Alternately, we move it into the docs.

Yes, I could see the changelog be on the docs site, with a thin shell `CHANGELOG.md` file at the root of the repo, much like for `CONTRIBUTING.md`.

StephenBrown2:
Or even perhaps a symlink one way or the other?

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.11.0 (January 9th, 2019)
```
dalf: I guess it should be 2020

tomchristie: 🤣

florimondmanca: Right! 🤪

PR #742, opened 2020-01-08
Unasync'ed base dispatchers

florimondmanca:
Refs #722 

First *smoll* step towards unasync'ing HTTPX…: unasync `AsyncDispatcher` into `SyncDispatcher`. Allows us to setup the tooling and avoid impacting any public piece of API.

Mostly done, but marking as a draft as this is probably only worth considering once 0.11 is out…


florimondmanca:
Just tried it out locally… the build works :+1:

```console
$ pip install "git+https://github.com/encode/httpx.git@unasynced-base-dispatchers"
$ python
Python 3.8.0 (default, Oct 16 2019, 18:21:21)
[Clang 11.0.0 (clang-1100.0.33.8)] on darwin
>>> import httpx
>>> r = httpx.get('https://www.google.com')
>>> httpx._sync.dispatch.base.SyncDispatcher
<class 'httpx._sync.dispatch.base.SyncDispatcher'>
>>> httpx.dispatch.base.SyncDispatcher
<class 'httpx._sync.dispatch.base.SyncDispatcher'>
>>>
```

requirements.txt
```diff
@@ -21,7 +21,9 @@ pytest-trio
 pytest-cov
 trio
 trustme
+unasync==0.4.*
```
florimondmanca: That's a bit funny, but listing `unasync` in the `build-system.requires` entry doesn't *actually* install it in a way that's available to the Python interpreter. I don't know when/where those dependencies are installed exactly — probably managed by pip…

florimondmanca:
Ah yes, the build will fail until https://github.com/python-trio/unasync/commit/2c2c9b7fdfd3aeb7cb44d4c1ac6f203500e5efae gets released. 😄

florimondmanca:
Closing, we can revisit if/when moving forward on this in `httpcore`…

PR #749, opened 2020-01-10
Add links to requests references

SiggyF:
Update README.md with links to requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Let's get started...
 
 ## Features
 
-HTTPX builds on the well-established usability of `requests`, and gives you:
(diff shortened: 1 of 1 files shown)


tomchristie:
I'd also be okay with us having all these link references so long as we're linking out to the other packages listed on the page, too. (It'd seem a bit odd to hyperlink to requests, but not to urllib3 in particular.)

florimondmanca:
Yes, that would sound good to me too. 👍

tomchristie:
Closing for now due to inactivity. We'd happily a PR here that interlinks to *all* other mentioned projects.

PR #750, opened 2020-01-11
Add an async example to intro snippets

florimondmanca:
A way to ensure newcomers are immediately exposed to the built-in async support provided by HTTPX. :-)


florimondmanca:
> Its a bit awkward to use the context managed style in the REPL, right?

Awkward but necessary…

```python
>>> import httpx
>>> client = httpx.AsyncClient()
>>> r = await client.get('https://www.example.org')
>>> exit()
An open stream object is being garbage collected; call "stream.close()" explicitly.
```

PR #753, opened 2020-01-11
Add docs and repo to project_urls metadata

hugovk:
For programmatic access to metadata on PyPI.

Also add the "Python 3 Only" Trove classifier, which can also be useful for programmatic use.


StephenBrown2:
Definitely a good idea to implement across all the encode projects.

tomchristie:
Lovely, yup!

PR #758, opened 2020-01-13
Support both zlib and deflate encodings

### LINKED ISSUE #756: DecodingError with zlib-compressed responses
Using the following minimal WSGI app:

```python
import zlib

def app(environ, start_response):
    start_response("200 OK", [("Content-Encoding", "deflate")])
    return [zlib.compress(b"hello world")]
```

This works fine in a web browser.

Requests is quite happy:

```
>>> import requests
>>> requests.get("http://localhost:9999")
<Response [200]>
>>> requests.get("http://localhost:9999").text
'hello world'
```

httpx says:

```
>>> httpx.get('http://localhost:9999')
  File "/Users/jamie/code/httpx/httpx/decoders.py", line 52, in decode
    return self.decompressor.decompress(data)
zlib.error: Error -3 while decompressing data: invalid stored block lengths

  File "/Users/jamie/code/httpx/httpx/api.py", line 170, in get
    trust_env=trust_env,
  File "/Users/jamie/code/httpx/httpx/api.py", line 96, in request
  File "/Users/jamie/code/httpx/httpx/client.py", line 568, in request
  File "/Users/jamie/code/httpx/httpx/client.py", line 593, in send
    response.read()
  File "/Users/jamie/code/httpx/httpx/models.py", line 900, in read
    self._content = b"".join([part for part in self.iter_bytes()])
  File "/Users/jamie/code/httpx/httpx/models.py", line 900, in <listcomp>
    self._content = b"".join([part for part in self.iter_bytes()])
  File "/Users/jamie/code/httpx/httpx/models.py", line 912, in iter_bytes
    yield self.decoder.decode(chunk)
  File "/Users/jamie/code/httpx/httpx/decoders.py", line 54, in decode
    raise DecodingError from exc
httpx.exceptions.DecodingError
```

The reason this is an issue and not a PR is because I don't understand the code in `decoders.py` - it very explicitly says `self.decompressor = zlib.decompressobj(-zlib.MAX_WBITS)`, which according to [the Python docs](https://docs.python.org/3/library/zlib.html#zlib.compressobj) means "Uses the absolute value of wbits as the window size logarithm, while producing a raw output stream with no header or trailing checksum." I don't know enough about zlib to understand this, but it looks very deliberate so I thought I'd better ask.

tomchristie:
Closes #756 

`DeflateDecoder` now attempt both "deflate" decoding and "zlib" decoding.

### DIFF
--- a/httpx/decoders.py
+++ b/httpx/decoders.py
@@ -45,12 +45,18 @@ class DeflateDecoder(Decoder):
     """
 
     def __init__(self) -> None:
-        self.decompressor = zlib.decompressobj(-zlib.MAX_WBITS)
+        self.first_attempt = True
+        self.decompressor = zlib.decompressobj()
 
     def decode(self, data: bytes) -> bytes:
+        was_first_attempt = self.first_attempt
+        self.first_attempt = False
         try:
             return self.decompressor.decompress(data)
         except zlib.error as exc:
+            if was_first_attempt:
+                self.decompressor = zlib.decompressobj(-zlib.MAX_WBITS)
+                return self.decode(data)
             raise DecodingError from exc
 
     def flush(self) -> bytes:
--- a/tests/test_decoders.py
+++ b/tests/test_decoders.py
@@ -18,8 +18,14 @@ REQUEST = httpx.Request("GET", "https://example.org")
 
 
 def test_deflate():
+    """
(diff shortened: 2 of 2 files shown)


j4mie:
Can confirm that this branch fixes my issue.

Issue #538, opened 2019-11-18
Key headers are lowercased before sent ?


manatlan:
Is it a normal behavior ?
From RFC2616, headers are case-insensitive ... but, why httpx/httpcore always lowercase them ?

It could be problematic for server/proxy/tools that don't respect rfc2616, and want case sensitive headers ... no ?

I understand the lowercasing for http/2. But for http/1 ... no ;-(

tomchristie:
> It could be problematic for server/proxy/tools that don't respect rfc2616

It's feasible that we *could* alter this behavior if we'd determined that in real-world usage it was problematic. Have you observed an issue with some specific tooling related to this?

Any idea what urllib3 / requests behavior is wrt. preserving header casing?

manatlan:
Sadly, I don't know any real-world usage with that problem, but I'm pretty sure : that should exist ;-(

At my work, in our intranet : I know (at least) one ...

BTW, I edit a [tool](https://reqman-docs.glitch.me/), based on httpx (httpcore) ... and I will need to find another http async lib ;-( ... which send headers1.X as is ;-(

manatlan:
btw, **requests** seems to do it

```
r=requests.get("https://httpbin.org/get",headers={"X-Hello":"yo"})
print(r.content)
```

manatlan:
BTW, **requests** seems to be ok
```
import asyncio
import httpx
import requests

asynk=httpx.AsyncClient()

url="http://headers.jsontest.com/"
myheaders={"idSPECIAL":"yo"}

r=requests.get(url,headers=myheaders)
print(r.content.decode())


async def test():
    r=await asynk.get(url,headers=myheaders)
    print(r.content.decode())

asyncio.run( test() )
```
will display : 

```
{
   "X-Cloud-Trace-Context": "e9c4e43c5980315155c2537d6e594d83/11304935074059934945",
   "Accept": "*/*",
   "idSPECIAL": "yo",
   "User-Agent": "python-requests/2.22.0",
   "Host": "headers.jsontest.com"
}

{
   "X-Cloud-Trace-Context": "c6119f95ab6f5e65fae2a10b5dcbea84/7193114510141344900",
   "host": "headers.jsontest.com",
   "idspecial": "yo",
   "user-agent": "python-httpx/0.7.7",
   "accept": "*/*"
}
```

manatlan:
[...]

PR #760, opened 2020-01-13
Preserve header casing

tomchristie:
Refs #538

This pull request would change `Headers` to be a case-preserving but case-insenstive mutli dict.

This would resolve HTTP header casing in the sync case, although not for the async case. (Since `h11` currently lowercase-normalizes headers in both directions)

### DIFF
--- a/httpx/auth.py
+++ b/httpx/auth.py
@@ -108,12 +108,12 @@ class DigestAuth(Auth):
             raise RequestBodyUnavailable("Request body is no longer available.")
         response = yield request
 
-        if response.status_code != 401 or "www-authenticate" not in response.headers:
+        if response.status_code != 401 or "WWW-Authenticate" not in response.headers:
             # If the response is not a 401 WWW-Authenticate, then we don't
             # need to build an authenticated request.
             return
 
-        header = response.headers["www-authenticate"]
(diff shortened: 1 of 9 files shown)


florimondmanca:
I merged `master` back in here just in case a week's worth of changes might break anything, though I don't think it will…

tomchristie:
So it occured to me that we ought to ensure that `.keys()` returned lowercased variants.

What's nice about that is we still have full header casing available in the `repr`, and available if users *really* want to see it in the raw header bytes, but otherwise we're keeping things properly constrained and exposing a strictly lowercase style.

So...

* `response.headers[X]` - Case insensitive lookups.
* `dict(response.headers)` - Returns lowercased variants, ensuring users can't easily write accidentally case-sensitive code.
* `repr(response.headers)` - Displays actual case variants.
* `response.headers.raw` - Access the list of byte tuples, in raw actual case.

tomchristie:
Might take another pass onto this at some point.

valiant1x:
Is it intended that async usage of http1.1 is still lowercasing all headers? In Python 3.8.2 with httpx 0.12.1 this is the behavior I see. How can we preserve proper header casing in HTTP/1.1 with `async`? @tomchristie @florimondmanca

PR #766, opened 2020-01-15
Fix for streaming a redirect response body with allow_redirects=False

tomchristie:
When making a request with `.stream(... allow_redirects=False)` any redirect response that is returned should be streamable, and not have already been read.

```python
>>> import httpx
>>> with httpx.stream("GET", "https://daringfireball.net/css/ie_sucks.php", allow_redirects=False) as response:
...     body = b''.join([chunk for chunk in response.iter_raw()])
... 
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in <listcomp>
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/httpx/models.py", line 939, in iter_raw
    raise StreamConsumed()
httpx.exceptions.StreamConsumed
```

### DIFF
--- a/httpx/client.py
+++ b/httpx/client.py
@@ -622,7 +622,8 @@ class Client(BaseClient):
             if not response.is_redirect:
                 return response
 
-            response.read()
+            if allow_redirects:
+                response.read()
             request = self.build_redirect_request(request, response)
             history = history + [response]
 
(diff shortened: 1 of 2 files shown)


tomchristie:
In the example given `https://daringfireball.net/css/ie_sucks.php` returns a 301 with an HTML body containing the redirect link.

Within the streaming block the response has already been read to completion, so attempting the read the raw bytes raises `StreamConsumed`.

Issue #779, opened 2020-01-18
When using httpx.Client with proxies, authentication fails


fbexiga:
When doing something like this

```
proxy = {
     'http' : 'http://user:pass@X.X.X.X:YY',
     'https' : 'http://user:pass@X.X.X.X:YY'
}

def run_sync(url):
    with httpx.Client(verify=False, proxies=proxy) as client:
        response = client.get(url)
        print(response)
```

```
>>> run_sync('http://www.google.com')
<Response [407 Proxy Authentication Required]>
```

For a private proxy that includes authentication, the request will fail with a "Proxy Authentication Required" error. I think httpx is doing something like this:

```
>>> from urllib3 import ProxyManager
>>> proxy = urllib3.ProxyManager('http://user:pass@x.x.x.x:yy')
>>> x = proxy.request('GET', 'http://google.com/')
>>> x.status
407
```

instead of this

```
>>> from urllib3 import ProxyManager, make_headers
>>> default_headers = make_headers(proxy_basic_auth='user:pass')
>>> proxy = urllib3.ProxyManager('http://x.x.x.x:yy', headers=default_headers)
>>> x = proxy.request('GET', 'http://google.com/')
>>> x.status
200
```

PR #780, opened 2020-01-19
Detect credentials in proxy URLs and create Proxy-authorization header

yeraydiazdiaz:
Fixes https://github.com/encode/httpx/issues/779

Use the [same logic from `proxy_http`](https://github.com/encode/httpx/blob/master/httpx/dispatch/proxy_http.py#L83-L96) in the `Proxy` config, e.g. detects if credentials are present in the proxy URL and adds the authorization header accordingly. Also remove the credentials for the URL.

### DIFF
--- a/httpx/config.py
+++ b/httpx/config.py
@@ -333,8 +333,7 @@ class Proxy:
             )
             # Remove userinfo from the URL authority, e.g.:
             # 'username:password@proxy_host:proxy_port' -> 'proxy_host:proxy_port'
-            credentials, _, authority = url.authority.rpartition("@")
-            url = url.copy_with(authority=authority)
+            url = url.copy_with(username=None, password=None)
 
         self.url = url
         self.headers = headers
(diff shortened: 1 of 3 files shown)


httpx/config.py
```diff
@@ -325,10 +326,25 @@ def __init__(
...
 
+        if url.username or url.password:
+            headers.setdefault(
+                "Proxy-Authorization",
+                self.build_auth_header(url.username, url.password),
+            )
+            # Remove userinfo from the URL authority, e.g.:
+            # 'username:password@proxy_host:proxy_port' -> 'proxy_host:proxy_port'
+            credentials, _, authority = url.authority.rpartition("@")
+            url = url.copy_with(authority=authority)
```
tomchristie: Perhaps replace these two lines with `ul = url.copy_with(username=None, password=None)`?

yeraydiazdiaz: That's definitely cleaner, good catch :+1:

The `copy_with` seems to add the default port to the URL though which forced me to change the tests a bit, but I think that's fine.

yeraydiazdiaz:
Awesome, thanks!

Issue #782, opened 2020-01-20
Drop `backend=...` argument.


tomchristie:
Now that we have async autodetection, I think we ought to consider dropping the `backend` argument from `AsyncClient.__init__(...)`, since it's not really required.

If anything it'd be better to find out if/when there are any cases that `sniffio` isn't able to correctly detect, and fix them, rather than having users be able to flick an override switch.

(It'd also potentially make it easier to have a clear split if we wanted to factor the dispatch out into an entirely independant package at some point, since we'd want the backend implementations to reside in the core networking package, rather than in `httpx`.)

dalf:
Note: I use a custom Backend to record the certificates of the HTTPS requetsts.
This backend behaves a proxy to a real backend, but record the ssl.SSLObject

Code:
* https://github.com/dalf/searx-stats2/blob/master/searxstats/common/ssl_info.py#L85
* https://github.com/dalf/searx-stats2/blob/master/searxstats/common/http.py#L70

tomchristie:
@dalf - That's a really useful (surprising) bit of feedback, thanks!

(And yes, you've got the right approach there, but make sure you're pinning your dependencies tightly if you're relying on private API!)

Issue #768, opened 2020-01-16
Transport API


tomchristie:
**What is the dispatcher API, and why is it useful?...**

The dispatcher is the part of the system that's responsible for actually sending the request and returning a response to the client. Having an API to override the dispatcher used by the client allows you to take complete control over that process, and switch out the default behaviour with something else.

For instance the "plug in to a WSGI app" and "plug in to an ASGI app" cases that we support are implemented as dispatcher classes, that send the request to a Python web application, rather than sending a network request.

**Currently the dispatchers are considered private API, but really we'd like a public API here at some point. This issue is for discussing what we think that API should be.**

Right now the API is...

```python
 def send(request: Request, timeout: Timeout = None) -> Response:
    ...

def close():
    ...
```

With`async` equivelents for the async case.

That's actually pretty okay, but if we're looking at a public API, we ought to figure out if it's actually what we want to expose. There's a case to be made for making the dispatcher API just take plain primitive datastructures, rather than dealing with `Request`/`Response` models. That'd make for a more portable interface, that could be used in other packages without importing `httpx`.

That might look *something* like this...

```python
def send(self, method: bytes, url: Typle[bytes, bytes, int, bytes], headers: List[Tuple[bytes, bytes]], body: ClosableIterator[bytes] = None, timeout: Timeout = None):
    ...
    return (status_code: int, http_version: bytes, reason_phrase: bytes, headers: List[Tuple[bytes, bytes]], body: ClosableIterator[bytes])

def close():
    ...
```

Both the request and response bodies would be byte-iterable, closable interfaces (or byte async-iterable, closable interfaces). Everything else would be plain ol' Python datastructures.

Thoughts:

* It's feasible that `url` *could* be defined as a tuple or url components instead, so that we can keep pass a *parsed* url objects from the client to a dispatcher interface?  **Update: I've tweaked the interface to describe it that way around, as "scheme", "host", "port", "full_path".
* We *do* need to have the timeout be part of the dispatch interface, since we want to be able to support configured-per-request timeout values - I'd guess we'd need to define that as an optional four-tuple of connect/read/write/pool float timeout values.
* We might want to think about defining expected exception types.
* There's no support for HTTP trailing headers in this interface. Perhaps we don't care about that, or could the `ContentStream.close()` perhaps optionally return a list of headers, maybe?

[...]

PR #784, opened 2020-01-20
Retries for connection failures

florimondmanca:
A simpler version of #778.

* Retries are **off-by-default**.
* They're still only triggered on **connection failures only**, and for **idempotent HTTP verbs only**.
* The only supported usage now is…

```python
client = httpx.Client(retries=<int>)
client = httpx.Client(retries=httpx.Retries(<int>[, backoff_factor=<float>])
```

* That's it!

Still TODO:

* [x] Add `retries=...` to the top-level API.
* [x] Add per-request `retries=...`

Note:

* There are no tests for sync client retries, mostly because they'd be duplicated versions of the async case… but ideally we *should* add these, as not having them makes coverage suffer (because the implementation *is* duplicated).
* There are no tests for using retries on the top-level API, because we can't switch for a mock `dispatcher` there, and adding fail-until-some-number-of-retries on the uvicorn server isn't straight-forward at all…

---

You can try this out with…

```python
# app.py
from starlette.responses import PlainTextResponse

app = PlainTextResponse("Hello, world!")
```

```python
# example.py
import asyncio
import httpx


async def main():
    retries = httpx.Retries(10, backoff_factor=0.5)
    async with httpx.AsyncClient(retries=retries) as client:
        r = await client.get("http://localhost:8000")
        print(r)


asyncio.run(main())
```

1. `$ python example.py`
2. Start the server.
3. HTTPX first fails to get a response, but eventually manages to get one when the server is started.

Example log output:

```console
DEBUG [2020-01-20 14:50:01] httpx.client - HTTP Request failed: NetworkError(OSError("Multiple exceptions: [Errno 61] Connect call failed ('::1', 8000, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 8000)"))
DEBUG [2020-01-20 14:50:01] httpx.client - Retrying in 0 seconds
DEBUG [2020-01-20 14:50:01] httpx.client - HTTP Request failed: NetworkError(OSError("Multiple exceptions: [Errno 61] Connect call failed ('::1', 8000, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 8000)"))
DEBUG [2020-01-20 14:50:01] httpx.client - Retrying in 0.5 seconds
DEBUG [2020-01-20 14:50:01] httpx.client - HTTP Request failed: NetworkError(OSError("Multiple exceptions: [Errno 61] Connect call failed ('::1', 8000, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 8000)"))
DEBUG [2020-01-20 14:50:01] httpx.client - Retrying in 1.0 seconds
DEBUG [2020-01-20 14:50:02] httpx.client - HTTP Request failed: NetworkError(OSError("Multiple exceptions: [Errno 61] Connect call failed ('::1', 8000, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 8000)"))
DEBUG [2020-01-20 14:50:02] httpx.client - Retrying in 2.0 seconds
DEBUG [2020-01-20 14:50:04] httpx.client - HTTP Request: GET http://localhost:8000 "HTTP/1.1 200 OK"
<Response [200 OK]>
```

For a sync client (had to make a few tweaks to our `Urllib3Dispatcher` here):

```python
# example.py
import httpx

retries = httpx.Retries(10, backoff_factor=0.5)
with httpx.Client(retries=retries) as client:
    r = client.get("http://localhost:8000")
    print(r)
```

```console
DEBUG [2020-01-20 14:59:36] httpx.client - HTTP Request failed: NetworkError(NewConnectionError('<urllib3.connection.HTTPConnection object at 0x102186310>: Failed to establish a new connection: [Errno 61] Connection refused'))
DEBUG [2020-01-20 14:59:36] httpx.client - Retrying in 0 seconds
DEBUG [2020-01-20 14:59:36] httpx.client - HTTP Request failed: NetworkError(NewConnectionError('<urllib3.connection.HTTPConnection object at 0x1032163d0>: Failed to establish a new connection: [Errno 61] Connection refused'))
DEBUG [2020-01-20 14:59:36] httpx.client - Retrying in 0.5 seconds
DEBUG [2020-01-20 14:59:37] httpx.client - HTTP Request failed: NetworkError(NewConnectionError('<urllib3.connection.HTTPConnection object at 0x103216990>: Failed to establish a new connection: [Errno 61] Connection refused'))
DEBUG [2020-01-20 14:59:37] httpx.client - Retrying in 1.0 seconds
DEBUG [2020-01-20 14:59:38] httpx.client - HTTP Request failed: NetworkError(NewConnectionError('<urllib3.connection.HTTPConnection object at 0x103216f90>: Failed to establish a new connection: [Errno 61] Connection refused'))
DEBUG [2020-01-20 14:59:38] httpx.client - Retrying in 2.0 seconds
DEBUG [2020-01-20 14:59:40] httpx.client - HTTP Request: GET http://localhost:8000 "HTTP/1.1 200 OK"
<Response [200 OK]>
```


docs/advanced.md
```diff
@@ -471,3 +471,50 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+When retries are enabled, HTTPX will retry sending the request up to the specified number of times. This behavior is restricted to **connection failures only**, i.e.:
+
+* Failures to establish or acquire a connection (`ConnectTimeout`, `PoolTimeout`).
+* Failures to keep the connection open (`NetworkError`).
+
+!!! important
+    HTTPX will **NOT** retry on failures that aren't related to establishing or maintaining connections. This includes in particular:
+
+    * Errors related to data transfer, such as `ReadTimeout` or `ProtocolError`.
+    * HTTP error responses (4xx, 5xx), such as `429 Too Many Requests` or `503 Service Unavailable`.
```
StephenBrown2: Hmm. I feel like retrying on status codes is still valuable for a retry config.

Could you add a doc on how to extend this to handle that, if it won't be available by default? (Those codes in particular I think are the primary use cases in fact)

florimondmanca: > Could you add a doc on how to extend this to handle that, if it won't be available by default?

Well so, the idea behind this PR is to not even provide the ability to extend *just yet*. I agree this can fall in scope for us though, as well as handling `Retry-After`…

florimondmanca: I'd rather we consider adding support for those cases and `Retry-After` as a follow-up task, in any case. :-) Let's get something really simple but API-proof in, first?

StephenBrown2: Yeah, agreed. I personally am looking for Retry support for those cases (well, server errors at least), so I'll be hoping that gets added as a follow-up. Definitely want something fairly bullet-proof to build on though!

docs/advanced.md
```diff
@@ -471,3 +471,56 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+
+* An integer, given as a required positional argument, representing the maximum number of connection failures to retry on.
+* `backoff_factor` (optional), which defines the increase rate of the time to wait between retries. By default this is `0.2`, which corresponds to issuing a new request after `(0s, 0.2s, 0.4s, 0.8s, ...)`. (Note that most connection failures are immediately resolved by retrying, so HTTPX will always issue the initial retry right away.)
+
+```python
+# Retry at most 5 times on connection failures,
+# and issue new requests after `(0s, 0.5s, 1s, 2s, 4s, ...)`.
+retries = httpx.Retries(5, backoff_factor=0.5)
+client = httpx.Client(retries=retries)
+```
```
yeraydiazdiaz: The docs are lovely, well done! 💯

httpx/config.py
```diff
@@ -352,6 +355,73 @@ def __repr__(self) -> str:
...
+            and self.backoff_factor == other.backoff_factor
+        )
+
+    @classmethod
+    def should_retry_on_exception(cls, exc: HTTPError) -> bool:
+        for exc_cls in cls._RETRYABLE_EXCEPTIONS:
+            if isinstance(exc, exc_cls):
+                break
+        else:
+            return False
```
yeraydiazdiaz: I'm always confused by `for..else` constructs, I always have look it up. In this case I feel this might be clearer:

```python
if any(isinstance(exc, exc_class) for exc_cls in cls._RETRYABLE_EXPECTIONS):
    return False
```

httpx/client.py
```diff
@@ -64,6 +69,7 @@ def __init__(
         headers: HeaderTypes = None,
         cookies: CookieTypes = None,
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
+        retries: RetriesTypes = DEFAULT_RETRIES_CONFIG,
```
tomchristie: Could we perhaps drop `DEFAULT_RETRIES_CONFIG` and instead use `retries: RetriesTypes = None` throughout?

httpx/client.py
```diff
@@ -1126,6 +1216,43 @@ def dispatcher_for_url(self, url: URL) -> AsyncDispatcher:
...
+        while True:
+            try:
+                return await self.send_handling_redirects(
+                    request,
+                    auth=auth,
+                    timeout=timeout,
+                    allow_redirects=allow_redirects,
+                )
+            except HTTPError as exc:
+                if not retries.limit or not retries.should_retry_on_exception(exc):
```
tomchristie: Should that be `if not retries_left`?

httpx/client.py
```diff
@@ -1105,17 +1188,24 @@ def dispatcher_for_url(self, url: URL) -> AsyncDispatcher:
...
         auth = self.build_auth(request, auth)
 
-        response = await self.send_handling_redirects(
-            request, auth=auth, timeout=timeout, allow_redirects=allow_redirects,
+        response = await self.send_handling_retries(
+            request,
+            auth=auth,
+            retries=retries,
+            timeout=timeout,
+            allow_redirects=allow_redirects,
```
tomchristie: Something that we *could?* choose to do would be something like...

```python
if retries is not None:
    response = await self.send_handling_retries(...)
else:
    response = await self.send_handling_redirects(...)
```

httpx/config.py
florimondmanca: Fair enough! Although in this case that'd have be `if not any(...)` :-)

httpx/client.py
florimondmanca: Sure yes! Makes sense given that the default behavior is to not retry, and `None` isn't a special value in this case.

httpx/client.py
florimondmanca: Yes, this ties into your comment on skipping the retries stuff altogether when retries are disabled. Addressed in the latest commit by not entering this loop at all when that's the case.

httpx/client.py
florimondmanca: Did this a bit differently because it's easier to have `retries` always be a `Retries` instance (i.e. `httpx.Retries()` means "retries disabled" the same way `httpx.Timeout()` refers to "timeouts disabled"). So I added this switch inside `.send_handling_redirects()` to make the flow clearer when retries are disabled.

florimondmanca:
Thanks for the reviews, @yeraydiazdiaz and @tomchristie! Addressed in the latest commits. Let me know what you think about these. :-)

httpx/config.py
```diff
@@ -352,6 +355,77 @@ def __repr__(self) -> str:
...
+
+        return True
+
+    def get_delays(self) -> typing.Iterator[float]:
+        """
+        Used by clients to determine how long to wait before issuing a new request.
+        """
+        yield 0  # Retry immediately.
+        for n in itertools.count(2):
```
tomchristie: Should this be `itertools.count(self.limit - 1)`?

florimondmanca:
> should we consider using private method names on the class?

Well, we could, but then we'd need to have the client call into those private methods… I suppose not documenting `Retries` as an extensible component (similar to how `Timeout` isn't advertised as extensible) is enough for now?

florimondmanca:
Stuff to consider as follow-ups:

- Add on-by-default support for [`Retry-After`](https://tools.ietf.org/html/rfc7231#section-7.1.3) headers. (We'll definitely need to have some design discussion there, as we probably don't want to let users shoot themselves in the foot by DoS their entire program because HTTPX decided to wait 30s before re-issuing a request — worse even when that's done in a sync context.)

florimondmanca:
@tomchristie I marked #782 and #785 as pending for v0.12. This one seems ready to be merged, so should we have it in for v0.12 as well?

StephenBrown2:
How is this looking for 0.12.0? Looks like some conflicts now because of the private name switch, but I tested a merge and it goes cleanly aside from that.

If it gets merged and I have time this weekend, I may attempt to build on it for status code retries.

florimondmanca:
@stephenbrown2 Personally I’d be happy to *consider* it for 0.12. The features so far are restricted in scope which is definitely a good thing... Though I’m still undecided on whether this should eventually be handled by a third-party package, if/when a middleware API such as in #783 is added. That’s why I might be tempted to mark this as « provisional », at least.

tomchristie:
So, this is looking great.

There's one implementation-level bit that we might want to consider...

Adding `.sleep()` to the backends, and using it from the client means that we've got some bit of coupling from the client to the backend implementations which doesn't currently exist.

That conflicts slightly with https://github.com/encode/httpx/issues/768 which drops backends into being a strictly dispatcher-only bit of implementation, and with the intent of #782, which removes `backend` from the API at the client level.

Something that we *could* consider here is treating `sleep` *not* as part of the backend API, but instead just implementing a plain old `async def sleep()` function. (which uses `sniffio` and defers to either `trio.sleep` or `asyncio.sleep`).

We could use that in the async case, and use the regular `time.sleep` in the sync case. (Rather than calling `lookup_backend`, and calling `backend.sleep()`.)

dimaqq:
Possibly OT question: does this handle server sending a GOAWAY frame in response to starting a new stream?
Explanation: https://serverfault.com/a/962761
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1050329#c14
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=681477

florimondmanca:
Closing this off as this is super stale now - lots of conflicts I'm not sure are easily resolvable.

_If_ we figure connection retries are something we'd want to have built in, it should be fairly easy to start off from this diff to create a fresh branch from `master`.

PR #785, opened 2020-01-20
Switch to private module names

### LINKED ISSUE #772: Private module names
I've noticed a couple of projects lately which use private module names inside the package...

* https://github.com/python-hyper/h11/tree/master/h11
* https://github.com/oremanj/tricycle/tree/master/tricycle

I'd initially found it an odd style, but actually it's a really smart approach, that helps you enforce that only the top-level API that you're exposing is actually public API.

For example, right now users might do something like...

```python
from httpx.config import SSLConfig
from httpx.models import Origin
from httpx.auth import FunctionAuth
```

None of which are actually considered public API by us, but that's not immediately obvious to the user.
 
We've mentioned in chat or elsewher (I don't recall) that we really want to ensure our users are only ever using `import httpx` or `from httpx import <SOMETHING>`, and *yes* we could simply mention that in our documentation. But *actually* it'd be far clearer if we actually used private module names, and *enforced* that users would be aware if they choose to import and use non-public API.

This isn't *strictly* a breaking API change, but we *would* only want to make it on a median version bump (0.12.0 or 1.0.0), since it's likely that at least *some* of our users are currently using sub-module import styles.

We'd also want to update our test cases so that we're using public API wherever possible, and so that it's clearer where there are cases that we're using private API (which would either indicate unit-testing to an extent, or that we've got some part of API that we're planning to expose, but haven't yet finalised on.)

I think we might also want to do a bit of rejigging of how we handle `__all__` with this too. We probably don't *actually* want to support `from httpx import *`.

Somewhat related to this is that we've got a bunch of methods on classes that we don't document, aren't treating as public API, and probably ought to move into private methods. (Eg. on the `Client` everything "below" `.build_request` and `.send` should probably be private API.) However, let's treat that as a seperate issue to be discussed.

florimondmanca:
Fixes #772 

Still need to confirm…

- References in docs


tomchristie:
Fantastic! I think we might also rejig how the imports are arranged, but I think we should probably leave that to a follow-up PR.

(Ie. drop `__all__` from the top-level. In cases where we're importing lots of classes from a module consider using eg. `from ._exceptions import *` in our top-level `__init.py`, with any Public API exceptions listed in `__all__` in the `_exceptions.py` module. I think it'd be a lot clearer that listing all the import exceptions in a seperate place from where we're defining them.)

How do we want to approach this all release-wise/timing-wise?

Presumably at least both this and #782 in a 0.12.0 release?

florimondmanca:
> Presumably at least both this and #782 in a 0.12.0 release?

Yup, that sounds good to me!

florimondmanca:
Alright, I updated the docs -- only a few references to internal modules in logs, but otherwise nothing indicating that we've got anything missing at the top-level. This should be ready for review! I'll issue a follow-up for reorganizing imports.

tests/test_multipart.py
```diff
@@ -7,10 +7,10 @@
 import pytest
 
 import httpx
-from httpx.config import CertTypes, TimeoutTypes, VerifyTypes
-from httpx.content_streams import encode
-from httpx.dispatch.base import AsyncDispatcher
-from httpx.utils import format_form_param
+from httpx._config import CertTypes, TimeoutTypes, VerifyTypes
```
tomchristie: Incidentally, it looks like we've erronously left the signature as `send(request, verify, cert, timeout)` in a bunch of these and other test cases, rather than switching to `send(request, timeout)` on the dispatch interface.

We're missing those because we're not type checking our test cases.

Not related to this PR, so doesn't neccessarily need to change here, but this helps highlight it, since we're pulling in more private API that we might otherwise expect.

tomchristie:
What do we reckon then? Hit "Go" on this one now, and then start drawing up whatever we want in a 0.12 release?

florimondmanca:
Yup, I think we’re ready for getting 0.12 together. :)

tomchristie:
Okay, let's roll this one now, then.

PR #786, opened 2020-01-21
Reorganize __init__ modules

florimondmanca:
Builds on #785 (which is currently the base here for the sake of review).

This PR moves the declarations from the top-level `__init__.py` to `__all__` variables in each module that contains public API elements.

I figured the consistency of adding an `__all__` everywhere (even though we only expose a small selection of elements) was nicer than mixing `from _foo import *`/`from _bar import A, B, C` in the top-level `__init__.py`. :-)

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -1,86 +1,15 @@
-from .__version__ import __description__, __title__, __version__
-from ._api import delete, get, head, options, patch, post, put, request, stream
-from ._auth import Auth, BasicAuth, DigestAuth
-from ._client import AsyncClient, Client
-from ._config import PoolLimits, Proxy, Timeout
-from ._dispatch.asgi import ASGIDispatch
-from ._dispatch.wsgi import WSGIDispatch
-from ._exceptions import (
-    ConnectionClosed,
-    ConnectTimeout,
-    CookieConflict,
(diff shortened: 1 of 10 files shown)


tomchristie:
> I figured the consistency of adding an __all__ everywhere (even though we only expose a small selection of elements) was nicer than mixing from _foo import */from _bar import A, B, C in the top-level __init__.py. :-)

I'd probably go the other way on that, since using `*` in all those cases makes it awkward for our users to figure out eg. "where is `httpx.Proxy` defined?", or "what's the complete list of public API?".

I think it's okay for us to do for the exceptions, since that's such a big list, and since they're fairly obviously exception classes, and since it's a big ol' chunk of stuff. I'd probably prefer if we either:

* Only use `*` for the exception classes.
* *Or:* Change tack and just pass on this. (Although perhaps? still dropping `__all__` from the top-level.)

florimondmanca:
Okay then, I’m not sure I understood the initial motivation. Is it just that we’d want to prevent users from doing `from httpx import *`? AFAIK we can’t, and dropping `__all__` will just drop us specifying what `import *` exposes.

tomchristie:
I think we're ducking out on this refactor, right?

PR #788, opened 2020-01-22
Fix `httpx.Proxy(..., mode="FORWARD_ONLY")` configuration.

bandoche:
FORWARD_ONLY seems correct

https://github.com/encode/httpx/blob/b23420392efdcc10f3d802f335739d9cb3d72d5c/httpx/dispatch/proxy_http.py#L63
and
https://github.com/encode/httpx/blob/b23420392efdcc10f3d802f335739d9cb3d72d5c/httpx/config.py#L326
seems different.


tomchristie:
Good catch, thanks!

PR #791, opened 2020-01-23
Drop backend parameter on AsyncClient

### LINKED ISSUE #782: Drop `backend=...` argument.

florimondmanca:
Fixes #782 

Worth noting that any `backend` parameter is solely scoped within the realm of dispatchers now (e.g. `ConnectionPool`, `HTTPConnection`, `HTTPProxy`). :-)

(Technically breaking since this was advertised in the docs, but very low risk of actually affecting anyone in practice since auto-detection is the most convenient and the default.)


tomchristie:
Needs conflicts resolving, since #785.

Issue #790, opened 2020-01-22
Support for Event Hooks


martsime:
I'm currently using [https://github.com/encode/requests-async](https://github.com/encode/requests-async) for one of my projects and now I wanted to switch to HTTPX, but I can't find out if it supports event hooks like this [https://requests.readthedocs.io/en/master/user/advanced/#event-hooks](https://requests.readthedocs.io/en/master/user/advanced/#event-hooks).

Or is there another easy way to add hooks onto an `AsyncClient` object to be called every request?

martsime:
We use hooks to collect metrics about all the requests, mostly for monitoring purposes. It would be nice if the  `AsyncClient` supported it similar to how `Session` in requests and requests_async does. So one could do something like this: 

```python3
def response_hook(response, *args, **kwargs): 
    # Log status of response, etc...
    ...

def get_client():
    client = httpx.AsyncClient()
    client.hooks["response"].append(response_hook) 
    return client
```
A more general middleware API could be a good idea and definitely solves the problem.

It's not a big deal to make a workaround for our specific case, but since the other libraries supported it I thought there might be another way of doing it in HTTPX which I couldn't find.

florimondmanca:
IMO a general middleware API like the one drafted in #783 would be the most versatile way to provide these "on request/on response" hooks…

```python
class MonitoringMiddleware:
    def __init__(self, parent):
        self.parent = parent

    def __call__(self, request, scope, timeout):
        request_hook(request, ...)
        response = yield from self.parent(request, scope, timeout)
        response_hook(response, ...)
        return response

client = httpx.Client()
client.add_middleware(MonitoringMiddleware)
```

florimondmanca:
Though we might want to provide a more streamlined API, that might just happen to be implemented as middleware, e.g...

```python
def monitor_requests(request):
    # Process request...
    response = yield request
    # Process response...
    
client = httpx.Client(hooks=[monitor_requests])
```

jbrockmendel:
Jumping on here instead of creating a new issue:

I've got some project code that patched both requests and the stdlib to allow for hooks that logged requests and redirects.  Went to go upstream it to requests and ended up here instead.  Is there enough interest here to make a proof of concept worthwhile?

florimondmanca:
Hey,

Yes, I think we could go for a PR.

Forget the middleware stuff I mentioned above at the time, I’d agree a simpler and more integrated « register request/response callbacks » API is probably a better first step.

[...]

PR #800, opened 2020-02-02
Middleware API design doc

florimondmanca:
Pushing a simplification of #783, focused on docs (and minimal code changes to allow example code snippets to work, i.e. hooking middleware into `Client.get()`.)

Not sure we'd want to move forward with this whole middleware idea *yet* but I wanted to make a case for it here.

I'm pretty much convinced this is the correct API if we wanted to provide a general-purpose extension mechanism for clients.

It allows the following:

- Intercepting/inspecting/modifying requests
- Intercepting/inspecting/modifying responses
- Sending multiple requests (w/o being coupled to the client or the sync/async environment).
- Handling exceptions in a predictable fashion using `try/except` blocks.

An extension mechanism has benefits in itself, even if we don't immediately use it internally (e.g. for auth/redirects as initially implemented in #783). Mainly ,it would greatly reduce any tendency towards feature creep, and help us keep our core API/feature set under control. (Many "so I have this particular feature request…" issues could be addressed as "consider building a middleware".)

For example, retries (#784) and hooks (#790) can be implemented as middleware, with a much lower impact for us, and most likely more value for users (as focused third-party packages can afford going much deeper into the feature set, whereas we'd have a tendency to keep things minimal to reduce maintenance overhead).

(Eg. it's really super interesting to me how more powerful and versatile a retries middleware built on top of [tenacity](https://github.com/jd/tenacity) would be, compared to us bringing some *very limited* built-in functionality via #784… See example in the docs here.)

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -471,3 +471,160 @@ If you do need to make HTTPS connections to a local server, for example to test 
 >>> r
 Response <200 OK>
 ```
+
+## Middleware
+
+Middleware is a general-purpose mechanism for extending the built-in functionality of a `Client`.
+
+### Using middleware
+
+Middleware generally comes in the form of classes. Middleware classes and their configuration parameters are meant to be passed as a list of `httpx.Middleware` instances to a client:
+
+```python
+from example.middleware import ExampleMiddleware
+import httpx
+
+middleware = [
+    httpx.Middleware(ExampleMiddleware, client_param="value", ...),
+]
+
+with httpx.Client(middleware=middleware) as client:
+    # This request will pass through ExampleMiddleware before
+    # reaching the core processing layers of the Client.
+    r = client.get("https://example.org")
+```
+
+### Writing middleware
+
+Middleware classes should accept a `get_response` parameter, that represents the inner middleware, as well as any keyword arguments for middleware-specific configuration options. They should implement the `.__call__()` method, that accepts a `request`, `timeout` configuration, and any per-request keyword arguments coming directly from keyword arguments passed to `client.get()`, `client.post()`, etc. The `__call__()` method should return a generator that returns a `Response` instance. To get a response from the inner middleware, use `response = yield from get_response(...)`.
+
+#### Basic example
+
+Here is a "do-nothing" middleware that sends the request unmodified, and returns the response unmodified:
+
+```python
+import httpx
+
+class PassThroughMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request, timeout, **kwargs):
+        return (yield from self.get_response(request, timeout, **kwargs))
+```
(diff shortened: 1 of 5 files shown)


StephenBrown2:
This is definitely cool. So this would only add on to the current "middleware" baked-in to httpx, and not replace it?

florimondmanca:
> So this would only add on to the current "middleware" baked-in to httpx, and not replace it?

It’s not required that we switch the implementation for redirect/auth to using this middleware API, but it is technically possible. (See the original #783 for how that might look. I left it out from this PR to focus on the API itself.)

docs/advanced.md
```diff
@@ -471,3 +471,160 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+Here is a "do-nothing" middleware that sends the request unmodified, and returns the response unmodified:
+
+```python
+import httpx
+
+class PassThroughMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request, timeout, **kwargs):
```
florimondmanca: It just occurred to me now that we might want to make it so that `timeout` is buried inside `**kwargs`. It’s not useful for the vast majority of middleware, so exposing it as a positional argument in the API is somewhat clunky.

docs/advanced.md
```diff
@@ -471,3 +471,160 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+    print("Got response")
+```
+
+Output:
+
+```console
+Client options: {'client_option': 'example'}
+Sending request...
+Request: <Request('GET', 'https://example.org')>
+Request options: {'request_option': 'example'}
```
florimondmanca: This might actually not be true, as `auth=None` and `allow_redirects=True` might end up in there.

StephenBrown2:
Yeah, I was thinking of the original #783 with that. It's good to be separated for now. In the future [middleware_stack](https://github.com/encode/httpx/pull/800/files#diff-eff812167840d2389ea5f0d809b30c32R494) would just need to be aware of the builtins and possibly allow for overriding, but that's for the future.

florimondmanca:
Just found out about this Elixir HTTP client that’s completely built around the concept of middleware: https://github.com/teamon/tesla

Middleware there only allows sending a single request, but there’s a notion of context that allows passing state between recursive calls.

In fact, even adding in headers or a `base_url` is handled via middleware. Probably over engineered for these simple cases, but it occurred to me that, well, yes we *could technically* achieve that here too. (But we probably shouldn’t!)

Anyway dropping it here ‘cause I think it’s another example that this whole middleware thingy probably makes sense to have in some day.

florimondmanca:
I guess I'll close this for now, as this is not in our priorities for 1.0. Let's reconsider after we reached that milestone. :-)

PR #810, opened 2020-02-10
Ignore PermissionError in netrc_info()

Jonney:
I'm running uWSGI, the user is  "nobody", the netrc_file should be '~/.netrc'.
But I don't know why, the location trunes out to be '/root/.netrc', so a PermissionError raised, it's better to ignore this error instead of to raise.


tomchristie:
Okay. It's not super clear if it'd be better for us to ignore this, or to raise it, to ensure the user is aware of the issue.

You can disable `.netrc` handling by using `trust_env=False`, which would be a workaround in the meantime.

I *suppose* we should probably follow requests' lead here and silently ignore `.netrc` in cases where it is malformed or causes any kind of IOError... https://github.com/psf/requests/blob/master/requests/utils.py#L209

The build on this pull request is currently failing due to linter errors - make sure to run `scripts/lint` to apply black and other linting checks.

httpx/_utils.py
```diff
@@ -134,9 +134,12 @@ def netrc_info(self) -> typing.Optional[netrc.netrc]:
             self._netrc_info = None
             for file_path in self.netrc_files:
                 expanded_path = Path(file_path).expanduser()
-                if expanded_path.is_file():
-                    self._netrc_info = netrc.netrc(str(expanded_path))
-                    break
+                try:
+                    if expanded_path.is_file():
+                        self._netrc_info = netrc.netrc(str(expanded_path))
```
florimondmanca: We should probably place the try/except block more precisely, ie around this particular line? (I assume the `.is_file` check isn’t what’s causing the `PermissionError`.)

httpx/_utils.py
```diff
@@ -134,9 +134,12 @@ def netrc_info(self) -> typing.Optional[netrc.netrc]:
...
+                        break
+                expect PermissionError:
+                    pass
```
florimondmanca: Skipping sounds reasonable, but should we at least log a debug message about this error?

(You can use `get_logger()` to instantiate a logger in this module, if there isn’t one yet. Might need to revise the order of functions if logging utils are later down this file.)

florimondmanca:
@jonney Also, were you able to debug this by using the `netrc` module directly in a REPL? There might be something set that would make `~` resolve  to the `root` user (although that seems unlikely).

httpx/_utils.py
tomchristie: I'd assume `is_file` probably *could* raise an IOError here if there's no permissions to that path?

florimondmanca: Hmm, yes, that’s something we’d need to double check — is that covered by `PermissionError`?

Jonney: Here is the trace info:

[2020-02-11 22:29:24,797] ERROR in app: Exception on /my/app/url/path [POST]
  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.7/dist-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/dist-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/path/to/my/app.py", line 28, in my_function
    my_variable=my_variable.my_function(my_variable,my_variable)
  File "/path/to/my/app.py", line 17, in my_function
    my_variable=my_variable.my_function()
  File "/path/to/my/app.py", line 73, in my_function
    r=httpx.get(url,cookies=self.cookies,headers=headers)
  File "/usr/local/lib/python3.7/dist-packages/httpx/api.py", line 170, in get
    trust_env=trust_env,
  File "/usr/local/lib/python3.7/dist-packages/httpx/api.py", line 96, in request
  File "/usr/local/lib/python3.7/dist-packages/httpx/client.py", line 573, in request
  File "/usr/local/lib/python3.7/dist-packages/httpx/client.py", line 590, in send
    auth = self.build_auth(request, auth)
  File "/usr/local/lib/python3.7/dist-packages/httpx/client.py", line 282, in build_auth
    credentials = self.netrc.get_credentials(request.url.authority)
  File "/usr/local/lib/python3.7/dist-packages/httpx/utils.py", line 145, in get_credentials
    if self.netrc_info is None:
  File "/usr/local/lib/python3.7/dist-packages/httpx/utils.py", line 137, in netrc_info
    if expanded_path.is_file():
  File "/usr/lib/python3.7/pathlib.py", line 1365, in is_file
    return S_ISREG(self.stat().st_mode)
  File "/usr/lib/python3.7/pathlib.py", line 1161, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/root/.netrc'

Jonney:
> You can disable `.netrc` handling by using `trust_env=False`, which would be a workaround in the meantime.

It's ugly to use trust_env=False **every time**, if so, I want to use Requests instead of httpx

StephenBrown2:
You can use a `Client`/`AsyncClient` instance, where `trust_env` can be set on the client itself, rather that on the request methods:

```py
client = httpx.Client(trust_env=False)
```

httpx/_utils.py
```diff
@@ -123,20 +123,27 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
             for file_path in self.netrc_files:
                 expanded_path = Path(file_path).expanduser()
-                if expanded_path.is_file():
-                    self._netrc_info = netrc.netrc(str(expanded_path))
-                    break
+                try:
+                    if expanded_path.is_file():
+                        self._netrc_info = netrc.netrc(str(expanded_path))
+                        break
+                expect (netrc.NetrcParseError, IOError, PermissionError):
```
StephenBrown2:
```suggestion
                except (netrc.NetrcParseError, IOError, PermissionError):
```

httpx/_utils.py
```diff
@@ -123,20 +123,27 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
...
-                    break
+                try:
+                    if expanded_path.is_file():
+                        self._netrc_info = netrc.netrc(str(expanded_path))
+                        break
+                except (netrc.NetrcParseError, IOError, PermissionError):
+                    # If there was a parsing error or a permissions issue reading the file,
+                    # we'll just skip netrc auth unless explicitly asked to raise errors.
+                    if self.raise_errors:
+                        raise
```
florimondmanca: Given that we don't actually allow passing `raise_errors=True` anywhere, I'd suggest we drop this parameter and just do:

```suggestion
                    # Issue while reading the netrc file, ignore...
                    pass
```

tomchristie:
Thanks so much! Now closed, see #1104

PR #819, opened 2020-02-20
Drop RedirectLoop exception

oczkers:
Simply drop RedirectLoop exception and instead rely on max number of redirects (and retries on future) like requests, urllib3 and many others do.
More info in previous closed pr https://github.com/encode/httpx/pull/801


tests/client/test_redirects.py
```diff
@@ -242,23 +241,6 @@ class MockDispatch(AsyncDispatcher):
             response = await response.anext()
 
 
-@pytest.mark.usefixtures("async_environment")
-async def test_redirect_loop():
-    client = AsyncClient(dispatch=MockDispatch())
-    with pytest.raises(RedirectLoop):
-        await client.get("https://example.org/redirect_loop")
```
tomchristie: I guess we *could* keep this test in, and just ensure that `TooManyRedirects` is raised?

oczkers: There is already test for TooManyRedirects with 21 redirects so i thought it's gonna be enough but sure more is better.

tomchristie:
Fab, thanks!

tomchristie:
Minor API change here, which is fine since we're already commited to the next release being 0.12.0

PR #820, opened 2020-02-20
When setting headers, fallback to `utf-8` if no encoding is specified yet.

oczkers:
Reproduce steps:
```python
>>> r = httpx.Client()
>>> r.headers['Velocity®Ultra'] = 'Velocity®Ultra'
  File "/usr/lib/python3.7/site-packages/httpx-0.11.1-py3.7.egg/httpx/_models.py", line 518, in __setitem__
    set_key = key.lower().encode(self.encoding)
UnicodeEncodeError: 'ascii' codec can't encode character '\xae' in position 8: ordinal not in range(128)
```

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -515,13 +515,8 @@ class Headers(typing.MutableMapping[str, str]):
         Set the header `key` to `value`, removing any duplicate entries.
         Retains insertion order.
         """
-        try:
-            set_key = key.lower().encode(self.encoding)
-            set_value = value.encode(self.encoding)
-        except UnicodeEncodeError:
-            self.encoding = "utf8"
-            set_key = key.lower().encode(self.encoding)
-            set_value = value.encode(self.encoding)
+        set_key = key.lower().encode(self._encoding or "utf-8")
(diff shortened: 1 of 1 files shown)


httpx/_models.py
```diff
@@ -515,8 +515,13 @@ def __setitem__(self, key: str, value: str) -> None:
...
         """
-        set_key = key.lower().encode(self.encoding)
-        set_value = value.encode(self.encoding)
+        try:
+            set_key = key.lower().encode(self.encoding)
+            set_value = value.encode(self.encoding)
+        except UnicodeEncodeError:
+            self.encoding = "utf8"
+            set_key = key.lower().encode(self.encoding)
+            set_value = value.encode(self.encoding)
```
tomchristie: Actually I think we need a slightly different approach on this, since eg. we don't want to just change `self.encoding` if it has been explicitly set.

How's something like this?...

```python
encoding = self._encoding or "utf-8"
set_key = key.lower().encode(encoding)
set_value = value.encode(encoding)
```

oczkers: Done :-)

tomchristie:
Great, thanks!

PR #835, opened 2020-02-29
Enable NO_PROXY environment variable support 

evanlurvey:
This PR addresses issue https://github.com/encode/httpx/issues/360. I found this issue by attempting to use httpx in a environment that had both endpoints that needed to be proxied and ones that needed to stay in the local network. That is when I found out that there was no "NO_PROXY" support.

### DIFF
--- a/tests/client/test_proxies.py
+++ b/tests/client/test_proxies.py
@@ -1,6 +1,7 @@
 import pytest
 
 import httpx
+from httpx import URL
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/client/test_proxies.py
```diff
@@ -87,3 +88,18 @@ def test_dispatcher_for_request(url, proxies, expected):
...
+    client = httpx.AsyncClient()
+    dispatcher = client.dispatcher_for_url(URL("http://google.com"))
+    assert dispatcher == client.dispatch
+
+
+def test_no_proxy_not_set_returns_correct_dispatcher(monkeypatch):
+    monkeypatch.setenv("HTTP_PROXY", "http://example.com")
+    client = httpx.AsyncClient()
+    dispatcher = client.dispatcher_for_url(URL("http://google.com"))
+    assert dispatcher == client.proxies["http"]
```
florimondmanca: I think we can use the style from the other test cases here, and parametrize this test to be able to easily add more test cases if needed (e.g. I added one below for the default "no env set" case), so…

```suggestion
@pytest.mark.parametrize(
    ["url", "env", "expected"],
    [
        ("http://google.com", {}, None),
        (
            "http://google.com",
            {"HTTP_PROXY": "http://example.com"},
            "http://example.com",
        ),
        (
            "http://google.com",
            {"HTTP_PROXY": "http://example.com", "NO_PROXY": "google.com"},
            None,
        ),
    ],
)
def test_proxies_environ(monkeypatch, url, env, expected):
    for name, value in env.items():
        monkeypatch.setenv(name, value)

    client = httpx.AsyncClient()
    dispatcher = client.dispatcher_for_url(httpx.URL(url))

    if expected is None:
        assert dispatcher == client.dispatch
    else:
        assert dispatcher.proxy_url == expected
```

tests/client/test_proxies.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 import httpx
+from httpx import URL
```
florimondmanca: `URL` is a public piece of API; in that case we tend to have tests stick to the `httpx.*` usage we recommend for users. :-)

```suggestion
```

PR #840, opened 2020-03-01
Document cleaning up streaming responses without block usage

### LINKED ISSUE #830: How about provide non-contextmanager api for stream?
I'm tring to get a stream using httpx and forward it using starlette, Here is the code I expect:
```python
    async with client.stream("GET", url) as r:
            return StreamingResponse(r.aiter_raw())
```
But, because the call of `r.aiter_raw` is after the call of stream context manager's `__aexit__`, the code above will raise a `ResponseClosed` exception.
I don't learn more about the design of httpx and starlette, and I think the simple way is providing non-contextmanager api for stream, just like `AsyncClient`. It can prevent we use dunder method directly:
```python
    ctxm = client.stream("GET", wos_url)
    response = await ctxm.__aenter__()
    return StreamingResponse(
        response.aiter_raw(),
        background=BackgroundTasks([BackgroundTask(ctxm.__aexit__)]),
    )
```

florimondmanca:
Fixes #830 

Update the documentation on streaming responses (both in the sync and async cases) to indicate how users can clean use the `.stream()` context manager when block usage via `with ...` is not possible.

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -78,7 +78,35 @@ The async response streaming methods are:
 * `Response.aiter_text()` - For streaming the response content as text.
 * `Response.aiter_lines()` - For streaming the response content as lines of text.
 * `Response.aiter_raw()` - For streaming the raw response bytes, without applying content decoding.
-* `Response.aclose()` - For closing the response. You don't usually need this, since `.stream` block close the response automatically on exit.
+* `Response.aclose()` - For closing the response. When using the context-managed syntax, `.stream()` will close the response automatically on exit.
+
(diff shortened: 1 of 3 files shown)


tomchristie:
I think using `AsyncExitStack` here is a bit of a hacky work around.

We *really* want to stear folks away from using `stream` without a context manager wherever possible, but I'd be okay with us explaining *why* we do that, but documenting how to do it in the cases where it really is needed. For example...

```python
response = client.send(..., stream=True)
try:
    ...
finally:
    response.close()
```

florimondmanca:
Closing for now, will need to take another shot at this. :-)

PR #851, opened 2020-03-05
Do not mark wheel as universal

### LINKED ISSUE #850: Wheel should not be tagged as universal
Currenlty ``httpx`` wheels are tagged as universal - ``py2.py3`` because of following configuration in ``setup.cfg``:
```
[bdist_wheel]
universal = 1
```
https://github.com/encode/httpx/blob/707e54c48415e98224ca74e67809a5593b41b03d/setup.cfg#L1

However this is incorrect because ``httpx`` uses modern Python syntax and features so its wheel must not be marked as universal.

daa:
``httpx`` requires Python > 3 so its wheel must not be tagged as universal.

Fixes #850

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[bdist_wheel]
-universal = 1
-
 [flake8]
 ignore = W503, E203, B305
 max-line-length = 88


tomchristie:
Great, thanks!

daa:
Thank you for work on ``httpx``!

PR #853, opened 2020-03-06
Use timeout when establishing tunnel connection through proxy

### LINKED ISSUE #845: AsyncClient should use timeout when establishing tunnel proxy connection
Currently ``httpx._dispatch.proxy_http.HTTPProxy.request_tunnel_proxy_connection()`` method doesn't support timeout which results in unexpectedly long waiting time when connection to proxy cannot be established. For example this happens when packets to proxy are dropped.

My suggestion is that ``tunnel_connection()`` method should pass ``timeout`` to ``request_tunnel_proxy_connection`` and the latter function should use it when sending request to proxy server.

https://github.com/encode/httpx/blob/707e54c48415e98224ca74e67809a5593b41b03d/httpx/_dispatch/proxy_http.py#L138
https://github.com/encode/httpx/blob/707e54c48415e98224ca74e67809a5593b41b03d/httpx/_dispatch/proxy_http.py#L121

daa:
Here I implemented the same logic as in ``connection_pool`` module in ``request_tunnel_proxy_connection()`` function to apply timeout where required.

Fixes #845.

### DIFF
--- a/httpx/_dispatch/proxy_http.py
+++ b/httpx/_dispatch/proxy_http.py
@@ -118,7 +118,7 @@ class HTTPProxy(ConnectionPool):
         connection = self.pop_connection(origin)
 
         if connection is None:
-            connection = await self.request_tunnel_proxy_connection(origin)
+            connection = await self.request_tunnel_proxy_connection(origin, timeout)
 
             # After we receive the 2XX response from the proxy that our
             # tunnel is open we switch the connection's origin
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks for this!

Given that our next release is likely to be a 0.13.dev0 which drops this module in favour of using `httpcore` for the networking layer, I'm not sure it'd make sense for us to pull this in.

It'd probably make more sense to check if we've got the equivelent issue in `httpcore` or not, and apply the fix there if it's an issue. https://github.com/encode/httpcore

PR #857, opened 2020-03-09
Implement streaming multipart uploads

### LINKED ISSUE #114: Design Multipart helper to stream by default

florimondmanca:
Fixes #114. ✨ 

Ensures we don't ever load any files or buffers in full into memory when uploading form data.

I've also refined tests to make sure we have full test coverage on the multipart implementation.


florimondmanca:
Interestingly this addresses issues people have been having in Requests/urllib3 due to urllib3 not supporting streaming in this case, eg:

- https://github.com/psf/requests/issues/4687: High resource usage of multipart in Requests
- https://github.com/psf/requests/issues/5024: Python bug when calling `.sendall()` with 2GB+ of content

Would be interesting to validate against a local but realistic setup though, eg streaming a 500MB+ file over to a local server, and see how memory usage behaves.

httpx/_utils.py
```diff
@@ -309,6 +310,37 @@ def unquote(value: str) -> str:
     return value[1:-1] if value[0] == value[-1] == '"' else value
 
 
+def guess_content_type(filename: typing.Optional[str]) -> typing.Optional[str]:
+    if filename:
+        return mimetypes.guess_type(filename)[0] or "application/octet-stream"
+    return None
+
+
+def get_filelike_length(stream: typing.IO) -> int:
```
tomchristie: Looks good! Based off of an existing implementation someplace, or figured out from scratch?

httpx/_utils.py
```diff
@@ -309,6 +310,37 @@ def unquote(value: str) -> str:
...
+    in number of bytes.
+    """
+    try:
+        # Is it an actual file?
+        fd = stream.fileno()
+    except OSError:
+        # No... is it something that supports random access, like `io.BytesIO`?
+        if not hasattr(stream, "seekable") or not stream.seekable():  # pragma: nocover
+            # Not even that? Sorry, we're doomed...
+            raise
```
tomchristie: We might need to think about how we handle this case a bit more I guess. I don't recall off the top of my head if individual multipart file fields *have* to include a content-length, or if they support framing the data when the content length is unkown up front?

tomchristie: Right, so yeah the content is delimited by the `boundary` markers, so files *without* a content-length header are perfectly okay. Eg in the MDN example here https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

```
POST /test.html HTTP/1.1
Host: example.org
Content-Type: multipart/form-data;boundary="boundary"

--boundary
Content-Disposition: form-data; name="field1"

value1
--boundary
Content-Disposition: form-data; name="field2"; filename="example.txt"

value2
--boundary--
```

I *guess* we ought to include the Content-Length where possible, but it's feasible that our policy could alternately be "let's *not* include the Content-Length". Perhaps it'd be good to take a steer on that from some prior art. Eg. what does the streaming multipart implementation in the requests toolbelt do?... https://toolbelt.readthedocs.io/en/latest/uploading-data.html#streaming-multipart-data-encoder

tomchristie: For the purposes of *this* function, I guess we'll want to return an `Optional[int]`, and return `None` when we cannot determine the content length.

httpx/_content_streams.py
```diff
@@ -207,17 +206,34 @@ def __init__(self, name: str, value: typing.Union[str, bytes]) -> None:
             self.name = name
             self.value = value
 
+        @functools.lru_cache(1)
         def render_headers(self) -> bytes:
             name = format_form_param("name", self.name)
             return b"".join([b"Content-Disposition: form-data; ", name, b"\r\n\r\n"])
 
+        @functools.lru_cache(1)
```
tomchristie: These functions are simple enough that I'd expect a `@functools.lru_cache(1)` is likely to introduce more overhead than it'll save. We *might* consider either:

* Dropping the `@functools.lru_cache(1)` on trivial method implementations.
* Dropping the `@functools.lru_cache` cases everywhere for this PR, focusing strictly on correctness, and *potentially* adding them in as an incremental improvement sperately. (It's feasible that there's stateful approaches that we might consider instead of `lru_cache` wrapping.)

Any thoughts?

httpx/_utils.py
florimondmanca: Basically a patchwork of info from here and there :-) eg the [io](https://docs.python.org/3/library/io.html) module docs, and SO for `os.fstat` and "how to get length of BytesIO/StringIO". :-)

httpx/_utils.py
florimondmanca: Ah interesting about `Content-Disposition`… I actually took heavy inspiration from this blog post: [Streaming Multipart Requests](http://khanlou.com/2018/11/streaming-multipart-requests/), and it also actually doesn't use the `Content-Disposition: form-data; ...` format too.

I'll try that out and do some validation against some live servers too. :-)

httpx/_content_streams.py
florimondmanca: I actually initially used private unset attributes + `if not hasattr(self, "_headers")`, but decided the `lru_cache` way was a bit cleaner while functionally identical.

I guess we can revert to that correct-first approach though, sure.

florimondmanca:
I ran a bunch of experiments against HTTPBin and a Django microserver to see how they'd behave to the presence/absence of `Content-Length` for multipart POST requests. They basically treat it as `Content-Length: 0` and don't even try to read the body. So it looks like in practice `Content-Length` _is_ required for multipart form data anyway…

Details below:

First, issuing requests against `https://httpbin.org/post`…

Including the `Content-Length` results in correct behavior:

```python
>>> import httpx
>>> url = "https://httpbin.org/post"
>>> lorem = open("lorem.txt")
>>> files = {"lorem": lorem}
>>> data = {"hello": "world"}
>>> r = httpx.post(url, data=data, files=files)
>>> r.status_code
200
>>> r.json()
{'args': {},
 'data': '',
 'files': {'lorem': '\n\nLorem ipsum <...Redacted for brievety...>'},
 'form': {'hello': 'world'},
 'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate, br',
  'Content-Length': '3261',
  'Content-Type': 'multipart/form-data; boundary=e7fa8325d92b926b35d877268b7b8435',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.12.0',
  'X-Amzn-Trace-Id': 'Root=1-5e69ebc0-2cde1d1a6d866d38fbb79c57'},
 'json': None,
 'origin': '92.169.139.224',
 'url': 'https://httpbin.org/post'}
```

If `Content-Length` isn't included, then HTTPBin infers it as 0 (although the request content wasn't empty: it contained data and files) and doesn't even read the request:

```python
>>> import httpx
>>> url = "https://httpbin.org/post"
>>> lorem = open("lorem.txt")
>>> files = {"lorem": lorem}
>>> data = {"hello": "world"}
>>> client = httpx.Client()
>>> request = client.build_request("POST", url, data=data, files=files)
>>> request.headers.pop("content-length")
'3261'
>>> r = client.send(request)
>>> r.status_code
200
>>> r.request.headers  # NOTE: no 'content-length' here.
Headers({'host': 'httpbin.org', 'user-agent': 'python-httpx/0.12.0', 'accept': '*/*', 'accept-encoding': 'gzip, deflate, br', 'connection': 'keep-alive', 'content-type': 'multipart/form-data; boundary=405abb1f7a6217c5c1bfbfbf8b96010e'})
>>> r.json()  # See how this is mostly empty?
{'args': {},
 'data': '',
 'files': {},
 'form': {},
 'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate, br',
  'Content-Length': '0',
  'Content-Type': 'multipart/form-data; boundary=405abb1f7a6217c5c1bfbfbf8b96010e',
  'Host': 'httpbin.org',
  'User-Agent': 'python-httpx/0.12.0',
  'X-Amzn-Trace-Id': 'Root=1-5e69eb31-ef22d77230959bd05da2aac4'},
 'json': None,
 'origin': '92.169.139.224',
 'url': 'https://httpbin.org/post'}
```

Now with `requests-toolbelt`…

First with `Content-Length` — works fine:

```python
>>> import requests
>>> from requests_toolbelt.multipart.encoder import MultipartEncoder
>>> lorem = open("lorem.txt", "rb")
>>> fields = {"lorem": ("lorem.txt", lorem, "text/plain"), "hello": "world"}
>>> m = MultipartEncoder(fields=fields)
>>> url = "https://httpbin.org/post"
>>> r = requests.post(url, data=m, headers={"Content-Type": m.content_type})
>>> r.status_code
200
>>> r.request.headers
{'User-Agent': 'python-requests/2.23.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Type': 'multipart/form-data; boundary=f34a54a283644e4da0f6cc94c9f11548', 'Content-Length': '3261'}
```

Now, dropping `Content-Length`, I actually had to go through hoops to even be able to send the request due to weird edge cases in `requests-toolbelt` and Requests, but eventually we get the same behavior as in HTTPX: HTTPBin will interpret as 'Content-Length: 0' and won't read the body.

```python
>>> import requests
>>> from requests_toolbelt.multipart.encoder import MultipartEncoder
>>> lorem = open("lorem.txt", "rb")
>>> fields = {"lorem": ("lorem.txt", lorem, "text/plain"), "hello": "world"}
>>> m = MultipartEncoder(fields=fields)
>>> url = "https://httpbin.org/post"
>>> s = requests.Session()
>>> # BUG: 'MultipartEncoder' is not iterable
>>> data = m.to_string()
>>> # NOTE: request won't be streamed now, but not the point here.
>>> req = requests.Request("POST", url, data=data, headers={"Content-Type": m.content_type})
>>> prepped = req.prepare()
>>> # BUG: 'int has no len()' (because iterating over bytes yields ints)
>>> prepped.body = prepped.body.splitlines()
>>> prepped.headers.pop("Content-Length")
'3261'
>>> r = s.send(prepped)
>>> r.status_code
200
>>> r.json()
{'args': {},
 'data': '',
 'files': {},
 'form': {},
 'headers': {'Content-Type': 'multipart/form-data; boundary=9515a8b9bea94739968c0976c03d8d63',
  'Host': 'httpbin.org',
  'X-Amzn-Trace-Id': 'Root=1-5e69eeeb-e2950740e34e2f3045592140'},
 'json': None,
 'origin': '92.169.139.224',
 'url': 'https://httpbin.org/post'}
```

Now, is this specific to HTTPBin? Let's try with a [Django microserver](https://github.com/wsvincent/django-microframework).

```python
# app.py
import json
from django.conf import settings
from django.core.handlers.wsgi import WSGIHandler
from django.http import HttpResponse
from django.urls import path

settings.configure(ROOT_URLCONF=__name__,)


def upload_file(request):
    content = {
        "method": request.method,
        "headers": dict(request.headers),
        "data": dict(request.POST),
        "files": dict(request.FILES),
    }
    return HttpResponse(json.dumps(content), content_type="application/json")


urlpatterns = [path("", upload_file)]

app = WSGIHandler()
```

```bash
uvicorn app:app --interface wsgi
```

```python
>>> import httpx
>>> url = "http://localhost:8000/"
>>> lorem = open("lorem.txt")
>>> files = {"lorem": lorem}
>>> data = {"hello": "world"}
>>> client = httpx.Client()
>>> request = client.build_request("POST", url, data=data, files=files)
>>> request.headers.pop("content-length")
'3261'
>>> r = client.send(request)
>>> r.status_code
200
>>> r.json()
{'method': 'POST',
 'headers': {'Content-Length': '0',
  'Host': 'localhost:8000',
  'User-Agent': 'python-httpx/0.12.0',
  'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate, br',
  'Connection': 'keep-alive',
  'Content-Type': 'multipart/form-data; boundary=65a4b7387ae71a5324d29bbaaa9375bd'},
 'data': {},
 'files': {}}
```

So, here again, Django interpreted the absence of `Content-Length` header as "there's no body, so I'm not reading it".

httpx/_utils.py
florimondmanca: See https://github.com/encode/httpx/pull/857#issuecomment-598063395: looks like in practice `Content-Length` is very much needed.

So my intuition here would be to fail loudly in case we can't compute the `Content-Length`.

What kind of exception would we like to throw? Surely bubbling up `OSError` isn't great UX. Would this be a new subclass of `StreamError`?

florimondmanca:
Ah so, actually, it's _us_ that default to setting `Content-Length` to `0` in the sync case currently…

https://github.com/encode/httpx/blob/a82adcc933345c6b8cb1623b031eb85723e7665b/httpx/_dispatch/urllib3.py#L94

I'll try this against #804 and see if we get the same results.

florimondmanca:
After experimenting some more with https://github.com/encode/httpx/pull/863, the `httpcore` interface does solve some problems (it doesn't necessarily force `Content-Length: 0`).

But the conclusion I reached is:

- We should include `Content-Length` if possible, as some servers will require it.
- If not possible (i.e. filelike is not `seekable()`, which is an extra edgy case anyway), include `Transfer-Encoding: chunked`, as some servers may support it.
- If not, well, let's have users wrap their weird files around `BytesIO`. :-)

florimondmanca:
@tomchristie I think this is ready for a re-review.

I've revised a bit my assumptions: it's probably okay to just load the entire contents into memory if we end up dealing with weird file-like objects that are readable but non-seekable. This means we can cover all sensible situations, with streaming effective in 99.99% of cases.

httpx/_utils.py
florimondmanca: Now stale, as we now handle this special 0.01%-of-cases situation by reading the entire file in memory as a fallback.

tests/test_multipart.py
```diff
@@ -146,38 +158,125 @@ def test_multipart_encode_files_allows_filenames_as_none():
...
+    # if it were to be consumed in full.
+    assert (
+        size_mb > max_allowed_memory - baseline
+    ), "Impact of loading entire file in memory wouldn't be detectable"
+
+    # Now verify memory usage.
+    assert all(memory < max_allowed_memory for memory in memory_per_line), (
+        max_allowed_memory,
+        memory_per_line,
+    )
```
yeraydiazdiaz: Wow, I haven't seen memory consumption being part of a test before, great job! 💯

tests/test_multipart.py
```diff
@@ -146,38 +158,125 @@ def test_multipart_encode_files_allows_filenames_as_none():
...
             "--{0}--\r\n"
             "".format(boundary).encode("ascii")
         )
+        assert stream.get_headers()["Content-Length"] == str(len(content))
+        assert b"".join(stream) == content
+
+
+def test_multipart_encode_non_seekable_filelike() -> None:
+    """
+    Test that special readable but non-seekable filelike objects are supported,
```
yeraydiazdiaz: Sorry if I missed something in the conversation but could you give an example of such case?

florimondmanca: This araised more as a theoretical aspect (a branch of the code that'd otherwise result in an exception being raised), but I've seen people being in this situation, eg on SO: [Stream a non-seekable file-like object to multiple sinks](https://stackoverflow.com/questions/39324151/stream-a-non-seekable-file-like-object-to-multiple-sinks) (unfortunately OP doesn't clarify what their non-seekable file-like object looks like). Basically this is a special edge case that theoretically we have to handle, so here we are. :-)

florimondmanca:
Hey @tomchristie, how's this looking? :-) Are we ready to merge yet?

tests/test_multipart.py
```diff
@@ -146,38 +158,125 @@ def test_multipart_encode_files_allows_filenames_as_none():
...
+
+    def bench() -> None:
+        files = {"file": open(path, "rb")}
+        stream = encode(files=files, boundary=b"+++")
+        # Consume the stream one chunk at a time.
+        for _ in stream:
+            pass
+
+    # Measure memory usage of `main()` -- one entry per LOC (plus init/teardown).
+    memory_per_line: List[float] = memory_profiler.memory_usage((bench, (), {}))
```
tomchristie: I'd suggest that we might not want to enforce memory profiling in the test case, as it might be a bit brittle, adds a requirement we could possibly do without.

florimondmanca: Yeah looking back at this PR, this test is *fancy* for sure. I still think it's nice to have some kind of proof that streaming works as expected (reduced memory consumption is the main motivation factor in the first place), but not absolute on keeping this either, since once we use generators there's not much risk we buffer everything up.

florimondmanca:
Alright, updated these on top of the HTTPCore usage that's now in `master`, should be ready for a new quick review. :+1:

florimondmanca:
Okay, thanks, merging now. :-)

PR #861, opened 2020-03-12
Fix multipart edge cases with `data={}` and/or `files={}`

### LINKED ISSUE #862: Redundant `boundary` if files/data is empty
If you try to send a post request with empty data and files — you will create a body with one `--boundary--` string, which can break the server.

It's because httpx create MultipartStream for `data={}, files={}`:
https://github.com/encode/httpx/blob/a82adcc933345c6b8cb1623b031eb85723e7665b/httpx/_content_streams.py#L332-L334

And this MultipartStream doesn't have special behavior for empty data/files:
https://github.com/encode/httpx/blob/a82adcc933345c6b8cb1623b031eb85723e7665b/httpx/_content_streams.py#L277-L283

Reproducible example:

```
from httpx.content_streams import encode

stream = encode(data={}, files={}, boundary=b"+++")
print(b"".join([part for part in stream]))
# b'--+++--\r\n'
```

I see two ways to solve it:
1. Check files not only for None but also for emptiness in `_content_streams.decode`
2. Check the count of fields in `MultipartStream.__init__` and doesn't write anything if fields are empty.

I already opened PR (#861) with a test example. If someone tells me which of these options is better, I will be happy to implement it  🙌

b0g3r:
Fix #862 

Could have conflict with #857

### DIFF
--- a/httpx/_content_streams.py
+++ b/httpx/_content_streams.py
@@ -330,7 +330,7 @@ def encode(
         else:
             return ByteStream(body=b"")
     elif isinstance(data, dict):
-        if files is not None:
+        if files:
             return MultipartStream(data=data, files=files, boundary=boundary)
         else:
             return URLEncodedStream(data=data)
--- a/httpx/_content_streams.py
+++ b/httpx/_content_streams.py
@@ -323,7 +323,7 @@ def encode(
     Handles encoding the given `data`, `files`, and `json`, returning
     a `ContentStream` implementation.
     """
-    if data is None:
+    if not data:
         if json is not None:
             return JSONStream(json=json)
         elif files:
--- a/tests/test_content_streams.py
+++ b/tests/test_content_streams.py
@@ -196,9 +196,7 @@ async def test_empty_request():
     async_content = b"".join([part async for part in stream])
 
     assert stream.can_replay()
-    assert stream.get_headers() == {
-        "Content-Length": "0",
-    }
+    assert stream.get_headers() == {}
     assert sync_content == b""
     assert async_content == b""
 
--- a/tests/test_content_streams.py
+++ b/tests/test_content_streams.py
@@ -198,7 +198,6 @@ async def test_empty_request():
     assert stream.can_replay()
     assert stream.get_headers() == {
         "Content-Length": "0",
-        "Content-Type": "application/x-www-form-urlencoded",
     }
     assert sync_content == b""
     assert async_content == b""
--- a/tests/test_content_streams.py
+++ b/tests/test_content_streams.py
@@ -190,15 +190,15 @@ async def test_multipart_data_and_files_content():
 
 
 @pytest.mark.asyncio
-async def test_empty_multipart():
-    stream = encode(data={}, files={}, boundary=b"+++")
+async def test_empty_request():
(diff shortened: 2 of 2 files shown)


florimondmanca:
Looks like we've got some unrelated type checking breakage here, issued #865 to fix it. :-)

florimondmanca:
Merged `master` in here, should resolve the build. :-)

httpx/_content_streams.py
```diff
@@ -331,7 +331,7 @@ def encode(
         else:
             return ByteStream(body=b"")
     elif isinstance(data, dict):
```
florimondmanca: Seems like we should also check for empty data here, so as to not send an empty `URLEncodedStream`, but instead an empty `ByteStream`:

```suggestion
    elif isinstance(data, dict) and data:
```

This would put us in line with what requests does:

```python
>>> import requests
>>> url = 'https://httpbin.org/post'
>>> r = requests.post(url, data={}, files={})
>>> r.request.headers
{'User-Agent': 'python-requests/2.23.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Length': '0'}
```

tests/test_content_streams.py
```diff
@@ -189,6 +189,21 @@ def hello_world():
...
+@pytest.mark.asyncio
+async def test_empty_request():
+    stream = encode(data={}, files={})
+    sync_content = b"".join([part for part in stream])
+    async_content = b"".join([part async for part in stream])
+
+    assert stream.can_replay()
+    assert stream.get_headers() == {
+        "Content-Length": "0",
+        "Content-Type": "application/x-www-form-urlencoded",
```
florimondmanca: (And so to match Requests there shouldn't be any `Content-Type` in the request. :-))

httpx/_content_streams.py
b0g3r: Implemented this idea in different line

tests/test_content_streams.py
b0g3r: I found that ByteStream.get_headers doesn't return Content-Length header with an empty body. Should I fix it?
https://github.com/encode/httpx/blob/43ec09c3cbb089edd2e83ccc717fb95beef0c189/httpx/_content_streams.py#L73-L75

b0g3r: Because the body is `bytes` (`Union[str, bytes]` transformed in `__init__` to `bytes`) this check looks redundant 🤔

florimondmanca: Looking at [RFC 7320](https://tools.ietf.org/html/rfc7230#section-3.3.2), it looks like for POST requests including a `Content-Length` always is sort of recommended, yes:

> A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body. **For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).**

So given that we only use `encode()` for multipart POST requests, I think it would make sense for us to do it. But let's treat that as a separate PR maybe? 😄

b0g3r: ✅ Done!

PR #865, opened 2020-03-13
Fix type checking breakage

florimondmanca:
Probably due to the mypy 0.770 release very recently (https://github.com/python/mypy/issues/8437)…

Discovered as this is affecting https://github.com/encode/httpx/pull/861, which was reproducible if upgrading to `mypy==0.770` locally.


florimondmanca:
Thanks all! 😅

PR #866, opened 2020-03-16
Add `Content-Length` header for empty bytestream

b0g3r:
Reference: https://github.com/encode/httpx/pull/861#discussion_r392840494


tomchristie:
I think we need to be more careful with this. Eg. we shouldn't be sending `Content-Length: 0` on GET requests.

I note the RFC text wrt *"A user agent SHOULD send a Content-Length in a request message when no Transfer-Encoding is sent and the request method defines a meaning for an enclosed payload body."* although even in that case I'd like us to *start* by validating what the equivelent behaviour in urllib3/requests would be there.

tomchristie:
I'd suggest we might want to start off with a PR reverting this change, since it improves the RFC behaviour in one case, but regresses it in another *"A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body."*

We can then take a second look at this.

tests/models/test_requests.py
```diff
@@ -10,7 +10,7 @@ def test_request_repr():
 
 def test_no_content():
     request = httpx.Request("GET", "http://example.org")
-    assert "Content-Length" not in request.headers
+    assert request.headers["Content-Length"] == "0"
```
b0g3r: We shouldn't send Content-Length with get requests
References:
 - https://github.com/encode/httpx/pull/866#issuecomment-605899948
 - https://tools.ietf.org/html/rfc7230#section-3.3.2 "A user agent SHOULD NOT ..."
 - `requests` implementation for non-stream requests https://github.com/psf/requests/blob/fd13816d015c4c90ee65297fa996caea6a094ed1/requests/models.py#L522-L533
 - `requests` implementation for stream requests https://github.com/psf/requests/blob/fd13816d015c4c90ee65297fa996caea6a094ed1/requests/models.py#L476-L479

PR #869, opened 2020-03-21
Rename `HTTPError` to `RequestError`, introduce `HTTPStatusError`

### LINKED ISSUE #867: Improving exception handling, HTTPError is misleading.
Hi,

`HTTPError` is the base class exception, it have `HTTP` while it also handle not HTTP issue, like Connection issue (NetworkError, ...) and so on.

On the other side, we have no exception defined for real HTTP client side error (4XX) and HTTP service side error (5XX).

For exemple, `requests` have `ConnectionError` (no status_code) and `HTTPError` (with status_code), both inherit from `RequestException`.
`aiohttp` offers even more precise exceptions, [all HTTP codes have an associated exceptions](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_exceptions.py):

The code end up often with ton of thing like this:

```
try:
   ...
exception httpx.HTTPError as e:
     if e.response and e.reponse.status_code:
         if e.reponse.status_code == 404:
             ... handle my expected error
         elif e.reponse.status_code == 422:
             ... handle my expected error
     raise
```

I think something like this would be clearer :
```
try:
   ...
exception httpx.HTTPNotFound as e:
     ... handle my expected error
exception httpx.HTTPUnprocessableEntity as e:
     ... handle my expected error
```

I will be happy to add this exception handling, are your interested in contribution like this ?

sileht:
## feat: introduce HTTPStatusError

Fixes #867

## feat: rename HTTPError to RequestError

Unlike HTTPStatusError, HTTPError name is not always related to HTTP status code >= 400.

The base class is renamed RequestError, to avoid confusion with
HTTPStatusError.

Related #867


sileht:
I don't see any automated CI here, there is plan to add one ? Can I help ?

florimondmanca:
@sileht Seems like [Travis is under scheduled maintenance right now](https://www.traviscistatus.com/):

>  We will be undergoing scheduled maintenance. During this time, your builds triggered via GitHub will still be accepted, but all public services (including api.travis-ci.org) cannot be reached. Any scheduled or running builds will be canceled and re-enqueued to begin running after the maintenance window. 

It seems your build should start automatically once maintenance is over. :-)

florimondmanca:
Going to try and trigger the Travis CI build...

sileht:
> Any places in the docs that need updating?

That's surprise me, but I didn't find any reference of HTTPError in the documentation.

httpx/_exceptions.py
```diff
@@ -17,10 +17,12 @@ def __init__(
         super().__init__(*args)
 
 
+HTTPError = RequestError  # For backward compatibility purpose
```
florimondmanca: Tiny copy nit:

```suggestion
HTTPError = RequestError  # For backward compatibility purposes.
```

sileht:
I added a commit with some basic doc.

florimondmanca:
> I added a commit with some basic doc.

Ideally, can we move this to a separate PR? In general we try to have one PR do one thing, but this one PR already does _two_ things… :-)

sileht:
I have split it.

httpx/_exceptions.py
```diff
@@ -59,22 +61,28 @@ class ProxyError(HTTPError):
...
-class DecodingError(HTTPError):
+class DecodingError(RequestError):
     """
     Decoding of the response failed.
     """
 
 
+class HTTPStatusError(RequestError):
+    """
+    HTTP Status Error.
```
florimondmanca: Probably worth being a bit more precise about *when* this might be raised?

```suggestion
    Called `.raise_for_status()` on an HTTP response with an error status code.
```

florimondmanca:
I'm going to close this for now — there's been some changes and improvements since March in the realm of the exception hierarchy, and I think we might need to properly discuss and agree on the changes proposed here beforehand. Filed #1064.

Either way, @sileht _thank you_ so much for putting time into this. This is definitely super valuable and I'm sure we'll be able to reuse this very soon!

PR #874, opened 2020-03-24
Added params json/data to .delete() method

sfilatov96:


florimondmanca:
Hi!

Thanks for this contribution, but we purposefully do not support request body parameters on DELETE due to non-standard semantics. This is documented here along with a workaround if you really need to send a body: https://www.python-httpx.org/compatibility/#request-body-on-http-methods. See also #408 for background.

Thanks!

PR #741, opened 2020-01-08
Add an Issue Template

StephenBrown2:
This is merely a starting suggestion to provide an issue template for the repository, based on https://github.com/encode/httpx/issues/740#issuecomment-572202571

What do y'all think?

PR #880, opened 2020-03-28
Add issue templates

florimondmanca:
A second iteration over #741, with templates for questions, bug reports and feature requests.

I think with these in place we can hope to reduce maintainer burden, as contributors will *think* more thoroughly about alternatives to opening an issue, possible solutions, etc.

This should help improve the contributing experience for all. :-)

(I based these off of the default GitHub templates, as well as those we can find on the [FastAPI repo](https://github.com/tiangolo/fastapi/issues/new/choose).)

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-question.md
+++ b/.github/ISSUE_TEMPLATE/1-question.md
@@ -1,0 +1,17 @@
+---
+name: Question
+about: Ask a question
+---
+
+### Checklist
+
+<!-- To help keep this issue tracker clean and focused, please make sure you tried *all* the following resources before submitting your question. -->
+
+- [ ] I looked through similar issues on GitHub, but didn't find anything.
+- [ ] I searched the [HTTPX documentation](https://www.python-httpx.org).
+- [ ] I looked up "How to do ... in HTTPX" on a search engine and didn't find any information.
+- [ ] I asked the [community chat](https://gitter.im/encode/community) for help but didn't get an answer.
+
+### Question
+
+<!-- Please ask your question here. -->
--- a/.github/ISSUE_TEMPLATE/1-question.md
+++ b/.github/ISSUE_TEMPLATE/1-question.md
@@ -7,8 +7,8 @@ about: Ask a question
 
 <!-- To help keep this issue tracker clean and focused, please make sure you tried *all* the following resources before submitting your question. -->
 
-- [ ] I looked through similar issues on GitHub, but didn't find anything.
 - [ ] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
+- [ ] I looked through similar issues on GitHub, but didn't find anything.
 - [ ] I looked up "How to do ... in HTTPX" on a search engine and didn't find any information.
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/1-question.md
```diff
@@ -0,0 +1,17 @@
...
+name: Question
+about: Ask a question
+---
+
+### Checklist
+
+<!-- To help keep this issue tracker clean and focused, please make sure you tried *all* the following resources before submitting your question. -->
+
+- [ ] I looked through similar issues on GitHub, but didn't find anything.
+- [ ] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
```
yeraydiazdiaz: Searching the docs should go first, I think.

```suggestion
- [ ] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
- [ ] I looked through similar issues on GitHub, but didn't find anything.
```

.github/ISSUE_TEMPLATE/2-bug-report.md
```diff
@@ -0,0 +1,86 @@
...
+```python
+import httpx
+
+r = httpx.get("http://localhost:8000")
+print(r)
+```
+
+4. Run the test script.
+5. It raises an error.
+6. But I expected it to successfully print the response.
```
yeraydiazdiaz: This is great but I wonder if we should add a link to these instructions instead, maybe in the docs?

I just worry it might be too much information for someone coming in.

florimondmanca: I'm really not sure where this would be in the docs… Also, having these instructions away from the issue template doesn't sound great in terms of keeping these up to date…

The motivation for providing an example of a repro case is because a lot of the bug reports we get assume that attaching a script is a repro example… skipping the important part that it shouldn't have any external dependencies.

I updated the template to drop the code example, and add a bit more prose with a focus on "keep dependencies at a minimum". :-)

.github/ISSUE_TEMPLATE/2-bug-report.md
```diff
@@ -0,0 +1,61 @@
...
+<!-- A clear and concise description of what the bug is. -->
+
+### To reproduce
+
+<!-- Provide a *minimal* example with steps to reproduce the bug locally.
+
+NOTE: try to keep any external dependencies *at an absolute minimum* (servers, proxies, certificates...).
+In other words, remove anything that doesn't make the bug go away.
+
+If you need a local server to replicate against, you can build one using eg. Starlette [0] and Uvicorn [1], or any tool you feel comfortable with. Remember to share setup instructions here. :-)
```
yeraydiazdiaz:
```suggestion
If you need a local server to replicate against, you can build one using eg. Starlette [0] and Uvicorn [1], or any tool you feel comfortable with. Check out other issues for examples and remember to share setup instructions here. :-)
```

.github/ISSUE_TEMPLATE/2-bug-report.md
yeraydiazdiaz: Yeah, the docs is probably not the right place for that. 👍

Issue #789, opened 2020-01-22
On type annotation style.


tomchristie:
In large portions of our Client code we're creating type shorthands like `TimeoutTypes` and then using those everywhere, rather than listing a more verbose form, such as `Union[None, int, Sequence[int], Timeout]`.

Although that gives us a more concise style I think it's also got some disadvantages...

It makes it harder to figure out what the type annotation *actually is*, and it leads to a bunch of type classes that are not public API, but which users might end up trying to treat as such.

I think it might well be worth us exploring what the codebase would look like if we tries to use more explicit annotation styles throughout.

```python
class BaseClient:
    def __init__(
        self,
        *,
        auth: AuthTypes = None,
        params: QueryParamTypes = None,
        headers: HeaderTypes = None,
        cookies: CookieTypes = None,
        timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
        max_redirects: int = DEFAULT_MAX_REDIRECTS,
        base_url: URLTypes = None,
        trust_env: bool = True,
    ):
```

Would be...

```python
DEFAULT_TIMEOUT_CONFIG = 5
DEFAULT_MAX_REDIRECTS = 20

class BaseClient:
    def __init__(
        self,
        *,
        auth: Union[Tuple[AnyStr, AnyStr], Callable[[Request], Request], Auth] = None,
        params: Union[AnyStr, Dict[AnyStr, Any], List[Tuple[AnyStr, Any]], QueryParams] = None,
        headers: Union[Dict[AnyStr, AnyStr], List[Tuple[AnyStr, AnyStr]], Headers] = None,
        cookies: Union[Dict[AnyStr, AnyStr], Cookies, CookieJar] = None,
        timeout: Union[None, float, Sequence[float], Timeout] = DEFAULT_TIMEOUT_CONFIG,
        max_redirects: int = DEFAULT_MAX_REDIRECTS,
        base_url: Union[AnyStr, URL] = None,
        trust_env: bool = True,
    ):
```

On one hand it's much more verbose, and we would end up needing to change annotations in multiple places when they change. On the other hand we're not having to do much tweaking now since we're fairly settled on the public API, and it at least makes the type annotations explcit up front and centre, without introducing any indirect "we're defining this this elsewhere" annotations.

[...]

PR #881, opened 2020-03-28
Move type definitions to a dedicated module

florimondmanca:
Refs https://github.com/encode/httpx/issues/789#issuecomment-578143085

Also improves/fixes the typing information in the multipart stream implementation, as some of it was actually wrong.


httpx/_content_streams.py
```diff
@@ -298,8 +281,8 @@ def iter_fields(
             else:
                 yield self.DataField(name=name, value=value)
 
-        for name, value in files.items():
-            yield self.FileField(name=name, value=value)
+        for name, filevalue in files.items():
+            yield self.FileField(name=name, value=filevalue)
```
florimondmanca: Different variable name to `value` is required of mypy would complain that the `value` in `data` doesn't have the same type as the `value` in `files` (but ideally we don't want to just use `Any` here).

florimondmanca:
Realizing that this might become largely reduced in scope once #804 is in, so let's wait…

florimondmanca:
Closed in favor of https://github.com/encode/httpx/pull/902

PR #882, opened 2020-03-28
Drop usage of `AnyStr`

florimondmanca:
Extracted of #881 to reduce its scope.

We're using `AnyStr` in several places to refer to "either bytes or str", i.e. `Union[str, bytes]`.

But `AnyStr` is not a synonym for `Union[str, bytes]`, so we should really just be using `Union[str, bytes]`.

(For details: `AnyStr` is meant to be used when we can accept either, but don't allow them to mix, so eg `(s1: AnyStr, s2: AnyStr) -> AnyStr)` would either accept str/str and return str, or accept bytes/bytes and return bytes. See [typing docs](https://docs.python.org/3/library/typing.html#typing.AnyStr).)

This PR introduces a `StrOrBytes` util type to use across the codebase.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -39,6 +39,7 @@ from ._exceptions import (
     StreamConsumed,
 )
 from ._status_codes import StatusCode
+from ._types import StrOrBytes
 from ._utils import (
     ElapsedTimer,
     flatten_queryparams,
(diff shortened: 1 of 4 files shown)


httpx/_models.py
```diff
@@ -67,8 +68,8 @@
 
 HeaderTypes = typing.Union[
     "Headers",
-    typing.Dict[typing.AnyStr, typing.AnyStr],
-    typing.List[typing.Tuple[typing.AnyStr, typing.AnyStr]],
+    typing.Dict[StrOrBytes, StrOrBytes],
+    typing.Sequence[typing.Tuple[StrOrBytes, StrOrBytes]],
```
florimondmanca: Note: this change from `List` to `Sequence` was required, or mypy would raise errors here about `List` being invariant (so not working well with unions of types):

https://github.com/encode/httpx/blob/21f533774f87292c2381376eb40f3216a48eee06/httpx/_models.py#L492

PR #883, opened 2020-03-28
Fix Client Docstring and add AsyncClient to API docs

### LINKED ISSUE #879: Sync client fails on ASGI apps with unhelpful traceback
I tried using httpx with with an ASGI app in the way that was implied by the docs.  It throws an error that feels like a problem with httpx, as FastAPI is currently being used in production in other ways.  If you confirm that this is a bug in httpx, I'm willing (and probably capable) to fix it and submit a patch.

```
import httpx
from fastapi import FastAPI

app = FastAPI()


@app.get("/users/{user_id}")
async def read_user(user_id: str):
    return {'user_id': user_id}

with httpx.Client(app=app) as client:
    response = client.get("http://www.example.com/users/foo")
```

Throws the error:
```
  File "temp.py", line 12, in <module>
    response = client.get("http://www.example.com/users/foo")
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 714, in get
    timeout=timeout,
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 571, in request
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 591, in send
    request, auth=auth, timeout=timeout, allow_redirects=allow_redirects,
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 618, in send_handling_redirects
    request, auth=auth, timeout=timeout, history=history
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 654, in send_handling_auth
    response = self.send_single_request(request, timeout)
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_client.py", line 678, in send_single_request
    response = dispatcher.send(request, timeout=timeout)
  File "/Users/singletoned/.envs/testino/lib/python3.7/site-packages/httpx/_dispatch/wsgi.py", line 90, in send
TypeError: __call__() missing 1 required positional argument: 'send'
```

Singletoned:
The Client docstring incorrectly says that it takes an ASGI app when it actually takes a WSGI app.

Also the AsyncClient's API isn't in the docs anywhere, so it is difficult to tell that you should use that instead.

It is arguable that the API of the AsyncClient should be on the Async client, and a link to it could be on the API page, but I think it makes more sense to have all the API in one place.  Happy to be corrected on this.

Fixes #879

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,6 +38,8 @@
     :docstring:
     :members: headers cookies params request get head options post put patch delete build_request send close
 
+## `AsyncClient`
+
 ::: httpx.AsyncClient
     :docstring:
     :members: headers cookies params request get head options post put patch delete build_request send
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,6 +38,11 @@
     :docstring:
     :members: headers cookies params request get head options post put patch delete build_request send close
 
+::: httpx.AsyncClient
+    :docstring:
+    :members: headers cookies params request get head options post put patch delete build_request send
+
+
 ## `Response`
 
 *An HTTP response.*
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -426,7 +426,7 @@ class Client(BaseClient):
     request URLs.
     * **dispatch** - *(optional)* A dispatch class to use for sending requests
     over the network.
-    * **app** - *(optional)* An ASGI application to send requests to,
+    * **app** - *(optional)* An WSGI application to send requests to,
     rather than sending actual network requests.
     * **trust_env** - *(optional)* Enables or disables usage of environment
     variables for configuration.


docs/api.md
```diff
@@ -38,6 +38,11 @@
     :docstring:
     :members: headers cookies params request get head options post put patch delete build_request send close
 
+::: httpx.AsyncClient
```
florimondmanca: Let's put this in a dedicated section maybe to make it more easily discoverable in the navigation?

```suggestion
## `AsyncClient`

::: httpx.AsyncClient
```

Screenshot 2020-03-28 at 15 44 38

florimondmanca:
Deployed :+1: https://www.python-httpx.org/api/#asyncclient

PR #884, opened 2020-03-28
doc: add some documentation about raised exceptions

sileht:


docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
 >>> httpx.get("https://example.com", auth=auth)
 <Response [200 OK]>
 ```
+
+## Errors and Exceptions
+
+In some situations, HTTPX can raise exceptions:
+
+* `NetworkError` in case of network problem (e.g. DNS failure, refused connection, ...)
```
florimondmanca:
```suggestion
* `NetworkError` in case of a network problem (e.g. DNS failure, refused connection, ...).
```

docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
...
+* `ConnectTimeout`, `ReadTimeout`, `WriteTimeout`, `PoolTimeout` in case of a configured timeout is reached. All of them inherit from `TimeoutException`
```
florimondmanca:
```suggestion
* `ConnectTimeout`, `ReadTimeout`, `WriteTimeout`, `PoolTimeout` in case a configured timeout is reached. All of them inherit from `TimeoutException`.
```

docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
...
+* `TooManyRedirects` if a request exceeds the configured number of maximum redirections
```
florimondmanca:
```suggestion
* `TooManyRedirects` if a request exceeds the configured number of maximum redirections.
```

docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
...
+
+Also when `Response.raise_for_status()` is called HTTPX will also `HTTPStatusError` if the HTTP request returned status code >= 400.
```
florimondmanca: We can probably include this one along with the others:

```suggestion
* `HTTPStatusError` when calling `.raise_for_status()` on a rsponse with status code 4xx or 5xx.

```

docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
...
+
+All exceptions that HTTPX explicitly raises inherit from httpx.RequestError and contains the request and the response if any.
```
florimondmanca: A suggestion to clarify that the request is _always_ included (but not the response):

```suggestion
All exceptions that HTTPX raises inherit from `httpx.RequestError` and contain the request, as well as any receved response.
```

docs/quickstart.md
```diff
@@ -451,3 +451,26 @@ as above:
...
+
+```python
+>>> try:
+>>>     httpx.get("https.example.com")
+>>> except httpx.HTTPStatusError as exc:
+>>>     print(f"The request `{exc.request}` return the status code {exc.request.status_code}")
+>>> except TimeoutException as exc:
+>>>     print(f"The request `{exc.request}` have reached the timeout")
+>>> except NetworkError as exc:
+>>>     print(f"The request `{exc.request}` have failed")
```
florimondmanca: Not sure we *need* the interpreter prompts in this snippet, WDYT?

```suggestion
try:
    httpx.get("https.example.com")
except httpx.HTTPStatusError as exc:
    print(f"The request `{exc.request}` return the status code {exc.request.status_code}")
except TimeoutException as exc:
    print(f"The request `{exc.request}` have reached the timeout")
except NetworkError as exc:
    print(f"The request `{exc.request}` have failed")
```

pawamoy:
If I understand correctly, half the exceptions are now in httpcore, so #869 and this PR should be rebased on master before being worked on more.

In the meantime, I have to catch `Exception` because httpcore's errors and httpx's errors don't share the same base class, right?

tomchristie:
Not quite, no.

HTTPX should only ever raise exceptions from its own package space.

We should always wrap httpcore exceptions, and if there’s anywhere we’re not currently doing that, then we should consider it a bug.

pawamoy:
Oh okay, great, thanks for the clarification!

florimondmanca:
Closing similarly to #869 — these changes are super valuable but right now they're assuming we'd have already made the switch to `RequestError` + `HTTPStatusError`. I think while #1064 gets sorted out we can very much have docs about the current state of things — which improved since this PR was opened — which is "`HTTPError` as base + specific exceptions for specific cases". Thanks!

PR #887, opened 2020-03-28
Fix support for generator-based WSGI apps

Singletoned:
The WSGI adapter was trying to take the status code from `start_response` straightaway, rather than waiting until the first non-empty chunk, as per the spec: https://www.python.org/dev/peps/pep-3333/#the-start-response-callable

I've made a small change that iterates through the response until it gets a non-empty chunk, and then passes everything on as normal.  I've also added a test for that scenario.


httpx/_dispatch/wsgi.py
```diff
@@ -7,6 +8,19 @@
 from .base import SyncDispatcher
 
 
+def _get_non_empty_chunk(body: typing.Iterable) -> typing.Iterable:
```
florimondmanca: Would this be a more appropriate name for what this does, i.e. skip any empty chunks in the `body`?

```suggestion
def _seek_non_empty_chunk(body: typing.Iterable) -> typing.Iterable:
```

httpx/_dispatch/wsgi.py
```diff
@@ -7,6 +8,19 @@
...
+    """
+    Get a non-empty chunk from body. This is needed because the status returned
+    by start_response shouldn't be used until the first non-empty chunk has been
+    served.
```
florimondmanca: Nit: can we move this comment to when we actually call this function? (What this function does is quite clear, but it might be better to keep the motivation close to the actual usage.)

tests/client/test_client.py
```diff
@@ -153,3 +153,54 @@ def test_elapsed_delay(server):
...
+    def application(env, start_response):
+        body = [b"", b"", b""]
+        start_response("200 OK", [("Content-Type", "text/html")])
+        for line in body:
+            yield line
+
+    with httpx.Client(app=application) as client:
+        response = client.get("http://example.com/")
+        assert response.status_code == 200
+        assert response.content == b""
```
florimondmanca: FYI, we already have a `tests/test_wsgi.py` file. :-)

It doesn't seem like we need more tests there — we'd just need to refactor `echo_body_with_response_stream` to be an actual generator (instead of calling `start_response` synchronously and *then* returning a generator), which would match eg the `test_wsgi_app_generator` case here:

https://github.com/encode/httpx/blob/fc980e779276e46ce2e2aad10a8ced1d50334306/tests/test_wsgi.py#L36-L50

httpx/_dispatch/wsgi.py
Singletoned: Sure.  The style where I work is to avoid comments, so it's just a habit.

tests/client/test_client.py
Singletoned: I'll move them.  Though I'm a little surprised the tests aren't in `tests/client/`.  I assumed the test structure would roughly match the code structure.

florimondmanca: Yeah it's a bit surprising. Personally I'd see them in `tests/dispatch/` since `asgi.py` and `wsgi.py` are in the `_dispatch` sub-package. :-) Happy to review any follow-up PR for this!

Singletoned:
I removed the "Content-length" header from the tests because it isn't actually used anywhere in the code, and if people look at the tests as an example, it might be misleading.

httpx/_dispatch/wsgi.py
```diff
@@ -7,6 +8,15 @@
 from .base import SyncDispatcher
 
 
+def _get_non_empty_chunk(body: typing.Iterable) -> typing.Iterable:
+    """Get a non-empty chunk from body."""
+    body = iter(body)
+    for chunk in body:
+        if chunk:
+            return itertools.chain([chunk], body)
+    return []
```
yeraydiazdiaz: From the PEP:

> In other words, response headers must not be sent until there is actual body data available, or until the application's returned iterable is exhausted. *(The only possible exception to this rule is if the response headers explicitly include a Content-Length of zero.)*

Is it worth explictly checking for the Content-Length to be zero inrrespective of the body also being empty?

I'm not sure if that's something that can ever happen though, not super familiar with WSGI myself.

Singletoned: My reading of this is that if a Content-Length of zero is set, then you CAN skip checking for body data, not that you SHOULD skip checking for body data (eg for performance reasons).  It's not something I've ever bothered to put in a WSGI adapter before, though my use cases are always quite lightweight (I mainly use it for testing web apps, which is what I'm doing with httpx).

It would mitigate against a situation where the app returned an infinite empty iterator, as long as it had a Content-Length of zero.  You would still have to take the first value of the iterator though, eg when a generator is being used.

Happy to add it, as it's a tiny amount of code.  It's really your decision over how you want the library to behave.

yeraydiazdiaz: I guess it might be worth it while we're at it but I'd like to hear @tomchristie and @florimondmanca's opinion.

Thanks again @Singletoned

florimondmanca: Feels like a super narrow edge case to me… I read the paragraph in the PEP as "in case of `Content-Length: 0` then you *don't have to* check for body data" too. So maybe we can leave things as they are for now, in the name of keeping things simple while covering 99.99% of use cases?

httpx/_dispatch/wsgi.py
```diff
@@ -7,6 +8,15 @@
 from .base import SyncDispatcher
 
 
+def _get_non_empty_chunk(body: typing.Iterable) -> typing.Iterable:
+    """Get a non-empty chunk from body."""
```
florimondmanca: I notice you reverted the name, but I still think the current "get chunk" naming is not conveying correctly what this function does… A "get chunk" function should return a `chunk`, but this returns a modified `body` iterable that skips any leading empty chunks.

So, maybe…?

```suggestion
def _skip_leading_empty_chunks(body: typing.Iterable) -> typing.Iterable:
```

httpx/_dispatch/wsgi.py
Singletoned: I guess a broader question here is, in general, do you want httpx to be:

  a) as strict as possible (tightly conforming to specs)
  b) as permissive as possible (eg handle broken responses, etc)
  c) as simple as possible (least code complexity)

Given that it will be used to fetch things on the web, I would guess "b" is the answer, but any of them are valid.  If you do have a clear answer, I'd suggest putting it in the "contributing" docs.

Under either "a" or "b" I'd suggest that this should be added.  It conforms to the spec, and it also allows broken cases to succeed.  Under "c" then it's not really needed.

tests/test_wsgi.py
```diff
@@ -93,3 +93,21 @@ def test_wsgi_exc():
...
         client.get("http://www.example.org/")
+
+
+def test_wsgi_generator():
+    output = [b"", b"", b"Some content", b" and more content"]
+    client = httpx.Client(app=application_factory(output))
+    response = client.get("http://www.example.org/")
+    assert response.status_code == 200
+    assert response.text == "Some content and more content"
+    assert response.content == b"Some content and more content"
```
florimondmanca: Do we actually *need* to assert both the `text` and `content`? We're not testing body decoding, so I feel like asserting the `text` (like in other tests here) should be enough, WDYT?

```suggestion
```

tests/test_wsgi.py
```diff
@@ -93,3 +93,21 @@ def test_wsgi_exc():
...
+
+
+def test_wsgi_generator_empty():
+    output = [b"", b"", b"", b""]
+    client = httpx.Client(app=application_factory(output))
+    response = client.get("http://www.example.org/")
+    assert response.status_code == 200
+    assert response.text == ""
+    assert response.content == b""
```
florimondmanca: Same note here:

```suggestion
```

httpx/_dispatch/wsgi.py
florimondmanca: I think for areas like these where neither urllib3 nor Requests support the feature, we're more on a pragmatic variant of c).

If there's something obviously broken we should fix it. But so far we haven't seen any user issues about "my WSGI server returns an infinite stream of empty chunks and HTTPX can't process it", which is why I'd be advocating for "let's fix this only if/when we have actual proof users are encountering it in the wild". :-)

httpx/_dispatch/wsgi.py
Singletoned: That was a regression on my part.  Happy to accept any name whatsoever.

florimondmanca: Sure, sorry if my wording sounded offensive if any way :) I meant to write "I notice the name was reverted".

httpx/_dispatch/wsgi.py
```diff
@@ -88,6 +97,9 @@ def start_response(
             seen_exc_info = exc_info
 
         result = self.app(environ, start_response)
+        # This is needed because the status returned by start_response
+        # shouldn't be used until the first non-empty chunk has been served.
+        result = _get_non_empty_chunk(result)
```
florimondmanca: Woops 😄  (Forgot to add this suggestion to my previous review!)

```suggestion
        result = _skip_leading_empty_chunks(result)
```

PR #888, opened 2020-03-28
Remove stitches in `requirements.txt`

florimondmanca:
We had a few hot fixes in our `requirements.txt` file due to transient failures in certain of our test dependencies — dropping them as these failures should be fixed by now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,3 +22,5 @@ trio
 trustme
 uvicorn
 seed-isort-config
+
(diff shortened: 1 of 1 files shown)


florimondmanca:
Tried to drop the `attrs` version pin *again* (already tried in #641), but the issue doesn't seem to be resolved yet. Still not sure what's causing it… Anyway, put it back for now.

PR #890, opened 2020-03-30
Test on Python 3.9-dev

hugovk:
Python 3.9 is due for beta release in May and full release in October:

* https://www.python.org/dev/peps/pep-0596/#schedule

Testing `3.9-dev` on Travis CI can help avoid any surprises.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ python:
   - 3.6
   - 3.7
   - 3.8
+  - 3.9-dev
 
 stages:
   - check


florimondmanca:
Thanks! Not sure what’s going on, but the Travis build is not being reported, but the output is here (and it’s ✅): https://travis-ci.org/github/encode/httpx/builds/668858500

.travis.yml
```diff
@@ -11,6 +11,7 @@ python:
   - 3.6
   - 3.7
   - 3.8
+  - 3.9-dev
```
florimondmanca: I’m actually wondering if we could do without this until 3.9 reaches beta? With 3mo until the final release we’ll have lots of time to report back any issues. At this point we’re trying to keep maintenance overhead as low as possible in this project. (We could add 3.9 to allow_failures but it would defeat the point as builds wouldn’t report as errors...)

hugovk: Sure, it can wait. At least it's done one run now and nothing untoward happened. Feel free to close this.

florimondmanca: Yes, it’s good to know that we’re not currently broken. :) Thank you!

PR #891, opened 2020-03-31
single input multiple files enabled

### LINKED ISSUE #777: Support for multiple files per POST field
```
async def http_upload():
    async with httpx.AsyncClient(verify=False) as client:
        data_path = os.path.join(PARENT_PATH, 'data_file/blob/data')
        index_path = os.path.join(PARENT_PATH, 'data_file/blob/index')
        files = {
            "upload_file": [
                open(data_path, 'rb'),
                open(index_path, 'rb'),
            ]
        }
        # files = [
        #     ("upload_file", open(data_path, 'rb')),
        #     ("upload_file", open(index_path, 'rb'))
        # ]
        params = {
            "requestId": "10001",
            "sdkVersion": "V1.2.3",
            "area": "CH"
        }

        response = await session.post(url, data=params, files=files)
```

```console
self = <httpx.content_streams.MultipartStream.FileField object at 0x7f503c22ea60>

        if isinstance(self.file, str):
            content = self.file
        else:
>           content = self.file.read()
E           AttributeError: 'list' object has no attribute 'read'

or
    def iter_fields(
        self, data: dict, files: dict
    ) -> typing.Iterator[typing.Union["FileField", "DataField"]]:
            if isinstance(value, (list, dict)):
                for item in value:
                    yield self.DataField(name=name, value=item)
            else:

>       for name, value in files.items():
E       AttributeError: 'list' object has no attribute 'items'
```

Excuse me, how should I use。

tegkhanna:
Added support for multiple files through a single input!

Fixes #777


tegkhanna:
Requesting review @florimondmanca

demonno:
I am having same issue, and saw there is some progress already, based on discussion here #777 problem is already clear and implementation [suggestion](https://github.com/encode/httpx/issues/777#issuecomment-584545078) is also made.

Expecting to get similar API to requests [post-multiple-multipart-encoded-files](https://2.python-requests.org/en/master/user/advanced/#post-multiple-multipart-encoded-files)

```python
>>> url = 'https://httpbin.org/post'
>>> multiple_files = [
...     ('images', ('foo.png', open('foo.png', 'rb'), 'image/png')),
...     ('images', ('bar.png', open('bar.png', 'rb'), 'image/png'))]
>>> r = httpx.post(url, files=multiple_files)
>>> r.text
```

@tegkhanna your [suggestion](https://github.com/encode/httpx/pull/891/files#diff-9f6d9e5c7bf266543d5eb915120ab285R153) in this Pull Request is different from above example.

 Why don't not you follow [this suggestion](https://github.com/encode/httpx/issues/777#issuecomment-584545078), did it have any issues?

httpx/_content_streams.py
```diff
@@ -276,7 +276,7 @@ def __init__(self, data: dict, files: dict, boundary: bytes = None) -> None:
         body = BytesIO()
         if boundary is None:
             boundary = binascii.hexlify(os.urandom(16))
-
+
```
florimondmanca, line 279: Careful:

```suggestion

```

(This made CI fail, you can run code formatting locally with `$ scripts/lint`.)

florimondmanca:
Went ahead and moved the `Fixes #777` note to the PR description, that way #777 is automatically linked. :-)

tegkhanna:
@florimondmanca Hi! I will commit the changes soon :) Actually, I was confused that should I add the list of tuples support as it was different from the core API. Should I add the support for `data = list of (fieldname, value) item` as well?



> Went ahead and moved the `Fixes #777` note to the PR description, that way #777 is automatically linked. :-)

Thanks for that. I thought it gets linked by adding to the `Fixes` keyword to the title!

florimondmanca:
@tegkhanna 

> Should I add the support for data = list of (fieldname, value) item as well?

I don't think so no, currently `data` only supports either string-like or a `dict`:

```python
RequestData = typing.Union[
    dict, str, bytes, typing.Iterator[bytes], typing.AsyncIterator[bytes]
]
```

Not obvious yet that we need anything else, but if we do then in any case this should be a different issue/PR (as this one is about files).

tegkhanna:
@florimondmanca One last question, should I then drop the support for multiple files if they are given as part of a dict `{fieldName : [list of file objects]}` or should I keep it as well?

florimondmanca:
Yes, let’s drop it in favor of the list of tuple style. :)

httpx/_content_streams.py
```diff
@@ -15,21 +15,44 @@
...
 
-RequestFiles = typing.Dict[
-    str,
-    typing.Union[
-        # file (or str)
-        typing.Union[typing.IO[str], typing.IO[bytes], StrOrBytes],
-        # (filename, file (or str))
-        typing.Tuple[
-            typing.Optional[str],
+RequestFiles = typing.Union[
```
florimondmanca, line 18: Let's maybe rewrite using a new type alias for the file details…

```python
FileEntry = typing.Union[
    # file (or str)
    # (filename, file (or str))
    # (filename, file (or str), content_type)
]

RequestFiles = typing.Union[
    typing.Dict[str, FileEntry],
    typing.List[typing.Tuple[str, FileEntry]]
]
```

tegkhanna:
@florimondmanca Sure I will make another PR for the data part! I have added the required change for filetype reusability to this PR :)

tegkhanna:
@florimondmanca Should I add the data list changes to this PR only? Because a separate PR will cause conflicts if the changes are based on the master

florimondmanca:
Yes I suppose so... sure, you can add those changes here if that helps. :)

tegkhanna:
The PR is complete. Files and data both now accepts a list of tuples along with a dict.

florimondmanca:
Closing this one in favor of #1032, thank you so much for the time you've put into this, @tegkhanna!

PR #892, opened 2020-04-04
Added adapter to convert requests.Session to httpx.AsyncClient

subhadeepdas91:
As httpx supports requests compatible API, it would be great to convert requests Session to AsyncClient, this will results in easy migrations / interoperability between these two libraries. 

To achieve this conversion this adapter will take requests session as argument as returns AsyncClient

following is a snippet of usage
```
async with httpx.from_requests(session) as client:
```

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -29,6 +29,7 @@ from ._exceptions import (
 )
 from ._models import URL, Cookies, Headers, QueryParams, Request, Response
 from ._status_codes import StatusCode, codes
+from ._requests import from_requests
 
 __all__ = [
     "__description__",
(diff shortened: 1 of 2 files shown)


florimondmanca:
Hi,

Thanks for opening this! While I appreciate the effort I’m personally tempted to push back a little on this...

Since this most likely wouldn’t be a « one size fits all » solution (eg what about the sync Client? what about retries? ...) this would most likely be better served and much easier to evolve and manage for everyone involved as a third-party package, rather than built into core.

Happy to hear @tomchristie’s and others’ thoughts on this, but that would be where I stand. :)

subhadeepdas91:
@florimondmanca Appreciate you decision, looking forward for future supports for requests Session, by the way this library will help a lot in migrating my project to asyncio, kudos to httpx.

PR #900, opened 2020-04-10
feat: allow to pass payload to GET/DELETE

sileht:
Even it's not common, there is no reason to forbid payload for DELETE
and GET.


florimondmanca:
Hi!

Thanks, but this is actually a deliberate choice. Request bodies on GET and DELETE are non-standard in HTTP. See eg. https://github.com/encode/httpx/issues/395#issuecomment-535975002.

As documented [here](https://www.python-httpx.org/compatibility/#request-body-on-http-methods), you should instead use:

```python
httpx.request('GET', data=...)
httpx.request('DELETE', data=...)
```

I'm giving here the same answer than in #874.

Mz49:
I understand it's not encouraged, but according to some docs I've read, notably [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE) and the relevant [RFC section](https://tools.ietf.org/html/rfc7231#section-4.3.5), it's still allowed and up to the server on how the response should be handled. While even to me it wouldn't make sense to send a request body on a GET request, I can see how a request body could be useful in DELETE requests. I'm glad this topic has been raised already, but I think having a request body allowed on the `delete()` method would be alright to have. If that's still a no, that's unfortunate but the `request()` will work for now.

PR #902, opened 2020-04-10
Move types definitions to a dedicated module

florimondmanca:
Refs https://github.com/encode/httpx/issues/789#issuecomment-578143085

Move all type definitions and aliases to a centralized `_types` module, instead of them being scattered across the code base. In line with what we do in HTTPCore. :-)

A simpler version of #881, branched off from `master`, and without any typing changes - only import changes.

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -1,17 +1,19 @@
 import typing
 
-from ._auth import AuthTypes
 from ._client import Client, StreamContextManager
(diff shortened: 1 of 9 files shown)


florimondmanca:
@yeraydiazdiaz Would you like to give this one some eyes? This is a bit of parallel work than what's been going on in HTTPCore. :-)

florimondmanca:
No worries, thanks!

PR #903, opened 2020-04-11
Headers dictionary views

### LINKED ISSUE #899: Should `Headers.keys` be a set-like keys view?
### Checklist

- [X] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
- [X] I looked through similar issues on GitHub, but didn't find anything.
- [ ] I looked up "How to do ... in HTTPX" on a search engine and didn't find any information.
- [ ] I asked the [community chat](https://gitter.im/encode/community) for help but didn't get an answer.

### Question

presently [`Headers.keys`](https://github.com/encode/httpx/blob/master/httpx/_models.py#L458) returns a `typing.List[str]` rather than a (set-like) `typing.KeysView[str]`, which appears to violate the `Mapping` protocol: https://docs.python.org/3/library/collections.abc.html#collections.abc.KeysView

this breaks code that expects set-like operations with `Headers.keys`, e.g.
```
    dst: Dict[str, str] = ...
    headers: httpx.Headers = ...
    keys: Set[str] = {"set-cookie", "set-cookie2"}
    dst.update({k: headers[k] for k in keys & headers.keys()})
```

is this an intentional design choice or simply an oversight?

florimondmanca:
Fixes #899

The goal of this PR is to modify `headers.keys()`, `headers.values()`, `headers.items()` so that they return [dictionary views](https://docs.python.org/3/library/stdtypes.html#dict-views), as per the `Mapping` protocol.

(Currently they return lists.)

This puts `Headers` in line with `QueryParams`, which already return views built from the underlying `dict`.

Header views are a bit more complex because we need to deal with multi-items and case-insensitive keys.


httpx/_models.py
```diff
@@ -401,6 +401,72 @@ class Headers(typing.MutableMapping[str, str]):
...
+
+        def __contains__(self, key: typing.Any) -> bool:
+            try:
+                normalized_key = key.lower().encode(self._encoding)
+            except AttributeError:
+                return False
+            return any(k == normalized_key for k, _ in self._raw)
+
+        def __iter__(self) -> typing.Iterator[str]:
+            return iter(k.decode(self._encoding) for k, _ in self._raw)
```
obataku, line 419: the generator `k.decode(self._encoding) for k, _ in self.raw` should be fine enough (it is already an `Iterator`)

httpx/_models.py
```diff
@@ -401,6 +401,72 @@ class Headers(typing.MutableMapping[str, str]):
...
+
+    class _HeadersValuesView(typing.ValuesView[str]):
+        def __init__(
+            self, raw: typing.List[typing.Tuple[bytes, bytes]], encoding: str
+        ) -> None:
+            self._raw = raw
+            self._encoding = encoding
+
+        def __contains__(self, value: typing.Any) -> bool:
+            return any(v.decode(self._encoding) == value for _, v in self._raw)
```
obataku, line 432: is there a reason we don't just try to normalize `value` first?

httpx/_models.py
```diff
@@ -401,6 +401,72 @@ class Headers(typing.MutableMapping[str, str]):
...
+
+        def __contains__(self, item: typing.Any) -> bool:
+            try:
+                key, value = item
+            except (TypeError, ValueError):
+                return False
+
+            try:
+                normalized_key = key.lower().encode(self._encoding)
+                normalized_value = value.encode(self._encoding)
```
obataku, line 455: is there a compelling reason not to use `normalize_header_key`, `normalize_header_value` from [`_utils.py`](https://github.com/encode/httpx/blob/7ab7efd5b04c00a4449c4ee89aea832286919666/httpx/_utils.py#L34) here?

florimondmanca:
@obataku Going to close this since I've lost a bit of context over this PR, and not quite sure I'd be able to provide the best implementation. Feel free to give this a shot though. :-)

PR #907, opened 2020-04-24
Test Suite on GitHub Actions

tomchristie:
Run test suite using GitHub actions.

https://github.com/encode/httpx/actions


tomchristie:
This currently does not:

* Run the tests on windows (I'm okay with that given that our travis tests on windows were in "allowed failure" mode)
* Run the docs build.
* Upload coverage results.

Also the linting and tests just run in the same environment, which I actually think is perfectly okay. They're nicely separated still, eg...

Screenshot 2020-04-24 at 16 22 34

My personal take on this would probably be that it's important enough for us to move away from travis since it's been pretty flaky about reporting to GitHub lately, and that we ought to get this in, and then iterate on any improvements. (But I'm open to review from @encode/maintainers)

tomchristie:
We've currently now got GitHub action test suites for `uvicorn` & `starlette`, plus pending PRs for them in `httpx` and `httpcore`. There's some minor differences between them that we ought to unify. Also it'll be a good step towards *also* using github actions for PyPI & docs deployments.

tomchristie:
We'd also want to update the docs and README badges.

PR #915, opened 2020-04-30
Version 0.13.0.dev0

tomchristie:
The 0.13.0.dev0 is a *pre-release* version. To install it, use `pip install httpx --pre`.

This release switches to `httpcore` for all the internal networking, which means:

* We're using the same codebase for both our sync and async clients.
* HTTP/2 support is now available with the sync client.
* We no longer have a `urllib3` dependancy for our sync client, although there is still an *optional* `URLLib3Dispatcher` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,25 @@ This release switches to `httpcore` for all the internal networking, which means
 
 It also means we've had to remove our UDS support, since maintaining that would have meant having to push back our work towards a 1.0 release, which isn't a trade-off we wanted to make.
 
+### Changed
+
+* Use `httpcore` for underlying HTTP transport. Drop `urllib3` requirement. (Pull #804)
+
+### Added
+
+* Added `URLLib3Dispatcher` class for optional `urllib3` transport support. (Pull #804)
+* Streaming multipart uploads. (Pull #857)
+
+### Fixed
+
+* Performance improvement in brotli decoder. (Pull #906)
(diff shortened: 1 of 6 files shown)


tomchristie:
Okay @encode/maintainers, having been kinda derailed over the last few weeks, I think we're ready to issue the 0.13 pre-release. We can sit on that for however long we feel is appropriate, and then promote it to a 0.13 release.

Once we've done that I think we're ready to start talking through whatever final API points we think need consideration before being confident about issuing a 1.0 release.

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
+
+The 0.13.0.dev0 is a *pre-release* version. To install it, use `pip install httpx --pre`.
+
+This release switches to `httpcore` for all the internal networking, which means:
+
+* We're using the same codebase for both our sync and async clients.
+* HTTP/2 support is now available with the sync client.
+* We no longer have a `urllib3` dependency for our sync client, although there is still an *optional* `URLLib3Dispatcher` class.
+
+It also means we've had to remove our UDS support, since maintaining that would have meant having to push back our work towards a 1.0 release, which isn't a trade-off we wanted to make.
```
florimondmanca, line 17: Should we include the rest of PRs that got in since 0.12.1? I believe there are a couple of them. We should also get them listed in the changelog.

tomchristie: Gotcha. Done!

PR #919, opened 2020-05-01
ASGI: Wait for response to complete before sending disconnect message

JayH5:
Second attempt at fixing the problem described in encode/starlette#908. More discussion of this in #909.


httpx/_concurrency.py
```diff
@@ -0,0 +1,31 @@
+import typing
```
JayH5, line 1: Let me know if this module and/or type checking stuff is over-the-top 🙃. The type checking does allow us to assert that `asyncio.Event` and `trio.Event` expose similar-enough interfaces.

tomchristie: I’d probably not bother with the Protocol level stuff here yeah.

JayH5: OK I've reduced the type-checking stuff and got rid of this new `_concurrency` module :+1:

httpx/_dispatch/asgi.py
```diff
@@ -1,9 +1,28 @@
...
 from typing import Callable, Dict, List, Optional, Tuple
 
 import httpcore
+import sniffio
 
 from .._content_streams import ByteStream
 
+if typing.TYPE_CHECKING:  # pragma: no cover
+    import asyncio
+    import trio
```
tomchristie, line 11: Okay, so `trio` might not be installed.

Perhaps we should be doing something like this?...

```python
try:
    import trio
    Event = typing.Union[asyncio.Event, trio.Event]
except ImportError:
    Event = asyncio.Event
```

Also it's not obvious to me that we need to bother with the `if typing.TYPE_CHECKING:` dance? We *usually* only use that if it's required to avoid otherwise circular imports.

JayH5: I tried what you're suggesting but unfortunately run into an error like:
```
error: Cannot assign multiple types to name "Event" without an explicit "Type[...]" annotation
```

Seems like type aliases need to be statically resolvable: https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

tomchristie: Okay. We can't rely on the `trio` package necessarily being installed tho.
(Even if we're limiting this to `if typing.TYPE_CHECKING:`)

One other option might be to just under-specify the type in this case, as a lazy `typing.Any`. We're not exposing the type info or using that function outside of this scope, so perhaps that's a sensibly practical option?

JayH5: The `if typing.TYPE_CHECKING` does actually help here. If `trio` (actually, `trio-typing`) is not installed, mypy is still happy because we use `--ignore-missing-imports` so it just ignores `import trio` and `trio.Event()`.

JayH5: @tomchristie still want me to make that a `typing.Any`?

httpx/_dispatch/asgi.py
```diff
@@ -76,23 +95,27 @@ async def request(
...
 
         request_body_chunks = stream.__aiter__()
 
         async def receive() -> dict:
-            nonlocal response_complete
+            nonlocal request_complete, response_complete
 
-            if response_complete:
+            if request_complete:
+                # Simulate blocking until the response is complete and then disconnect
```
tomchristie, line 111: I'd tweak this comment. We're not "simulating" blocking until the response is complete, we *are* blocking until the response is complete. :)

JayH5: Took out "streaming" and then found the comment a bit redundant 🤔

requirements.txt
```diff
@@ -26,6 +26,7 @@ pytest-asyncio
 pytest-trio
 pytest-cov
 trio
+trio-typing
```
tomchristie, line 29: Do we need to add this dependency?

tomchristie:
Curious what the other @encode/maintainers think to this? Also, would we want it in a 0.13.dev2 release, along with https://github.com/encode/httpcore/pull/81 perhaps?

Seems decent @JayH5 - Are you able to confirm that it resolves https://github.com/encode/starlette/issues/908?

requirements.txt
JayH5: For any trio types to be type-checked this must be installed (the only type being checked is `trio.Event`). This package also has a mypy plugin that can be enabled for extra functionality, which I did enable in encode/httpcore#69, but I didn't bother here.

JayH5:
@tomchristie I can confirm that this fixes encode/starlette#908 for me 😄.

Would love for this to be in a released version of httpx. Currently stuck with starlette <0.13.3 in some places until the issue is fixed.

PR #923, opened 2020-05-02
Mark `Response.json()` as returning `Any`

florimondmanca:
Anytime I process the returned `json()` data from an API call, I run into mypy issues because the return type is currently `Union[dict, list]`:

```python
data = response.json()
_ = data["items"]
```

```console
debug/client.py:4: error: No overload variant of "__getitem__" of "list" matches argument type "str"
debug/client.py:4: note: Possible overload variants:
debug/client.py:4: note:     def __getitem__(self, int) -> Any
debug/client.py:4: note:     def __getitem__(self, slice) -> List[Any]
```

The workaround is this… (Note that marking as `: dict` is not enough, as mypy just won't let us down-grade a `Union` via a simple annotation.)

```
data: dict= response.json()  # type: ignore
```

But it's a pain, so this PR just marks the return type as `Any` — which is in line with `json.loads()`. This means we can freely annotate the data as seems fit, eg `list`, `dict`, etc.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -838,7 +838,7 @@ class Response:
             message = message.format(self, error_type="Server Error")
             raise HTTPError(message, response=self)
 
-    def json(self, **kwargs: typing.Any) -> typing.Union[dict, list]:
+    def json(self, **kwargs: typing.Any) -> typing.Any:
         if self.charset_encoding is None and self.content and len(self.content) > 3:
             encoding = guess_json_utf(self.content)
             if encoding is not None:


httpx/_models.py
```diff
@@ -838,7 +838,7 @@ def raise_for_status(self) -> None:
             message = message.format(self, error_type="Server Error")
             raise HTTPError(message, response=self)
 
-    def json(self, **kwargs: typing.Any) -> typing.Union[dict, list]:
+    def json(self, **kwargs: typing.Any) -> typing.Any:
```
yeraydiazdiaz, line 841: I'm surprised by:

> note that marking as : dict is not enough

Do you get the same error?

florimondmanca: I get a different, and confusing one:

```python
data: dict = httpx.get("http://example.com").json()
```

```console
error: Incompatible types in assignment (expression has type "Union[Dict[Any, Any], List[Any]]", variable has type "Dict[Any, Any]")
```

I assume it's because mypy doesn't allow just down-sizing a union like that - you'd need to make a proper `assert`, or `cast()`.

yeraydiazdiaz: Ugh, yeah, sounds like `Any` is the way to go here 👍

PR #930, opened 2020-05-05
Export more private modules

lepture:
1. class in `_content_streams` are very useful, we should allow people to use them.
2. class in `_types` are helpful for type hint, they should be public to developers.

Related issues:

- https://github.com/lepture/authlib/pull/209
- https://github.com/lepture/authlib/pull/219


tomchristie:
I've made some comments on https://github.com/lepture/authlib/pull/209/files re. how you'd be able to tweak things around so that you're only using public API from `httpx`.

PR #932, opened 2020-05-06
Add Request.content setter

lepture:
Continue solution for discussion here https://github.com/lepture/authlib/pull/209/files

cc @tomchristie your solution doesn't work, there is no `body` or `content` parameter in class `Request`. The only possible key could be `data=..` from my understanding of the Request class.
But it still has a problem, because in the Authlib case, if `rebuild_request` returns a new request instance, it would raise error when accessing `.content`, since the new request instance hasn't called `.read()` or `.aread()` yet.

If you don't want to export more modules #930, my suggestion is to add a setter for `.content`.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -644,6 +644,11 @@ class Request:
             raise RequestNotRead()
         return self._content
 
+    @content.setter
+    def content(self, value: bytes) -> None:
+        self._content = value
+        self.stream = ByteStream(self._content)
+
     def read(self) -> bytes:
         """
         Read and return the request content.


lepture:
ping @tomchristie could you check this change?

tomchristie:
Yeah, I'm not sure if we want to take this approach, or if there's actually already sufficient public API here, and we instead need to rejig the auth flow in `authlib`.

I've issued https://github.com/lepture/authlib/pull/232 as a preliminary bit of ensuring authlib works against the `httpx` public API. Once that's all resolved I'll take a look at the content streams / auth flow bit.

tomchristie:
I think https://github.com/lepture/authlib/pull/233 should resolve the remaining issues here.

I'm going to close this off, since setting the `.content` directly is awkward, given that you'll also want to make sure you've got the right headers, such as `Content-Length` that are set in `__init__`.

Issue #929, opened 2020-05-04
Regression with 0.13.dev0: Missing mandatory Host: header


andrewshadura:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

With an HTTP proxy enabled, the HTTP/1.1 implementation raises a `LocalProtocolError` exception:
```python
  File "/usr/local/lib/python3.8/dist-packages/obs_proxy-0-py3.8.egg/obs_proxy/repserver.py", line 100, in proxy_request
    resp = await client.send(req, stream=True, auth=build_auth(), timeout=HTTP_TIMEOUT)
  File "/usr/local/lib/python3.8/dist-packages/httpx-0.13.dev0-py3.8.egg/httpx/_client.py", line 1137, in send
    response = await self.send_handling_redirects(
  File "/usr/local/lib/python3.8/dist-packages/httpx-0.13.dev0-py3.8.egg/httpx/_client.py", line 1164, in send_handling_redirects
    response = await self.send_handling_auth(
  File "/usr/local/lib/python3.8/dist-packages/httpx-0.13.dev0-py3.8.egg/httpx/_client.py", line 1201, in send_handling_auth
    response = await self.send_single_request(request, timeout)
  File "/usr/local/lib/python3.8/dist-packages/httpx-0.13.dev0-py3.8.egg/httpx/_client.py", line 1233, in send_single_request
    ) = await dispatcher.request(
  File "/usr/local/lib/python3.8/dist-packages/httpcore-0.8.2-py3.8.egg/httpcore/_async/http_proxy.py", line 76, in request
    return await self._tunnel_request(
  File "/usr/local/lib/python3.8/dist-packages/httpcore-0.8.2-py3.8.egg/httpcore/_async/http_proxy.py", line 145, in _tunnel_request
    proxy_response = await proxy_connection.request(
  File "/usr/local/lib/python3.8/dist-packages/httpcore-0.8.2-py3.8.egg/httpcore/_async/connection.py", line 69, in request
    return await self.connection.request(method, url, headers, stream, timeout)
  File "/usr/local/lib/python3.8/dist-packages/httpcore-0.8.2-py3.8.egg/httpcore/_async/http11.py", line 52, in request
    await self._send_request(method, url, headers, timeout)
[...]

PR #933, opened 2020-05-06
Document dev proxy setup

yeraydiazdiaz:
Prompted by https://github.com/encode/httpx/issues/929#issuecomment-624131087

I thought it'd be good to have it in the docs so we can point people to it as well.


yeraydiazdiaz:
Probably should mention this is only testing in Mac OS X 😅 

It's all Python tho so it's probably fine 😉

Issue #943, opened 2020-05-12
Fix flaky `test_write_timeout` test in CI


yeraydiazdiaz:
[One of our tests](https://github.com/encode/httpx/blob/master/tests/test_timeouts.py#L15-L22) has decided to start failing across different Python versions:

https://github.com/encode/httpx/pull/941/checks?check_run_id=666427391
https://github.com/encode/httpx/runs/666405754
https://github.com/encode/httpx/runs/666709254

I could not reproduce the error locally on 3.8 or 3.6 on my Mac.

florimondmanca:
No idea yet, but maybe a change in one of our dependencies...?

- pytest-asyncio 0.12, released on May 4th?
- Regression in httpcore?

tomchristie:
#946 might be a bit more reliable here?

Really we ought to also look at...

* Pinning the dependencies installed for the test suite.
* Using dependabot to update those as separate PRs.

And then...

* Cache the virtualenv directory between builds.

That'd help us with debugging stuff like this, even in cases where it's not obvious that it *is* related to any library changes, and could well just be an environmental issue, at least we'd have a better view onto that.

tomchristie:
Closed via #946

PR #945, opened 2020-05-12
Pin httpcore==0.8.3

yeraydiazdiaz:
https://github.com/encode/httpx/issues/943 maybe...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         "chardet==3.*",
         "idna==2.*",
         "rfc3986>=1.3,<2",
-        "httpcore==0.8.4",
+        "httpcore==0.8.3",
     ],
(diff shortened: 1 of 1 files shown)


yeraydiazdiaz:
Running tests one more time, for safety

yeraydiazdiaz:
Looks like we have a winner...

I guess we should merge this to unblock other PRs?

tomchristie:
And *ping*. A failure against 0.8.3 too.

tomchristie:
So ATM we've no evidence that the flakiness is an `httpcore` regression, but if it was it'd be something introduced in 0.8.3 or prior.

yeraydiazdiaz:
Yep, makes sense I could not see anything that would explain this. 👍

PR #946, opened 2020-05-12
Attempt to resolve test flakiness

tomchristie:
Refs #943

* We've already got a test that ensures `response.elapsed > datetime.timedelta(0)`, so let's not try to enforce a "this should end up as more than one second" case.
* Let's bump the slow response default to a more reliably noticable 1 second delay.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -102,12 +102,7 @@ async def hello_world(scope, receive, send):
 
 
 async def slow_response(scope, receive, send):
-    delay_ms_str: str = scope["path"].replace("/slow_response/", "")
-    try:
-        delay_ms = float(delay_ms_str)
-    except ValueError:
-        delay_ms = 100
-    await sleep(delay_ms / 1000.0)
+    await sleep(1.0)
     await send(
         {
             "type": "http.response.start",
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -102,12 +102,7 @@ async def hello_world(scope, receive, send):
 
 
 async def slow_response(scope, receive, send):
-    delay_ms_str: str = scope["path"].replace("/slow_response/", "")
-    try:
-        delay_ms = float(delay_ms_str)
-    except ValueError:
-        delay_ms = 100
-    await sleep(delay_ms / 1000.0)
+    await sleep(1.0)
```
yeraydiazdiaz, line 105: This feels a bit too long, each test that uses `slow_response` will take a full second to complete.

Maybe we could keep the above logic and set only the offending test to use a 1 second delay?

tomchristie: So there's two timeout tests that use this, and I don't think they *do* take a full second to complete, because the client raises a timeout exception in each case.

The delay needs to not be ridiculously excessive, or I think the tests will hang at the end of the suite, when the server is shutting down.

yeraydiazdiaz: Right, makes sense 👍

PR #955, opened 2020-05-15
Enforce coverage

tomchristie:
Enforce test coverage when running the test suite.

Refs #316.

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -1,0 +1,11 @@
+#!/bin/sh -e
+
+export PREFIX=""
+if [ -d 'venv' ] ; then
+    export PREFIX="venv/bin/"
+fi
+export SOURCE_FILES="httpx tests"
+
+set -x
+
+${PREFIX}coverage report --show-missing --skip-covered --fail-under=97
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,6 @@ line_length = 88
 multi_line_output = 3
 
 [tool:pytest]
-addopts = --cov=httpx --cov=tests --cov-report=term-missing -rxXs
+addopts = --cov=httpx --cov=tests -rxXs
 markers =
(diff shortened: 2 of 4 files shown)


scripts/coverage
```diff
@@ -0,0 +1,11 @@
...
+
+export PREFIX=""
+if [ -d 'venv' ] ; then
+    export PREFIX="venv/bin/"
+fi
+export SOURCE_FILES="httpx tests"
+
+set -x
+
+${PREFIX}coverage report --show-missing --skip-covered --fail-under=97
```
yeraydiazdiaz, line 11: I can't find `show-missing` on the coverage.py help or docs?

setup.cfg
```diff
@@ -16,6 +16,6 @@ line_length = 88
 multi_line_output = 3
 
 [tool:pytest]
-addopts = --cov=httpx --cov=tests --cov-report=term-missing -rxXs
+addopts = --cov=httpx --cov=tests -rxXs
```
yeraydiazdiaz, line 19: Should we include `--branch` as well?

StephenBrown2: I am in favor of this.

scripts/coverage
tomchristie: In the docs it’s only referred to by `-m`

yeraydiazdiaz: Ah, I see 👍

setup.cfg
tomchristie: I’m personally a pretty hard no on branch coverage.

Keeping things simple and pinning at 100% line coverage seems to absolutely be the sweet spot for code quality, and is a clear baseline that we can aim to enforce across all encode projects.

yeraydiazdiaz: Personally I think if we are aiming for 100% we'd need branch coverage and rely on explicitly ignoring them in cases where it's not possible.

That being said, I'm not going to die on this particular hill 😅

PR #963, opened 2020-05-20
Transport API

yeraydiazdiaz:
Closes https://github.com/encode/httpx/issues/768

~I did some `dispatch > transport` renaming but did not rename the `_dispatch` package and the `*Dispatch` classes in it. I considered renaming the package to `_transports` and all classes to `*Transport` with `*Dispatch` aliases but thought it might be better to check first.~

Edit: Pipelines failed for some reason so I went for it while I gave GH actions some time to sort themselves out.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -628,8 +628,8 @@ sending of the requests.
 
 These Transport objects must implement some methods from
 [`httpcore`'s API](https://www.encode.io/httpcore/api/), either
-`httpcore.AsyncHTTPTransport` if you're using the asynchronous `Client`, or
-`httpcore.SyncHTTPTransport` if you're using a synchronous one.
+`httpcore.AsyncHTTPTransport` if you're using `AsyncClient`, or
+`httpcore.SyncHTTPTransport` if you're using `Client`.
 
 Specifically you *MUST* implement `request`, and `close` or `aclose` depending
 on the type of client you're using.
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -626,14 +626,11 @@ HTTPX's `Client` also accepts a `transport` argument. This argument allows you
 to provide a custom Transport object that will be used to perform the actual
 sending of the requests.
 
-These Transport objects must implement at least one method from
-[`httpcore`'s API](https://www.encode.io/httpcore/api/), either
-`httpcore.AsyncHTTPTransport` if you're using `AsyncClient`, or
-`httpcore.SyncHTTPTransport` if you're using `Client`.
-
-Specifically you *MUST* implement `request`. If you need additional logic for
-closing the transport you can also implement `close` or `aclose` depending on
-the type of client you're using.
+A transport instance must implement the Transport API defined by
+[`httpcore`](https://www.encode.io/httpcore/api/). You
+should either subclass `httpcore.AsyncHTTPTransport` to implement a transport to
+use with `AsyncClient`, or subclass `httpcore.SyncHTTPTransport` to implement a
+transport to use with `Client`.
 
 For example, HTTPX ships with a transport that uses the excellent
 [`urllib3` library](https://urllib3.readthedocs.io/en/latest/):
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -639,8 +639,7 @@ For example, HTTPX ships with a transport that uses the excellent
 
 ```python
 >>> import httpx
->>> from httpx._dispatch.urllib3 import URLLib3Dispatcher
->>> client = httpx.Client(transport=URLLib3Dispatcher())
+>>> client = httpx.Client(transport=httpx.URLLib3Transport())
 >>> client.get("https://example.org")
 <Response [200 OK]>
 ```
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -626,13 +626,14 @@ HTTPX's `Client` also accepts a `transport` argument. This argument allows you
 to provide a custom Transport object that will be used to perform the actual
 sending of the requests.
 
-These Transport objects must implement some methods from
+These Transport objects must implement at least one method from
 [`httpcore`'s API](https://www.encode.io/httpcore/api/), either
 `httpcore.AsyncHTTPTransport` if you're using `AsyncClient`, or
 `httpcore.SyncHTTPTransport` if you're using `Client`.
 
-Specifically you *MUST* implement `request`, and `close` or `aclose` depending
-on the type of client you're using.
+Specifically you *MUST* implement `request`. If you need additional logic for
+closing the transport you can also implement `close` or `aclose` depending on
+the type of client you're using.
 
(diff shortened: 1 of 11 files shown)


yeraydiazdiaz:
Hmmm, not sure what's going on here, all checks are "cancelled" and there's no log at all so I guess I'll try again later 🤷‍♂️

docs/advanced.md
```diff
@@ -619,3 +619,28 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+
+Specifically you *MUST* implement `request`, and `close` or `aclose` depending
+on the type of client you're using.
+
+For example, HTTPX ships with a transport that uses the excellent
+[`urllib3` library](https://urllib3.readthedocs.io/en/latest/):
+
+```python
+>>> import httpx
+>>> from httpx._dispatch.urllib3 import URLLib3Dispatcher
```
tomchristie, line 642: We'll want to rename this to `URLLib3Transport`, and promote it to public API, so `httpx.URLLib3Transport(...)`.

httpx/_client.py
```diff
@@ -417,7 +418,10 @@ class Client(BaseClient):
     that should be followed.
     * **base_url** - *(optional)* A URL to use as the base when building
     request URLs.
-    * **dispatch** - *(optional)* A dispatch class to use for sending requests
+    * **transport** - *(optional)* A transport class to use for sending requests
+    over the network.
+    * **dispatch** - *(optional)* A deprecated alias for transport.
+    * **transport** - *(optional)* A transport class to use for sending requests
```
tomchristie, line 424: `transport` is duplicated here. :)

tomchristie:
Looks like a linter failure on import sorting... https://github.com/encode/httpx/pull/963/checks?check_run_id=693477366

docs/advanced.md
tomchristie: I guess we also ought to note that `urllib3` needs to be installed to use this transport.

We *might* need to look at if we need an `assert` check in the codebase too, since `urllib3` is no longer going to be a core dependancy of `httpx`, it'll be an optional from 13.0 onwards.

yeraydiazdiaz: Yep, that makes sense 👍

docs/advanced.md
```diff
@@ -619,3 +619,28 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+
+## Custom Transports
+
+HTTPX's `Client` also accepts a `transport` argument. This argument allows you
+to inject a custom Transport objects that will be used to perform the actual
+sending of the requests.
+
+These Transport objects must implement some methods from
+[`httpcore`'s API](https://www.encode.io/httpcore/api/), either
+`httpcore.AsyncHTTPTransport` if you're using the asynchronous `Client`, or
```
JayH5, line 631: Did you mean `AsyncClient`?

yeraydiazdiaz: Indeed, thanks!

httpx/_client.py
```diff
@@ -456,16 +459,24 @@ def __init__(
 
         proxy_map = self.get_proxy_map(proxies, trust_env)
 
-        self.dispatch = self.init_dispatch(
+        if dispatch is not None:
+            warnings.warn(
```
JayH5, line 463: Do we want to make the `category` of these warnings `DeprecationWarning` (like #908)?

httpx/_transports/wsgi.py
```diff
@@ -131,3 +132,20 @@ def start_response(
...
+
+class WSGIDispatch(WSGITransport):
+    def __init__(
+        self,
+        app: typing.Callable,
+        raise_app_exceptions: bool = True,
+        script_name: str = "",
+        remote_addr: str = "127.0.0.1",
+    ) -> None:
+        warnings.warn("WSGIDispatch is deprecated, please use WSGITransport")
```
florimondmanca, line 145: `DeprecationWarning` is missing here as well - let's review the `warnings.warn` additions to make sure we use it everywhere?

httpx/_client.py
```diff
@@ -456,16 +459,25 @@ def __init__(
...
```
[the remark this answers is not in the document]
yeraydiazdiaz: I can add that, makes sense 👍

httpx/_transports/wsgi.py
yeraydiazdiaz: I'll change it to use the new util function 👍

httpx/_client.py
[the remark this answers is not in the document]
florimondmanca: Just pushed https://github.com/encode/httpx/pull/964 😄

yeraydiazdiaz:
> I assume the docs would need updating too - we're still using the Dispatch phrasing there as of the current state of this PR. Do we want to update it in this PR or as a follow-up?

Probably best to change the docs now, there's quite a few code snippets using `Dispatch` as well.

docs/advanced.md
```diff
@@ -619,3 +619,27 @@ If you do need to make HTTPS connections to a local server, for example to test
...
 Response <200 OK>
 ```
+
+## Custom Transports
+
+HTTPX's `Client` also accepts a `transport` argument. This argument allows you
+to inject a custom Transport objects that will be used to perform the actual
+sending of the requests.
+
+These Transport objects must implement some methods from
```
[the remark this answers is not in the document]
yeraydiazdiaz: Well, since you've written it already I might as well include it! 😄

docs/advanced.md
```diff
@@ -619,3 +619,59 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+sending of the requests.
+
+These Transport objects must implement at least one method from
+[`httpcore`'s API](https://www.encode.io/httpcore/api/), either
+`httpcore.AsyncHTTPTransport` if you're using `AsyncClient`, or
+`httpcore.SyncHTTPTransport` if you're using `Client`.
+
+Specifically you *MUST* implement `request`. If you need additional logic for
+closing the transport you can also implement `close` or `aclose` depending on
+the type of client you're using.
```
tomchristie, line 636: I think I'd probably rephrase this... "must implement at least one method from httpcore`'s API [...] Specifically you *MUST* implement `request`." scans a bit confusingly to me.

Perhaps...

"A transport instance must implement the Transport API defined by `httpcore`. You should either subclass `httpcore.AsyncHTTPTransport` to implement a transport to use with `AsyncClient`, or subclass `httpcore.SyncHTTPTransport` to implement a transport to use with `Client`."

yeraydiazdiaz: Yeah it reads a bit odd, your text is much clearer.

tomchristie:
Great, once everyone's happy let's get this in as it is.

Couple of follow ups I can think of, which I'll open individual issues for.

yeraydiazdiaz:
Thanks all! 🎉

PR #967, opened 2020-05-21
Updates for httpcore 0.9 interface.

tomchristie:
Needs to pin to httpcore 0.9 once available.

* Pass proxy_url instead of proxy_origin.
* Update signature of transport classes to have `port` as an `Optional[int]`.


tomchristie:
Pulling this in now, to fix the builds.

Issue #966, opened 2020-05-21
Consider `URLLib3Transport` interface, since it's becoming public API.


tomchristie:
Follow on from #963

There's a couple of things we could consider in the `URLLib3Transport` interface, now that it's becoming public API.

* We don't really want to pass `pool_limits`, since it doesn't quite fit the parameters that `urllib3` uses to configure it's PoolManager. We *needed* to pass it before, since we were using the class for our sync implementation, and needed the consistency between our async and sync cases, but now that we're providing it separately we should just mirror their native `num_pools` and `maxsize`.
* With `httpcore` we provide different classes for the proxy and the plain-connection-pool transports. We *could?* choose to do the same with the `urllib3` transport, rather than providing a single class with an optional `proxy=...` argument.

tomchristie:
Also we *could* just drop the `URLLib3Dispatch` compat shim, since it's only ever existed in the pre-releases.

florimondmanca:
The only reason I'd be in favor of keeping `URLLib3Transport` is to make it easier for folks that use Requests to migrate over without having to deal with slight behavior differences right away.

I'm actually not _certain_ that's what people would do in practice - there are already lots of behavior differences that come from switching from _Requests_ anyway, so not sure if switching to our transport is a big leap.

Another argument in favor of dropping it before 1.0 lands would be reducing maintenance burden - now that the transport interface is public it would be pretty easy for someone to maintain this as a 3rd party component (esp given that we have a working implementation in 0.12.x for people to take inspiration from).

yeraydiazdiaz:
I'm 👍 on dropping `URLLib3Transport` completely tbh.

I might be missing something but it seems atm it feels like the remnants of our old sync flow rather than a core piece of HTTPX that users would reach for. It already has our [lowest test coverage percentage](https://github.com/encode/httpx/runs/696413093?check_suite_focus=true#step:8:13), feels like it will eventually break for someone since it's now in the public API.

tomchristie:
Okay, well let's drop the `URLLib3Transport`/`URLLib3ProxyTransport`  convo from the 0.13 milestone so we can get this thing released, and *keep* the idea of `URLLib3Transport` as a third-party-package as an open option.

(Personally I *probably* think it's a pretty useful thing to have available to us, but I'm not absolute on it.)

tomchristie:
Remaining points for 1.0 consideration...

[...]

PR #968, opened 2020-05-21
Rename pool limit options

### LINKED ISSUE #948: Deprecate ProxyLimits naming in favour of max_connections and max_keepalive
Prompted by https://github.com/encode/httpx/pull/942#discussion_r424341967

The `PoolLimits` class uses `soft_limit` and `hard_limit` whereas in httpcore we use `max_connections` and `max_keepalive` which are clearer.

tomchristie:
Closes #948


tomchristie:
The URLLib3 changes here will become redundant after #966, but I'm going to ignore that here to keep this PR tightly scoped.

httpx/_config.py
```diff
@@ -288,29 +289,47 @@ class PoolLimits:
...
+            warnings.warn(
+                "'soft_limit' is deprecated. Use 'max_keepalive' instead.",
+                DeprecationWarning,
+            )
+        if hard_limit is not None:  # pragma: nocover
+            self.max_connections = hard_limit
+            warnings.warn(
+                "'hard_limit' is deprecated. Use 'max_connections' instead.",
+                DeprecationWarning,
+            )
```
yeraydiazdiaz, line 319: You can use the new [`warn_deprecated` util function.](https://github.com/encode/httpx/blob/master/httpx/_utils.py#L406-L407)

httpx/_config.py
```diff
@@ -288,29 +288,43 @@ class PoolLimits:
...
+        self.max_keepalive = max_keepalive
+        self.max_connections = max_connections
+        if soft_limit is not None:  # pragma: nocover
+            self.max_keepalive = soft_limit
+            warn_deprecated("'soft_limit' is deprecated. Use 'max_keepalive' instead.",)
+        if hard_limit is not None:  # pragma: nocover
+            self.max_connections = hard_limit
+            warn_deprecated(
+                "'hard_limit' is deprecated. Use 'max_connections' instead.",
+            )
```
florimondmanca, line 314: I _assume_ eventually one of the linters is going to cough on the extra commas here, so just a heads up:

```suggestion
        if soft_limit is not None:  # pragma: nocover
            self.max_keepalive = soft_limit
            warn_deprecated("'soft_limit' is deprecated. Use 'max_keepalive' instead.")
        if hard_limit is not None:  # pragma: nocover
            self.max_connections = hard_limit
            warn_deprecated(
                "'hard_limit' is deprecated. Use 'max_connections' instead."
            )
```

tomchristie:
Merged #967 into this as we're dependant on it.

httpx/_config.py
florimondmanca: Actually not, feel free to ignore!

PR #970, opened 2020-05-21
Use URLLib3 configuration options for URLLib3 transport

tomchristie:
Refs #966

Instead of using `pool_limits` with `URLLib3Transport` this PR switches us to instead using their native pool limiting configuration options. I've taken the external naming and defaults in line with how `requests` adapter classes use `urllib3`.

One other thing we could consider here is if we'd like the external interface to be `URLLib3Transport` and `URLLib3ProxyTransport` for symmetry with the `httpcore` classes, and with the `urllib3` interface itself, bu t let's consider that independently.

### DIFF
--- a/httpx/_transports/urllib3.py
+++ b/httpx/_transports/urllib3.py
@@ -1,11 +1,10 @@
-import math
 import socket
 import ssl
 from typing import Dict, Iterator, List, Optional, Tuple, Union
 
 import httpcore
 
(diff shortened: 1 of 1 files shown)


florimondmanca:
I like this - less indirection both for us and users.

tomchristie:
Great, let's get this slice in then. 👍

PR #985, opened 2020-05-23
Added type annotation for tests/client/test_async_client.py

hramezani:
Refs https://github.com/encode/httpx/issues/650

We have still 3 issues in this file:
```
tests/client/test_async_client.py:29: error: Argument 1 to "update" of "Headers" has incompatible type "Dict[str, str]"; expected "Union[Headers, Dict[Union[str, bytes], Union[str, bytes]], Sequence[Tuple[Union[str, bytes], Union[str, bytes]]], None]"
tests/client/test_async_client.py:148: error: Argument "headers" to "post" of "AsyncClient" has incompatible type "Dict[str, str]"; expected "Union[Headers, Dict[Union[str, bytes], Union[str, bytes]], Sequence[Tuple[Union[str, bytes], Union[str, bytes]]], None]"
```
In above two test we defined headers like:
`headers = {"Custom-header": "value"}`
`headers = {"Expect": "100-continue"}`
which have type `Dict[str, str]` and I think they are valid type based on [HeaderTypes definition](https://github.com/encode/httpx/blob/a4145ce2fba4c273702aff2cc4da85b8acc6455e/httpx/_types.py#L40). we can fix them by replacing them by List of Tuples like `headers = [("Expect", "100-continue")]`. TBH, can not figure out why they are incompatible.

```
tests/client/test_async_client.py:100: error: "raise_for_status" of "Response" does not return a value
```
What should we do with this error :thinking:? Shall we ignore checking for this line?

I don't have much experience in type annotation, that's why I am asking this questions.

### DIFF
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -95,7 +95,7 @@ async def test_raise_for_status(server):
                     response.raise_for_status()
                 assert exc_info.value.response == response
             else:
-                assert response.raise_for_status() is None
+                assert response.raise_for_status() is None  # type: ignore
 
 
 @pytest.mark.usefixtures("async_environment")
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -110,7 +110,7 @@ def test_raise_for_status(server):
                     response.raise_for_status()
                 assert exc_info.value.response == response
             else:
-                assert response.raise_for_status() is None
+                assert response.raise_for_status() is None  # type: ignore
 
 
 def test_options(server):


florimondmanca:
Hey, thanks for this! Definitely useful to expand our type coverage in tests to discover oddities like this.

> TBH, can not figure out why they are incompatible.

I think it’s yet another case of « Dict is invariant in values », so the Union’s in the values make mypy say « I see those are `str`, but I am *really* expecting `Union[str, bytes]`, so those don’t match, nope ».

TL;DR I think we’d need to update the `HeaderTypes` to use `Mapping` (or most likely `MutableMapping`) instead of `Dict`.

florimondmanca:
As for…

> What should we do with this error 🤔? Shall we ignore checking for this line?

Yes I think it makes sense to `# type: ignore` that particular assertion. :)

hramezani:
Thanks @florimondmanca, I changed the `HeaderTypes` and used `MutableMapping` instead of `Dict` in a separate commit.

httpx/_types.py
```diff
@@ -24,7 +26,7 @@
     from ._config import Proxy, Timeout  # noqa: F401
     from ._models import URL, Cookies, Headers, QueryParams, Request  # noqa: F401
 
-StrOrBytes = Union[str, bytes]
+StrOrBytes = TypeVar("StrOrBytes", str, bytes)
```
florimondmanca, line 29: Was this change required somehow? We specifically don't want to use type vars/generics for `HeaderTypes` and others — we initially used that but realized a `Union` better fits what we accept (i.e. "`str` _or_ `bytes`, I don't care").

We initially used `typing.AnyStr` as our "str or bytes" annotation (which this is equivalent to), but changed it because

tests/client/test_async_client.py
```diff
@@ -1,12 +1,14 @@
...
 import pytest
+from uvicorn.main import Server
 
 import httpx
 
 
 @pytest.mark.usefixtures("async_environment")
-async def test_get(server):
+async def test_get(server: Server) -> None:
     url = server.url
```
florimondmanca, line 12: I'm a bit curious about why this works at all — Uvicorn's `Server` doesn't have a `.url` property, but our `TestServer` does. 🤔 I assume it's because Uvicorn doesn't ship type annotations so mypy really just treats `Server` as `Any`? (A `reveal_type(server)` could help confirm it.)

I don't think importing from `conftest` is common practice? If not we might need to move the `TestServer` (and some other items) to `utils.py`. And we might as well rename `TestServer` as `Server`? (Any such changes should  probably go into their own PRs too, right?)

(I realize there are some rough edges to clean up here and there for us to be able to roll out type hints to the tests code more fluently in the future; I think it's okay to do that small bit of work first.)

httpx/_types.py
```diff
@@ -38,7 +40,9 @@
 ]
 
 HeaderTypes = Union[
-    "Headers", Dict[StrOrBytes, StrOrBytes], Sequence[Tuple[StrOrBytes, StrOrBytes]],
+    "Headers",
+    MutableMapping[StrOrBytes, StrOrBytes],
```
florimondmanca, line 44: Should we consider submitting this change as its own "fix headers type annotations" PR? Very similar to #976. (We tend to strongly go for "one PR = one change", mostly because then it's easier to decide what goes into release changelogs and what don't.)

httpx/_types.py
hramezani: I tried `MutableMapping[StrOrBytes, StrOrBytes]`, but I had the incompatibility error. that's why I decided to change the `StrOrBytes`.
What should we do? split  `Dict[StrOrBytes, StrOrBytes]` in `HeaderTypes` to `MutableMapping[str, str], MutableMapping[bytes, bytes]`?

httpx/_types.py
hramezani: Yes, sure.

florimondmanca: FYI I'm in the process of opening a PR for this that also resolves the odd mypy error you mentioned here https://github.com/encode/httpx/pull/985#discussion_r429624707 :-)

tests/client/test_async_client.py
hramezani: > I'm a bit curious about why this works at all — Uvicorn's Server doesn't have a .url property, but our TestServer does. thinking I assume it's because Uvicorn doesn't ship type annotations so mypy really just treats Server as Any? (A reveal_type(server) could help confirm it.)

Yes, you are right. its type is `Any`.

> I don't think importing from conftest is common practice? If not we might need to move the TestServer (and some other items) to utils.py. And we might as well rename TestServer as Server? (Any such changes should probably go into their own PRs too, right?)

I will do this in separate PR as well

florimondmanca: What do you think about #991?

I'm thinking adding the `server: Server` annotations everywhere is super cumbersome, and doesn't add a lot of value. We can have mypy type check the inside of our test functions anyway, which will save us a big amount of work. If we go for that approach, then we actually don't really need to move `TestServer` anywhere (since we won't really need to import it for type hints purposes).

Besides, note that I initally submitted #990 but closed it as I think #991 is a better way forward.

httpx/_types.py
florimondmanca: Yeah, it's odd. I actually traced that back to a mypy limitation in `Mapping`/`MutableMapping` (the key type is invariant). Opened #993 with a tight fix inspired by yours here. :-)

hramezani: Nice!

httpx/_types.py
hramezani: Thanks for the PR :+1: . Yes, I agree with you. this change will save our time and we can go faster.

tests/client/test_async_client.py
hramezani: Thanks for the PR :+1:  . Yes, I agree with you. this change will save our time and we can go faster.

hramezani:
@florimondmanca I will wait for #991 and #993 to be merged and then I can rebase this PR.

hramezani:
@florimondmanca, I rebased it. we just have one error in `tests/client/test_async_client.py ` and one in `tests/client/test_client.py ` :+1:

Issue #978, opened 2020-05-21
Memory leak when creating lots of AsyncClient contexts


Recursing:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

After creating an AsyncClient context (with async with), it does not seem to be garbage collected, that can be a problem for very long running services that might create a bunch of them and eventually run out of memory

### To reproduce
```
import httpx
import gc
import asyncio

print(f"httpx version: {httpx.__version__}")


async def make_async_client():
    async with httpx.AsyncClient() as client:
        await asyncio.sleep(10)


async def main(n):
    tasks = []
    for _ in range(n):
        tasks.append(make_async_client())
    print(f"Creating {n} contexts, sleeping 10 secs")
    await asyncio.wait(tasks)


asyncio.run(main(2000))
print("Finished run, still using lots of memory")
gc.collect()
input("gc.collect() does not help :(")
```

Comparison with `aiohttp`
```
import aiohttp
import asyncio

print(f"aiohttp version {aiohttp.__version__}")


async def make_async_client():
    async with aiohttp.ClientSession() as client:
        await asyncio.sleep(10)


async def main(n):
    tasks = []
    for _ in range(n):
        tasks.append(make_async_client())
    print(f"Creating {n} contexts and sleeping")
    await asyncio.wait(tasks)


asyncio.run(main(200000))
input("Finished run, all memory is freed")
```

### Expected behavior

Memory gets freed, after exiting the async context, like for `aiohttp`

### Actual behavior
Memory does not get freed, even after explicitly calling `gc.collect()`

### Environment

- OS: Linux (many versions)
- Python version: 3.8.3
- HTTPX version: both 0.12.1 and `master`
- Async environment: both `asyncio` and `trio`
- HTTP proxy: no
- Custom certificates: no

### Additional context

I understand typically you need to have only one async ClientSession, but it shouldn't leak memory anyway, for very long running processes it can be a problem

Thanks for this great library! If you're interested I can try to debug this issue and send a PR

tomchristie:
That’s *super* surprising if so. Can you test against master and see if you’re getting the same results?

There’s not anywhere I can think of on a first pass where that could be an issue for us, so yeah any digging into it would be most welcome.

What are you using to measure the memory usage?

Recursing:
I did test against master, can you reproduce?
I'm just measuring system memory usage with `top` and `htop`

[...]

Issue #983, opened 2020-05-22
Standalone `create_ssl_context()` function?


tomchristie:
Now that we've got a nicely spec'ed Transport API we can give our users some pretty neat flexibility, but I think there's also one place where we've maybe got a gap we'd like to plug?...

Something that's important for us is to be able to easily demonstrate to our users how to drop down to lower levels of abstraction in the API when needed, for example it'd be really helpful how to instantiate a client directly with an `httpcore` connection pool...

```python
transport = httpcore.AsyncConnectionPool(...)
client = httpx.AsyncClient(transport=transport)
```

That's really valuable because it *helps* an understanding of how the pieces fit together, and also allows us to do stuff like add some low-level connection options on `httpcore`, without *necessarily* having to expose those all the way through to `httpx`. (Eg https://github.com/encode/httpcore/issues/88)

However, the one bit that's missing is that there currently isn't any *convenient* way for our users to generate an appropriately configured SSLContext without digging into the gnarly details.

I think we might well want to expose a stand alone function to provide users with our "generate an SSL context" functionality, like so...

```python
ssl_context = httpx.create_ssl_context()  # Accepts the standard 'verify', 'cert', trust_env' arguments.
```

Which then gives our users everything they need to directly plug in a properly configured `httpcore` connection pool...

```python
ssl_context = httpx.create_ssl_context()
transport = httpcore.AsyncConnectionPool(ssl_context=ssl_context)
client = httpx.AsyncClient(transport=transport)
```

Implementation-wise this'd just be...

```python
def create_ssl_context(...):
    return SSLConfig(...).ssl_context
```

And then expose the function at the top-level init.

*Incidentally I think our `SSLConfig` instances our now a bit of a vestige of past designs, and we could probably internally use the same `create_ssl_context` inside the client implementation, rather than hanging on to the intermediary config instance.*

florimondmanca:
I _think_ this should have been closed by https://github.com/encode/httpx/pull/996? :-)

PR #986, opened 2020-05-23
Moved  create and verify ssl context function into the utils  and included  lru cache.

cansarigol:
Refs #978 

_globally caching our SSL contexts_ suggestion of @tomchristie .

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -8,27 +8,13 @@ import certifi
 
 from ._models import URL, Headers
 from ._types import CertTypes, HeaderTypes, TimeoutTypes, URLTypes, VerifyTypes
-from ._utils import get_ca_bundle_from_env, get_logger, warn_deprecated
-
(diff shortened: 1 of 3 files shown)


florimondmanca:
Hi @cansarigol, thanks for tackling this!

From what I can see, this PR tries to address both #978 and #983. It's probably best that we keep fixes for these separate. Most likely #983 is higher priority, as it would make `SSLConfig` go away (and this PR currently uses it for the other part).

cansarigol:
Hi @florimondmanca , I thought that we can handle in a single PR and I can include a test about `SSLConfig().ssl_context == httpx.create_ssl_context()` .

tomchristie:
Closing in favour of #996

tomchristie:
Incidentally let's *not* jump straight in with implementing SSLContext caching once #996 is in. I think that aspect is worth *considering* but I don't think it's *necessarily* a given.

PR #990, opened 2020-05-24
Move test server to a separately importable module

florimondmanca:
Refs #650, prompted by https://github.com/encode/httpx/pull/985#discussion_r429614983

Being able to import the test `Server` is required* for annotating tests.

*At least, it's required with our _current_ test setup. Dunno if we might want to consider tweaking our mypy configuration with `--check-untyped-defs`, so that the inside of test functions is type checked even though their signature is not annotated. In general it's really the inside of tests (calls to HTTPX) that we want to make sure provide a good typing experience — type hinting the test function signatures doesn't provide a huge gain (but on the contrary a non-negligible amount of work is required if we did want to annotate them all).

So, marking this as a draft for now…

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,5 @@
-import asyncio
 import json
 import os
-import threading
-import time
 import typing
 
 import pytest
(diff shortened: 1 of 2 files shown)


florimondmanca:
Closing as I think #991 might be a better approach for us - one that will not even _require_ adding `server: Server` annotations everywhere just to have mypy type check the inside of test functions.

PR #992, opened 2020-05-24
Fix handling of None values in timeout tuple type hint

florimondmanca:
Fixes an issue with our `TimeoutTypes` that makes mypy complain if users pass `None`'s as part of a timeout tuple (although `None` are accepted, meaning "no timeout"):

```python
# example.py
import httpx

_ = httpx.Timeout((1.0, None, None, None))
```

```console
$ mypy example.py
example.py:3: error: Argument 1 to "Timeout" has incompatible type "Tuple[float, None, None, None]"; expected "Union[float, Tuple[float, float, float, float], Timeout, None]"
```

### DIFF
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -45,7 +45,11 @@ CookieTypes = Union["Cookies", CookieJar, Dict[str, str]]
 
 CertTypes = Union[str, Tuple[str, str], Tuple[str, str, str]]
 VerifyTypes = Union[str, bool, ssl.SSLContext]
-TimeoutTypes = Union[None, float, Tuple[float, float, float, float], "Timeout"]
+TimeoutTypes = Union[
+    Optional[float],
+    Tuple[Optional[float], Optional[float], Optional[float], Optional[float]],
+    "Timeout",
+]
 ProxiesTypes = Union[URLTypes, "Proxy", Dict[URLTypes, Union[URLTypes, "Proxy"]]]
 
 AuthTypes = Union[


httpx/_types.py
```diff
@@ -45,7 +45,11 @@
 
 CertTypes = Union[str, Tuple[str, str], Tuple[str, str, str]]
 VerifyTypes = Union[str, bool, ssl.SSLContext]
-TimeoutTypes = Union[None, float, Tuple[float, float, float, float], "Timeout"]
+TimeoutTypes = Union[
+    Optional[float],
+    Tuple[Optional[float], Optional[float], Optional[float], Optional[float]],
```
yeraydiazdiaz, line 50: `Timeout` seems to require [the first two elements of the tuple to be not `None`](https://github.com/encode/httpx/blob/master/httpx/_config.py#L229-L230), do we want to keep those as not optional?

florimondmanca: It seems actually we _do_ allow those to be `None` (which makes sense - don't think there's a reason we'd want to prevent people from disabling connect/read timeouts):

https://github.com/encode/httpx/blob/440b5ab95faf94f56472e08c8c5dbf59b2175fc6/httpx/_config.py#L210-L211

(However for some reason we allow any of a 2-, 3- or 4- tuple. 🤔  Also we don't document the tuple usage in [Fine-tuning timeouts](https://www.python-httpx.org/advanced/#fine-tuning-the-configuration), instead hinting people to use kwargs, eg `Timeout(read_timeout=...)`. Is that for Requests compatibility purposes?)

yeraydiazdiaz: Ah sorry, I misread that 👍 

I don't think there's a lot of benefit from the tuple form FWIW.

PR #993, opened 2020-05-24
Tweak headers type hints

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/985#discussion_r429615097

Fixes a type hints itch that our users can come across currently, similarly to https://github.com/python/mypy/issues/8477. This happens when providing headers from a separate variable, like we do in some of our tests:

https://github.com/encode/httpx/blob/82dc6f32f864d26677df8fc61249a7fc74c687e8/tests/models/test_responses.py#L40-L48

```python
# example.py
headers = {"content-type": "example"}
httpx.Headers(headers)
```

```console
$ mypy example.py
example.py:5: error: Argument 1 to "Headers" has incompatible type "Dict[str, str]"; expected "Union[Headers, Dict[Union[str, bytes], Union[str, bytes]], Sequence[Tuple[Union[str, bytes], Union[str, bytes]]], None]"
```

This error is super confusing, so we should fix it. (It's due to a limitation in mypy: `Mapping` is invariant in key (see https://github.com/python/mypy/issues/1114) — so `Union`'s don't typically work very well.)

This PR fixes it in the form of a surgical tweak to our `HeaderTypes` type alias, based on @hramezani's fix here: https://github.com/encode/httpx/pull/985#discussion_r429615097.

### DIFF
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -7,6 +7,7 @@ from http.cookiejar import CookieJar
 from typing import (
     IO,
     TYPE_CHECKING,
+    AnyStr,
     AsyncIterator,
     Callable,
     Dict,
@@ -38,7 +39,14 @@ QueryParamTypes = Union[
 ]
 
 HeaderTypes = Union[
-    "Headers", Dict[StrOrBytes, StrOrBytes], Sequence[Tuple[StrOrBytes, StrOrBytes]],
+    "Headers",
+    # NOTE: Mapping is invariant in key (https://github.com/python/mypy/issues/1114).
+    # So using `StrOrBytes` as the mapping key type would result in
+    # users come across https://github.com/python/mypy/issues/8477.
+    # The suggested solution is to use a generic key type (defined as a `TypeVar`),
+    # hence `AnyStr` here.
+    Mapping[AnyStr, StrOrBytes],
+    Sequence[Tuple[StrOrBytes, StrOrBytes]],
 ]
 
 CookieTypes = Union["Cookies", CookieJar, Dict[str, str]]


httpx/_types.py
```diff
@@ -38,7 +39,14 @@
...
 HeaderTypes = Union[
-    "Headers", Dict[StrOrBytes, StrOrBytes], Sequence[Tuple[StrOrBytes, StrOrBytes]],
+    "Headers",
+    # NOTE: Mapping is invariant in key (https://github.com/python/mypy/issues/1114).
+    # So using `StrOrBytes` as the mapping key type would result in
+    # users come across https://github.com/python/mypy/issues/8477.
+    # The suggested solution is to use a generic key type (defined as a `TypeVar`),
+    # hence `AnyStr` here.
+    Mapping[AnyStr, StrOrBytes],
+    Sequence[Tuple[StrOrBytes, StrOrBytes]],
```
tomchristie, line 49: Okay, so let's dig into something here...

Should we just use `AnyStr` everywhere here instead?

If we're using `Mapping[AnyStr, AnyStr]` I get that it implies that the type of the key must be the same throughout, and that the type of the value must be the same throughout. Does it also imply that the key and value types must also match?

In any case, why *not* adopt an invariant constraint on the types a user can properly use here. Either "all bytes" or "all str" is likely to result in tighter, better code.

florimondmanca: Honestly not sure about this - maybe it would be okay to enforce « either str or bytes, not a mix of both », and have users encode/decode explicitly. But at the same time it would require some runtime checks to enforce this behavior (not just at the type hints level), yet the current runtime behavior has been working well enough it seems.

In any case I’m tempted to push forward for this fix as currently we have a UX bug regardless of what we decide about this. Thoughts?

yeraydiazdiaz: > Does it also imply that the key and value types must also match?

Apparently not.

```python
import typing

AnyStrMapping = typing.Mapping[typing.AnyStr, typing.AnyStr]
MustMatchMapping = typing.Union[
    typing.Mapping[str, str], typing.Mapping[bytes, bytes],
]
m1: AnyStrMapping = {b"foo": "bar"}  # ok
m2: MustMatchMapping = {"foo": "bar"}  # ok
m3: MustMatchMapping = {b"foo": b"bar"}  # ok
m4: MustMatchMapping = {b"foo": "bar"}  # mypy error
```

I'm 👍  on merging this for the time being and discuss this further in another issue (FWIW I prefer the must-match version).

tomchristie: So I'd be okay with this *if* it switches to `AnyStr` throughout those *four* cases, rather than just the one key there. And at that point I think we can drop the comment too.

Alternately #999, which is similar, but also introduces some tighter checking on similar signatures in `_utils.py`

florimondmanca:
Closing in favor of #999

PR #995, opened 2020-05-25
Add "Content-Length: 0" on POST, PUT, PATCH if required.

### LINKED ISSUE #987: Content-Length header is missing in POST/DELETE/PUT/PATCH requests without body
### Checklist

- [x] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
- [x] I looked through similar issues on GitHub, but didn't find anything.
- [x] I looked up "How to do ... in HTTPX" on a search engine and didn't find any information.
- [x] I asked the [community chat](https://gitter.im/encode/community) for help ~~but didn't get an answer.~~ [and decided to create an issue. :)](https://gitter.im/encode/community?at=5ec95c46778fad0b13201d03)

### Environment
**httpx version**: `0.13.1`
**Python version**: `3.7.3`
**OS**: Windows 7

### Question
I am not sure how to classify this issue: it could be a bug, since it is an incompatible with `requests`, on the other hand, it could be feature request, if you consider it is not that important. So, I decided to go with question. :D

Was replacing `requests` with `httpx` in my small script and received an error from the server `httpx._exceptions.HTTPError: 411 Client Error: Length Required for url: ...` while doing a `POST`  request **without a body** (same behavior is applicable for `PUT` and `PATCH` requests too).

*Steps to reproduce:*
```python
import requests
import httpx

for client in (requests, httpx):
    for method in ("get", "head", "delete", "post", "put", "patch"):
        r = client.request(method, f"https://httpbin.org/headers")
        print(
            f"[{client.__name__}] method={method.upper()} "
            f'Content-Length={r.request.headers.get("Content-Length")}'
        )
```

`requests` adds `Content-Length` header in every possible http method, except `GET` and `HEAD`.
`httpx` does not add `Content-Length` header at all, **unless requests has a body**.

#866 added `Content-Length` to all methods, but since it is not needed in `GET` and `HEAD` methods, it was reverted.

I assume, it should be handled during the `Request` building.
`httpx` decides whether to add `Content-Length` depending on the steam type, which is handled in [_content_streams.py#L314](https://github.com/encode/httpx/blob/aa630d36c22642d7004a57abd222a14461ee4d77/httpx/_content_streams.py#L372).

### Thoughts on how to solve it
To be honest, I am not sure what is the proper way to deal with this. :-(

My first idea was to use `method` inside the [stream](https://github.com/encode/httpx/blob/aa630d36c22642d7004a57abd222a14461ee4d77/httpx/_content_streams.py#L49):
```python3
class ByteStream(ContentStream):
    def __init__(self, body: typing.Union[str, bytes], method: str) -> None:
        self.body = body.encode("utf-8") if isinstance(body, str) else body
        self.method = method

    def get_headers(self) -> typing.Dict[str, str]:        
        if self.method in ("GET", "HEAD"):
            return {}
        content_length = str(len(self.body)) if self.body else "0"
        return {"Content-Length": content_length}
```

But this approach is not great, since it requires to pass `method` from `Request` to `encode` and then to `ByteStream` inside it. Also, it is not clear what `method` should I use when building stream in [`read`](https://github.com/encode/httpx/blob/aa630d36c22642d7004a57abd222a14461ee4d77/httpx/_models.py#L656) methods.

Please, let me know what you think. Thanks!

tomchristie:
Closes #987

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -37,7 +37,7 @@ def test_build_request(server):
     assert response.json()["Custom-header"] == "value"
 
 
-def test_build_port_request(server):
+def test_build_post_request(server):
     url = server.url.copy_with(path="/echo_headers")
     headers = {"Custom-header": "value"}
 
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -37,6 +37,22 @@ def test_build_request(server):
     assert response.json()["Custom-header"] == "value"
 
 
+def test_build_port_request(server):
+    url = server.url.copy_with(path="/echo_headers")
+    headers = {"Custom-header": "value"}
+
+    with httpx.Client() as client:
+        request = client.build_request("POST", url)
+        request.headers.update(headers)
+        response = client.send(request)
+
+    assert response.status_code == 200
+    assert response.url == url
+
+    assert response.json()["Content-length"] == "0"
+    assert response.json()["Custom-header"] == "value"
+
+
 def test_post(server):
     with httpx.Client() as client:
         response = client.post(server.url, data=b"Hello, world!")
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -616,6 +616,9 @@ class Request:
         auto_headers: typing.List[typing.Tuple[bytes, bytes]] = []
 
         has_host = "host" in self.headers
+        has_content_length = (
+            "content-length" in self.headers or "transfer-encoding" in self.headers
+        )
         has_user_agent = "user-agent" in self.headers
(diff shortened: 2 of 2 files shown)


tests/client/test_client.py
```diff
@@ -37,6 +37,22 @@ def test_build_request(server):
     assert response.json()["Custom-header"] == "value"
 
 
+def test_build_port_request(server):
```
florimondmanca, line 40: Tiny typo :-)

```suggestion
def test_build_post_request(server):
```

PR #996, opened 2020-05-25
Included create_ssl_context function to create the same context with SSLConfig and serve as API

cansarigol:
Refs #983


docs/advanced.md
```diff
@@ -577,13 +577,11 @@ import httpx
...
 ```
 
-Alternatively, you can pass a standard library `ssl.SSLContext`.
+Alternatively, you can use `httpx.create_ssl_context`.
 
 ```python
->>> import ssl
 >>> import httpx
->>> context = ssl.create_default_context()
->>> context.load_verify_locations(cafile="/tmp/client.pem")
```
tomchristie, line 586: I wouldn't replace this section, instead I think we should leave it as is, but then supplement it by following it with something like this...

---

We also include a helper function for creating properly configured `SSLContext` instances...

```python
>>> context = httpx.create_ssl_context()
```

The `create_ssl_context` function accepts the same set of SSL configuration arguments as `httpx.Client`...

* `verify` - Blah blah blah
* `cert` - Blah blah blah
* `trust_env` - Blah blah blah
* `http2` - Blah blah blah

---

That'd help make it clear that you can either generate the SSLContext instance entirely yourself, or else use the helper function.

docs/advanced.md
```diff
@@ -623,6 +621,21 @@ If you do need to make HTTPS connections to a local server, for example to test
...
+```python
+>>> import httpx
+>>> ssl_context = httpx.create_ssl_context()
+```
+
+```python
+ssl_context = httpx.create_ssl_context()
+transport = httpcore.AsyncConnectionPool(ssl_context=ssl_context)
+client = httpx.AsyncClient(transport=transport)
+```
```
tomchristie, line 637: We haven't yet introduced the "transport" API, so this doesn't feel quite right here.

tests/test_config.py
```diff
@@ -38,10 +36,11 @@ def test_load_ssl_config_verify_env_file(https_server, ca_cert_pem_file, config)
         else str(Path(ca_cert_pem_file).parent)
     )
     ssl_config = SSLConfig(trust_env=True)
```
tomchristie, line 38: It'd be *great* as part of this change if we dropped using the private `SSLConfig` from the tests completely.
Not *necessarily* a blocker to getting this PR in, but might be worth a go?

tests/test_config.py
```diff
@@ -94,17 +90,16 @@ def test_load_ssl_config_cert_without_key_raises(cert_pem_file):
...
 def test_load_ssl_context():
     ssl_context = ssl.create_default_context()
-    ssl_config = SSLConfig(verify=ssl_context)
+    context = httpx.create_ssl_context(verify=ssl_context)
 
-    assert ssl_config.ssl_context is ssl_context
+    assert context is ssl_context
 
 
 def test_ssl_repr():
```
tomchristie, line 105: We could probably drop the `__repr__` and `__eq__` methods from `SSLConfig`, and then drop these two tests.
Given that we're not exposing the instance anyplace they're pretty much superfluous.

PR #997, opened 2020-05-25
Tighten public client methods

tomchristie:
For 1.0 we've got a bunch of methods that are nominally public, in that they're not currently underscore, but they really aren't - they're undocumented, not expected to be used, and we don't particularly want them as public API.

This is an initial PR at pulling a bunch of the internal methods on `Client` into a proper formally private API space.

I haven't necessarily tried to capture everything here. Right now `init_transport` and `init_proxy_transport` have still been left in the public API space, as has `transport_for_url`.

We wouldn't want to merge any of this direct to master, since that'd make it awkward for us to issue any 0.13.x releases, so perhaps we should have a longer-lived branch for any API tightening that requires a 0.14 or 1.0 release?


florimondmanca:
@tomchristie I see this is still a draft - is this ready for review, though?

I'm :100: on this any case. :+1:

I'm not too sure on maintaining a separate "next" branch, but we can try and see if we keep track of things. If we merge master into it quite often (every other bug fix PR) we should be fine.

jcugat:
What about making them all private, but still keeping their public counterpart with a warning? Something similar to what's already done with [ASGIDispatch](https://github.com/encode/httpx/blob/093cb4250006dc0b73af986f7e6537b8c4395dcf/httpx/_transports/asgi.py#L167) & [WSGIDispatch](https://github.com/encode/httpx/blob/093cb4250006dc0b73af986f7e6537b8c4395dcf/httpx/_transports/wsgi.py#L145).
That way we avoid having to maintain a separate branch, and at the same time we don't disrupt too much users of the 0.13/0.14 versions that are already using them.

tomchristie:
Ready for a review, so we can push towards a 0.14 release.

PR #999, opened 2020-05-26
Use AnyStr where appropriate

tomchristie:
Refs #993, #985 

Makes more careful use of `typing.Union[str, bytes]` vs. `typing.AnyStr`.

Use `AnyStr` where we'd expect the types to be the same across everything in the signature. Eg. `typing.Mapping[AnyStr, AnyStr]` means "choose either str or bytes here, but be consistent.".

Resolves an issues where mypy errors with a really simple bit of `httpx` usage...

**example.py**

```python
import httpx

headers = {"content-type": "example"}
httpx.Headers(headers)
```

**MyPy**

```shell
$ venv/bin/mypy example.py 
example.py:5: error: Argument 1 to "Headers" has incompatible type "Dict[str, str]"; expected "Union[Headers, Dict[Union[str, bytes], Union[str, bytes]], Sequence[Tuple[Union[str, bytes], Union[str, bytes]]], None]"
```

### DIFF
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -7,6 +7,7 @@ from http.cookiejar import CookieJar
 from typing import (
     IO,
     TYPE_CHECKING,
+    AnyStr,
     AsyncIterator,
     Callable,
     Dict,
@@ -24,7 +25,6 @@ if TYPE_CHECKING:  # pragma: no cover
     from ._config import Proxy, Timeout  # noqa: F401
     from ._models import URL, Cookies, Headers, QueryParams, Request  # noqa: F401
 
-StrOrBytes = Union[str, bytes]
 
 PrimitiveData = Optional[Union[str, int, float, bool]]
 
@@ -38,7 +38,7 @@ QueryParamTypes = Union[
 ]
 
 HeaderTypes = Union[
-    "Headers", Dict[StrOrBytes, StrOrBytes], Sequence[Tuple[StrOrBytes, StrOrBytes]],
+    "Headers", Dict[AnyStr, AnyStr], Sequence[Tuple[AnyStr, AnyStr]],
 ]
 
 CookieTypes = Union["Cookies", CookieJar, Dict[str, str]]
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -7,7 +7,6 @@ from http.cookiejar import CookieJar
 from typing import (
     IO,
     TYPE_CHECKING,
-    AnyStr,
     AsyncIterator,
     Callable,
     Dict,
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -38,7 +38,11 @@ QueryParamTypes = Union[
 ]
 
 HeaderTypes = Union[
-    "Headers", Dict[AnyStr, AnyStr], Sequence[Tuple[AnyStr, AnyStr]],
+    "Headers",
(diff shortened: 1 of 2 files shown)


tomchristie:
After this change...

```shell
$ venv/bin/mypy example.py 
Success: no issues found in 1 source file
```

yeraydiazdiaz:
> Eg. typing.Mapping[AnyStr, AnyStr] means "choose either str or bytes here, but be consistent.".

I don't think that's the case, either using `Mapping` or `Dict`, mypy is fine with the code below:

```python
import typing

AnyStrDict = typing.Dict[typing.AnyStr, typing.AnyStr]
AnyStrMapping = typing.Mapping[typing.AnyStr, typing.AnyStr]
d: AnyStrDict = {b"foo": "bar"}
m: AnyStrMapping = {b"foo": "bar"}
```

tomchristie:
Gotcha. It’s at least “consistent within this part of the type” tho right? Eg. Keys as str or Keys as bytes, not keys as str or bytes.

yeraydiazdiaz:
Apparently not 😕 , again mypy is completely fine with the code below:

```python
import typing

AnyStrDict = typing.Dict[typing.AnyStr, typing.AnyStr]
AnyStrMapping = typing.Mapping[typing.AnyStr, typing.AnyStr]
d: AnyStrDict = {"foo": "bar", b"bar": "baz"}
m: AnyStrMapping = {b"foo": "bar", "foo": "baz"}
```

Looks like the guarantees only hold in the context of a function. From the [docs](https://docs.python.org/3/library/typing.html#typing.AnyStr):

> It is meant to be used for functions that may accept any kind of string without allowing different kinds of strings to mix.

httpx/_types.py
```diff
@@ -38,7 +38,7 @@
 ]
 
 HeaderTypes = Union[
-    "Headers", Dict[StrOrBytes, StrOrBytes], Sequence[Tuple[StrOrBytes, StrOrBytes]],
+    "Headers", Dict[AnyStr, AnyStr], Sequence[Tuple[AnyStr, AnyStr]],
```
florimondmanca, line 41: Based off @yeraydiazdiaz's observations it seems this might be closer to what we need:

```suggestion
    "Headers",
    Dict[str, str],
    Dict[bytes, bytes],
    Sequence[Tuple[str, str]],
    Sequence[Tuple[bytes, bytes]],
```

It would work (i.e. tell users "use either str/str or bytes/bytes, but don't mix") even outside of functions :+1:

tomchristie:
Interestingly `MyPy`'s behaviour varies depending on if the annotation is used directly or indirectly, eg...

```python
from typing import Union, Dict, Sequence, Tuple, AnyStr

HeaderTypes = Union[
    Dict[AnyStr, AnyStr], Sequence[Tuple[AnyStr, AnyStr]],
]

def foo(h: HeaderTypes) -> None:
    print(h)


def bar(h: Union[Dict[AnyStr, AnyStr], Sequence[Tuple[AnyStr, AnyStr]]]) -> None:
    print(h)


headers = {b"content-type": b"example", "accept": "*/*"}
foo(headers)  # Ok
bar(headers)  # Error 
```

httpx/_types.py
tomchristie: Good call, folks.

PR #1002, opened 2020-05-27
Improve error when redirecting with custom schemes

### LINKED ISSUE #822: KeyError occurs when redirect to custom scheme(ex. market://)
# Information

OS platform : mac OS
Python version : 3.7.2
Installed dependencies and versions : `httpx==0.9.3`
Code snippet
```
    @property
    def port(self) -> int:
        port = self._uri_reference.port
        if port is None:
            return {"https": 443, "http": 80}[self.scheme]
        return int(port)
```

Error traceback
```
[2020-02-24 14:57:08 +0900] [82150] [ERROR] Exception
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/sanic/testing.py", line 120, in _collect_response
    method, url, *request_args, **request_kwargs
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/sanic/testing.py", line 41, in _local_request
    url, verify=False, *args, **kwargs
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 671, in get
    trust_env=trust_env,
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 268, in request
    trust_env=trust_env,
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 410, in send
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 478, in send_handling_redirects
    request = self.build_redirect_request(request, response)
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 500, in build_redirect_request
    headers = self.redirect_headers(request, url, method)
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/client.py", line 555, in redirect_headers
    if url.origin != request.url.origin:
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/models.py", line 215, in origin
    return Origin(self)
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/models.py", line 287, in __init__
    self.port = url.port
  File "/Users/sym/.pyenv/versions/airbridge-ads-was/lib/python3.7/site-packages/httpx/models.py", line 165, in port
    return {"https": 443, "http": 80}[self.scheme]
KeyError: 'market'
[2020-02-24 14:57:08 +0900] [82150] [INFO] Starting worker [82150]
[2020-02-24 14:57:08 +0900] [82150] [INFO] Stopping worker [82150]
[2020-02-24 14:57:08 +0900] [82150] [INFO] Server Stopped
```

# Description
i'm using sanic and sanic uses httpx to test web request.
when i make a redirect response which goes to "market://details?id=~~" (android market url)", KeyError occurred.
I think it is associated with `port` property method.
Is this the intended behavior?

Thank you.

jcugat:
Fixes #822

I used `Scheme "blah" not supported` as the error message instead of `URL scheme must be "http" or "https".` used in other parts of the code. It makes more sense to show the actual invalid scheme since it's coming from the server and not something that the user specified.

I also tried the same test with `allow_redirects=False` but unfortunately when the response object is built, it needs a port which we can't derive from the custom scheme. Not sure how to solve this without changing the port definition of `Url` to allow a `None` port, ideas?

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -322,6 +322,10 @@ class BaseClient:
 
         url = URL(location, allow_relative=True)
 
+        # Check that we can handle the scheme
+        if url.scheme and url.scheme not in ("http", "https"):
+            raise InvalidURL(f'Scheme "{url.scheme}" not supported.')
+
         # Handle malformed 'Location' headers that are "absolute" form, have no host.
         # See: https://github.com/encode/httpx/issues/771
         if url.scheme and not url.host:
--- a/tests/client/test_redirects.py
+++ b/tests/client/test_redirects.py
@@ -8,6 +8,7 @@ import pytest
 from httpx import (
     URL,
     AsyncClient,
+    InvalidURL,
     NotRedirectResponse,
     RequestBodyUnavailable,
     TooManyRedirects,
@@ -139,6 +140,17 @@ class MockDispatch(httpcore.AsyncHTTPTransport):
                 )
             else:
                 return b"HTTP/1.1", 200, b"OK", [], ByteStream(b"Hello, world!")
+
+        elif path == b"/redirect_custom_scheme":
+            status_code = codes.MOVED_PERMANENTLY
+            headers = [(b"location", b"market://details?id=42")]
+            return (
+                b"HTTP/1.1",
+                status_code,
+                b"Moved Permanently",
+                headers,
+                ByteStream(b""),
+            )
 
         return b"HTTP/1.1", 200, b"OK", [], ByteStream(b"Hello, world!")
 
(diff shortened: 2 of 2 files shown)


httpx/_client.py
```diff
@@ -322,6 +322,10 @@ def redirect_url(self, request: Request, response: Response) -> URL:
 
         url = URL(location, allow_relative=True)
 
+        # Check that we can handle the scheme
+        if url.scheme and url.scheme not in ("http", "https"):
```
jcugat, line 326: This pattern `("http", "https")` (or something similar) is repeated in a few places, like:

https://github.com/jcugat/httpx/blob/9160631d7fb957a319ce8985c97a858f8eab49a3/httpx/_client.py#L609
https://github.com/jcugat/httpx/blob/9160631d7fb957a319ce8985c97a858f8eab49a3/httpx/_models.py#L105
https://github.com/jcugat/httpx/blob/9160631d7fb957a319ce8985c97a858f8eab49a3/httpx/_models.py#L132

It would probably make sense to unify it in a single place.

tomchristie: In these kinds of cases, if it's a single-liner I'd often prefer *not* to introduce indirection. It's as clear as it can possibly be, and it's not a complicated expression.

PR #1003, opened 2020-05-27
Increase test coverage, document Headers.copy(), misc cleanup

jcugat:
Small work towards https://github.com/encode/httpx/issues/316, added a few tests to increase the coverage.

Also removed a couple of bits that were not used anymore.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -481,9 +481,6 @@ class Headers(typing.MutableMapping[str, str]):
         for header in headers:
             self[header] = headers[header]
 
-    def copy(self) -> "Headers":
-        return Headers(self.items(), encoding=self.encoding)
-
     def __getitem__(self, key: str) -> str:
         """
         Return a single header value.
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -481,6 +481,9 @@ class Headers(typing.MutableMapping[str, str]):
         for header in headers:
             self[header] = headers[header]
 
+    def copy(self) -> "Headers":
+        return Headers(self.items(), encoding=self.encoding)
+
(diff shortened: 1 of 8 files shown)


httpx/_models.py
```diff
@@ -481,9 +481,6 @@ def update(self, headers: HeaderTypes = None) -> None:  # type: ignore
         for header in headers:
             self[header] = headers[header]
 
-    def copy(self) -> "Headers":
-        return Headers(self.items(), encoding=self.encoding)
-
```
tomchristie, line 486: Did you mean to remove this? Evidently we're not using it right now, but we *probably* still want it as part of the public API. (Or at least, if we *did* want to remove it, then it's a *small* breaking change, so we'd need to do so in a major version bump, so we can call it out in the CHANGELOG)

yeraydiazdiaz: I'd prefer to keep it altogether, I feel a lot of our users to header manipulation and rely on this method.

yeraydiazdiaz:
No further objections from me. Thanks again @jcugat 💯

httpx/_models.py
jcugat: Hey @tomchristie, yes, I removed it on purpose since it isn't documented and I thought it was just used internally: https://www.python-httpx.org/api/#url

jcugat: Also I noticed the same thing can be done already with the class constructor, but happy to roll this change back if you prefer 🙂 
https://github.com/encode/httpx/blob/21d7e16559d9360ae3a5c5cfd23bab8bb85ee4a8/httpx/_models.py#L391

jcugat: I've added the method back and documented & tested it.

PR #1010, opened 2020-05-30
Replace remaining occurrences of dispatch with transport

jcugat:
Since there were still a few uses of `dispatch` instead of `transport` in tests, pytest complained about deprecation warnings.

Same thing happened with `hard_limit` and `max_connections` from `httpx.PoolLimits`.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -54,9 +54,6 @@ from ._utils import (
     warn_deprecated,
 )
 
-if typing.TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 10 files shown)


httpx/_models.py
```diff
@@ -54,9 +54,6 @@
     warn_deprecated,
 )
 
-if typing.TYPE_CHECKING:  # pragma: no cover
-    from ._dispatch.base import AsyncDispatcher  # noqa: F401
```
jcugat, line 58: Seems a leftover from #804

PR #1012, opened 2020-05-31
Increase test coverage - take 2

jcugat:
More work towards #316, continuation of #1003

After this is merged, the only remaining untested part will be `URLLib3Transport`.

Small note about `StreamContextManager`. [The only line not tested](https://github.com/encode/httpx/blob/8c84210555294ce10c681edae2df71627095a4e9/httpx/_client.py#L1543) is because the only usage of the context manager in an async environment is from [BaseClient.stream](https://github.com/jcugat/httpx/blob/e4048964d40fb1aec627c77d1c0adcd509fc1975/httpx/_client.py#L171) where it doesn't use the `close_client` parameter (and by default is `false`). The only other usage is from the [api.stream](https://github.com/jcugat/httpx/blob/e4048964d40fb1aec627c77d1c0adcd509fc1975/httpx/_api.py#L100) method but that is always called from a sync context, so the `__aexit__` will never have `close_client=True`. Should it be removed?

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -28,50 +28,45 @@ def get_header_value(headers, key, default=None):
     return default
 
 
-class AsyncMockTransport(httpcore.AsyncHTTPTransport):
+class MockTransport:
     def __init__(self, auth_header: bytes = b"", status_code: int = 200) -> None:
         self.auth_header = auth_header
         self.status_code = status_code
 
+    def _request(
+        self,
+        method: bytes,
+        url: typing.Tuple[bytes, bytes, int, bytes],
+        headers: typing.List[typing.Tuple[bytes, bytes]],
+        stream: ContentStream,
+        timeout: typing.Dict[str, typing.Optional[float]] = None,
+    ) -> typing.Tuple[
+        bytes, int, bytes, typing.List[typing.Tuple[bytes, bytes]], ContentStream
+    ]:
+        authorization = get_header_value(headers, "Authorization")
+        response_headers = (
+            [(b"www-authenticate", self.auth_header)] if self.auth_header else []
+        )
+        response_stream = JSONStream({"auth": authorization})
+        return b"HTTP/1.1", self.status_code, b"", response_headers, response_stream
+
+
+class AsyncMockTransport(MockTransport, httpcore.AsyncHTTPTransport):
     async def request(
-        self,
-        method: bytes,
-        url: typing.Tuple[bytes, bytes, int, bytes],
-        headers: typing.List[typing.Tuple[bytes, bytes]],
-        stream: ContentStream,
-        timeout: typing.Dict[str, typing.Optional[float]] = None,
+        self, *args, **kwargs
     ) -> typing.Tuple[
         bytes, int, bytes, typing.List[typing.Tuple[bytes, bytes]], ContentStream
     ]:
-        authorization = get_header_value(headers, "Authorization")
-        response_headers = (
-            [(b"www-authenticate", self.auth_header)] if self.auth_header else []
-        )
-        response_stream = JSONStream({"auth": authorization})
-        return b"HTTP/1.1", self.status_code, b"", response_headers, response_stream
+        return self._request(*args, **kwargs)
 
 
-class SyncMockTransport(httpcore.SyncHTTPTransport):
-    def __init__(self, auth_header: bytes = b"", status_code: int = 200) -> None:
-        self.auth_header = auth_header
-        self.status_code = status_code
-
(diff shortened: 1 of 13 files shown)


tests/client/test_auth.py
```diff
@@ -50,6 +51,29 @@ async def request(
         return b"HTTP/1.1", self.status_code, b"", response_headers, response_stream
 
 
+class SyncMockTransport(httpcore.SyncHTTPTransport):
```
jcugat, line 54: I didn't manage to reuse the code from `AsyncMockTransport` without copy&pasting it.

tests/client/test_redirects.py
```diff
@@ -155,9 +156,43 @@ async def body():
         return b"HTTP/1.1", 200, b"OK", [], ByteStream(b"Hello, world!")
 
 
+class SyncMockTransport(httpcore.SyncHTTPTransport):
```
jcugat, line 159: Same here.

tests/client/test_auth.py
```diff
@@ -114,12 +138,58 @@ def challenge_send(
         return b"HTTP/1.1", 401, b"", headers, ContentStream()
 
 
+class RepeatAuth(Auth):
```
jcugat, line 141: Moved from below, no code changes.

tests/client/test_auth.py
```diff
@@ -114,12 +138,58 @@ def challenge_send(
...
+            request.headers["Authorization"] = f"Repeat {index}"
+            response = yield request
+            nonces.append(response.headers["www-authenticate"])
+
+        key = ".".join(nonces)
+        request.headers["Authorization"] = f"Repeat {key}"
+        yield request
+
+
+class ResponseBodyAuth(Auth):
```
jcugat, line 167: Moved from below, no code changes.

tests/concurrency.py
```diff
@@ -10,6 +10,6 @@
 
 async def sleep(seconds: float) -> None:
     if sniffio.current_async_library() == "trio":
-        await trio.sleep(seconds)
+        await trio.sleep(seconds)  # pragma: nocover
```
jcugat, line 13: This `sleep()` method is only used from `conftest.TestServer`, but the server always runs under the `asyncio` loop so the `trio` branch is never executed.

tests/client/test_auth.py
yeraydiazdiaz: I would naively suggest moving the body of `request` to a separate function? Seems like it's all synchronous code, but it's not a big deal anyway.

tests/client/test_redirects.py
yeraydiazdiaz: I would suggest the same as above, this one is a bit longer so it adds a bit more noise, but again no big deal.

tomchristie:
Fantastic work, thanks *so much*!

Going to leave this one with @yeraydiazdiaz since he's got a better handle on it, but tested locally and looking really nice. Will try to help with the follow-up once it's in.

tests/client/test_auth.py
tomchristie: Personally I'm usually okay with a bit of carefully applied light duplication like this, but happy enough either ways. 😄

jcugat: I've moved it into a class instead, this way the `__init__` is also reused.

tests/client/test_redirects.py
jcugat: For this one it's a bit harder, since there are a couple of `part async for part in stream` inside. I tried changing those to `part for part in stream` but it fails with `RuntimeError: Attempted to call a sync iterator on an async stream.`.

It's only really used in `redirect_body_target` (in the other usages the content is discarded). Is there any way to fetch the stream independently from the context (async/sync) we're in?

jcugat: Nevermind, I used the example in [custom transports](https://www.python-httpx.org/advanced/#custom-transports) and it's working now :)

tests/client/test_redirects.py
```diff
@@ -348,19 +380,19 @@ async def test_can_stream_if_no_redirect():
 
 @pytest.mark.usefixtures("async_environment")
 async def test_cannot_redirect_streaming_body():
-    client = AsyncClient(transport=MockTransport())
+    client = AsyncClient(transport=AsyncMockTransport())
     url = "https://example.org/redirect_body"
 
     async def streaming_body():
-        yield b"Example request body"
+        yield b"Example request body"  # pragma: nocover
```
jcugat, line 387: Had to add this since we need a streaming body for this test, although is never consumed.

PR #1016, opened 2020-06-08
Add support for async multipart file contents

### LINKED ISSUE #1015: Streaming Get and Post of a File
### Checklist

- [x] I searched the [HTTPX documentation](https://www.python-httpx.org) but couldn't find what I'm looking for.
- [x] I looked through similar issues on GitHub, but didn't find anything.
- [x] I looked up "How to do ... in HTTPX" on a search engine and didn't find any information.
- [x] I asked the [community chat](https://gitter.im/encode/community) for help.

### Question

I would like to use httpx to perform a streaming get of a file and then stream that file to another service.  Is that possible with httpx?  Would it require receiving the entire file before making a streaming post to the other service?  It might be a useful example to have within the documentation. Thanks.

florimondmanca:
Fixes #1015 

Add an `AsyncMultipartContent` helper that requires passing a `length` hint, and review our multipart implementation to expect that in the `__aiter__()` code path (which we had already implemented to allow using multipart with `AsyncClient`).

Still needs a bit of refining, but submitting this for comments already…

The idea behind a dedicated wrapping helper is that there's no official "here's what async file I/O looks like" in Python (mostly because file I/O in kernels is almost always synchronous anyway, and eg `aiofiles` achieves async by deffering to threads). The only blocker to allowing any arbitrary async iterator is that we need to be able to peek the content length, otherwise we'd have to read everything, and lose the streaming behavior. (I'm actually thinking we might want this potentially-unsafe default to be opt-in, eg "raise exception unless user has set an allow flag", but that's something else.) So, hence this wrapper for "give me an aiterator + its expected total content length".

### DIFF
--- a/httpx/_content_streams.py
+++ b/httpx/_content_streams.py
@@ -173,6 +173,35 @@ class URLEncodedStream(ContentStream):
         yield self.body
 
 
+class AsyncMultipartContent(ContentStream):
+    """
+    Async streaming multipart request content.
+    """
+
+    # NOTE: this helper is part of the public API.
+
+    def __init__(
+        self, aiterator: typing.AsyncIterator[bytes], content_length: int
+    ) -> None:
+        self._aiterator = aiterator
+        self._content_length = content_length
+
+    def can_replay(self) -> bool:
+        return False
+
+    def get_headers(self) -> typing.Dict[str, str]:
+        return {"Transfer-Encoding": "chunked"}
+
+    def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        return self._aiterator.__aiter__()
+
+    def __iter__(self) -> typing.Iterator[bytes]:
+        raise RuntimeError("Attempted to call an async iterator on an sync stream.")
+
+    def length(self) -> int:
+        return self._content_length
+
+
 class MultipartStream(ContentStream):
     """
     Request content as streaming multipart encoded form data.
@@ -222,6 +251,10 @@ class MultipartStream(ContentStream):
             yield self.render_headers()
             yield self.render_data()
 
+        async def arender(self) -> typing.AsyncIterator[bytes]:
+            for chunk in self.render():
+                yield chunk
+
     class FileField:
         """
         A single file field item, within a multipart form field.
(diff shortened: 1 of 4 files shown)


httpx/_content_streams.py
```diff
@@ -173,6 +173,35 @@ async def __aiter__(self) -> typing.AsyncIterator[bytes]:
...
+    ) -> None:
+        self._aiterator = aiterator
+        self._content_length = content_length
+
+    def can_replay(self) -> bool:
+        return False
+
+    def get_headers(self) -> typing.Dict[str, str]:
+        return {"Transfer-Encoding": "chunked"}
+
```
florimondmanca, line 194: I can probably drop these two + `__iter__()` and not make this derive from any base class - it's not actually used as a proper content stream in the multipart impl.

florimondmanca:
Closing for now since as much as I'd like to see it possible to pass an arbitrary async iterable to HTTPX, I'm not convinced this is the best approach, and there's probably some initial discussion to be had first.

samuelcolvin:
Is there any interest in reopening this or accepting a PR to implement something similar? aiofiles support in httpx for multipart file uploads seems a natural fit for httpx.

florimondmanca:
@samuelcolvin Hello,

I think we could have a discussion around async file support. It's just not obvious what we should decide to do. You say asyncio seems natural, but HTTPX aims at being async agnostic: what would we do for trio or curio? Instead, currently my best intuition would be a custom interface that accepts just what we need (probably an async iterator and a file size hint). But happy to discuss further, yes. :)

samuelcolvin:
(Think you mean aiofiles, not asyncio, but it's more or less the same thing since I think aiofiles only supports asyncio)

I see three possible ways forward (these would all solve my problem, but perhaps not those of others):
* async iterator and a file size, as you suggest
* Your own logic which duplicate a little of aiofiles (e.g. uses a thread pool executor to read the file) to allow a `Path` to be passed to `files=` which httpx then reads without blooking the event loop, not sure how hard this would be to do in an event loop agnostic way
* breaking out the multipart form building logic so I could use it myself with aiofiles and then stream the resulting bytes using `data=`

PR #1021, opened 2020-06-13
Refactor ASGITransport.request()

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/998#issuecomment-643623769, small bit of refactoring for `ASGITransport` to help reduce the diff there. I think these are valid readability improvements any case.

- Unify `typing` import style.
- Declare related variables in logical groups.
- Cleanup any unnecessary `nonlocal`'s
- Add some comments on how everything is split.

### DIFF
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -1,5 +1,4 @@
-import typing
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Tuple, Union
 
 import httpcore
 import sniffio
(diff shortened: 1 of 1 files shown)


httpx/_transports/asgi.py
```diff
@@ -144,7 +148,7 @@ async def send(message: dict) -> None:
         try:
             await self.app(scope, receive, send)
         except Exception:
-            if self.raise_app_exceptions or not response_complete:
+            if self.raise_app_exceptions or not response_complete.is_set():
```
florimondmanca, line 151: I _think_ this was a typo before.

JayH5: Think that was my mistake. Missed this when changing `response_complete` from a bool to an event. Also didn't realise the `nonlocal` could be removed 😬

PR #1026, opened 2020-06-16
Tweak .stream() API

florimondmanca:
Refs #830 

A proof-of-concept (but working) PR to illustrate https://github.com/encode/httpx/issues/830#issuecomment-644974594

- Add context manager capabilities to `Response`. (This brings us closer to what Requests does.)
- Make `.stream()` return a `Response`, allowing both context-managed and non-context-managed usage with an equivalent level of user experience.

The high-level `httpx.stream()` is kept to context-managed only because we also need to close the client on exit (and it would be awkward to eg implement a close callback API on the `Response` class just for this internal use case).

The breaking change here is that users would need to use `async with await client.stream()` in the async context-managed use case (extra `await`). This usage might be surprising but in line with other things such as [trio's `open_file()` API](https://trio.readthedocs.io/en/stable/reference-io.html#trio.open_file).

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -62,13 +62,44 @@ await client.aclose()
 
 ### Streaming responses
 
-The `AsyncClient.stream(method, url, ...)` method is an async context block.
+The `await AsyncClient.stream(method, url, ...)` method can be used as async context block:
 
 ```python
->>> client = httpx.AsyncClient()
->>> async with client.stream('GET', 'https://www.example.com/') as response:
->>>     async for chunk in response.aiter_bytes():
->>>         ...
+async with httpx.AsyncClient() as client:
+    async with await client.stream('GET', 'https://www.example.com/') as r:
+        async for chunk in r.aiter_bytes():
+            ...
+```
+
+!!! note
+    The "`async with await`" here is not a typo: you must first call `await client.stream(...)`, and then use the result as an async context manager.
+
+Alternatively, use `r = await client.stream(method, url, ...)` if you want to close the response explicitly:
+
+```python
+client = httpx.AsyncClient()
+
+r = await client.stream('GET', 'https://www.example.com/')
+
+try:
+    async for chunk in r.aiter_bytes():
+        ...
+finally:
+    await r.aclose()
+```
+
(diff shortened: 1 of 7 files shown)


florimondmanca:
Thanks!

I need to do a pass on docs (and highlight that the extra await there isn’t a typo 🤣).

tomchristie:
Grand. Will aim to take a look this week. *Potentially* some mixed thoughts about it, but needs a proper look! 🙂

docs/async.md
```diff
@@ -62,13 +62,43 @@ await client.aclose()
...
 
 ```python
->>> client = httpx.AsyncClient()
->>> async with client.stream('GET', 'https://www.example.com/') as response:
->>>     async for chunk in response.aiter_bytes():
->>>         ...
+async with httpx.AsyncClient() as client:
+    async with await client.stream('GET', 'https://www.example.com/') as r:
+        async for chunk in r.aiter_bytes():
+            ...
```
florimondmanca, line 71: Removed the prompt symbols to align with the rest of the snippets in this page.

tomchristie:
So my reservations around this are to do with the `.stream` API already being deliberately designed to ensure that it's not possible that incorrectly use it, and fail to close connections as a result.

Changing things this way around would mean that...

* It becomes easy for developers to mis-use the `.stream` API, by using the sequential style, without an enforced final close.
* It becomes possible for developers to use context managed style on the plain calls, eg `async with await client.get() as response` which is superfluous and misleading.

So I'm not yet convinced that it's clear if we shouldn't just stick with helping developers down the straight and narrow path, while allowing `.build_request` and `.send` if you *really* need to break out of that.

The pointer to trio's `open_file` API is a decent counter-point although there's a somewhat different set of concerns there, in that they'll want to be mirroring stdlib's `open` style as closely as possible, so it naturally leads to distinct `async with` and `await` components.

florimondmanca:
Those are valid points @tomchristie, thanks. Opened #1046 to explore the "just document how to use manual streaming" path, and I think it's the most convincing option. Closing this in favor of #1046 for now…

PR #1032, opened 2020-06-24
Add support for multiple files per POST field

### LINKED ISSUE #777: Support for multiple files per POST field

euri10:
fixes #777 
inspired by great work done on #891 
separates files and data as per @florimondmanca suggesiton in gitter
I commented below a mypy error I ignored because I'm clueless on how to deal with it to be honest


httpx/_content_streams.py
```diff
@@ -328,7 +328,8 @@ def _iter_fields(
             else:
                 yield self.DataField(name=name, value=value)
 
-        for name, value in files.items():
+        file_items = files.items() if isinstance(files, dict) else files
+        for name, value in file_items:  # type: ignore
```
euri10, line 332: No idea how to fix the mypy error on this line `httpx/_content_streams.py:332: error: Unpacking a string is disallowed
`

florimondmanca: I think it’s because the type hints is using Mapping but we’re only matching against dict (a specific type of mapping), so mypy assumes that the other branch could also receive other types of mappings, and iterating over a mapping yields its keys (strings).

It’s actually possible to `isinstance(obj, typing.Mapping)` (since it’s the same Mapping class than in collections.abc), so I think we can go for that?

Alternatively, switch the `if` to match a tuple (and treat the default case as mapping) - probably a bit less surprising.

tests/test_content_streams.py
```diff
@@ -204,3 +204,50 @@ async def test_empty_request():
...
+
+@pytest.mark.asyncio
+async def test_multipart_multiple_files_single_input_content():
+    files = [
+        ("file", io.BytesIO(b"<file content 1>")),
+        ("file", io.BytesIO(b"<file content 2>")),
+    ]
+    stream = encode(files=files, boundary=b"+++")
+    sync_content = b"".join([part for part in stream])
+    async_content = b"".join([part async for part in stream])
```
florimondmanca, line 217: Is there value in testing for the sync and async content concatenation here? I think we’re using async for in the other test cases.

(If there’s a missing test case for “async for yields the same than for” then we could add it, but separately to this PR imo.)

docs/advanced.md
```diff
@@ -464,6 +464,16 @@ MIME header field.
     It is safe to upload large files this way. File uploads are streaming by default, meaning that only one chunk will be loaded into memory at a time.
  
  Non-file data fields can be included in the multipart form using by passing them to `data=...`.
+ 
+You can also send multiple files in one go with a multiple file field form.
+To do that you'll need to pass a list of tuples that looks like `(form_field, (2_or_3_elements_tuple))` where the `2_or_3_elements_tuple` is described above and the `form_field` is the name of your form field.
```
florimondmanca, line 469: Little copy suggestion that _might_ be easier to understand?

```markdown
To do that, pass a list of `(field, <file>)` items instead of a dictionary, allowing you to pass multiple items with the same `field`.
```

docs/advanced.md
```diff
@@ -464,6 +464,16 @@ MIME header field.
...
+For instance this request sends 2 files, `foo.png` and `bar.png` in one request on the `images` form field: 
+
+```python
+>>> multiple_files = [('images', ('foo.png', open('foo.png', 'rb'), 'image/png')),
```
florimondmanca, line 473: Should we stick to naming this variable `files` for consistency with the other examples?

tests/test_content_streams.py
euri10: I basically took the previous PR test and didn't think about it but looking all tests inside the module all have the very same joins on both sync and async.

httpx/_content_streams.py
euri10: fixed !

docs/advanced.md
euri10: implemented

docs/advanced.md
euri10: yep, done

tests/test_content_streams.py
florimondmanca: Ah, right, sorry then I didn’t actually check the code before writing that up. 😅 Either way works then, and we can clean up the tests later.

docs/advanced.md
florimondmanca: Sorry, I meant renaming `multiple_files` to `files` (the `images` field name is very nice considering the example deals with sending images).

euri10: too many files ! :) I changed back and updated the one you had in mind !

PR #1035, opened 2020-06-26
Non relative dir tests

euri10:
Current tests works fine and all pass running the script.

However it assumes the working directory under which pytest runs is `httpx`

When using various IDE, in my case PyCharm you run tests by default using `httpx/tests` as the working directory, which is why the 4 tests "fixed" will fail because of the files/directories discovery. You can modify this of course, but the reason why the tests are failing is not immediately obvious.

So long story short it's a non-essential PR but I scratched my head on this long enough and this would avoid someone willing to contribute losing time on tests failing because of his configuration

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -1,6 +1,7 @@
 import hashlib
 import os
 import typing
+from pathlib import Path
 
 import httpcore
 import pytest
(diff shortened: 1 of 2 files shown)


jcugat:
The pattern I usually do is have a `tests/utils.py` file with something like:

```py
from pathlib import Path

ROOT_PATH = Path(__file__).parents[1]
FIXTURES_PATH = ROOT_PATH / "tests/fixtures/"
```

And then it could be used like this from tests:

```py
os.environ["NETRC"] = FIXTURES_PATH / ".netrc"
```

In some cases I also put some more utils functions there:

```py
def get_fixture_bytes(path: str):
    return Path(FIXTURES_PATH / path).read_bytes()


def get_fixture_text(path: str):
    return Path(FIXTURES_PATH / path).read_text()


def get_fixture_json(path: str):
    with (FIXTURES_PATH / path).open() as f:
        response = json.load(f)
    return response
```

But I don't think this last part is needed in this project.

florimondmanca:
Hey,

Closing this one in favor of #1052 for now given the comments above - happy to have :eyes: on that one!

PR #1036, opened 2020-06-29
Improve list/dict comprehensions

hukkinj1:
- Use `list(iterable)` for converting to a list rather than `[item for item in iterable]`
- Use dict comprehension rather than a list comprehension wrapped in `dict()`


hukkinj1:
I added a further commit that simplifies the `b"".join([item for item in iterable])` lines to the form `b"".join(iterable)`

I.e. I removed the any-iterable-to-list conversion altogether. It seems to me there is no need for it, seeing that `bytes.join()` can take any iterable just the same way that `list()` can.

PR #1037, opened 2020-06-29
Remove .encode("ascii") call from null byte

hukkinj1:


florimondmanca:
Heya,

Can you clarify the motivation for this change? Doesn’t look obvious that this wouldn’t at least require another change in our decoding code...

hukkinj1:
Hey there! No major motivation. Just that AFAIK `"\x00".encode("ascii")` should always (in Python 3) be equal to the zero byte `b"\x00"` and I think writing `b"\x00"` is shorter, more explicit, and faster at runtime.

Also the comment seems outdated to me if Python2 is not supported.

florimondmanca:
Ah, I didn’t see the `b` prefix was added! Thanks. :)

Issue #1029, opened 2020-06-18
TimeoutException missing from package


HenrikOssipoff:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

[According to the documentation](https://www.python-httpx.org/advanced/#timeout-configuration), there should be an `httpx.TimeoutException` class that you can catch if you want to catch any timeout related exception. However, such exception doesn't exist.

### To reproduce

```python
import httpx
httpx.TimeoutException
```

### Additional context

I'm assuming it's intended behaviour, and it's a case of the docs not being updated when introducing httpcore over urllib3. However, I'm unsure about what the correct approach now is; should one catch HTTPError instead, or are we forced to catch each individual timeout related exception?

HenrikOssipoff:
I see now that the `TimeoutException` is living in `httpcore` since 0.13.
I feel like this should be documented (did I miss it?).

Would it maybe make more sense to wrap these exceptions in `httpx` instead, so that the user doesn't have to worry about the underlying library?

johnanthonyowens:
This is covered in #949.

florimondmanca:
Hey,

As mentioned above this is part of a known issue tracked via #949, going to close this as a duplicate.

Thanks

Issue #949, opened 2020-05-13
Exceptions - Documentation, Hierarchy & Naming.


tomchristie:
We ought to comprehensively revisit our exceptions for 1.0, and make sure we're super happy we've got a public API we're happy to stick by for 1.0.

* Naming.
* The hierarchy.
* Decent documentation.
* `.request`/`.response` attributes (can we make either of these mandatory in some cases?)

See also: #940, #884, #869.

tomchristie:
Current hierarchy looks like this...

## Transport Errors, mapped from `httpcore`

We're currently synonym'ing these, but we *might* want to properly *map* them instead, so we can subclass our base class, and so that we can include a mandatory `request=...` parameter.

Also, we might want to additionally subclass the `stdlib` TimeoutError for the timeout types.

* `ProtocolError`
* `ProxyError`
* `TimeoutException` *Exists in `httpcore`, but not in `httpx`*
  * `PoolTimeout`
  * `ConnectTimeout`
  * `ReadTimeout`
  * `WriteTimeout`
* `NetworkError`
  * `ConnectError`
  * `ReadError`
  * `WriteError`
  * `CloseError`

## Content Errors

* `DecodingError`

## Redirect Errors

* `RedirectError`
  * `TooManyRedirects`
  * `NotRedirectResponse`

## Stream Errors

We might want to additionally subclass the `stdlib` RuntimeError for (some of?) these types.

It's actually possible that we might want to just drop some of these from the API completely, and instead use `RuntimeError`, since not all of these ought to actually be handled exception cases, but are instead unconditionally a "the developer has broken something" case.

* `StreamError`
  * `RequestBodyUnavailable`
  * `StreamConsumed`
  * `ResponseNotRead`
  * `RequestNotRead`
  * `ResponseClosed`

Other cases...

* `InvalidURL`
* `CookieConflict`

johnanthonyowens:
+1 in favor of mapping the httpcore exceptions to a unified hierarchy of httpx exceptions. I have a library that I recently updated from httpx 0.12 to 0.13 and found that getting the equivalent of `except HTTPError` was tricky due to the fragmented exception hierarchy.

MicahLyle:
> 
> 
> +1 in favor of mapping the httpcore exceptions to a unified hierarchy of httpx exceptions. I have a library that I recently updated from httpx 0.12 to 0.13 and found that getting the equivalent of `except HTTPError` was tricky due to the fragmented exception hierarchy.

[...]

PR #1044, opened 2020-07-02
Map HTTPCore exceptions

### LINKED ISSUE #1023: 0.13: some `httpcore` exceptions are missing from top level package
### Checklist

### Describe the bug
I believe that `HTTPError` is expected to be the base exception for all exceptions that httpx may raise? Since 0.13 this is no longer true and comments in [`_exceptions.py`](https://github.com/encode/httpx/blob/master/httpx/_exceptions.py#L11) indicate this is a bug. In real world use I have at least one bit of code that has failed due to this. I could see an argument for trying to catch a more specific error, but I think this case `HTTPError` was really nice to use.

Even if users should be using more specific exceptions having a single base exception is an incredibly useful feature for any library and it'd be great if we could ensure this behavior is set for the upcoming 1.0 release.

### To reproduce
The following code assumes localhost does not have any webserver running on port 80. Running on httpx 0.12 it correctly catches the `HTTPError`, running on 0.13 it does not catch the exception (httpcore.ConnectError, which doesn't appear to be properly exported in [`\_\_init\_\_.py](https://github.com/encode/httpx/blob/master/httpx/__init__.py), though it is aliased in `_exceptions.py`).
```python
import httpx
try:
    httpx.get('http://localhost')
except httpx.HTTPError:
    print('There was an httpx error')
```

### Expected behavior
If `HTTPError` is the base exception for httpx I'd expect catching `HTTPError` to actually catch all errors httpx may raise.

### Actual behavior
Instead of catching `HTTPError` it appears that httpcore errors bubble up. This breaks exception handling that expects `HTTPError` to be the base exception.

### Environment
- OS: macOS 10.15.5
- Python version: 3.8.3
- HTTPX version: 0.13
- Async environment: asyncio
- HTTP proxy: No
- Custom certificates: No

florimondmanca:
Fixes #1023, refs #1029, refs #949

Map any `httpcore` exceptions to (new) HTTPX-specific exceptions, so that we can guarantee that this kind of usage will catch _any_ exception raised by HTTPX:

```python
try:
    ...
except httpx.HTTPError:
    ...
```


florimondmanca:
#1045 just helped me catch that `TimeoutException` hadn't been added to `httpx.*`. ;-)

PR #1048, opened 2020-07-04
Remove needless whitespace strips

hukkinj1:
Dont strip whitespace from base64 strings. Whitespace characters are not included in base64 character set, so AFAIK it should be impossible for `b64encode` output to contain them.


hukkinj1:
CI fails because isort v5 was released. Merging https://github.com/encode/httpx/pull/1050 should fix this.

PR #1049, opened 2020-07-04
Simplify bytes to hex string conversion

hukkinj1:


hukkinj1:
CI fails because isort v5 was released. Merging https://github.com/encode/httpx/pull/1050 should fix this.

PR #1052, opened 2020-07-05
Use relative tests directory references

florimondmanca:
Alternative to #1035

cc @jcugat @euri10


euri10:
lgtm ! Sorry couldn't handle more of it, holidays 😁

tests/common.py
```diff
@@ -0,0 +1,5 @@
+import pathlib
+
+TESTS_DIR = pathlib.Path(__file__).parent
+TESTS_DIR = TESTS_DIR.relative_to(TESTS_DIR.parent)  # Ensure relative to project root.
```
jcugat, line 4: Is there any benefit of using a relative path? I think all the usages admit a full one.

florimondmanca: Mostly so that the existing assertions in `tests/test_utils.py` don't have to be changed (eg `assert ... == 'tests'`). Otherwise we'd be setting env variables to eg `/Users/florimond/.../tests/.netrc` and so we'd have to assert against that, and I realized it's not trivial due to trailing slashes etc. So, went for keeping things simple here.

jcugat: Good point, I missed this part (and learned about `relative_to()` in the way!).

jcugat: After thinking about it again, it seems that using the relative path will still have the same problems that were reported in https://github.com/encode/httpx/pull/1035

So running the tests from inside the `httpx/tests/` will fail since it won't find the correct fixtures path.

An alternative would be to remove line 4 from `tests/common.py` and update the test to check `assert get_ca_bundle_from_env().endswith("tests")` (or similar) instead.

florimondmanca: Good call @jcugat, updated the PR. Running `pytest` from inside `httpx/tests/` runs tests fine now. :-)

florimondmanca:
@tomchristie Ping for ✅? :)

PR #1050, opened 2020-07-04
Make isort configuration compatible with isort v5

hukkinj1:
See https://timothycrosley.github.io/isort/CHANGELOG/

isort removed the flags `--recursive` and `--apply`. They are the default behavior in isort v5.

PR #1053, opened 2020-07-05
Pin isort version

jcugat:
While reviewing https://github.com/encode/httpx/pull/1050 I noticed that `isort` is not pinned to the major version like it was done in httpcore's PR https://github.com/encode/httpcore/pull/111. I think it's a good idea to have it in both.

Related to this, should we pin all the development dependencies to strict versions and let someting like dependabot keep them updated? That way we would avoid breakages like what happened here: https://github.com/encode/httpx/pull/1048#issuecomment-653781722

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ cryptography
 flake8
 flake8-bugbear
 flake8-pie
-isort
+isort==5.*
 mypy
 pytest
 pytest-asyncio


florimondmanca:
> Related to this, should we pin all the development dependencies to strict versions and let someting like dependabot keep them updated? That way we would avoid breakages like what happened here: #1048 (comment)

I think there's a wider discussions to be had with @tomchristie about how dev dependencies should be managed and pinned. I do remember we've been affected by breaking changes in those in the past (though I don't remember _how_ exactly), and I'd agree a more strict pinning of them could be beneficial. But managing and updating those pins manually in a consistent way across repos sounds difficult.

So dependabot could be a nice way forward… I think it would just need some playing around to see if it could work for us w/o too much overhead. Eg maybe enable it here first and see how that goes? :-)

jcugat:
@florimondmanca Good point about the dependencies. I've updated my fork to [pin all the dependencies](https://github.com/jcugat/httpx/commit/af456d11f7f0d3dcd4bf6f0b83e3786732fcb986) and also [enabled dependabot to bump them automatically](https://github.com/jcugat/httpx/commit/65afad0fd48abb0cda8fbc6284b6b06754b7095c). I'll report in a few days (probably by opening a separate issue) to see how it goes.

Issue #1043, opened 2020-07-02
Considering non env-var 'no_proxy' support?


florimondmanca:
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

Hey all,

So while looking into an issues with `no_proxy` in Requests (https://github.com/psf/requests/issues/5000) I tried one of the failing cases with HTTPX, and ended up realizing that we don't take into account `no_proxy` in the `proxies={...}` usage. We only account for the `no_proxy`/`NO_PROXY` environment variable.

Eg this will fail as the request will be routed through the proxy:

```python
import httpx

proxies = {'https': 'http://no.where.com', 'no_proxy': 'google.com'}
client = httpx.Client(proxies=proxies)
client.get('https://google.com')  # ConnectError
```

However, this use case is covered by Requests (although with rough edges - see https://github.com/psf/requests/issues/5000).

```python
import os
import requests

os.environ['HTTPS_PROXY']='https://b.r.o.k.e.n.com'
proxies = {'no_proxy': 'google.com'}
r = requests.get('https://google.com', proxies=proxies)
```

## Describe the solution you would like.

The Requests way of passing a `no_proxy` key in the dict is somewhat clunky tbh, and also not compatible with us allowing users to pass an `httpx.Proxy` config object, or an URL, instead of a dict.

So I'd be thinking about a dedicated `no_proxy` option on clients, which accepts a list of domains or IPs (similar to how the env variable accepts a comma-separated string):

```python
httpx.Client(proxies=..., no_proxy=['<domain>', '<ip>', ...])
```

In terms of implementation, we could easily store this as an extra attribute, and then pass it when checking for "should we use proxies for this URL", without interfering with the proxy map building logic:

https://github.com/encode/httpx/blob/620b0670db07ae844aa6fae40d9209a856440c9d/httpx/_client.py#L555-L556

The priority order compared to env vars would be:

1. `no_proxy=[...]`
2. Environment variables (`no_proxy`/`NO_PROXY`)

Happy to discuss use cases and API details further!

tomchristie:
So a consistent way to approach this would be to allow the values in the `proxies=...` dictionary to be `None`, so eg...

`proxies={'all': 'http://127.0.0.1:1234', 'all://google.com': None}`

tomchristie:
[...]

Issue #1039, opened 2020-06-30
ASGI requests are routed through proxy


marco-neumann-by:
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Using the `AsyncClient` with an ASGI app seems to go through the proxy. Either this is is bug or I don't understand under which conditions this would make sense.

### To reproduce

```python
import asyncio
import os
import httpx


async def hello_world(scope, receive, send):
    status = 200
    output = b"Hello, World!"
    headers = [(b"content-type", "text/plain"), (b"content-length", str(len(output)))]

    await send({"type": "http.response.start", "status": status, "headers": headers})
    await send({"type": "http.response.body", "body": output})


async def test_asgi_with_env():
    os.environ["http_proxy"] = "http://localhost:1111"

    client = httpx.AsyncClient(app=hello_world)
    response = await client.get("http://www.example.org/")
    assert response.status_code == 200
    assert response.text == "Hello, World!"


asyncio.run(test_asgi_with_env())
```

### Expected behavior

Test should pass, no proxy interaction should be made.

Note that passing `trust_env=False` to `AsyncClient` disables the proxy and the test passes. However, this interaction seems weird.

### Actual behavior

```text
httpcore._exceptions.ConnectError: Multiple exceptions: [Errno 61] Connect call failed ('::1', 1111, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 1111
```

### Environment

- OS: macOS
- Python version: 3.8.3
- HTTPX version: 0.13.2 and on master (`cfd07cc290cc610fceb8d9f025c6bebc1d4b3275`)
- Async environment: asyncio, but also tested with trio where the issue also occurs.
- HTTP proxy: see description
- Custom certificates: no

### Additional context
https://github.com/tiangolo/fastapi/pull/1619

florimondmanca:
Might be worth looking if there’s any past discussion about whether we should make `trust_env` default to `False`... I think we’ve had strange UX like these in the past due to HTTPX using env cars when it didn’t really make sense.

Alternatively, on a smaller scope, we can look at the auto proxies setup code, see if it makes sense to bypass it if an `app` was passed... (Noting that it’s fixes the behavior in this particular case only, eg custom dispatchers where mount-by-scheme doesn’t really make sense could also experience this bug in the future. But you know, YAGNI...)

tomchristie:
[...]

Issue #977, opened 2020-05-21
Mount API


tomchristie:
Somewhat related to #927, #954 

We now have a really nicely defined Transport API. We also have the ability to mount proxies under particular schemes, or particular hosts.

Really we ought to be generalising that into a "mount a transport onto this scheme/domain/...".

Eg...

```python
client.mount('http:', ...)  # Mount a particular transport instance to handle all "http" scheme requests.
```

There's an equivalent API in requests... https://requests.readthedocs.io/en/master/user/advanced/#transport-adapters

```python
>>> s = requests.Session()
>>> s.mount('https://github.com/', MyAdapter())
```

Right now the closest we have is `proxies={<mount-point>: <Proxy or Transport>}`. However the set of mount points that are supported doesn't feel super well defined, and it *nominally* only supports proxies (even through they can actually be *any* transport instance).

Here's a bit of a proposal for how a more consistent, and generalised mount API might look...

* Support a `.transports = {'http': ..., 'https': ...}` read-only property on the client.
* The default `transport` sets `'http'` and `'https'`.
* Transports can only be mounted while a client is closed. We'll track that status in `.send`/`.close`.
* Closing a client closes all the transports.
* We'd support `.mount(prefix, transport)`.

Glossing over a bunch of stuff here since I just want to get the idea out there.

Ideally we'd also move *away* from having proxy transport instance be supported via `proxies = ...`, and instead only support the *simple* proxy config style for `proxies=...`, with *any* more specific transport mounting handled via the `.mount` interface so that there's only one way to do things.

tomchristie:
Although this is new proposed API, I think it falls under the 1.0 cutoff, because it *impacts* stuff we want to resolve for 1.0.

tomchristie:
Since #1098, #1099, #1103, this has now become more obvious.

I don't think we should support a `.mount()` method - that's awkward because really you only want to mount or unmount transports on instantiating the client, so that you can ensure that the transport lifespans match the client lifespan.

Instead the API ought to look like this...

```python
client = httpx.Client(mounts={"http://www.example.com": WSGITransport(...)})
client = httpx.Client(mounts={"http://": MyCustomTransport()})
client = httpx.Client(mounts={"http://": MyCustomTransport(), "https://": MyCustomTransport()})
client = httpx.Client(mounts={"all://": MyCustomTransport()})
```

Additionally we could extend our URL pattern matching to also support the following...

```python
[...]

PR #1059, opened 2020-07-15
Mount API

tomchristie:
Working towards having the proxy and transport configuration as part of a more generic URL mounting system.

Here's what I've done so far...

* Make `.transport` and `.proxies` private. We don't really want those as public API.
* A bit of cleanup in the `getproxies` code, for simplicity.
* A bit of refactoring in `transport_for_url` for simplicity.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -470,7 +470,7 @@ class Client(BaseClient):
 
         proxy_map = self.get_proxy_map(proxies, trust_env)
 
-        self.transport = self.init_transport(
+        self._transport = self.init_transport(
             verify=verify,
             cert=cert,
             http2=http2,
@@ -479,7 +479,7 @@ class Client(BaseClient):
             app=app,
             trust_env=trust_env,
         )
-        self.proxies: typing.Dict[str, httpcore.SyncHTTPTransport] = {
+        self._proxies: typing.Dict[str, httpcore.SyncHTTPTransport] = {
             key: self.init_proxy_transport(
                 proxy,
                 verify=verify,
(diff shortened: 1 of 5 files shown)


tomchristie:
Right, worth talking about this a bit, since it's a really fiddly one.

This PR drops `.transport` and `.proxies` in favour of a generic `._mounts`, which maps URL keys onto transport instances.

A URL like `https://www.google.com:1234` will attempt to match against the following URL keys, with this precedence...

* `https://www.google.com:1234`
* `all://www.google.com:1234`
* `https://www.google.com`
* `all://www.google.com`
* `https://*.google.com`
* `all://*.google.com`
* `https://*.com`
* `all://*.com`
* `https`
* `all`

The motivation of this change is that it...

Closes #1043, by supporting eg. `proxies={'all': '1.2.3.4', 'all://localhost': None}`
Closes #1039

florimondmanca:
To be honest, dunno about this.

I think I need to read through #977 again to get the gist of the "mounting" idea.

- What would be a general design rule for mount keys?
- Currently we only have proxies and the connection pool as mounted transports, correct?
- Is there any similar idea implemented in an existing HTTP client? (Perhaps a non-Python even…)

tomchristie:
> But to be honest, I'm in a "dunno about this" state for now.

I can understand that, yup.

Let's take this another way around then. A starting point for thinking about this is... https://requests.readthedocs.io/en/master/user/advanced/#transport-adapters

We'd like to support an equivalent functionality, although it's possible(?) that we want something slightly broader than straight-up prefix matching, since it'd be nice and consistent if we were able to fold the `NO_PROXY` behaviour into an equivalent mount-this-case-as-an-exclusion.

tomchristie:
I'm going to see if I can take some smaller iterations onto *parts* that this is attempting to resolve, eg. looking at closing #1062 without introducing so much thrashing.

tomchristie:
Superseeded by #1098, #1099, #1103

Issue #1061, opened 2020-07-16
AttributeError exception if header value is None


tector:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Header normalization fails if header value is None. 

### To reproduce

Send a request with a header key but give it no value:

```python
async with https.AsyncClient as client:
    r = client.get(URL, headers={'foo': None})
```

### Expected behavior

Ignore the header with None value (like 'requests' lib handle it)

### Actual behavior

Gives AttributeError: 'NoneType' object has no attribute 'encode'

### Debugging material

```bash
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_client.py", line 1305, in get
    return await self.request(
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_client.py", line 1137, in request
    request = self.build_request(
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_client.py", line 195, in build_request
    headers = self.merge_headers(headers)
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_client.py", line 242, in merge_headers
    merged_headers.update(headers)
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_models.py", line 480, in update
    headers = Headers(headers)
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_models.py", line 393, in __init__
    self._list = [
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_models.py", line 394, in <listcomp>
    (normalize_header_key(k, encoding), normalize_header_value(v, encoding))
  File "/home/ubuntu/.pyenv/versions/<REDACTED>/lib/python3.8/site-packages/httpx/_utils.py", line 53, in normalize_header_value
    return value.encode(encoding or "ascii")
AttributeError: 'NoneType' object has no attribute 'encode'
```

### Environment

- OS: Ubuntu 18.04.4 LTS
- Python version: 3.8.2
- HTTPX version: 0.13.3
- Async environment: asyncio
- HTTP proxy: yes
- Custom certificates: no

### Additional context

[...]

PR #1063, opened 2020-07-16
Allow and ignore `None` header values

varunu28:
Bug fix for #1061 

 - Allowing `Optional` values for `HeaderTypes`
 - Drop any `None` values from `Headers`
 - Adding tests on both header and client layer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ venv*/
 .python-version
 build/
 dist/
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+.idea/
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -41,6 +41,8 @@ HeaderTypes = Union[
     "Headers",
     Dict[str, str],
     Dict[bytes, bytes],
+    Dict[str, Optional[str]],
+    Dict[bytes, Optional[bytes]],
     Sequence[Tuple[str, str]],
     Sequence[Tuple[bytes, bytes]],
 ]
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -43,8 +43,8 @@ HeaderTypes = Union[
     Dict[bytes, bytes],
     Dict[str, Optional[str]],
     Dict[bytes, Optional[bytes]],
-    Sequence[Tuple[str, str]],
-    Sequence[Tuple[bytes, bytes]],
+    Sequence[Tuple[str, Optional[str]]],
+    Sequence[Tuple[bytes, Optional[bytes]]],
 ]
 
 CookieTypes = Union["Cookies", CookieJar, Dict[str, str]]
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -39,10 +39,8 @@ QueryParamTypes = Union[
 
 HeaderTypes = Union[
     "Headers",
-    Dict[str, str],
-    Dict[bytes, bytes],
-    Dict[str, Optional[str]],
-    Dict[bytes, Optional[bytes]],
+    Mapping[str, Optional[str]],
+    Mapping[bytes, Optional[bytes]],
     Sequence[Tuple[str, Optional[str]]],
     Sequence[Tuple[bytes, Optional[bytes]]],
 ]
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+.idea/
```
florimondmanca, line 13: Tip: you can put this in a [global `~/.gitignore`](http://egorsmirnov.me/2015/05/04/global-gitignore-file.html). :-) 

```suggestion
```

httpx/_types.py
```diff
@@ -41,6 +41,8 @@
     "Headers",
     Dict[str, str],
     Dict[bytes, bytes],
+    Dict[str, Optional[str]],
+    Dict[bytes, Optional[bytes]],
```
florimondmanca, line 45: Would this be enough? A `Dict[str, str]` is _also_ a `Dict[str, Optional[str]]` (as in "it's a sub-type").

```suggestion
    Dict[str, Optional[str]],
    Dict[bytes, Optional[bytes]],
```

tests/client/test_async_client.py
```diff
@@ -155,3 +156,10 @@ async def test_100_continue(server):
 
     assert response.status_code == 200
     assert response.content == data
+
+
+@pytest.mark.usefixtures("async_environment")
+async def test_request_with_none_header(server):
+    value: Optional[str] = None
+    async with httpx.AsyncClient() as client:
+        await client.get(server.url, headers={"foo": value})
```
florimondmanca, line 165: (Nit) It looks like this test is redundant with the one already added in `test_headers.py`. I think we can drop it. (The `typing` important can then be removed too.)

```suggestion
```

tests/client/test_headers.py
```diff
@@ -181,3 +181,22 @@ async def test_host_with_non_default_port_in_url():
             "authorization": "Basic dXNlcm5hbWU6cGFzc3dvcmQ=",
         }
     }
+
+
+@pytest.mark.asyncio
+async def test_none_headers():
```
florimondmanca, line 187: Should we add a small docstring to explain what this test is about? I think embedding the "Requests compatibility" motivation is important here.

```suggestion
async def test_none_headers():
    """
    For compatibility with Requests, `None` header values are allowed but ignored.
    """
```

tests/client/test_headers.py
```diff
@@ -181,3 +181,22 @@ async def test_host_with_non_default_port_in_url():
...
+    url = "http://example.org/echo_headers"
+    value: typing.Optional[str] = None
+    headers = {"foo": value}
```
florimondmanca, line 190: (Nit) Does this simplify to…? (Also a suggestion to use a perhaps more explanatory header name, to suggest "this header will be ignored".)

```suggestion
    headers = {"x-ignore": None}
```

httpx/_types.py
```diff
@@ -41,6 +41,8 @@
...
     Sequence[Tuple[str, str]],
     Sequence[Tuple[bytes, bytes]],
```
florimondmanca, line 47: I _think_ (?) we also need to update these:

```suggestion
    Sequence[Tuple[str, Optional[str]]],
    Sequence[Tuple[bytes, Optional[bytes]]],
```

httpx/_types.py
varunu28: mypy complaints here due to invariance. More info: https://mypy.readthedocs.io/en/latest/common_issues.html#variance

We have to use explicit types in this case or create a custom type and then use it in replacement of `Dict[str, str]`

httpx/_types.py
varunu28: Adding this. Also added a check under model layer to ignore None values for Sequence.

httpx/_types.py
florimondmanca: …Or should we switch to `Mapping` instead of `Dict`? I believe it's covariant in value types.

varunu28: Switching to `Mapping` helps. I believe it's a small change.

florimondmanca:
Closing given https://github.com/encode/httpx/issues/1061#issuecomment-660077975, thanks so much for putting some time into this @varunu28 :-)

PR #1070, opened 2020-07-20
Drop private Origin model

tomchristie:
Since we don't expose the `Origin` model, and only use it in one place, this PR now switches that single usage to a plain util function.


tests/test_utils.py
```diff
@@ -294,3 +295,15 @@ def test_should_not_be_proxied(url, no_proxy, expected):
...
     assert should_not_be_proxied(parsed_url) == expected
+
+
+def test_same_origin():
+    origin1 = httpx.URL("https://example.com")
+    origin2 = httpx.URL("HTTPS://EXAMPLE.COM:443")
+    assert same_origin(origin1, origin2)
+
+
+def test_now_same_origin():
```
florimondmanca, line 306:
```suggestion
def test_not_same_origin():
```

Issue #954, opened 2020-05-15
Supporting alternate schemes.


tomchristie:
Custom transport classes might want to support alternate schemes in addition to "http" and "https".
Issue #931 includes a couple of examples here...

* Use of a `test:` scheme in unit tests for session handling.
* Use of `http+thing:` custom schemes that enable additional behaviours at the transport level.

The question is how would we best want to support this. One thing that might be sensible here is to change the Transport API slightly, so that the `url` signature is `Tuple[bytes, bytes, Optional[int], bytes]` instead of the current `Tuple[bytes, bytes, int, bytes]`.

We would then pass *only* any explicitly included port in the URL, and leave any protocol->port mapping to the responsibility of the transport class. The `AsyncConnectionPool` and `SyncConnectionPool` classes would enforce `scheme=[b'http'|b'https']`, but alternate transport classes could provide other behaviour.

A first step here would to to review *where* in the `httpx` codebase we're currently accessing the `URL` port.

yeraydiazdiaz:
I think we should scope this issue a bit.

1. Are we strictly refering to choosing a port based the scheme?

If that's the case then we can change how port is derived from the scheme in the [`URL` model](https://github.com/encode/httpx/blob/0.13.0.dev2/httpx/_models.py).

Or, 

2. Are we including the selection of a transport as well? Which I think it's the objective behind #931 makes things more complicated. To add some context, currently:

- The client creates a default transport if the user does not provide one,
- Depending on proxy configurations additional transports may be created,
- Upon sending a request [`dispatch_for_url`](https://github.com/encode/httpx/blob/0.13.0.dev2/httpx/_client.py#L528-L551) decides which transport to use.

Supporting arbitrary schemes introduces quite a few questions:

- Which schemes and therefore transports do we create at client init time?
- Do we create more transports at request time based on the scheme of the request?
- Do we allow the user to provide their own scheme-to-transport map?
- Transports are currently in the realm of httpcore, do we allow users to extend those or do we provide a higher level abstraction?

tomchristie:
To rephrase this slightly, it's not exactly about "supporting alternate schemes" so much as "Enforcing http/https and determining default port mappings should instead be the responsibility *of the transport layer*."

* The transport API would be better off with the URL signature being `Tuple[bytes, bytes, Optional[int], bytes]`.
* `URL.port` should return an `Optional[int]` instead of `int`. Don't return a default-port-for-the-scheme if there's no port in the URL.
[...]

PR #1073, opened 2020-07-20
Drop `URL(allow_relative=bool)`

tomchristie:
This PR pushes all `url.scheme` and `url.host` checking down into the `._transport_for_url` method.

This allows us to drop the `allow_relative` flag on the `URL` class, and instead just always allow relative URLs, up until the point that we're looking up a transport instance, at which point we enforce HTTP/HTTPS.

There are two benefits here...

* `allow_relative=...` wasn't really ever intended as a bit of public API, we just happened to be relying on it ourselves. It's cleaner for our API for allow eg `httpx.URL("")` as a valid relative URL, without having to explicitly enable it. But we *do* still want to raise an error if you attempt to issue a request with it.
* This change is a necessary pre-requisite for #954, since we don't necessarily want to hard-code enforce `http`/`https` schemes only, once we support mounting other scheme types too. Dropping the enforcement logic into the transport lookup means we'll be able to instead error if there's no appropriately mounted transport.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -24,7 +24,6 @@ from ._decoders import (
 from ._exceptions import (
     CookieConflict,
     HTTPStatusError,
-    InvalidURL,
     NotRedirectResponse,
     RequestNotRead,
     ResponseClosed,
@@ -55,12 +54,7 @@ from ._utils import (
 
 
 class URL:
-    def __init__(
-        self,
-        url: URLTypes,
-        allow_relative: bool = False,
-        params: QueryParamTypes = None,
-    ) -> None:
+    def __init__(self, url: URLTypes, params: QueryParamTypes = None,) -> None:
         if isinstance(url, str):
             self._uri_reference = rfc3986.api.iri_reference(url).encode()
         else:
@@ -79,13 +73,6 @@ class URL:
             else:
                 query_string = str(QueryParams(params))
             self._uri_reference = self._uri_reference.copy_with(query=query_string)
-
-        # Enforce absolute URLs by default.
-        if not allow_relative:
-            if not self.scheme:
-                raise InvalidURL("No scheme included in URL.")
-            if not self.host:
-                raise InvalidURL("No host included in URL.")
 
         # Allow setting full_path to custom attributes requests
         # like OPTIONS, CONNECT, and forwarding proxy requests.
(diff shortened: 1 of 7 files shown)


httpx/_models.py
```diff
@@ -55,12 +54,7 @@
 
 
 class URL:
-    def __init__(
-        self,
-        url: URLTypes,
-        allow_relative: bool = False,
-        params: QueryParamTypes = None,
-    ) -> None:
+    def __init__(self, url: URLTypes, params: QueryParamTypes = None,) -> None:
```
florimondmanca, line 57: (Nit)

```suggestion
    def __init__(self, url: URLTypes, params: QueryParamTypes = None) -> None:
```

tests/client/test_client.py
```diff
@@ -22,6 +22,16 @@ def test_get(server):
...
 
 
+def test_get_invalid_url(server):
+    with httpx.Client() as client:
+        with pytest.raises(httpx.InvalidURL):
+            client.get("invalid://example.org")
+        with pytest.raises(httpx.InvalidURL):
+            client.get("://example.org")
+        with pytest.raises(httpx.InvalidURL):
+            client.get("http://")
```
florimondmanca, line 32: (Nit) Considering using a parametrized test style? I enjoy the documentation purpose of `id=...`, and that each case runs as its own test...

Then should we back-port this to `test_async_client.py` as well?

```suggestion
@pytest.mark.parametrize(
    "url",
    [
        pytest.param("invalid://example.org", id="scheme-not-http(s)"),
        pytest.param("://example.org", id="no-scheme"),
        pytest.param("http://", id="no-host"),
    ],
)
def test_get_invalid_url(server, url):
    with httpx.Client() as client:
        with pytest.raises(httpx.InvalidURL):
             client.get(url)
```

tests/client/test_async_client.py
```diff
@@ -23,6 +23,10 @@ async def test_get_invalid_url(server):
     async with httpx.AsyncClient() as client:
         with pytest.raises(httpx.InvalidURL):
             await client.get("invalid://example.org")
+        with pytest.raises(httpx.InvalidURL):
+            await client.get("://example.org")
+        with pytest.raises(httpx.InvalidURL):
+            await client.get("http://")
```
florimondmanca, line 29:
```python
@pytest.mark.parametrize(
    "url",
    [
        pytest.param("invalid://example.org", id="scheme-not-http(s)"),
        pytest.param("://example.org", id="no-scheme"),
        pytest.param("http://", id="no-host"),
    ],
)
async def test_get_invalid_url(server, url):
        with pytest.raises(httpx.InvalidURL):
             await client.get(url)
```

PR #1076, opened 2020-07-21
Refactor map_exceptions

tomchristie:
Refactors `map_exceptions` to also allow optional `**kwargs` which are passed to the new exception.

This is useful for us because the `httpcore` exception classes don't have an associated `request` because the model doesn't exist at that layer, but the equivelent `httpx` exception classes all do have an associated `request`.

As a result we can now do...

```python
with map_exceptions(HTTPCORE_EXC_MAP, request=request):
    http_version, status_code, reason_phrase, headers, stream = transport.request(...)
```

Without having to do any extra dancing around re-catching the raised exception and setting the request.

### DIFF
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -216,7 +216,8 @@ class CookieConflict(HTTPError):
 
 @contextlib.contextmanager
 def map_exceptions(
-    mapping: typing.Mapping[typing.Type[Exception], typing.Type[Exception]]
+    mapping: typing.Mapping[typing.Type[Exception], typing.Type[Exception]],
+    **kwargs: typing.Any,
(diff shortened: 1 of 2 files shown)


httpx/_exceptions.py
```diff
@@ -235,7 +236,8 @@ def map_exceptions(
         if mapped_exc is None:
             raise
 
-        raise mapped_exc(exc) from None
+        message = str(exc)
+        raise mapped_exc(message, **kwargs) from None  # type: ignore
```
tomchristie, line 240: We have to `type: ignore` here, because `**kwargs` isn't *generally* appropriate to any exception class, so it's up to us to ensure that calling code uses it appropriately with whichever exceptions are being mapped too.

PR #1078, opened 2020-07-21
Prefer defaulting to utf-8, rather than using chardet autodetect

### LINKED ISSUE #1018: Replace chardet with plain UTF-8 detection
I'd suggest replacing `chardet` with dead-simple decoding logic:

```python
encoding = "UTF-8"
try:
    text = databytes.decode(encoding)
except UnicodeDecodeError:
    encoding = "ISO-8859-1"
    text = databytes.decode(encoding)
```

It is *fairly* safe to assume that any valid UTF-8 sequence is in fact UTF-8, and should be decoded as such.

Otherwise ISO-8859-1 (as specified in HTTP RFC) should be used instead of trying to detect languages, which is always quite unreliable. Further, ISO-8859-1 maps **all** possible byte values directly to Unicode code points 0-255, so the final decoding step in the above code will never raise UnicodeDecodeError.

Since the chardet module has been around a long time and is known as the standard go-to solution for charset detection, many modules depend on it, and this is a frequent source of dependency conflicts such as the one reported here: https://github.com/huge-success/sanic/issues/1858

A more serious problem is that chardet is notoriously unreliable. Especially its tendency to misdetect anything as Turkish is well known:

```python
>>> chardet.decode("Testiä".encode("UTF-8"))   # "Testing" in Finnish
{'encoding': 'ISO-8859-9',
 'confidence': 0.5521177026603239,
 'language': 'Turkish'}

>>> "Testiä".encode("UTF-8").decode("ISO-8859-9")  # Yields gibberish, not Turkish
'TestiÃ¤'
```

tomchristie:
Closes #1018

Not marking this as a "draft" since it *is* ready for review, but we shouldn't merge it without prior discussion.
Drops `chardet` autodetection in favour of `utf-8` as the default.

This changes our behaviour from:

* Use any explicitly set `.encoding`.
* Then use any `charset` included on the Content-Type.
* Then use use iso-8859-1 for any `text/*` Content-Type without a charset.
* Then use chardet autodetection.

To...

* Use any explicitly set `.encoding`.
* Then use any `charset` included on the Content-Type.
* Then use use iso-8859-1 for any `text/*` Content-Type without a charset.
* Then use utf-8.

We would probably want to document how to use chardet for autodetection, at least for the simpler non-streaming case.
Users can replicate the previous behaviour with...

```python
response = client.get(...)
if response.charset_encoding is None:
    response.encoding = chardet.detect(self.content)["encoding"]
response.text
```

We *might* also consider a TextDecoder API at some point in order to allow users to customize text decoding, including the streaming case, which would allow us to support eg. chardet or HTML meta tag detection options.

What are our thoughts on switching to utf-8 as the fallback case, rather than using chardet-autodetection?

### DIFF
--- a/tests/models/test_responses.py
+++ b/tests/models/test_responses.py
@@ -48,25 +48,15 @@ def test_response_content_type_encoding():
     assert response.encoding == "latin-1"
 
 
-def test_response_autodetect_encoding():
+def test_response_fallback_to_utf8():
     """
-    Autodetect encoding if there is no charset info in a Content-Type header.
-    """
-    content = "おはようございます。".encode("EUC-JP")
-    response = httpx.Response(200, content=content, request=REQUEST)
-    assert response.text == "おはようございます。"
-    assert response.encoding == "EUC-JP"
-
-
-def test_response_fallback_to_autodetect():
-    """
-    Fallback to autodetection if we get an invalid charset in the Content-Type header.
+    Fallback to utf-8 if we get an invalid charset in the Content-Type header.
     """
     headers = {"Content-Type": "text-plain; charset=invalid-codec-name"}
-    content = "おはようございます。".encode("EUC-JP")
+    content = "おはようございます。".encode("utf-8")
     response = httpx.Response(200, content=content, headers=headers, request=REQUEST)
     assert response.text == "おはようございます。"
-    assert response.encoding == "EUC-JP"
+    assert response.encoding == "utf-8"
 
 
 def test_response_default_text_encoding():
@@ -84,10 +74,11 @@ def test_response_default_text_encoding():
 
 def test_response_default_encoding():
     """
-    Default to utf-8 if all else fails.
+    Default to utf-8 for decoding.
     """
-    response = httpx.Response(200, content=b"", request=REQUEST)
-    assert response.text == ""
+    content = "おはようございます。".encode("utf-8")
+    response = httpx.Response(200, content=content, request=REQUEST)
+    assert response.text == "おはようございます。"
     assert response.encoding == "utf-8"
 
 
@@ -98,7 +89,7 @@ def test_response_non_text_encoding():
     headers = {"Content-Type": "image/png"}
     response = httpx.Response(200, content=b"xyz", headers=headers, request=REQUEST)
     assert response.text == "xyz"
-    assert response.encoding == "ascii"
+    assert response.encoding == "utf-8"
 
 
 def test_response_set_explicit_encoding():
@@ -129,7 +120,7 @@ def test_read():
 
     assert response.status_code == 200
     assert response.text == "Hello, world!"
-    assert response.encoding == "ascii"
+    assert response.encoding == "utf-8"
     assert response.is_closed
 
     content = response.read()
@@ -145,7 +136,7 @@ async def test_aread():
 
     assert response.status_code == 200
     assert response.text == "Hello, world!"
-    assert response.encoding == "ascii"
+    assert response.encoding == "utf-8"
     assert response.is_closed
 
     content = await response.aread()
--- a/README.md
+++ b/README.md
@@ -112,7 +112,6 @@ The HTTPX project relies on these excellent libraries:
   * `h11` - HTTP/1.1 support.
   * `h2` - HTTP/2 support.
 * `certifi` - SSL certificates.
-* `chardet` - Fallback auto-detection for response encoding.
 * `hstspreload` - determines whether IDNA-encoded host should be only accessed via HTTPS.
 * `idna` - Internationalized domain name support.
 * `rfc3986` - URL parsing & normalization.
--- a/docs/index.md
+++ b/docs/index.md
@@ -110,7 +110,6 @@ The HTTPX project relies on these excellent libraries:
   * `h11` - HTTP/1.1 support.
   * `h2` - HTTP/2 support.
 * `certifi` - SSL certificates.
-* `chardet` - Fallback auto-detection for response encoding.
 * `hstspreload` - determines whether IDNA-encoded host should be only accessed via HTTPS.
 * `idna` - Internationalized domain name support.
 * `rfc3986` - URL parsing & normalization.
--- a/httpx/_decoders.py
+++ b/httpx/_decoders.py
@@ -7,8 +7,6 @@ import codecs
 import typing
 import zlib
 
-import chardet
-
 from ._exceptions import DecodingError
 
 try:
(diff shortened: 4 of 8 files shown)


StephenBrown2:
Would you want to make the encoder pluggable (as suggested in https://github.com/encode/httpx/issues/1018#issuecomment-659233983), so that one could add it in the Client settings or such, something like `encoding=`, `decoder=`, or `charset_detector=`:

```py
import charset_normalizer

normalize_encoding = lambda content: charset_normalizer.detect(content)["encoding"]

with Client(encoding=normalize_encoding) as client:
	response = client.get(...)
print(response.text)
```
This could short-curcuit the logic above, or just be added as a fallback if the rest fail, depending on how much trust/speed trade-off there should be.

Admittedly, this could be done as you stated above, like:

```py
import charset_normalizer

response = client.get(...)
if response.charset_encoding is None:
    response.encoding = charset_normalizer.detect(self.content)["encoding"]
print(response.text)
```

but having it on the Client would be a nice-to-have to avoid repeating the same if block everywhere.

tests/models/test_responses.py
```diff
@@ -48,25 +48,15 @@ def test_response_content_type_encoding():
...
-    Fallback to autodetection if we get an invalid charset in the Content-Type header.
+    Fallback to utf-8 if we get an invalid charset in the Content-Type header.
     """
     headers = {"Content-Type": "text-plain; charset=invalid-codec-name"}
-    content = "おはようございます。".encode("EUC-JP")
+    content = "おはようございます。".encode("utf-8")
     response = httpx.Response(200, content=content, headers=headers, request=REQUEST)
     assert response.text == "おはようございます。"
-    assert response.encoding == "EUC-JP"
+    assert response.encoding == "utf-8"
```
florimondmanca, line 59: So, what if we _do_ get a `"おはようございます。".encode("EUC-JP")`, and the server fails to include `; encoding=EUC-JP`?

I assume we'll try to `.decode("utf-8")` — and that it will fail?

```python
>>> text = "おはようございます。".encode("EUC-JP")
>>> text.decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa4 in position 0: invalid start byte
```

In that case I assume we want to allow the user to provide some kind of manual encoding (eg if they _know_ it's some exotic encoding).

Most likely it would just be a recommendation to use something like this...

```python
text = response.content.decode("EUC-JP")
```

This is mostly fine I suppose.

But then, we probably also want to provide more helpful error messaging than the `UnicodeDecodeError` above, correct? For example, a `DecodingError` with an error message saying "consider decoding `.content` manually" (we may have this already, but not sure?).

tomchristie:
So after having a dig into this I think we clearly want to update our strategy here.

For one thing the `text/*` -> `iso-8859-1` rule that we currently have in place is there because of an obsolete part of the HTTP spec.

For example, HTML5 [now recommends](https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding) cp1252 ("Windows 1252") as the fallback encoding. It's a good choice because like iso-8859-1 it also encompasses the full byte range, but it's printable characters are a *superset* of the printable characters in iso-8859-1.

I asked for a bit of [dev feedback on twitter](https://twitter.com/_tomchristie/status/1286248713673945089) mostly to try to get an idea about which non utf-8 charsets we might expect. There's also [some useful statistical info, here](https://w3techs.com/technologies/overview/character_encoding).

Screenshot 2020-07-23 at 12 27 37

Note that windows 1251 is more prevalent than windows 1252, but it's not a good fallback because the later is aligned with ISO-8859-1.

I can think of 3 possible strategies that we might want to adopt here...

1. charset from "Content-Type" header.
2. UTF-8.
3. Windows 1252.

Which would be nice and simple. Users can anyways use `response.encoding = ...` if still required.

1. charset from "Content-Type" header.
2. UTF-8.
3. Autodetect using chardet, if installed.
4. Windows 1252.

Bit fancy. Potentially awkward in streaming text cases, since chardet will buffer up before returning a charset.

1. charset from "Content-Type" header.
2. UTF-8.
3. User locale if not UTF-8 or Windows 1252.
4. Windows 1252.

Could bit more resilient than the first case, since machine locale *may* sometimes be a decent proxy onto possible text encodings of the URL being accessed by the client. But might be a bit of a non-obvious behaviour difference between devs testing stuff locally vs. deployed to server, which may well not have the same locale.

Any thoughts?

tomchristie:
Here's a bit of provisional documentation for how things might look if we adopted the fist of those three options, but also allowed the text encodings to be configurable...

## Text Encoding

When accessing `response.text`, the byte content in the response needs to be decoded into a unicode string, using some particular encoding. The precedence rules that are used for `httpx` here are...

* If the response includes a Content-Type header with a "charset=" attribute, then use that.
* Try to decode as UTF-8.
* Decode as Windows 1252.

UTF-8 is now very widely adopted, so we always attempt that as our first choice.

The HTML5 spec recommends decoding as Windows 1252 when the encoding is not known, since it is a superset of ISO-8859-1, and includes the full byte range, so will always successfully return a unicode text string, for any input.

In some cases you might need to override the default behaviour, to deal with resources that are not including character set information, and are not utf-8 encoded. Here are some scenarios for how you might override this behaviour:

Explicitly override the encoding...

```python
response = client.get(...)
response.encoding = ...
print(response.text)
```

Explicitly override the encoding, if no charset info is included...

```python
response = client.get(...)
if response.charset_encoding is None:
    response.encoding = ...
print(response.text)
```

Auto-determine the encoding using `chardet`, if no charset info is included...

```python
response = client.get(...)
if response.charset_encoding is None:
    response.encoding = chardet.detect(self.content)["encoding"]
print(response.text)
```

Use a different fallback encoding...

```python
# Configure client to expect Russian language resources.
# Try UTF-8 first, then fallback to Window 1251.
client = Client(text_encodings=['utf-8', 'cp1251'])
response = client.get(...)
print(response.text)
```

Use a fallback encoding based on the user-locale...

```python
# Configure client to try the user locale encoding,
# if it is not already one of the two defaults options.
locale_lang, locale_encoding = locale.getdefaultlocale()
text_encodings = ['utf-8', 'cp1252']
if locale_encoding.lower() not in ['utf-8', 'cp1252']:
    text_encodings = ['utf-8', locale_encoding, 'cp1252']

client = Client(text_encodings=text_encodings)
response = client.get(...)
print(response.text)
```

tests/models/test_responses.py
tomchristie: Rather than...

```python
text = response.content.decode("EUC-JP")
print(text)
```

You'd use..

```python
response.encoding = "EUC-JP"
print(response.text)
```

florimondmanca:
I think first option + detailed docs for more exotic cases (exactly like the above) should get us covered, right?

In particular I like how the default behavior would be really simple while accommodating most situations, and then how the "manual mode" docs seem to allow dealing with _any_ special cases.

tomchristie:
Also useful context... https://github.com/psf/requests/issues/2086

StephenBrown2:
I'm in favor of Option 2, but in the name of speed + simplicity, Option 1 + detailed docs is the best.

Unfortunately the docs you wrote aren't quite copy-and-pastable, as there's a `self.content` where I think it should be `response.content`:
```diff
-    response.encoding = chardet.detect(self.content)["encoding"]
+    response.encoding = chardet.detect(response.content)["encoding"]
```

You may also want to list a couple of the charset-detection modules to give users more choice (those especially interested in speed may not prefer `chardet` if they see something else exists).

tomchristie:
Switching this to a draft since I don't think it's *quite* the right implementation.

PR #1079, opened 2020-07-23
Parameterise invalid URL tests

tomchristie:
Follow up to #1073.
I marked the conversation as resolved without having realised that I hadn't yet committed this follow up.
Thanks so much for the thorough reviews! ✨

### DIFF
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -18,15 +18,19 @@ async def test_get(server):
     assert response.elapsed > timedelta(seconds=0)
 
 
+@pytest.mark.parametrize(
+    "url",
+    [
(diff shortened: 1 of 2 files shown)


tests/client/test_async_client.py
```diff
@@ -18,15 +18,19 @@ async def test_get(server):
     assert response.elapsed > timedelta(seconds=0)
 
 
+@pytest.mark.parametrize(
+    "url",
+    [
+        pytest.param("invalid://example.org", id="scheme-not-http(s)"),
```
jcugat, line 24: Just learned about `id=` 🤯

tomchristie:
Thanks folks!

PR #1080, opened 2020-07-23
URL.port becomes Optional[int]

tomchristie:
This PR changes `URL.port` from `int` to `Optional[int]`.

Previously any non http/https URL without an explicit port component would raise an error if `url.port` was accessed.
Now `url.port` is optional, and only returns any explicitly included port in the URL, without also trying to use a scheme-default if one is not present.

This is important because...

1. URL is more generically useful. Eg. `URL('ftp://www.example.com').port == None` which makes more sense than `KeyError("ftp")`.
2. This is a precursor to #954, since the only point we're now enforcing http/https schemes on URLs is strictly inside `_transport_for_url`, rather than at any point in the codebase that accesses `url.port`.

(Yes, it's a little bit fiddly)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -543,10 +543,13 @@ class Client(BaseClient):
         enforce_http_url(url)
 
         if self._proxies and not should_not_be_proxied(url):
-            is_default_port = (url.scheme == "http" and url.port == 80) or (
-                url.scheme == "https" and url.port == 443
+            default_port = {"http": 80, "https": 443}[url.scheme]
+            is_default_port = url.port is None or url.port == default_port
+            hostname = (
+                f"{url.host}:{default_port}"
+                if url.port is None
+                else f"{url.host}:{url.port}"
             )
-            hostname = f"{url.host}:{url.port}"
             proxy_keys = (
                 f"{url.scheme}://{hostname}",
                 f"{url.scheme}://{url.host}" if is_default_port else None,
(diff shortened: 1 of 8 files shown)


httpx/_client.py
```diff
@@ -543,10 +543,13 @@ def _transport_for_url(self, url: URL) -> httpcore.SyncHTTPTransport:
         enforce_http_url(url)
 
         if self._proxies and not should_not_be_proxied(url):
-            is_default_port = (url.scheme == "http" and url.port == 80) or (
-                url.scheme == "https" and url.port == 443
+            default_port = {"http": 80, "https": 443}[url.scheme]
```
StephenBrown2, line 546: Should this be a `.get()` to avoid `KeyError: 'ftp'`?
```suggestion
            default_port = {"http": 80, "https": 443}.get(url.scheme)
```

httpx/_client.py
```diff
@@ -543,10 +543,13 @@ def _transport_for_url(self, url: URL) -> httpcore.SyncHTTPTransport:
...
+            is_default_port = url.port is None or url.port == default_port
+            hostname = (
+                f"{url.host}:{default_port}"
+                if url.port is None
+                else f"{url.host}:{url.port}"
             )
```
StephenBrown2, line 552: Could this be simplified:
```suggestion
			port = url.port or default_port
            hostname = f"{url.host}:{port}"
```

httpx/_client.py
```diff
@@ -1076,10 +1079,13 @@ def _transport_for_url(self, url: URL) -> httpcore.AsyncHTTPTransport:
...
```
StephenBrown2, line 1088: Same comments as above

httpx/_utils.py
```diff
@@ -273,12 +273,20 @@ def enforce_http_url(url: "URL") -> None:
         raise InvalidURL('URL scheme must be "http" or "https".')
 
 
+def port_or_default(url: "URL") -> typing.Optional[int]:
```
StephenBrown2, line 276: Maybe this should be used above?

httpx/_client.py
tomchristie: Actually no, because we've called `enforce_http_url(url)` before doing anything else, so `url.scheme` will always be `http|https` here, and `default_port` will be an `int` rather than an `Optional[int]`.

(Agree that it's a bit futzy, and I think it'd potentially clean up a bit if we rethought the proxy_keys/mount API.)

httpx/_client.py
tomchristie: Great call yup.

PR #1084, opened 2020-07-24
Tighten public API on auth classes

tomchristie:
The only public API on any of our auth classes should be the `.auth_flow(...)` method.

This refactoring strictly moves methods/attributes to private names, as well as refactoring the `DigestAuth` implementation slightly so that `_DigestAuthChallenge` is now just a simple data holder.

From a user POV this is really just a private bit of refactoring, but it's good to be 100% clear that the internals of any of our auth implementations should be considered strictly opaque.

### DIFF
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -14,6 +14,9 @@ from ._utils import to_bytes, to_str, unquote
 class Auth:
     """
     Base class for all authentication schemes.
+
+    To implement a custom authentication scheme, subclass `Auth` and override
+    the `.auth_flow()` method.
     """
 
     requires_request_body = False
@@ -51,10 +54,10 @@ class FunctionAuth(Auth):
     """
 
     def __init__(self, func: typing.Callable[[Request], Request]) -> None:
-        self.func = func
+        self._func = func
 
     def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
-        yield self.func(request)
+        yield self._func(request)
 
 
 class BasicAuth(Auth):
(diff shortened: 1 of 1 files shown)


florimondmanca:
Planning to do a pass over https://pypi.org/project/httpx-auth/ (which I just found out about) to see if this would break anything for them... :)

tomchristie:
Good call.

httpx/_auth.py
```diff
@@ -51,10 +54,10 @@ class FunctionAuth(Auth):
     """
 
     def __init__(self, func: typing.Callable[[Request], Request]) -> None:
-        self.func = func
+        self._func = func
```
florimondmanca, line 57: Curious if we'd want to more clearly separate out "the Auth interface" from our default "function-based Auth" implementation - ie separate Auth from FuncAuth?

tomchristie: I'm not sure *exactly* what you mean here, but... We don't expose `FunctionAuth` publicly anywhere, but we *might* want to have a property setter on Client, as we do with several of the other attributes on it, so that eg...

```
The following are all valid...

client.auth = None
client.auth = ('username', 'password')
client.auth = httpx.DigestAuthentiction(...)
client.auth = simple_function_auth
client.auth = CustomAuthClass()

print(client.auth)  # Always returns an instance of `Auth`. Our instances don't expose any additional public API beyond that.
```

florimondmanca: Ah okay, nevermind — the GitHub UI tricked me. I thought this diff was on the `Auth` class, but it's on `FunctionAuth`. So it's all as I was meaning to think it should be. :-)

PR #1086, opened 2020-07-24
Full coverage, with exception of URLLib3Transport

tomchristie:
* Add coverage onto missing line in `Response.next()`
* Exclude `URLLib3Transport` from coverage.
* Enforce 100% coverage elsewhere.

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -8,4 +8,4 @@ export SOURCE_FILES="httpx tests"
 
 set -x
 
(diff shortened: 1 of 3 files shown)


tomchristie:
We're currently mentioning "99% coverage" in the docs, which I guess we should keep as it is, until `URLLib3Transport` is dropped out into an external package.

jcugat:
I wanted to keep working on this after https://github.com/encode/httpx/pull/1012 but unfortunately other things got in the way. So great to finally see this 🎉

PR #1087, opened 2020-07-24
Keep Proxy._build_auth_header as a private method

tomchristie:
Tiny refactoring. `Proxy._build_auth_header` should be a private method.

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -330,7 +330,7 @@ class Proxy:
         if url.username or url.password:
             headers.setdefault(
                 "Proxy-Authorization",
-                self.build_auth_header(url.username, url.password),
+                self._build_auth_header(url.username, url.password),
             )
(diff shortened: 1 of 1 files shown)


jcugat:
Small note, might be worth moving the code from `BasicAuth.build_auth_header` and `Proxy._build_auth_header` to a utils function (since it's essentially the same).

florimondmanca:
@tomchristie - Should we add @jcugat to the Maintainers (don't have permissions to do this myself) so his reviews are well accounted for by the GitHub UI checks? :-)

tomchristie:
@florimondmanca @jcugat 💯 👍  Good call. Done.

florimondmanca:
@tomchristie Not sure which comment you were answering to but I don't see Josep in the members list yet: https://github.com/orgs/encode/teams/maintainers/members

PR #1088, opened 2020-07-24
Single consistent name for status codes

tomchristie:
For consideration.

Currently we're exposing both `httpx.StatusCodes` and `httpx.codes` for working with status codes.

We're [only documenting `httpx.codes`](https://www.python-httpx.org/quickstart/#response-status-codes), and we probably only want a single canonical name here, right?

The documentation could also do with being expanded, at the very least in the API section.

Usages are like...

```python
>>> httpx.codes.OK
<codes.OK: 200>
>>> httpx.codes.OK == 200
True
>>> httpx.codes.get_reason_phrase(200)
'OK'
>>> httpx.codes.is_error(400)
True
>>> httpx.codes.is_redirect(400)
False
>>> httpx.codes.is_client_error(400)
True
>>> httpx.codes.is_server_error(400)
False
>>> httpx.codes(200).phrase
'OK'
>>> httpx.codes(200).value
200
```

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -32,7 +32,7 @@ from ._exceptions import (
     WriteTimeout,
 )
 from ._models import URL, Cookies, Headers, QueryParams, Request, Response
-from ._status_codes import StatusCode, codes
+from ._status_codes import codes
 from ._transports.asgi import ASGITransport
 from ._transports.urllib3 import URLLib3ProxyTransport, URLLib3Transport
 from ._transports.wsgi import WSGITransport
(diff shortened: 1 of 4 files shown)


jcugat:
There are some usages in the wild of `StatusCode`, but I guess we are fine breaking compatibility with this before 1.0 right?

florimondmanca:
I'm not absolute on either option though and pretty much bikeshedding at this point... The important bit is that we should have a single status codes enum, that I agree. :)

tomchristie:
I've updated this with a more gentle deprecation of `httpx.StatusCode`.

PR #1095, opened 2020-07-27
Exception hierarchy

tomchristie:
Refs #949

* Decommissions the too-broad to be useful for anything `HTTPError`, in favour of an actually meaningful `RequestError`.
* Rearranges the hierarchy as follows, dropping the `RedirectError` which encompassed two totally different kinds of error (one a coding error, the other a request error), and introducing `TransportError`.
* Ensures consistently enforced attributes / `__init__` parameters on all exceptions.

---

* RequestError *Exposes `.request`*
  + TransportError
    - TimeoutException
      · ConnectTimeout
      · ReadTimeout
      · WriteTimeout
      · PoolTimeout
    - NetworkError
      · ConnectError
      · ReadError
      · WriteError
      · CloseError
    - ProxyError
    - ProtocolError
  + DecodingError
  + TooManyRedirects
  + RequestBodyUnavailable
  + InvalidURL
* HTTPStatusError *Exposes `.request`, `.response`*
* NotRedirectResponse
* CookieConflict
* StreamError
  + StreamConsumed
  + ResponseNotRead
  + RequestNotRead
  + ResponseClosed


httpx/_decoders.py
```diff
@@ -44,7 +50,8 @@ class DeflateDecoder(Decoder):
     See: https://stackoverflow.com/questions/1838699
     """
 
-    def __init__(self) -> None:
+    def __init__(self, request: "Request") -> None:
+        self.request = request
```
florimondmanca, line 54: ? (Applicable to all other decoders here)

```suggestion
        super().__init__(request)
```

httpx/_exceptions.py
```diff
@@ -1,3 +1,33 @@
+"""
+Our exception heirachy:
```
florimondmanca, line 2:
```suggestion
Our exception hierarchy:
```

httpx/_exceptions.py
```diff
@@ -7,30 +37,26 @@
...
+    Base class for all exceptions that may occur when issuing a `.request()`.
     """
 
-    def __init__(
-        self, *args: typing.Any, request: "Request" = None, response: "Response" = None
-    ) -> None:
-        super().__init__(*args)
-        self._request = request or (response.request if response is not None else None)
-        self.response = response
+    def __init__(self, message: str, *, request: "Request",) -> None:
```
florimondmanca, line 45:
```suggestion
    def __init__(self, message: str, *, request: "Request") -> None:
```

httpx/_exceptions.py
```diff
@@ -100,76 +126,103 @@ class CloseError(NetworkError):
...
+    """
+
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
 
 # Stream exceptions...
 
+# These may occur as the result of a programming error, by accessing
+# the request/respose stream in an invalid manner.
```
florimondmanca, line 213:
```suggestion
# the request/response stream in an invalid manner.
```

httpx/_exceptions.py
```diff
@@ -100,76 +126,103 @@ class CloseError(NetworkError):
...
 
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+
+
+class CookieConflict(Exception):
+    """
+    Attempted to lookup a cookie by name, but multiple cookies existed.
+
+    Can occur when calling `response.cookies.get(...)`
```
florimondmanca, line 203: (I'm reviewing assuming these docstrings will end up in the public Developer Interface sometime in the near future 😉)

```suggestion
    Can occur when calling `response.cookies.get(...)`.
```

httpx/_decoders.py
tomchristie: Yeah a guess a lot of other Python developers would *tend* always just rely on "hey I'm supposed to call `super` right". Personally, I think explicit is preferable to indirect here. Calling into `super` isn't always necessary or inherently the "right" thing to do.

florimondmanca: I see, well then I guess `Decoder` would really be better served by a `Protocol`, rather than a full-fledged base class… But we only have them on Py38+, which is unfortunate.

```python
class Decoder(Protocol):
    request: Request

    def decode(self, data: bytes) -> bytes:
        ...

    def flush(self) -> bytes:
        ...
```

So maybe we could then also use the `request: Request` annotation on the `Decoder` interface, and _also_ add an explicit constructor for `IdentityDecoder`?

cdeler: `MultiDecoder` inherits the `Decoder` interface, but do not have this property

PR #1096, opened 2020-07-28
Use public API only from transports

tomchristie:
Prompted by https://github.com/encode/httpx/pull/998#issuecomment-664703591

Our `WSGITransport` and `ASGITransport` implementations now only use public API, which is beneficial for other folks implementing transports, and basing their work on these.

One thing that we *could* choose to further tweak here would be... `httpcore.SyncByteStream` and `httpcore.AsyncByteStream` provide a simple base implementation that allows users to pass `iterator`/`close` or `aiterator`/`aclose` parameters. That's actually a bit fiddly if you've just got a non-iterating bytestring that you want to pass. We might choose to also support passing plain bytes to the base implementation. So...

```python
    """
    The base interface for request and response bodies.
    Concrete implementations should subclass this class, and implement
    the `\\__aiter__` method, and optionally the `close` method.
    """

    def __init__(
        self, content: Union[bytes, AsyncIterator[bytes]] = b"", aclose_func: Callable = None,
    ) -> None:
        self.content = content
        self.aclose_func = aclose_func

    async def __aiter__(self) -> AsyncIterator[bytes]:
        """
        Yield bytes representing the request or response body.
        """
        if isinstance(self.content, bytes):
            yield self.content
        else:
            async for chunk in self.content:
                yield chunk

    async def aclose(self) -> None:
        """
        Must be called by the client to indicate that the stream has been closed.
        """
        if self.aclose_func is not None:
            await self.aclose_func()
```

That'd allow us to then just use eg. `httpcore.AsyncByteStream(b"")` rather than the little `async_byte_iterator` dance we've got at the moment.

### DIFF
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -1,9 +1,16 @@
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    AsyncIterator,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import httpcore
 import sniffio
-
-from .._content_streams import ByteStream
 
 if TYPE_CHECKING:  # pragma: no cover
     import asyncio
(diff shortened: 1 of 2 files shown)


httpx/_transports/asgi.py
```diff
@@ -24,6 +31,10 @@ def create_event() -> "Event":
         return asyncio.Event()
 
 
+async def async_byte_iterator(bytestring: bytes) -> AsyncIterator[bytes]:
+    yield bytestring
```
florimondmanca, line 35: I'd agree that having this built-in supported by HTTPCore would be beneficial. I hit this myself before too and having to write a dedicated async generator function each time is a bit cumbersome as I think you noted. :-)

Issue #458, opened 2019-10-06
Fix tests for Windows.


cansarigol:
- [x] Add Windows to the build matrix. (Closes by #457)
- [x] Skip tests that fail on Windows.(Closes by #402, Also Closes #191 )
- [ ] Remove lowercased environment variables test cases from the proxies tests. (Closes by #456 )

After complete these:

- [ ] Remove `os: windows` from `allow_failures` into `travis.yml`
- [ ] Fix `xfail`s for `AsyncioBackend` fails.

tomchristie:
I'm not really sure what state the tests are in for windows at the moment.
Is anyone keen on digging into this? Really we oughta get them back on track, and then turn off "allowed failure" for them so we don't slip back into this.

tomchristie:
We've recently added windows tests on the Uvicorn repo thanks to @euri10 - that'd be a good place to start when looking at how to add them here.

PR #1097, opened 2020-07-29
CI for windows

euri10:
WIP plugged from https://github.com/encode/httpx/issues/458


euri10:
it seems CI on windows works fine, I just had to slightly modify a test that was assuming path were posix-like bu that's pretty much it.

however the `test_write_timeout[asyncio]` fails and I have not the slightest idea as to why, any idea ?


```
================================== FAILURES ===================================
_________________________ test_write_timeout[asyncio] _________________________

server = <tests.conftest.TestServer object at 0x000001BFE7A5E358>

    async def test_write_timeout(server):
        timeout = httpx.Timeout(write_timeout=1e-6)
    
        async with httpx.AsyncClient(timeout=timeout) as client:
            with pytest.raises(httpx.WriteTimeout):
                data = b"*" * 1024 * 1024 * 100
                response = await client.put(
                    server.url.copy_with(path="/slow_response"), data=data
                )
                print("TIMEOUT")  # pragma: no cover
>               print(response)  # pragma: no cover
E               Failed: DID NOT RAISE <class 'httpx._exceptions.WriteTimeout'>

tests\test_timeouts.py:26: Failed
```

florimondmanca:
Hi @euri10 

> I just had to slightly modify a test that was assuming path were posix-like bu that's pretty much it

I think we can have a separate PR that makes those changes and get them independently, since they mean that Windows users aren't currently able to develop on HTTPX (since the test suite wouldn't pass for them).

Still no strong idea about why the write timeout test is failing, though…

tomchristie:
Going to close this off as stale

PR #1098, opened 2020-07-29
URL matching

### LINKED ISSUE #1062: Don't call `should_not_be_proxied` on each request
Currently we're calling `should_not_be_proxied` on each request, which is problematic for two reasons...

* It's ignoring the `trust_env`, and relying on an environment variable unilaterally.
* It's preforming an extra chunk of work on each request that we'd rather not do.

tomchristie:
This pull request refactors our URL matching for proxy lookups.

It adds a utility class `URLMatcher` that handles matching against the [proxy keys that we currently support](https://www.python-httpx.org/advanced/#http-proxying).

It's best explained through some examples...

```python
>>> from httpx._utils import URLMatcher  # Private API, intended for internal usage only.
>>> import httpx

>>> pattern = URLMatcher("all")
>>> pattern.matches(httpx.URL("http://example.com"))
True

# Witch scheme matching...
>>> pattern = URLMatcher("https")
>>> pattern.matches(httpx.URL("https://example.com"))
True
>>> pattern.matches(httpx.URL("http://example.com"))
False

# With domain matching...
>>> pattern = URLMatcher("https://example.com")
>>> pattern.matches(httpx.URL("https://example.com"))
True
>>> pattern.matches(httpx.URL("http://example.com"))
False
>>> pattern.matches(httpx.URL("https://other.com"))
False

# Wildcard scheme, with domain matching...
>>> pattern = URLMatcher("all://example.com")
>>> pattern.matches(httpx.URL("https://example.com"))
True
>>> pattern.matches(httpx.URL("http://example.com"))
True
>>> pattern.matches(httpx.URL("https://other.com"))
False

# With port matching...
>>> pattern = URLMatcher("https://example.com:1234")
>>> pattern.matches(httpx.URL("https://example.com:1234"))
True
>>> pattern.matches(httpx.URL("https://example.com"))
False
```

Here's why we care about it...

* The implementation in `transport_for_url` is much more clear now.
* This initial pass doesn't also hook in the no_proxies support, but we'll be able to extend this slightly to deal with matching against `NO_PROXY`, and allow us to resolve #1062 neatly.
* We'll be able to use this as the basis for our Mount API. #977

Note that some of the cases in `test_proxies_parameter` *have* changed, but they're all odd cases, where the desired behaviour is ambiguous, and I don't think we're actually regressing in any of those case.

* If a user explicitly uses `proxies={"http://example.com:80": httpx.Proxy(...)}`, then we now only match if the requested URL includes an explicit port 80. That seems pretty reasonable. It's an odd key to use in any case.
* If a user uses `proxies={"http://example.com": httpx.Proxy(...)}`, then we now match both `example.com` and `example.com:<some port>` to the proxy. That seems like an improvement in expected behaviour.
* A malformed request URL of "http://example.com:443" *will* now match against "http". Pretty ambiguous what you'd expect there, but if anything it seems reasonable. Likely you'll end up with a transport error later down the line, but that's to be expected. (Or not, depending on what the heck you're actually doing there with the kludgy scheme/port mismatch there.)

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -425,5 +425,45 @@ class ElapsedTimer:
         return timedelta(seconds=self.end - self.start)
 
 
+class URLMatcher:
+    def __init__(self, pattern: str) -> None:
+        from ._models import URL
+
+        if pattern and ":" not in pattern:
+            pattern += "://"
+
+        url = URL(pattern)
+        self.pattern = pattern
+        self.scheme = "" if url.scheme == "all" else url.scheme
+        self.host = url.host
+        self.port = url.port
+
+    def matches(self, other: "URL") -> bool:
+        if self.scheme and self.scheme != other.scheme:
+            return False
+        if self.host and self.host != other.host:
+            return False
+        if self.port is not None and self.port != other.port:
+            return False
+        return True
+
+    @property
+    def priority(self) -> tuple:
+        """
+        The priority allows URLMatcher instances to be sortable, so that
+        if we can match from most specific to least specific.
+        """
+        port_priority = -1 if self.port is not None else 0
+        host_priority = -len(self.host)
+        scheme_priority = -len(self.scheme)
+        return (port_priority, host_priority, scheme_priority)
+
+    def __lt__(self, other: "URLMatcher") -> bool:
+        return self.priority < other.priority
+
+    def __eq__(self, other: typing.Any) -> bool:
+        return isinstance(other, URLMatcher) and self.pattern == other.pattern
+
+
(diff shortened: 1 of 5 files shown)


httpx/_utils.py
```diff
@@ -429,5 +429,89 @@ def elapsed(self) -> timedelta:
...
+            return False
+        if self.port is not None and self.port != other.port:
+            return False
+        return True
+
+    @property
+    def priority(self) -> tuple:
+        """
+        The priority allows URLMatcher instances to be sortable, so that
+        if we can match from most specific to least specific.
```
florimondmanca, line 499:
```suggestion
        we can match from most specific to least specific.
```

PR #1103, opened 2020-07-30
Cleaner no proxy support

### LINKED ISSUE #1062: Don't call `should_not_be_proxied` on each request

tomchristie:
Builds on #1099, and #1098.
Closes #1062.

Refactors our NO_PROXY logic, so that instead of calling `should_not_be_proxied` we instead setup the `proxies` infomation in such a way that it has any exclusions within it, eg...

```
# The following environment...

# ALL_PROXY=http://localhost:1234
# NO_PROXY=example.com

# Is equivalent to...
proxies={"all": httpx.Proxy("http://localhost:1234"), "all://*example.com": None}
```

Note that this PR also introduces wildcard domain prefixing for proxy configuration, since it's required in order to handle `NO_PROXY` equivalents correctly, so we now support...

```python
# Match any domain
proxies={"all://*": httpx.Proxy("http://localhost:1234")}

# Match "example.com" exactly.
proxies={"all://example.com": httpx.Proxy("http://localhost:1234")}

# Match "www.example.com" and "example.com". (But not "wwwexample.com")
proxies={"all://*example.com": httpx.Proxy("http://localhost:1234")}

# Match "www.example.com" (But not "example.com")
proxies={"all://*.example.com": httpx.Proxy("http://localhost:1234")}
```


httpx/_client.py
```diff
@@ -90,7 +89,7 @@ def _get_proxy_map(
         if proxies is None:
             if trust_env:
                 return {
-                    key: Proxy(url=url)
+                    key: None if url is None else Proxy(url=url)
```
tomchristie, line 92: Previously `get_environment_proxies()` did not return `None` values.
Now it does.

httpx/_utils.py
```diff
@@ -496,10 +512,13 @@ def matches(self, other: "URL") -> bool:
     def priority(self) -> tuple:
         """
         The priority allows URLMatcher instances to be sortable, so that
-        we can match from most specific to least specific.
+        if we can match from most specific to least specific.
```
florimondmanca, line 515:
```suggestion
        we can match from most specific to least specific.
```

tests/test_utils.py
```diff
@@ -228,75 +227,85 @@ def test_obfuscate_sensitive_headers(headers, output):
...
         ),  # whole-domain matching
         (
             "https://www.example3.com",
-            {"NO_PROXY": "*"},
+            {"ALL_PROXY": "http://localhost:123", "NO_PROXY": "*"},
             True,
         ),  # wildcard * means nothing proxied
     ],
 )
 def test_should_not_be_proxied(url, no_proxy, expected):
```
florimondmanca, line 295: Should we move this test to `test_proxies.py` now that the `should_not_be_proxied` util is gone?

tomchristie: Good call

tomchristie: I've merged these tests into the existing `test_proxies_environ` case now.
Looks much nicer!

PR #1104, opened 2020-07-30
Ignore PermissionError in netrc_info()

tomchristie:
Clean PR bringing in the work in #810.

(Because, build failures that I'd not quite figured out)


tomchristie:
Not sure what's going on with the CI checks on #810, but let's get it in.

Issue #345, opened 2019-09-15
Public middleware API


florimondmanca:
**Problem**
We currently have an internal middleware API that was brought in via #295 and #268.

We should be looking at making that API (or a different form of it?) public so that users can extend the functionality of the client to fit their needs.

For example, a retry functionality (discussed in https://github.com/encode/httpx/issues/108 and drafted in https://github.com/encode/httpx/pull/134 then abandoned) could probably be implemented as a middleware (basic idea here: https://github.com/encode/httpx/pull/268#issuecomment-526909874).

**Questions to be answered**

- Is a public middleware API still a good idea at all? (IMO, it is.)
- We already have a middleware-side interface: `request, get_response -> response`. What's the client-side API going to look like? Probably something like…

```python
client = httpx.Client()
client.add_middleware(RetryMiddleware, max_retries=3)
```

- Should we apply existing internal middleware such as auth and redirect by default?

IMO, yes: treat those as core middleware, and add custom middleware on top. It might depend on the use case, but we can start off with that.

- Should users be able to we control at what level a new middleware is applied, and how? 

@sethmlarson proposed an idea inspired by Pyramid here: https://github.com/encode/httpx/issues/295#issuecomment-526933309

- Should middleware be able  to store session-wide data, and how?

This was originally mentioned in https://github.com/encode/httpx/issues/295#issuecomment-526933309, and would be used to store state worth remembering such as cookies or `Alt-Svc` headers for HTTP/3 (see https://github.com/encode/httpx/issues/275#issuecomment-531067555).

IMO the use case still needs to be refined for us to clearly see what needs to be done.

**What needs to be done**

*This is a proposal.* :)

- Answer the questions above and add the public middleware API.
- Document how to write middleware (Advanced section?). We learnt by experience (e.g. https://github.com/encode/httpx/pull/332#discussion_r322816118) that middleware instances shouldn't hold any mutable state, so let's make sure we mention that here.
- Document how to use middleware (Advanced section?).
- (Later) Figure out middleware ordering.
- (Later) Figure out client-side storage.

All thoughts welcome!

kesavkolla:
Yes this would be a great help.  Specially I'm looking for an usecase for supporting OAuth2.  I want the middleware ability so that I can add logic for getting OAuth token and attach required bearer header to actual request.

florimondmanca:
[...]

PR #1110, opened 2020-08-01
Drop HSTS Preloading

### LINKED ISSUE #896: Document HSTS is still honored if allow_redirects=False
I tried to disable the redirection with latest httpx but it is always redirection the URL:

```python
>>> import httpx
>>> r = httpx.head("http://www.python.org",  allow_redirects=False)
>>> r.url
URL('https://www.python.org')
```

status code is also 200 and not 301 as expected.

### LINKED ISSUE #1102: Consider dropping HSTS preloading
Prompted by https://github.com/encode/httpx/issues/1025#issuecomment-661269344

Issues like #1025 should prompt us to reconsider whether using HSTS preloading from a server-side client (introduced via #151) is a sensible thing to do at all.

Key points of reasoning here:

- While it mimics the way browsers behave, and [hstspreload](https://github.com/sethmlarson/hstspreload) is a super valuable library, from a UX perspective it can lead to ambiguous behaviors (#896) and cases when you _really_ don't want HSTS preload to kick in (like in #1025).
- Crucially, from #124 it seems that HSTS preloading was introduced as a way to tackle "schema-less URLs". As of today HTTPX _requires_ a schema on URLs, so if it was the initial motivation then it is now irrelevant. Besides, we can assume that developers are generally very aware of their choices to use `http://` vs `https://` on the server-side (as opposed to browser users just entering the domain name without a schema).

Some options are:

- Add a toggle so that HSTS preload becomes opt-in or opt-out.
- Drop HSTS functionality entirely.

Given all the context points above, we might want to just go with option 2).

florimondmanca:
Maybe? See the rationale in #1102

Fixes #1102, closes #896

Essentially:

```python
>>> import httpx
>>> r = httpx.get('http://www.paypal.com')  # -> Now kept over HTTP rather than forced to HTTPS.
```

I suppose if we want to move forward with this we'd want it in 0.14, rather than 1.0, since it might be a small breaking change?

Since this was an always-on feature not controlled by any options, I can't think of a smooth deprecation path, but any ideas welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ The HTTPX project relies on these excellent libraries:
   * `h2` - HTTP/2 support.
 * `certifi` - SSL certificates.
 * `chardet` - Fallback auto-detection for response encoding.
-* `hstspreload` - determines whether IDNA-encoded host should be only accessed via HTTPS.
 * `idna` - Internationalized domain name support.
 * `rfc3986` - URL parsing & normalization.
 * `sniffio` - Async library autodetection.
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,7 +111,6 @@ The HTTPX project relies on these excellent libraries:
   * `h2` - HTTP/2 support.
 * `certifi` - SSL certificates.
 * `chardet` - Fallback auto-detection for response encoding.
-* `hstspreload` - determines whether IDNA-encoded host should be only accessed via HTTPS.
 * `idna` - Internationalized domain name support.
 * `rfc3986` - URL parsing & normalization.
 * `sniffio` - Async library autodetection.
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -2,7 +2,6 @@ import functools
 import typing
 from types import TracebackType
 
-import hstspreload
 import httpcore
 
 from ._auth import Auth, BasicAuth, FunctionAuth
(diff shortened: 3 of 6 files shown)


StephenBrown2:
Perhaps `hstspreload`ing could be kept, but not enforced by default, and a `TRACE` or `DEBUG` message emitted when a host is found on the list and doesn't match the preferred scheme?

tomchristie:
So, it's a wooly issue, but I think yes let's get this in and treat HSTS as appropriate as a browser feature, but not *necessarily* desirable in client library.

StephenBrown2:
I would hate for the great work in the `hstspreload` module to go unused. Would it be good to include a brief mention in the docs for those that do want that functionality?

florimondmanca:
I'm trying to think of how people could enable this without us having to support it in core with a flag…

I don't think a custom transport would be the way to go, since transports don't deal with client logic like "optionally modify the request URL".

Probably more like a client subclass, then…?

```python
import httpx
import hstspreload

class HSTSPreloadMixin:
    def build_request(self, *args, **kwargs):
        request = super().build_request(*args, **kwargs)

        url = request.url

        if (
            url.scheme == "http"
            and hstspreload.in_hsts_preload(url.host)
            and len(url.host.split(".")) > 1
        ):
            port = None if url.port == 80 else url.port
            request.url = url.copy_with(scheme="https", port=port)

        return request

class AsyncClient(HSTSPreloadMixin, httpx.AsyncClient):
    pass

class Client(HSTSPreloadMixin, httpx.Client):
    pass
```

Incidentally I think this could also fit in a "middleware" kind of concept (#345), but that's definitely not something we'll have 1.0.

StephenBrown2:
Possibly, but I was thinking of and even simpler approach/demonstration:

`util.py`
```py
from typing import Union

from httpx import URL
from hstspreload import in_hsts_preload

def check_hsts(url: Union[str, URL]):
    if isinstance(url, str):
        url = URL(url)
    if in_hsts_preload(url.host):
        return url.copy_with(scheme="https")
    return url
```
...
`main.py`
```py
import httpx

from util import check_hsts

    client.get(check_hsts(the_url))
```
or with `base_url`:
```py
with httpx.Client(base_url=check_hsts(the_url)) as client:
    client.get(path)
```

PR #1111, opened 2020-08-01
Switch to more concise Timeout parameters

### LINKED ISSUE #1093: More concise Timeout parameters.
Follow up to #1085

Since we're on out final 1.0 polishing run-in I think we should place the `httpx.Timeout(5.0, connect_timeout=60.0)` style on a gentle deprecation with 0.14, in favour of the more concise style of `httpx.Timeout(5.0, connect=60.0)`.

I'm milestoning this for 0.14 so that we can track the remaining items for it, and aim for a release sometime this week.

florimondmanca:
Fixes #1093 

```python
httpx.Timeout(read_timeout=1, write_timeout=1, connect_timeout=5, pool_timeout=10)
```

Becomes:

```python
httpx.Timeout(read=1, write=1, connect=5, pool=10)
```

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -374,16 +374,16 @@ HTTPX also allows you to specify the timeout behavior in more fine grained detai
 There are four different types of timeouts that may occur. These are **connect**,
 **read**, **write**, and **pool** timeouts.
 
-* The **connect timeout** specifies the maximum amount of time to wait until
+* The **connect** timeout specifies the maximum amount of time to wait until
 a connection to the requested host is established. If HTTPX is unable to connect
 within this time frame, a `ConnectTimeout` exception is raised.
-* The **read timeout** specifies the maximum duration to wait for a chunk of
+* The **read** timeout specifies the maximum duration to wait for a chunk of
 data to be received (for example, a chunk of the response body). If HTTPX is
 unable to receive data within this time frame, a `ReadTimeout` exception is raised.
-* The **write timeout** specifies the maximum duration to wait for a chunk of
+* The **write** timeout specifies the maximum duration to wait for a chunk of
 data to be sent (for example, a chunk of the request body). If HTTPX is unable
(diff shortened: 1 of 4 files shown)


httpx/_config.py
```diff
@@ -251,17 +288,11 @@ def __init__(
                 )
                 timeout = None
             self.connect_timeout = (
```
florimondmanca, line 290: Note that I haven't changed the `Timeout.<xyz>_timeout` attributes yet. I _could_, but I don't actually think they're public API, are they? If not, then I'd like to leave that renaming for a follow-up cleanup PR.

docs/advanced.md
```diff
@@ -374,16 +374,16 @@ HTTPX also allows you to specify the timeout behavior in more fine grained detai
 There are four different types of timeouts that may occur. These are **connect**,
 **read**, **write**, and **pool** timeouts.
 
-* The **connect timeout** specifies the maximum amount of time to wait until
+* The **connect** timeout specifies the maximum amount of time to wait until
```
florimondmanca, line 377: This change in emphasis is so that it's clearer now that the usage is `connect=...`. :-)

httpx/_config.py
tomchristie: I think we'd *potentially?* be okay without a deprecation on those, but happy either way.

(Given that they're very unlikely to be used, aren't documented, and that we have a clear policy on 0.x version changes / pinning expectations and nice, clear versioning notes)

florimondmanca: That's my feeling as well… :-) If so I can make the change in this PR, since it's contained to the `Timeout` class (transports expect a timeout `dict` and those keys don't change), and maybe some tests.

docs/advanced.md
tomchristie: Nicely done, yes.

PR #1113, opened 2020-08-01
Rename PoolLimits to Limits

### LINKED ISSUE #1094: Renaming PoolLimits -> Limits.
Another 1.0 final run-in bit of polishing here...

I think we should probably renaming `pool_limits=...` to `limits=...` and `httpx.PoolLimits(...)` to `httpx.Limits(...)`.
If we *are* going to do this, then we should do so with a gentle deprecation, and we should target 0.14.

There's two motivations here...

* Prefer a more concise style.
* At some point we may have limits that are *not* strictly to do with the connection pool. (For example, it'd be helpful particularly with the AsyncClient to be able to enforce a client-wide max concurrent outgoing requests. Which we might just want to unilaterally apply across *all* mounted transports/proxies)

I'm open to push back on this if anyone thinks this is a bit over-zealous, and I *am* mindful that it introduces a bit of thrash. On the flip side, if there's a time to do this, then pre-1.0 is the time.

florimondmanca:
Fixes #1094

Change the following with gentle deprecation:

- `httpx.PoolLimits(...)` becomes `httpx.Limits(...)`
- `Client(pool_limits=...)` becomes `Client(limits=...)`
- `AsyncClient(pool_limits=...)` becomes `AsyncClient(limits=...)`

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -8,10 +8,10 @@ import httpcore
 from ._auth import Auth, BasicAuth, FunctionAuth
 from ._config import (
     DEFAULT_MAX_REDIRECTS,
-    DEFAULT_POOL_LIMITS,
+    DEFAULT_LIMITS,
     DEFAULT_TIMEOUT_CONFIG,
     UNSET,
-    PoolLimits,
+    Limits,
     Proxy,
     Timeout,
     UnsetType,
@@ -50,6 +50,7 @@ from ._utils import (
     get_logger,
     same_origin,
     should_not_be_proxied,
+    warn_deprecated,
 )
 
 logger = get_logger(__name__)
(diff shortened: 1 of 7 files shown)


httpx/_client.py
```diff
@@ -418,8 +419,7 @@ class Client(BaseClient):
     URLs.
     * **timeout** - *(optional)* The timeout configuration to use when sending
     requests.
-    * **pool_limits** - *(optional)* The connection pool configuration to use
-    when determining the maximum number of concurrently open HTTP connections.
+    * **limits** - *(optional)* The limits configuration to use.
```
florimondmanca, line 422: Haven't found a better description of the parameter that's not redundant with the info in the `Limits` docstring. I don't think it's included in the Developer Interface yet so that could be a nice follow-up, and then cross-link it here?

tomchristie: I guess we could say something like... "Allows for configuration options, such as the maximum number of open connections to allow." Just to give a bit of an indication what we mean by "limits" here(?)

tomchristie: Let's pull this in for now as it is. Welcome to follow up with docstring improvements later.

PR #1089, opened 2020-07-24
Consistent multidict methods

tomchristie:
Nice sensible symmetry and expected behaviours across each of the following...
 
* `Headers.get`
* `Headers.get_list`
* `Headers.items`
* `Headers.multi_items`
* `QueryParams.get`
* `QueryParams.get_list`
* `QueryParams.items`
* `QueryParams.multi_items`

The following have been renamed, but the synonyms remain, and are now pending deprecation...

* `Headers.getlist()` -> `Headers.get_list()`
* `QueryParams.getlist()` -> `QueryParams.get_list()`

PR #1114, opened 2020-08-01
Make Headers.keys(), Headers.values() and Headers.items() return set-like views instead of lists

### LINKED ISSUE #899: Should `Headers.keys` be a set-like keys view?

florimondmanca:
Fixes #899

`Headers` inherits from `MutableMapping`, and part of implementing this protocol is returning "set-like views" in `.keys()`, `.values()` and `.items()` (similarly to how dicts do). Previously it was difficult (see #903), but since #1089 we have an internal `Headers._dict` which we can use to build and return native views.

Now `Headers` behave even more like a native mapping. :-)

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -405,23 +405,25 @@ class Headers(typing.MutableMapping[str, str]):
         """
         return self._list
 
-    def keys(self) -> typing.List[str]:  # type: ignore
-        return [key.decode(self.encoding) for key in self._dict.keys()]
+    def keys(self) -> typing.KeysView[str]:
+        return {key.decode(self.encoding): None for key in self._dict.keys()}.keys()
 
-    def values(self) -> typing.List[str]:  # type: ignore
-        return [value.decode(self.encoding) for value in self._dict.values()]
+    def values(self) -> typing.ValuesView[str]:
+        return {
+            key: value.decode(self.encoding) for key, value in self._dict.items()
+        }.values()
 
-    def items(self) -> typing.List[typing.Tuple[str, str]]:  # type: ignore
+    def items(self) -> typing.ItemsView[str, str]:
         """
-        Return a list of `(key, value)` pairs of headers. Concatenate headers
+        Return `(key, value)` items of headers. Concatenate headers
         into a single comma seperated value when a key occurs multiple times.
         """
-        return [
-            (key.decode(self.encoding), value.decode(self.encoding))
+        return {
+            key.decode(self.encoding): value.decode(self.encoding)
             for key, value in self._dict.items()
-        ]
+        }.items()
 
-    def multi_items(self) -> typing.List[typing.Tuple[str, str]]:  # type: ignore
+    def multi_items(self) -> typing.List[typing.Tuple[str, str]]:
         """
         Return a list of `(key, value)` pairs of headers. Allow multiple
         occurences of the same key without concatenating into a single
@@ -470,7 +472,7 @@ class Headers(typing.MutableMapping[str, str]):
             self[header] = headers[header]
 
     def copy(self) -> "Headers":
-        return Headers(self.items(), encoding=self.encoding)
+        return Headers(dict(self.items()), encoding=self.encoding)
 
     def __getitem__(self, key: str) -> str:
         """
--- a/tests/models/test_headers.py
+++ b/tests/models/test_headers.py
@@ -14,9 +14,9 @@ def test_headers():
     assert h.get("a") == "123, 456"
     assert h.get("nope", default=None) is None
     assert h.get_list("a") == ["123", "456"]
-    assert h.keys() == ["a", "b"]
-    assert h.values() == ["123, 456", "789"]
-    assert h.items() == [("a", "123, 456"), ("b", "789")]
+    assert list(h.keys()) == ["a", "b"]
+    assert list(h.values()) == ["123, 456", "789"]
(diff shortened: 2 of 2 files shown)


Ousret:
Hi @florimondmanca 

I think that you should consider reverting this PR. Here's why: This implementation does not reflect how ItemsView and ValuesView works/can be advantageous. In this patch, you are instancing a new dict and the View point to this newly created dict and not your Headers instance and that creates confusion.

```python
from typing import Dict, ItemsView, ValuesView
from httpx._models import Headers

if __name__ == '__main__':

    k: Dict = {"a": "0", "b": "1", "c": "2"}

    iv: ItemsView = k.items()
    vv: ValuesView = k.values()

    print("-----------------before")

    print(iv)
    print(vv)

    del k["a"]

    print("-----------------after")

    print(iv)
    print(vv)

    print("HTTPX Case")

    print("-----------------before")

    headers: Headers = Headers({"a": "0", "b": "0", "c":"0"})

    iv_httpx: ItemsView = headers.items()
    vv_httpx: ValuesView = headers.values()

    print(iv_httpx)
    print(vv_httpx)

    print("-----------------after")

    del headers["a"]

    print(iv_httpx)
    print(vv_httpx)
```

Output : 
```plain
-----------------before
dict_items([('a', '0'), ('b', '1'), ('c', '2')])
dict_values(['0', '1', '2'])
-----------------after
dict_items([('b', '1'), ('c', '2')])
dict_values(['1', '2'])

HTTPX Case
-----------------before
dict_items([('a', '0'), ('b', '0'), ('c', '0')])
dict_values(['0', '0', '0'])
-----------------after
dict_items([('a', '0'), ('b', '0'), ('c', '0')])
dict_values(['0', '0', '0'])
```

Either : 
- You should do something (some ingenious trick) to make it synced with the self::Headers instance
- Or using the previous implementation was perfectly fine

You once told me that open source is not about pleasing every request. Sometimes we have to say no.
I've seen how the instances ItemsView and ValuesView work and it is not flexible like httpx would need it to be. Mostly because of `key: value.decode(self.encoding) for key, value in self._dict.items()` I guess.

Thank you.
And as always thank you for your contributions!

florimondmanca:
@ourset Great call, and you're absolutely right, there's something missing in this PR that I looked over completely which is "dict views should reflect changes to the underlying mapping".

This PR does return "set-like views", but they're snapshots over the headers contents at a given time. If the headers change, the views should reflect that, but right now they don't.

I'm open to reverting and wait to provide a more fully-thought-out implementation. (At first glance that might not obvious. Previously `.keys()`, `.values()` and `.items()` returned lists, which were _also_ one-off snapshots of headers contents. So they didn't completely match this part of the mapping protocol either.)

But to further understand your point:

> I've seen how the instances ItemsView and ValuesView work and it is not flexible like httpx would need it to be.

Not sure I understand this - Would you say that _if_ `Headers` views did reflect the changes to the underlying headers object, they'd be interesting to have?

Ousret:
Re:

> which were also one-off snapshots of headers contents
I agree with you, but at the previous state, we were not expecting anything else as the returned hint was `typing.List`. 

> I'm open to reverting and wait to provide a more fully-thought-out implementation.
Maybe, just maybe I think that you are not supposed to comply with the `MutableMapping` protocol when not required by the core python engine. 

> Not sure I understand this - Would you say that if Headers views did reflect the changes to the underlying headers object, they'd be interesting to have?

What I am saying is that when you are looking to reflect perfectly the `MutableMapping` on `Headers` you will catch the expert eye, some will notice the `typing.List` to `typing.**View` and may expect it to reflect at any time the target `Headers` instance.

Now, the simpler way to implement it correctly is : 
```python
from collections import ItemsView, ValuesView, KeysView

class Headers(...):
....
    def keys(self) -> typing.KeysView[str]:
        return KeysView(self)

    def values(self) -> typing.ValuesView[str]:
        return ValuesView(self)

    def items(self) -> typing.ItemsView[str, str]:
        """
        Return `(key, value)` items of headers. Concatenate headers
        into a single comma seperated value when a key occurs multiple times.
        """
        return ItemsView(self)
```

But it will certainly not do the decode task for you... `key.decode(self.encoding): value.decode(self.encoding) ... for key, value in self._dict.items()`

I myself tried to implement it in the kiss-headers project, but it ended up being ugly as `from collections import ItemsView, ValuesView, KeysView` are merely some definition that link to a c implemented feature of Python.

PR #1115, opened 2020-08-01
Allow disabling auth per-request using `auth=None`

### LINKED ISSUE #1004: Clearing authorization at the request level.
### Checklist

### Describe the bug

Merging of request-level auth *unset* does not take priority over Client level value.

### To reproduce

Try to 'unset' the client set authorization values.

```python
with httpx.Client(auth=('tom', 'mot123')) as client:
    r = client.get('http://httpbin.org', auth=None)
    
assert 'Authorization' not in r.request.headers['Authorization']
```

### Expected behavior

The documentation for [Merging of configuration](https://www.python-httpx.org/advanced/#merging-of-configuration) states `For all other parameters, the request-level value takes priority.` and shows this example:

```python
with httpx.Client(auth=('tom', 'mot123')) as client:
    r = client.get('https://example.com', auth=('alice', 'ecila123'))
_, _, auth = r.request.headers['Authorization'].partition(' ')
```

I would expect that by setting `auth=None` in the request level that the individual request would be sent without authorization.

### Actual behavior

In reality, the authorization is not overridden. This is due to this line: https://github.com/encode/httpx/blob/master/httpx/_client.py#L258

Overriding the auth value via this method works:

```python
with httpx.Client(auth=('tom', 'mot123')) as client:
    r = client.get('http://httpbin.org', auth=Auth())
    
assert 'Authorization' not in r.request.headers['Authorization']
```

Personally, I found passing in `auth=None` to be more intuitive than `auth=Auth()` to indicate that no authorization should be used.

florimondmanca:
Fixes #1004 

Switches request and verb handlers to default to `auth=UNSET`, so that `auth=None` can be treated as "no auth on this request" (instead of "use default auth from the client").

Right now this is only possible using an empty `Auth`instance:

```python
r = client.get(url, auth=httpx.Auth())
```

which I'd agree is not intuitive (and it also works "by chance" because `Auth` happens to be a "do-nothing" implementation, rather than a strict base class interface that raises `NotImplementedError`).

Note that this now differs from the behavior of `requests`, which treats `auth=None` as "use the default" (see https://github.com/encode/httpx/issues/1004#issuecomment-636193515).


florimondmanca:
Closed in favor of #1116

PR #1116, opened 2020-08-01
Improve docs on client/request configuration merging

### LINKED ISSUE #1004: Clearing authorization at the request level.

florimondmanca:
Fixes #1004 

Update the [Merging configuration](https://www.python-httpx.org/advanced/#merging-of-configuration) docs to resolve these items:

- Document what happens when mergeable configuration (headers, params, cookies) is passed `None` at the request level. Improve example snippet by using `X-Client-Header` and `X-Request-Header` example headers.

Screenshot 2020-08-01 at 13 16 13

- Document the gotcha that request-level `auth=None` doesn't disable auth, even though request-level `auth` is advertised as generally taking precedence over client-level `auth`.

Screenshot 2020-08-01 at 13 25 23

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -97,21 +97,20 @@ For example, to apply a set of custom headers _on every request_:
 
 When a configuration option is provided at both the client-level and request-level, one of two things can happen:
 
-- For headers, query parameters and cookies, the values are combined together. For example:
+- For headers, query parameters and cookies, the values are combined together. Passing `None` is treated as "do nothing" (i.e. "use the client value unchanged"). For example:
 
 ```python
->>> headers = {'X-Auth': 'from-client'}
+>>> headers = {'X-Client-Header': 'from-client'}
 >>> params = {'client_id': 'client1'}
 >>> with httpx.Client(headers=headers, params=params) as client:
-...     headers = {'X-Custom': 'from-request'}
-...     params = {'request_id': 'request1'}
-...     r = client.get('https://example.com', headers=headers, params=params)
+...     headers = {'X-Request-Header': 'from-request'}
+...     r = client.get('https://example.com', headers=headers, params=None)
 ...
 >>> r.request.url
-URL('https://example.com?client_id=client1&request_id=request1')
->>> r.request.headers['X-Auth']
+URL('https://example.com?client_id=client1')
+>>> r.request.headers['X-Client-Header']
 'from-client'
->>> r.request.headers['X-Custom']
+>>> r.request.headers['X-Request-Header']
 'from-request'
 ```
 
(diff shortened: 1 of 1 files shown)


tomchristie:
An alternative here would be to do the UNSET dance on each of these parameters, so that they *do* have the expected behaviour when passing `params=None`, `headers=None`, `auth=None`, allowing them to override the client default.

Would that make sense?

florimondmanca:
Well, hmm, I don't know. We already provide the "build a request and edit its headers/params/cookies to remove items" solution for these cases. Plus, I don't know if the use case for "send absolutely no headers/params/cookies on this request" is that convincing. On the contrary there _might_ a drawback which is that this change would be breaking to people that _do_ pass `None` to these params. Not sure there are many situations, I'd suppose most likely in a kind of "wrapper" situation, but not sure.

On the other hand, the argument for `auth` is a bit more ambiguous, we'd agree? So we either change it - #1115, or keep it for consistency while documenting the gotcha - this PR.

tomchristie:
So, I *think* I'm in favour of #1115 for auth, right? It's come up as an issue for us, and it's come up as an issue for requests, plus I can fairly easily see the use cases for it.

But yes, otherwise I guess we can be okay with `None` meaning "use the client value" for params, headers, and cookies.

florimondmanca:
Yup, that's my sentiment as well. Will reopen #1115.

PR #1118, opened 2020-08-02
Support bare hostname in proxy

### LINKED ISSUE #1082: Should we support bare hostnames in `ALL_PROXY`,`HTTP_PROXY`, `HTTPS_PROXY`?
### Checklist

### Describe the bug

I am a new user for httpx, and I just install it, then try to run it in python REPL.
I type `httpx.get('https://www.github.com')`  and then got the "No scheme included in URL".
So I try to debug it, and finally I found that because I set the `export ALL_PROXY=127.0.0.1:7890` in the shell, the proxy does not have a scheme. 
This proxy set is correct for when I use curl、brew and many other programe, So it should also work fun in httpx?
for any other reason httpx must have the scheme, I think it should prompt "a proxy set error" at least.

### Debugging material

```
~/projects » python3                                                                                                       hejl@hejldeIMAC
Python 3.8.4 (default, Jul 14 2020, 02:58:48) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
>>> import httpx
>>> httpx.get("https://www.github.com")
  File "/usr/local/lib/python3.8/site-packages/httpx/_api.py", line 159, in get
    return request(
  File "/usr/local/lib/python3.8/site-packages/httpx/_api.py", line 83, in request
    with Client(
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 467, in __init__
    proxy_map = self.get_proxy_map(proxies, trust_env)
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 87, in get_proxy_map
    return {
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 88, in <dictcomp>
    key: Proxy(url=url)
  File "/usr/local/lib/python3.8/site-packages/httpx/_config.py", line 335, in __init__
    url = URL(url)
  File "/usr/local/lib/python3.8/site-packages/httpx/_models.py", line 90, in __init__
    raise InvalidURL("No scheme included in URL.")
httpx._exceptions.InvalidURL: No scheme included in URL.
```

### Environment

- OS:  MacOS
- Python version: 3.8.4
- HTTPX version: 0.13.3
- Async environment: no
- HTTP proxy: yes  `127.0.0.1:7890`
- Custom certificates: no

j178:
Closes #1082 

- Auto prepend `http` scheme for bare hostname in proxy
- Fix two internal use of deprcating `.getlist()`

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -49,7 +49,6 @@ from ._utils import (
     get_environment_proxies,
     get_logger,
     same_origin,
-    should_not_be_proxied,
     warn_deprecated,
 )
 
@@ -95,7 +94,7 @@ class BaseClient:
         if proxies is None:
             if trust_env:
                 return {
-                    key: Proxy(url=url)
+                    key: None if url is None else Proxy(url=url)
                     for key, url in get_environment_proxies().items()
                 }
             return {}
@@ -558,10 +557,9 @@ class Client(BaseClient):
         url = request.url
         enforce_http_url(request)
 
-        if self._proxies and not should_not_be_proxied(url):
-            for pattern, transport in self._proxies.items():
-                if pattern.matches(url):
-                    return self._transport if transport is None else transport
+        for pattern, transport in self._proxies.items():
+            if pattern.matches(url):
+                return self._transport if transport is None else transport
 
         return self._transport
 
@@ -1089,10 +1087,9 @@ class AsyncClient(BaseClient):
         url = request.url
         enforce_http_url(request)
 
-        if self._proxies and not should_not_be_proxied(url):
-            for pattern, transport in self._proxies.items():
-                if pattern.matches(url):
-                    return self._transport if transport is None else transport
+        for pattern, transport in self._proxies.items():
+            if pattern.matches(url):
+                return self._transport if transport is None else transport
 
         return self._transport
 
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -299,42 +299,39 @@ def same_origin(url: "URL", other: "URL") -> bool:
     )
 
 
-def should_not_be_proxied(url: "URL") -> bool:
-    """
-    Return True if url should not be proxied,
-    return False otherwise.
-    """
-    no_proxy = getproxies().get("no")
-    if not no_proxy:
-        return False
-    no_proxy_list = [host.strip() for host in no_proxy.split(",")]
-    for name in no_proxy_list:
-        if name == "*":
-            return True
-        if name:
-            name = name.lstrip(".")  # ignore leading dots
-            name = re.escape(name)
-            pattern = r"(.+\.)?%s$" % name
-            if re.match(pattern, url.host, re.I) or re.match(
-                pattern, url.authority, re.I
-            ):
-                return True
-    return False
-
-
-def get_environment_proxies() -> typing.Dict[str, str]:
+def get_environment_proxies() -> typing.Dict[str, typing.Optional[str]]:
     """Gets proxy information from the environment"""
 
     # urllib.request.getproxies() falls back on System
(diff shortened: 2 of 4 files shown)


tests/models/test_headers.py
```diff
@@ -14,6 +14,7 @@ def test_headers():
     assert h.get("a") == "123, 456"
     assert h.get("nope", default=None) is None
     assert h.get_list("a") == ["123", "456"]
+    assert h.getlist("a") == ["123", "456"]
```
tomchristie, line 17:
```suggestion
```

tests/models/test_queryparams.py
```diff
@@ -20,6 +20,7 @@ def test_queryparams(source):
     assert q.get("a") == "456"
     assert q.get("nope", default=None) is None
     assert q.get_list("a") == ["123", "456"]
+    assert q.getlist("a") == ["123", "456"]
```
tomchristie, line 23:
```suggestion
```

tomchristie:
Fantastic, loving this, thanks! 😃

One interesting question here...
I *think* we probably only want to support this case wrt. bare environment variables, rather than supporting it in `Proxy` itself. (?)

Eg. If `get_environement_proxies` was changed from...

```python
for scheme in ("http", "https", "all"):
    if proxy_info.get(scheme):
        mounts[scheme] = proxy_info[scheme]
```

to instead be this...

```python
for scheme in ("http", "https", "all"):
    if proxy_info.get(scheme):
        hostname = proxy_info[scheme]
        mounts[scheme] = hostname if "://" in hostname else f"http://{hostname}"
```

Then we wouldn't change the code in `Proxy` at all, and our programmatic usage would remain a bit neater and more strict.

```python
client = httpx.Client(proxies="http://localhost:1234")  # Okay
client = httpx.Client(proxies="localhost:1234")         # Invalid, raises an error
```

While still handling the fuzzy `ALL_PROXY=localhost:1234` case gracefully.

Thoughts?

httpx/_models.py
```diff
@@ -281,7 +281,7 @@ def update(self, params: QueryParamTypes = None) -> None:
 
         params = QueryParams(params)
         for param in params:
-            item, *extras = params.getlist(param)
+            item, *extras = params.get_list(param)
```
tomchristie, line 284:
```suggestion
            item, *extras = params.getlist(param)
```

httpx/_models.py
```diff
@@ -795,7 +795,7 @@ def decoder(self) -> Decoder:
         """
         if not hasattr(self, "_decoder"):
             decoders: typing.List[Decoder] = []
-            values = self.headers.getlist("content-encoding", split_commas=True)
+            values = self.headers.get_list("content-encoding", split_commas=True)
```
tomchristie, line 798:
```suggestion
            values = self.headers.getlist("content-encoding", split_commas=True)
```

florimondmanca:
Great stuff! Agreed with ^. :-) Also, if we go the "env vars only" path, we can probably update the [Environment Variables: HTTP_PROXY, HTTPS_PROXY, ALL_PROXY](https://www.python-httpx.org/environment_variables/#http_proxy-https_proxy-all_proxy) docs as part of this PR?

tomchristie:
Great catch on the `get_list`/`getlist` - I've dropped those changes out of this PR.

We can treat them separately (Because I'd probably prefer that we `pragma: nocover` the to-be-deprecated lines, rather than include them in the test cases like that.)

tomchristie:
@florimondmanca Yes, good call. General note: We're also missing `NO_PROXY` there.

florimondmanca:
Yup. I'm working on a revamp of proxy docs anyway so I can include that stuff there as well.

PR #1119, opened 2020-08-02
Use get_list consistently

tomchristie:
Pulled through from a nice catch in #1118

I'd assumed that we ought to pragma: nocover these cases, rather than test them, but on a second look this *does* seem to be the pragmatic option, so let's just roll with it.

Thanks to @j178 for this!

### DIFF
--- a/tests/models/test_headers.py
+++ b/tests/models/test_headers.py
@@ -13,6 +13,7 @@ def test_headers():
     assert h["a"] == "123, 456"
     assert h.get("a") == "123, 456"
     assert h.get("nope", default=None) is None
+    assert h.getlist("a") == ["123", "456"]
     assert h.get_list("a") == ["123", "456"]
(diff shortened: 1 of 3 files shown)


tests/models/test_headers.py
```diff
@@ -13,6 +13,7 @@ def test_headers():
     assert h["a"] == "123, 456"
     assert h.get("a") == "123, 456"
     assert h.get("nope", default=None) is None
+    assert h.getlist("a") == ["123", "456"]
```
florimondmanca, line 16: Do we want a `with pytest.warns(DeprecatedWarning)` here, so that these known warnings don't show up in the pytest output?

tomchristie: 💯 - Yes, thanks! Which helped catch that they needed to be `DeprecationWarning` in line with everything else (not PendingDeprecationWarning).

PR #1121, opened 2020-08-02
Use named keyword arguments on `httpcore` streams

tomchristie:
Use explicitly named keyword arguments rather than unnamed args, in order to allow us to issue an `httpcore` 0.10 release that includes https://github.com/encode/httpcore/pull/127, without breaking anything in `httpx`.

A later follow up can use `content=...` to tidy things up here.

### DIFF
--- a/httpx/_content_streams.py
+++ b/httpx/_content_streams.py
@@ -18,6 +18,13 @@ from ._utils import (
 
 
 class ContentStream(httpcore.AsyncByteStream, httpcore.SyncByteStream):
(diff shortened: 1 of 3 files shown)


tomchristie:
Closing in favour of switching to the concrete ByteStream implementations hopefully coming in https://github.com/encode/httpcore/pull/133

Resolved this side in #1126 eg see https://github.com/encode/httpx/pull/1126/files#diff-0746bb9fa941d952b3980a91e14f9d7dR79

PR #1122, opened 2020-08-02
Setting `app=...` or `transport=...` should bypass environment proxies.

### LINKED ISSUE #1039: ASGI requests are routed through proxy

tomchristie:
If either `app=...` or `transport=...` are passed explicitly, then no env-var proxies should be used.

Closes #1039

Eg.

```python
# Disregard any `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` environment settings in these contexts...
client = httpx.Client(app=my_flask_app)
client = httpx.Client(transport=httpx.AsyncTransport(my_starlette_app)
```

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -473,7 +473,10 @@ class Client(BaseClient):
             )
             limits = pool_limits
 
-        proxy_map = self._get_proxy_map(proxies, trust_env)
+        if app is None and transport is None:
+            proxy_map = self._get_proxy_map(proxies, trust_env)
+        else:
+            proxy_map = {}
 
         self._transport = self._init_transport(
             verify=verify,
@@ -1003,7 +1006,10 @@ class AsyncClient(BaseClient):
             )
             limits = pool_limits
 
-        proxy_map = self._get_proxy_map(proxies, trust_env)
+        if app is None and transport is None:
+            proxy_map = self._get_proxy_map(proxies, trust_env)
+        else:
+            proxy_map = None
 
         self._transport = self._init_transport(
             verify=verify,
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -89,10 +89,10 @@ class BaseClient:
         return self._trust_env
 
     def _get_proxy_map(
-        self, proxies: typing.Optional[ProxiesTypes], trust_env: bool,
+        self, proxies: typing.Optional[ProxiesTypes], allow_env_proxies: bool,
     ) -> typing.Dict[str, typing.Optional[Proxy]]:
         if proxies is None:
-            if trust_env:
+            if allow_env_proxies:
                 return {
                     key: None if url is None else Proxy(url=url)
                     for key, url in get_environment_proxies().items()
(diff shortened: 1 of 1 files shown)


MarkWine:
Sorry pop this back up, and this probably just reflects may lack of understanding of proxies, but I was wondering why using a transport also disregards the proxy.  Looking at the requests HTTPAdapter (which seems to be similar in most regards), environmental proxies are maintained, so in the following example: 
```
import asyncio
import os

import httpx
from requests import Session
from requests.adapters import HTTPAdapter
from urllib3.util import Retry

os.environ["https_proxy"] = "http://127.0.0.1:8080/"

async def test_transport_proxy():
    base_client = httpx.AsyncClient()
    transport_client = httpx.AsyncClient(transport=httpx.AsyncHTTPTransport(retries=2))
    await base_client.get("https://example.org/1")
    await transport_client.get("https://example.org/2")


def test_adapter_proxy():
    base_session = Session()
    transport_session = Session()
    transport_session.mount("https://", HTTPAdapter(max_retries=Retry(total=3)))
    base_session.get("https://example.org/3")
    transport_session.get("https://example.org/4")


asyncio.run(test_transport_proxy_a())
test_adapter_proxy()
```

The proxy is used in all but case 2. I ran into this when we migrated an app to an environment needing proxies, and found everything worked based on environment proxy settings except the one case in which we added retries when making a large number of calls to a somewhat flaky external service. It was straightforward to add proxy rules directly, so no complaints per se, but I was wondering on the rationale behind this.

Issue #1105, opened 2020-07-31
On `proxies={"http": ...}` vs. `proxies={"http://": ...}`


tomchristie:
Currently we're in line with `requests` in supporting "all", "http", and "https" as valid proxy keys.

Internally we handle those as URL patterns that we can match against, and eg. `"http"` is *actually* mapped to `"http://"`.

I'm wondering if we might have better internal consistency by...

* Raising a warning on base usage of "all", "http", "https", later escalating this to an error.
* Instead requiring an explicit "all://", "http://", "https://", all of which properly parse directly as URLs.

My feeling is that doing so would create a nicer consistency, and also has a subtle indication to developers about what is actually going on here.

For example...

`proxies={"http://": httpx.Proxies(...)}` neatly implies that `proxies={"http://www.example.com": httpx.Proxies(...)}` might also be a valid thing to do, which indeed it is.

Allowing `proxies={"http": httpx.Proxies(...)}` makes it less obvious that the key is a general purpose URL pattern.

~~Additionally this change means that other usages will end up with more expected behaviors, eg. `proxies={"www.example.com": ...}` will do what you'd expect, rather than mapping to `proxies={"www.example.com://": ...}` which is what we currently do.~~ - Not true, that'd parse as scheme='', host='', path='www.example.com'. Instead we'd probably want to be strict and ensure that only absolute URLs are allowed as pattern keys. (While still including a graceful deprecation for plain "http", "https", "all")

This becomes doubly relevant once we introduce a mount API, which will use the same URL pattern style for it's API, as current `proxies=...` argument.

cdeler:
Hello @tomchristie ,

Are you talking about the changes in `httpx._client.BaseClient._get_proxy_map` (here: [link](https://github.com/encode/httpx/blob/master/httpx/_client.py#L101))
Also, what should we do with `get_environment_proxies`? ([link](https://github.com/encode/httpx/blob/master/httpx/_utils.py#L302)) This function requires envs like `HTTPS_PROXY` etc.

PR #1123, opened 2020-08-02
Revamp proxies documentation

florimondmanca:
Promted by https://github.com/encode/httpx/pull/927#issuecomment-667642611

Our current proxies documentation is a bit lacking, and not necessarily nudging users to use the most simple solution whenever possible (i.e. `Client(proxies="http://...")`), so I dived my head into the world of web proxies and came up with this updated documentation, which outlines:

- Basic usage examples.
- Authentication support.
- Detailed docs about how proxies are routed, and what we support.
- Advanced discussions on forwarding vs tunneling. (Which I _think_ basically boils down to "HTTPX does the right thing for 99% of use cases, and if you know what you're doing here's how you can force things".) I removed the "hide traffic from proxies" example because I don't think we're confident enough to provide any kind of security advice in these docs. Plus I'm not 100% sure that example is a generally advisable thing anyway (eg if you're using your company proxy, then allowing it to intercept the traffic might actually be a corporate security policy requirement).

Rendered preview:

proxies

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -262,68 +262,228 @@ client = httpx.Client(trust_env=False)
 
 ## HTTP Proxying
 
-HTTPX supports setting up HTTP proxies the same way that Requests does via the `proxies` parameter.
-For example to forward all HTTP traffic to `http://127.0.0.1:3080` and all HTTPS traffic
-to `http://127.0.0.1:3081` your `proxies` config would look like this:
+HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization.
+
+_Note: SOCKS proxies are not supported yet._
+
+<div align="center">
+    <img src="https://upload.wikimedia.org/wikipedia/commons/thumb/2/27/Open_proxy_h2g2bob.svg/480px-Open_proxy_h2g2bob.svg.png"/>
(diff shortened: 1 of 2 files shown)


tomchristie:
👀 Wow, this is fantastic!

Since it ties in here, what are your thoughts on https://github.com/encode/httpx/issues/1105?

I think the docs read even more nicely if we're starting with `"all://"` progressing to `"http://"`, `"https://"`, and then on to `"all://example.com"` and so on...

We could already move to using that style in these docs, since it works like that already, we just happen to be special casing the plain `"all"`/`"http"`/`"https"` type cases, but we could gracefully deprecate that. (?)

tomchristie:
Other thoughts...

* We should remember to update the "See documentation on `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` ..." to also include `NO_PROXY` once that's in there.
* I think we could possibly(?) switch the "Routing algorithm overview" to something a bit more general, that makes it clear that more specific patterns are matched in preference to less specific patterns, without getting into talking about "algorithms" or being quite so specific about implementation details.

tomchristie:
In any case, going to approve this, since it's *fantastic*, and we can move on whichever of those items you think is valid (or not).

florimondmanca:
Absolutely agree yes, the progression from `all://` to `http://` to `http://domain` to `http://domain:port` would be super sweet! :+1: Still need to take a closer look at #1105 but from afar I like the idea…

Issue #1108, opened 2020-07-31
Should HTTPError be a base class for RequestError and HTTPStatusError?


iwoloschin:
I'm going through some of my projects preparing them for the 0.14 release and it occurred to me that the following line may be confusing and trip folks up.

https://github.com/encode/httpx/blob/9409900898f40ed855e9c7a7664b7e068599a07d/httpx/_exceptions.py#L282

This will break something like the following since the HTTPStatusError is now no longer caught, it'd just raise straight up and crash the program.
```python
try:
    r = httpx.get('https://httpbin.org/status/400')
    r.raise_for_status()
except httpx.HTTPError as error:
    print(error)
```

Instead, can we change the line to:
```python
HTTPError = (HTTPStatusError, RequestError)
```

That way at least something close to (or maybe exactly like?) the original functionality is maintained, at least until it is formally deprecated?

StephenBrown2:
Would need to include all the top-level exceptions from https://github.com/encode/httpx/pull/1095, so:

```python
HTTPError = (HTTPStatusError, RequestError, NotRedirectResponse, CookieConflict, StreamError)

tomchristie:
That's a really interesting idea, thanks!

An alternative might be to keep it as an exception class, as a base class of `RequestError` and `HTTPStatusError`?

StephenBrown2:
> An alternative might be to keep it as an exception class, as a base class of RequestError and HTTPStatusError?

I'd support this option.

iwoloschin:
Actually, yeah, an exception class is probably better, you could probably do something to print out a deprecation warning that way?

tomchristie:
I guess theres a question here, re: if we should actually prefer to have that as a base class permanently, or if we'd only be doing this to capture both cases during a gentle deprecation.

Incidentally I'm not sure there is any way we'd be able to raise a deprecation warning to users importing and using an exception class in an `except` statement.

florimondmanca:
> Incidentally I'm not sure there is any way we'd be able to raise a deprecation warning to users importing and using an exception class in an except statement.

Actually I think there's a way: [Module `__getattr__`](https://www.python.org/dev/peps/pep-0562/). :-) (It's 3.7+ only, but can we afford a hard-break on 3.6 maybe…?)

```python
# httpx/__init__.py

def __getattr__(name: str):  # type: ignore  # Called if `httpx.<name>` was not found
    if name == "HTTPError":
        from ._utils import warn_deprecated

[...]

PR #1083, opened 2020-07-24
Version 0.14.0

tomchristie:
## 0.14.0 (August 7th, 2020)

The 0.14 release includes a range of improvements to the public API, intended on preparing for our upcoming 1.0 release.

* Our HTTP/2 support is now fully optional. **You now need to use `pip install httpx[http2]` if you want to include the HTTP/2 dependancies.**
* Our HSTS support has now been removed. Rewriting URLs from `http` to `https` if the host is on the HSTS list can be beneficial in avoiding roundtrips to incorrectly formed URLs, but on balance we've decided to remove this feature, on the principle of least surprise. Most programmatic clients do not include HSTS support, and for now we're opting to remove our support for it.
* Our exception hierarchy has been overhauled. Most users will want to stick with their existing `httpx.HTTPError` usage, but we've got a clearer overall structure now. See https://www.python-httpx.org/exceptions/ for more details.

When upgrading you should be aware of the following public API changes. Note that deprecated usages will currently continue to function, but will issue warnings.

* You should now use `httpx.codes` consistently in favour of `httpx.StatusCodes`.
* Usage of `httpx.Timeout()` should now always include an explicit default. Eg. `httpx.Timeout(None, pool=5.0)`.
* When using `httpx.Timeout()`, we now have more concisely named keyword arguments. Eg. `read=5.0`, instead of `read_timeout=5.0`.
* Use `httpx.Limits()` instead of `httpx.PoolLimits()`, and `limits=...` instead of `pool_limits=...`.
* The `httpx.Limits(max_keepalive=...)` argument is now deprecated in favour of a more explicit `httpx.Limits(max_keepalive_connections=...)`
* Keys used with `Client(proxies={...})` should now be in the style of `{"http://": ...}`, rather than `{"http": ...}`.
* The multidict methods `Headers.getlist()` and `QueryParams.getlist()` are deprecated in favour of more consistent `.get_list()` variants.
* The `URL.is_ssl` property is deprecated in favour of `URL.scheme == "https"`.
* The `URL.join(relative_url=...)` method is now `URL.join(url=...)`. This change does not support warnings for the deprecated usage style.

One notable aspect of the 0.14.0 release is that it tightens up the public API for `httpx`, by ensuring that several internal attributes and methods have now become strictly private.

The following previously had nominally public names on the client, but were all undocumented and intended solely for internal usage. They are all now replaced with underscored names, and should not be relied on or accessed.

These changes should not affect users who have been working from the `httpx` documentation.

[...]

PR #1125, opened 2020-08-03
Keep HTTPError as a base class for .request() and .raise_for_status()

tomchristie:
Closes https://github.com/encode/httpx/issues/1108, and following rationale in https://github.com/encode/httpx/pull/1083#issuecomment-667934688, which I'll repeat here...

I think there's probably also a decent case to make for *not* deprecating it at all, and instead keeping it as a base class, but *only* for `RequestError` and `HTTPStatusError`, so that...

```python
try:
    response = httpx.get("https://www.example.com/")  # May raise a RequestError subclass
    response.raise_for_status()  # May raise HTTPStatusError
    pass
```

Which means existing code all keeps working as expected, while still providing finer-grained classes where needed.

It's also more concise than the alternative...

```python
try:
    response = httpx.get("https://www.example.com/")  # May raise a RequestError subclass
    response.raise_for_status()  # May raise HTTPStatusError
except (httpx.RequestError, httpx.HTTPStatusError) as exc:
    pass
```

This hadn't really occurred to me when looking at https://github.com/encode/httpx/pull/1095.

The typing is also still nice and clear here. Both `HTTPStatusError` and `RequestError` provide `.request`, so we can suitably annotate that on the base class. However accessing `.response` on `httpx.HTTPError` would be a typing error, since it could either be a `RequestError` (which isn't associated with a response) or a `HTTPStatusError` (which is).

The following are all valid...

```python
response = httpx.get("https://www.example.com/")  # May raise a RequestError subclass
try:
    response.raise_for_status()  # May raise HTTPStatusError
    print(f"Unexpected status code in response: {exc.response.status_code}")
```

```python
try:
    response = httpx.get("https://www.example.com/")  # May raise a RequestError subclass
    response.raise_for_status()  # May raise HTTPStatusError
except httpx.RequestError as exc:
    print(f"Request to {exc.request.url} failed: {exc.message}")
    print(f"Unexpected status code in response: {exc.response.status_code}")
```

```python
try:
    response = httpx.get("https://www.example.com/")  # May raise a RequestError subclass
    response.raise_for_status()  # May raise HTTPStatusError
    if isinstance(exc, HTTPStatusError):
        # The following line will type-check properly now.
        print(f"Unexpected status code in response: {exc.response.status_code}")
    else:
        print(f"Request to {exc.request.url} failed: {exc.message}")
```

I think this is probably all a usability win for our users given...

* `except httpx.HTTPError as exc:` is more concise than `except (httpx.RequestError, httpx.StatusCodeError) as exc:`
* It's a closer match for requests codebases that are migrating.
* We're not breaking anything for our existing users here.

How do we feel about this?

### DIFF
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -1,32 +1,33 @@
 """
 Our exception hierarchy:
 
-* RequestError
-  + TransportError
-    - TimeoutException
-      · ConnectTimeout
-      · ReadTimeout
-      · WriteTimeout
-      · PoolTimeout
-    - NetworkError
-      · ConnectError
-      · ReadError
-      · WriteError
-      · CloseError
-    - ProxyError
-    - ProtocolError
-  + DecodingError
-  + TooManyRedirects
-  + RequestBodyUnavailable
-  + InvalidURL
-* HTTPStatusError
+* HTTPError
+  x RequestError
+    + TransportError
+      - TimeoutException
+        · ConnectTimeout
+        · ReadTimeout
+        · WriteTimeout
+        · PoolTimeout
+      - NetworkError
+        · ConnectError
+        · ReadError
+        · WriteError
+        · CloseError
+      - ProtocolError
+      - ProxyError
+    + DecodingError
+    + TooManyRedirects
+    + RequestBodyUnavailable
+    + InvalidURL
+  x HTTPStatusError
 * NotRedirectResponse
 * CookieConflict
 * StreamError
-  + StreamConsumed
-  + ResponseNotRead
-  + RequestNotRead
-  + ResponseClosed
+  x StreamConsumed
+  x ResponseNotRead
+  x RequestNotRead
+  x ResponseClosed
 """
 import contextlib
 import typing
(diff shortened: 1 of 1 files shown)


iwoloschin:
This looks great, with one question, is there a reason you're explicitly not using `HTTPError` as a base exception for all possible httpx exceptions?

tomchristie:
> is there a reason you're explicitly not using HTTPError as a base exception for all possible httpx exceptions?

Absolutely, yup. `RequestError` and `HTTPStatusError` are categories of exception that you may want to catch and handle.

The remaining cases are due to programmatic errors. It's not typically useful to catch and handle those cases, and including them in `HTTPError` means we'd be over-generalising a class that's *really* just intended for the particularly useful and common class of errors that are "could not make the request, or the response was 4xx/5xx".

Of the remaining cases...

* `NotRedirectResponse` occurs if you're calling `response.next()` without properly checking `.is_redirect_response` first.
* The `StreamError` cases occur if you're accessing request/response streams in some broken way, eg. trying to stream the body twice, trying to stream the body after having closed it etc.
* `CookieConflictError` occurs if you need to be more specific in your `response.cookies.get()` access, and narrow down which domain/path you want a cookie for. (Granted, this is a *bit* more ambiguous than the other two cases.)

iwoloschin:
Ah, well, that explains why I've never really needed those other exceptions! I think that's a reasonable explanation and I'd recommend that something similar (or even just what you wrote) definitely be included in the documentation.

michaeloliverx:
The `.request` attribute was removed from `HTTPError` in https://github.com/encode/httpx/pull/1522, was this intended?

PR #1126, opened 2020-08-03
Update to httpcore 0.10

tomchristie:
* Add `UnsupportedProtocol`, `LocalProtocolError`, `RemoteProtocolError`, remove `InvalidURL`.
* Use `timeout: Mapping[str, Optional[float]` instead of `timeout: Dict[str, Optional[float]`.
* Remove `enforce_http_url` and defer all scheme checking to the transport layer.
* Use `httpcore.PlainByteStream`/`httpcore.IteratorByteStream`.
* Use `max_keepalive_connections` not `max_keepalive`

(Also, builds on #1125)

Once httpcore 0.10 is released, we should update the httpcore version pin in this PR, ensure this all passes, and get it pulled in.


httpx/_exceptions.py
```diff
@@ -132,9 +155,35 @@ class ProxyError(TransportError):
     """
 
 
+class UnsupportedProtocol(TransportError):
+    """
+    Attempted to make a request to a no-supported protocol.
```
StephenBrown2, line 160:
```suggestion
    Attempted to make a request to an unsupported protocol.
```

httpx/_exceptions.py
```diff
@@ -319,6 +319,7 @@ def __init__(self) -> None:
         super().__init__(message)
 
 
+<<<<<<< HEAD
```
StephenBrown2: Whoops

PR #1127, opened 2020-08-04
Raise warning if proxy key is eg. "all" instead of "all://".

### LINKED ISSUE #1105: On `proxies={"http": ...}` vs. `proxies={"http://": ...}`

cdeler:
Closes #1105

### DIFF
--- a/tests/client/test_proxies.py
+++ b/tests/client/test_proxies.py
@@ -20,16 +20,19 @@ def url_to_origin(url: str):
 @pytest.mark.parametrize(
     ["proxies", "expected_proxies"],
     [
-        ("http://127.0.0.1", [("all", "http://127.0.0.1")]),
-        ({"all": "http://127.0.0.1"}, [("all", "http://127.0.0.1")]),
+        ("http://127.0.0.1", [("all://", "http://127.0.0.1")]),
+        ({"all://": "http://127.0.0.1"}, [("all://", "http://127.0.0.1")]),
         (
-            {"http": "http://127.0.0.1", "https": "https://127.0.0.1"},
-            [("http", "http://127.0.0.1"), ("https", "https://127.0.0.1")],
+            {"http://": "http://127.0.0.1", "https://": "https://127.0.0.1"},
+            [("http://", "http://127.0.0.1"), ("https://", "https://127.0.0.1")],
         ),
-        (httpx.Proxy("http://127.0.0.1"), [("all", "http://127.0.0.1")]),
+        (httpx.Proxy("http://127.0.0.1"), [("all://", "http://127.0.0.1")]),
         (
-            {"https": httpx.Proxy("https://127.0.0.1"), "all": "http://127.0.0.1"},
-            [("all", "http://127.0.0.1"), ("https", "https://127.0.0.1")],
+            {
+                "https://": httpx.Proxy("https://127.0.0.1"),
+                "all://": "http://127.0.0.1",
+            },
+            [("all://", "http://127.0.0.1"), ("https://", "https://127.0.0.1")],
         ),
     ],
 )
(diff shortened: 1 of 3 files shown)


cdeler:
I see the tests, should be changed, also I can add tests for the new warning appearing. @tomchristie let me know if I can continue with that

tomchristie:
Thanks for taking a go at this!

So, I'd change the implementation here, for simplicity.

Instead of changing *anything* in the Client, I'd add a single line here... https://github.com/encode/httpx/blob/78cf16ace92b42de5972a490787002631d558b32/httpx/_utils.py#L481 
that raises a warning such as `f"Proxy keys should use proper URL forms rather than plain scheme strings. Instead of {pattern}, use {pattern}://"` - that ought to be enough to deal with this more neatly.

You probably also want to then update the docs in https://github.com/encode/httpx/blob/master/docs/advanced.md to use the preferred styles.

tests/client/test_proxies.py
```diff
@@ -244,3 +247,21 @@ def test_proxies_environ(monkeypatch, client_class, url, env, expected):
...
+    ],
+)
+def test_for_deprecated_proxy_params(proxies, expected_scheme):
+    with pytest.deprecated_call() as block:
+        httpx.AsyncClient(proxies=proxies)
+
+    warning_message = str(block.pop(DeprecationWarning))
+
+    for scheme in expected_scheme:
+        assert scheme in warning_message
```
florimondmanca, line 267: Could this be simplified to this by converting `expected_scheme` to a plain string?

```suggestion
    assert expected_scheme in warning_message
```

tomchristie: Yeah I think that was walking through the *characters* in the string. :)

tests/client/test_proxies.py
```diff
@@ -244,3 +247,20 @@ def test_proxies_environ(monkeypatch, client_class, url, env, expected):
...
     else:
         assert transport.proxy_origin == url_to_origin(expected)
+
+
+@pytest.mark.parametrize(
+    ["proxies", "expected_scheme"],
+    [
+        ({"http": "http://127.0.0.1"}, ["http://"]),
+        ({"https": "http://127.0.0.1"}, ["https://"]),
+        ({"all": "http://127.0.0.1"}, ["all://"]),
```
florimondmanca, line 257:
```suggestion
        ({"http": "http://127.0.0.1"}, "http://"),
        ({"https": "http://127.0.0.1"}, "https://"),
        ({"all": "http://127.0.0.1"}, "all://"),
```

tests/client/test_proxies.py
florimondmanca: No, it was correct (I expect CI to fail now! ;-)), but `expected_scheme` was unecessarily a 1-item list. Sent a patch…

PR #1129, opened 2020-08-05
Clean up keyword argument name, using URL.join(url=...), not URL.join(relative_url=...).

tomchristie:
Another niggly bit of API cleanup, that's a hangover from past implementation details.

Now that the `URL()` class is agnostic to relative/absolute URLs, the keyword argument naming in `URL.join(relative_url=...)` isn't appropriate.

I think we can get away with a unilateral rename, given that it's not *really* a very end-user facing bit of API, and it'll be called out in the 0.14 release notes.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -208,7 +208,7 @@ class BaseClient:
         Merge a URL argument together with any 'base_url' on the client,
         to create the URL used for the outgoing request.
         """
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Superceded by #1130?

tomchristie:
No, just different issues.

PR #1117, opened 2020-08-01
Fixed base_url behavior when base_url has a path and also the URL has own relative path.

cansarigol:
Closes #846

PR #1130, opened 2020-08-05
Base URL improvements

### LINKED ISSUE #846: base_url strips relative path
Hi,
I've noticed that if I do something like:
```
http_client = https.AsyncClient(base_url="https://my-service/api/v1")
```
for further requests (e.g. `/messages`) `/api/v1` will be stripped.
Is it expected behavior?
Tested on httpx 0.11.1
Thanks!

tomchristie:
* Support no-argument invocation of `httpx.URL()`.
* Support `client.base_url = ...` as a property.
* Resolve `base_url` joining behaviour.

So, an alternative implementation that closes #846, prompted by work in #1117.

The implementation here handles this by...

* Always ensuring that `base_url` includes a trailing slash on the path. Eg. `https://www.example.com/subpath/`.
* Always stripping any leading `/` when joining relative URLs.

This ends up given the behaviours you'd naturally expect from `base_url` usage, and also supports dotted paths, eg `../foo/bar`

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -66,13 +66,10 @@ class BaseClient:
         cookies: CookieTypes = None,
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
-        base_url: URLTypes = None,
+        base_url: URLTypes = "",
         trust_env: bool = True,
     ):
-        if base_url is None:
-            self.base_url = URL("")
-        else:
-            self.base_url = URL(base_url)
+        self.base_url = URL(base_url)
 
         self.auth = auth
         self._params = QueryParams(params)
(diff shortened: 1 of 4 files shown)


httpx/_client.py
```diff
@@ -107,6 +110,17 @@ def _get_proxy_map(
             proxy = Proxy(url=proxies) if isinstance(proxies, (str, URL)) else proxies
             return {"all": proxy}
 
+    @property
+    def base_url(self) -> URL:
+        """
+        Base URL to use when sending requests with relative URLs.
+        """
+        return self._base_url
```
StephenBrown2, line 118: Where is `self._base_url` set/created? I see `self.base_url` on [Line 72](https://github.com/encode/httpx/pull/1130/files#diff-eff812167840d2389ea5f0d809b30c32R72), perhaps that should be privatized?

tomchristie: Thanks yup. Should have been `self._base_url` in `__init__`.
It was still working (because there's a property setter) but it's neater now.

florimondmanca:
@tomchristie Should we pull the docs added in https://github.com/encode/httpx/pull/1117 here as well?

tomchristie:
Yes, but I'm not 100% about the wording. Also I figure this is a bugfix either way, so let's get it in, and treat extra docs here as a nice-to-have.

Issue #984, opened 2020-05-22
More advanced Transport API example.


tomchristie:
Now that we've got a *really* tightly spec'ed Transport API, one useful thing we can do for the ecosystem is show our users how to do nifty things like adding in extra behaviour.

For example, here's a simple implementation of a throttle...

```python
class ThrottleTransport:
    """
    A custom httpx transport that adds some basic rate-throttling functionality.

    transport = ThrottleTransport('100/minute', **connection_pool_options)
    client = httpx.Client(transport=transport)
    """

    def __init__(self, throttle, **kwargs):
        self.pool = httpcore.SyncConnectionPool(**kwargs)
        self.history = []
        
        # Parse the thottle, which should be a string, like '100/minute'.
        count, _, duration = throttle.partition('/')
        self.max_in_history = int(count)
        self.cutoff = {'second': 1.0, 'minute': 60.0, 'hour': 3600.0}[duration]

    def request(method, url, headers, stream, timeout):
        now = time.time()

        while len(self.history) >= self.max_in_history:
            expiry = now - self.cutoff

            # Expiry old entries in the history.
            self.history = [timestamp for timestamp in self.history if timestamp > expiry]
            
            # Sleep for a bit if we've exceeded the throttle rate.
            if len(self.history) >= self.max_in_history:
                time.sleep(0.1)
                now = time.time()

        self.history.append(now)
        return self.pool.request(method, url, headers, stream, timeout)
    
    def close(self):
        self.pool.close()
```

I've not tested this out, just looks about right to me. Might well be worth making sure I've got this right, and then adding it as an example in the docs?

Slightly dependant on #983, since *really* users need that in order to configure the `httpcore` pool nicely.

florimondmanca:
I'm thinking it might also be possible to build features like retries (#784) and others using this sort of style.

I'm still not certain these shouldn't belong to some sort of "transport wrapper" concept…

E.g. throttling (or retries, or any sort of "modify the request/response flow" thing) should be fairly independent of how the actual request is made - you might _also_ want to throttle requests to ASGI apps, over UDS, to a proxy, etc.

(And… this is when I come back again with my "middleware" idea? 😅 #345)

lazka:
> I'm thinking it might also be possible to build features like retries (#784) and others using this sort of style.

Transparent caching would also be nice imho, using HTTP cache headers and connecting it to something like https://github.com/argaen/aiocache

PR #1133, opened 2020-08-05
Middleware API

florimondmanca:
And here I am pitching the idea of a middleware API again… :-)

As I mentioned in https://github.com/encode/httpx/issues/984#issuecomment-633234552, and more recently in https://github.com/encode/httpx/pull/1110#issuecomment-669450028, there is a variety of use cases that don't fall in the "make a custom transport" domain, the "subclass the client without using private API" domain. In general this is transport-agnostic feature that intercepts the request/response cycle, for example...

- Caching: look up the request key in a cache, and optionally return early with a response w/o hitting the transport.
- Throttling: defer the sending of the request until a throttle policy would be satisfied.
- HSTS (?): modify the request to use HTTPS if the requested website is on the HSTS Preload List.

After playing with ideas, this PR is all we would need to support a middleware API that would support the above use cases and, I believe, many more.

Core ideas are…

- Requests are passed through a stack of middleware. The processing stack is:

```console
# Process request...
middleware_1
    ... middleware_n
        send_handling_redirects() 
        # <Ultimately we dispatch to the transport>
        # Process the response...
    ... middleware_n
middleware_1
```

-

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -32,6 +32,7 @@ from ._exceptions import (
     WriteError,
     WriteTimeout,
 )
+from ._middleware import AsyncMiddleware, Middleware
 from ._models import URL, Cookies, Headers, QueryParams, Request, Response
 from ._status_codes import StatusCode, codes
 from ._transports.asgi import ASGITransport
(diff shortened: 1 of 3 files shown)


florimondmanca:
🤔 -> #1134

Issue #1066, opened 2020-07-17
Public API tightening


tomchristie:
Following on from #1065

There's a few attributes that are set in `BaseClient.__init__` that look like contenders here for moving into the private API space. Here's what's remaining there...

- [x] `self.base_url` - Potentially modify this to a property setter / getter style, but not 100% obvious?
- [ ] `self.auth` - Potentially modify this to a property setter / getter style, but not 100% obvious? - _See #1185_
- [x] `self.timeout` - Looks like it should *definitely* move to a property getter / setter style, as per params/headers/cookies.
- [x] `self.max_redirects` - I think this one is okay as a public attribute, since it's just a plain int.
- [x] `self.trust_env` - Setting this after `__init__` isn't really valid (eg. the transport is already initialized) so we could move it to a read-only property? _See #1112_
- [x] `self.netrc` - Ought to be private. _See #1071_

tomchristie:
Dropping the milestone on this from 0.14 to 1.0, since the remaining items here are pretty minor, but we should still track them.

PR #1135, opened 2020-08-06
Add `Client.timeout` property.

cdeler:
The PR is for #1066

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -75,7 +75,7 @@ class BaseClient:
         self._params = QueryParams(params)
         self._headers = Headers(headers)
         self._cookies = Cookies(cookies)
-        self.timeout = Timeout(timeout)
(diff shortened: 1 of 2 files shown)


tomchristie:
Ah, thanks good stuff.
So I'd issue separate pull requests for auth and for timeouts here, because...

* It think the Timeout PR is a clear-cut "yes please!"
* It's not 100% clear to me if we want `client.auth` to return `Optional[AuthTypes]` or if it should instead always return the *auth class instance*. This one probably needs some explaining / discussion.

Perhaps have this PR but *just* tackling the timeouts, and we can think about auth seperately.

cdeler:
> It's not 100% clear to me if we want client.auth to return Optional[AuthTypes] or if it should instead always return the auth class instance. This one probably needs some explaining / discussion.
yes, I'll re-edit the PR

PR #1138, opened 2020-08-06
Document exceptions

### LINKED ISSUE #1106: Document the exception hierarchy
We've put a stack of great work into this for 0.14 - let's also show it off.

tomchristie:
Closes #1106

Screenshot 2020-08-06 at 13 38 09

Etc...


tomchristie:
Note: Once https://github.com/encode/httpx/pull/1126 is in, there are 3 exception classes that we can uncomment in the docs here, so that they get pulled into the autodocs.

PR #1148, opened 2020-08-08
Fix typo in 0.14.0 changelog

xrmx:


xrmx:
As a non native, "in favour of" reads like the opposite meaning of "instead
of" :)

Il sab 8 ago 2020, 09:38 Tom Christie <notifications@github.com> ha scritto:

tomchristie:
Yeah, I can see how that might be. The plainer style is nicer. Thank you.

PR #1152, opened 2020-08-09
Expand client docstrings

### LINKED ISSUE #397: Improve the API Reference
From the README: "The [Developer Interface](https://www.encode.io/httpx/api/) provides a comprehensive API reference.".

I'll be blunt: I'm not even sure its useful in its current form, let alone comprehensive. The API reference is always my pathway to learning a library. Much knowledge was just inherited/assumed from ``requests``. Most of the rest I've gone to the code for. To be fair, this isn't a major hassle for me; I like code diving. But all users aren't like me. Prose docs are good for some things and up to a point. But many times, what someone needs is best handled by a good API reference.

I love mkdocs. I much prefer it for most parts of documentation to Sphinx. And I know mkdocs2 may become a thing in the future, but that's too far out.

I know its been discussed at least a bit in the 1.0 blockers issue, but I think the API reference/autodoc story is a much bigger deal than that. So that's my reasoning for opening this issue.

So, my question is... are there any suggestions/constraints on a possible way forward on httpx's API reference/autodoc story in the near future? I haven't yet played with [pydoc-markdown](https://github.com/NiklasRosenstein/pydoc-markdown), but that may be one reasonable way.

felix-hilden:
Hi, from a short discussion in Gitter, I expanded some docstrings in client methods. This is perhaps continuation to the closed #397.

I decided to offer a minimal change, as I've not really had the chance of asking your opinion on anything. I tried to stay close to the style you already have in place. That being said, I have some questions.
- Should the verb docstrings refer to the documentation of the corresponding verb in parameters instead of `request`, or should they contain full duplicate parameter lists? Or would you rather have no documentation in them?
- Should the docstrings be in root form (imperative? sorry, I'm not a native English speaker) or not? HTTPX uses both, I decided on root, which I do prefer, i.e. "send a request" instead of "sends".

Some additional points that I came across. Perhaps they should be dealt with in separate issues. Feel free to simply reject them, as some are opinionated.
- Many docstrings that have a timeout parameter seem to cut oddly like `..., timeout=)`, most likely because of the `UNSET` value. But it looks odd.
- Stream-related things are still undocumented, even the top level `httpx.stream()`, but I can't claim to know enough of them quite yet
- Some client methods could be converted to static, as in they have an unused `self`
- Parameter lists are a bit off-putting, mainly because of the asterisks. Maybe this is because of the way mkdocs needs to handle them, but still, `help(func)` isn't exactly the easiest to read. But this is a minor inconvenience.
- `help` displays references to objects as they are in the file tree, e.g. `transport: httpcore._sync.base.SyncHTTPTransport = None` and some default values as references, e.g. `<httpx._config.UnsetType object at 0x0000028FDB590748>`. Two things could be done to improve the situation: "relocate" objects as they are imported, and define proper `repr`s for default values. By relocating I mean editing the `module` attribute of objects, as [here](https://github.com/felix-hilden/tekore/blob/master/tekore/__init__.py#L120). They are accessed from the top namespace after all.

Thank you for your consideration, and for the awesome package!

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -206,6 +206,14 @@ class BaseClient:
     ) -> Request:
         """
         Build and return a request instance.
+
+        * The `params`, `headers` and `cookies` arguments
+        are merged with any values set on the client.
+        * The `url` argument is merged with any `base_url` set on the client.
+
+        See also: [Request instances][0]
+
+        [0]: /advanced/#request-instances
         """
         url = self._merge_url(url)
         headers = self._merge_headers(headers)
@@ -596,6 +604,22 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Build and send a request.
+
+        Roughly equivalent to:
+
+        ```python
+        request = client.build_request(...)
+        response = client.send(request, ...)
+        ```
+
+        See `Client.build_request()`, `Client.send()` and
+        [Merging of configuration][0] for how the various parameters
+        are merged with client-level configuration.
+
+        [0]: /advanced/#merging-of-configuration
+        """
         request = self.build_request(
             method=method,
             url=url,
@@ -619,6 +643,19 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a request.
+
+        The request is sent as-is, unmodified.
+
+        Typically you'll want to build one with `Client.build_request()`
+        so that any client-level configuration is merged into the request,
+        but passing an explicit `httpx.Request()` is supported as well.
+
+        See also: [Request instances][0]
+
+        [0]: /advanced/#request-instances
+        """
         timeout = self.timeout if isinstance(timeout, UnsetType) else Timeout(timeout)
 
         auth = self._build_auth(request, auth)
@@ -752,6 +789,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `GET` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "GET",
             url,
@@ -774,6 +816,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send an `OPTIONS` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "OPTIONS",
             url,
@@ -796,6 +843,11 @@ class Client(BaseClient):
         allow_redirects: bool = False,  # NOTE: Differs to usual default.
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `HEAD` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "HEAD",
             url,
@@ -821,6 +873,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `POST` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "POST",
             url,
@@ -849,6 +906,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `PUT` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "PUT",
             url,
@@ -877,6 +939,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `PATCH` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "PATCH",
             url,
@@ -902,6 +969,11 @@ class Client(BaseClient):
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a `DELETE` request.
+
+        **Parameters**: See `httpx.request`.
+        """
         return self.request(
             "DELETE",
             url,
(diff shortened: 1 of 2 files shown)


felix-hilden:
Based on your comment in Gitter, @florimondmanca, should I then separate the verb docstrings from `Client.close` and `AsyncClient.aclose`? I think it's a quite manageable diff, but I want to go by your conventions here.

httpx/_client.py
```diff
@@ -1154,6 +1209,11 @@ async def send(
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Send a pre-built request.
+
+        Does not merge parameters with the client.
+        """
```
florimondmanca, line 1216:
```suggestion
        """
        Send a request.

        The request is sent as-is, unmodified.

        Typically you'll want to get it from `Client.build_request()` (so that any
        client-level configuration is merged into the request), but passing
        an explicit `httpx.Request()` is supported as well.

        See also: [Request instances][0]

        [0]: /advanced/#request-instances
        """
```

httpx/_client.py
```diff
@@ -1130,6 +1180,11 @@ async def request(
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Build and send a request.
+
+        Merges parameters with the client.
+        """
```
florimondmanca, line 1187:
```suggestion
        """
        Build and send a request.

        Roughly equivalent to:

        ```python
        response = client.send(request, ...)
        ```

        See `Client.build_request()`, `Client.send()` and [Merging of configuration][0]
        for how the various parameters are merged with client-level configuration.

        """
```

florimondmanca:
Also addressing your questions…

> Many docstrings that have a timeout parameter seem to cut oddly like ..., timeout=), most likely because of the UNSET value. But it looks odd.

Yes indeed, I think it's a bug in [mkautodoc](https://github.com/tomchristie/mkautodoc). (@tomchristie - Should we move that repo to `encode` so that anyone within the Maintainers team can work on the repo? I remember I have an old "add support for `async`" change that's still in the release pipeline. :-))

> Stream-related things are still undocumented, even the top level `httpx.stream()`, but I can't claim to know enough of them quite yet

Yes, agreed we could include some usage notes in these docstrings as well (top-level and client-level). Separately?

> Some client methods could be converted to static, as in they have an unused self

Maybe - which ones are these exactly? In any case let's treat any such changes separately as well.

> Parameter lists are a bit off-putting, mainly because of the asterisks. Maybe this is because of the way mkdocs needs to handle them, but still, help(func) isn't exactly the easiest to read. But this is a minor inconvenience.

Not sure I understand this one. Are you referring to the `*` (kwarg-only delimiter)? I also agree that eg `help(httpx.Client.get)` isn't super easy to read due to the type annotations, but what can we do…

```python
>>> import httpx
>>> help(httpx.Client.get)
Help on function get in module httpx._client:

get(self, url: Union[ForwardRef('URL'), str], *, params: Union[ForwardRef('QueryParams'), Mapping[str, Union[str, int, float, bool, NoneType, Sequence[Union[str, int, float, bool, NoneType]]]], List[Tuple[str, Union[str, int, float, bool, NoneType]]], str, NoneType] = None, headers: Union[ForwardRef('Headers'), Dict[str, str], Dict[bytes, bytes], Sequence[Tuple[str, str]], Sequence[Tuple[bytes, bytes]]] = None, cookies: Union[ForwardRef('Cookies'), http.cookiejar.CookieJar, Dict[str, str]] = None, auth: Union[Tuple[Union[str, bytes], Union[str, bytes]], Callable[[ForwardRef('Request')], ForwardRef('Request')], ForwardRef('Auth')] = None, allow_redirects: bool = True, timeout: Union[float, NoneType, Tuple[Union[float, NoneType], Union[float, NoneType], Union[float, NoneType], Union[float, NoneType]], ForwardRef('Timeout'), httpx._config.UnsetType] = <httpx._config.UnsetType object at 0x10414bb20>) -> httpx._models.Response
```

> help displays references to objects as they are in the file tree, e.g. transport: httpcore._sync.base.SyncHTTPTransport = None and some default values as references, e.g. <httpx._config.UnsetType object at 0x0000028FDB590748>. Two things could be done to improve the situation: "relocate" objects as they are imported, and define proper reprs for default values. By relocating I mean editing the `__module__` attribute of objects, as here. They are accessed from the top namespace after all.

Would probably result in another big list of references that must be kept in sync in `__init__.py` :-) but interesting!

Note that some of the types that appear in there (such as `UnsetType`) are effectively private, so those shouldn't be _relocated_. (In essence, only public API items, i.e. those listed in `httpx/__init__.py`, are allowed to be relocated.)

felix-hilden:
Thanks!

I agree with handling streaming and the static methods in another PR. My bad, I wasn't clear in my wording: I meant the parameter lists in documentation. Due to bolding and lists there are a ton of asterisks when compared to, say, numpy-style documentation:

```
Parameters
----------
auth
    optional, an authentication class to use when sending requests
params
    optional, query parameters to include in request URLs,
    as a string, dictionary or list of two-tuples
...
```

But that may be a necessary evil, I'm not the one to tell you.

Yeah, I agree that modifying the module attribute results in a maintenance overhead. It's a matter of how you guys see the cost-benefit. Of course with maturity and stability comes less changes in the public API, but that's one more place to forget to change things in. And yeah, I'm not entirely sure what would be the appropriate thing to do with private types, maybe leave them as they are, as you said. I think there's an argument to be had whether it should be private given that it is visible. But I don't know exactly how you need to use the `UnsetType`, so I can't say much.

I'll make the requested changes to this PR, open another one for static methods, and an issue for streaming things. Maybe the relocation thing needs to be discussed further before you can make a decision.

httpx/_client.py
felix-hilden: Added this with a minor rewording to the new version!

httpx/_client.py
felix-hilden: Added this to the new version!

felix-hilden:
I made the requested modifications to `Client` and `AsyncClient`, these lengthier versions do indeed look a lot nicer to read! I think I have nothing to add really. Is it good to go?

httpx/_client.py
```diff
@@ -1130,6 +1205,22 @@ async def request(
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
     ) -> Response:
+        """
+        Build and send a request.
+
+        Roughly equivalent to:
```
tomchristie, line 1211: Should this just be "Equivalent to:"?

felix-hilden: Fair point, because currently *it is*. But do you want to have some leeway in case it stops being, for some reason?

felix-hilden:
LGTM, if my assumption is correct that you, Tom, have some sort of final say here. Thanks! You all made it easy to contribute and communicate 👍

florimondmanca:
@felix-hilden FYI, Tom is the one that's most often around here, but everyone from the `@encode/maintainers` team can review, accept and merge PRs. :-)

Issue #1151, opened 2020-08-08
Response cookies are not ingested properly in 0.14.0


djigit0s:
Hi. After last updatet when loading cookies from a file, it gives an error

**http/cookiejar.py", line 1947, in _really_load
domain_specified = domain.startswith(".")
AttributeError: 'NoneType' object has no attribute 'startswith'**

```python
import os
from http.cookiejar import LWPCookieJar
import httpx

h = httpx.Client(cookies=LWPCookieJar('test'))
if os.path.isfile('test'):
    h.cookies.jar.load()
h.get('https://www.google.ru')
h.cookies.jar.save()
```

**Cookies file - httpx 0.13.3**
#LWP-Cookies-2.0
Set-Cookie3: 1P_JAR="2020-08-08-21"; path="/"; domain=".google.ru"; path_spec; domain_dot; secure; expires="2020-09-07 21:47:19Z"; version=0
Set-Cookie3: NID="204=CQ898UjmES*****************************ctxJAE99ngLlZcojaaWLrqex-td7X00PooPgmVnW61VrtwCHIPSb_CPR8Ic6dpD_gRY_pBteV_4YSGU7Vu58V7uHPwNcPxZ6PMCAHt-RZSPgnxcoM"; path="/"; domain=".google.ru"; path_spec; domain_dot; expires="2021-02-07 21:47:19Z"; HttpOnly=None; version=0

**Cookies file - httpx 0.14.0**
#LWP-Cookies-2.0
Set-Cookie3: 1P_JAR="2020-08-08-21"; path="/"; domain=".google.ru"; path_spec; domain_dot; expires="2020-09-07 21:50:09Z"; HttpOnly=None; Secure, NID="204=NCk0**********************xnr8aaHwCrC-zMdQQ9bAS5QAp6rfv-yOiuwkYDhiN7kTyDASjlZVIJ0Zase-JQMH_YwCvaiS9I232kIRu36O684kwHD_nRXHeUVpTJTlS7mbniFmebf5nhE"; version=0

tomchristie:
Heya, first things...

1. Could you provide the full traceback?
2. Could you double check and confirm the exact behaviour on 0.13?
3. I don't recall us making any changes in our cookie handling? Both versions are tagged on GitHub so you could review if there are any changes?

djigit0s:
**1. Traceback, script runned on version 0.14.0**
**(manual demand for payment) djigit0s@R:/mnt/d/prj/manual demand for payment$ python test.py
[...]

PR #1153, opened 2020-08-09
response cookies fix

djigit0s:
Fix response cookies in 0.14.0. Didn't return all headers. This error is described here #1151 

I apologize if I did not correctly formulate, this is my first PR


tomchristie:
Thanks for having a go at this! Your PR is absolutely fine, but I've drafted up another so that we've got a test case to prevent regressions here... https://github.com/encode/httpx/pull/1156

Really appreciate you raising this as an issue.

PR #1155, opened 2020-08-10
Set `__module__ = 'httpx'` on everything we expose via the public API.

tomchristie:
Set `__module__ = 'httpx'` on all our public API.

Primarily useful for classes that haven't declared a `__repr__` method, but might also improve the output of other code-introspection tooling?

Before:

```python
>>> import httpx
>>> httpx.DigestAuth(username='', password='')
<httpx._auth.DigestAuth object at 0x106682850>
```

After:

```python
>>> import httpx
>>> httpx.DigestAuth(username='', password='')
<httpx.DigestAuth object at 0x103529650>
```

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -114,4 +114,4 @@ __all__ = [
 _locals = locals()
 for name in __all__:
     if not name.startswith("__"):
-        setattr(_locals[name], "__module__", "httpx"). # noqa
+        setattr(_locals[name], "__module__", "httpx")  # noqa
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -109,3 +109,9 @@ __all__ = [
     "DigestAuth",
     "WSGITransport",
 ]
+
+
+_locals = locals()
+for name in __all__:
+    if not name.startswith('__'):
+        setattr(_locals[name], '__module__', 'httpx')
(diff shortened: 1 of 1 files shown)


httpx/__init__.py
```diff
@@ -109,3 +109,9 @@
     "DigestAuth",
     "WSGITransport",
 ]
+
+
+_locals = locals()
+for name in __all__:
+    if not name.startswith("__"):
+        setattr(_locals[name], "__module__", "httpx")
```
florimondmanca, line 117:
```suggestion
        setattr(_locals[name], "__module__", "httpx"). # noqa
```

httpx/__init__.py
```diff
@@ -109,3 +109,9 @@
     "DigestAuth",
     "WSGITransport",
 ]
+
+
+_locals = locals()
+for name in __all__:
+    if not name.startswith("__"):
+        setattr(_locals[name], "__module__", "httpx"). # noqa
```
florimondmanca, line 117: Oops, sorry!

```suggestion
        setattr(_locals[name], "__module__", "httpx")  # noqa
```

felix-hilden:
Glad to see this was useful! And using locals is a neat way of doing it. Should a similar issue be opened in `httpcore` as well, seeing that it could apply, e.g. the example in #1152?

(Also apologies if I'm intrusive, not really sure what conversation should be on Gitter or here...)

tomchristie:
Yeah good call re. applying the same to `httpcore`. 👍

PR #1160, opened 2020-08-10
Remove all unused self params and convert to staticmethods

felix-hilden:
From the discussion in #1152, this PR converts all methods with an unused `self` to staticmethods.

This was not exactly agreed upon, so feel no obligations towards merging this if it's not to your taste. There is a case to be made for ignoring insignificant changes to reduce noise and keep history, but I do feel that small details add up. Anyway, thanks for having a look!

### DIFF
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -75,8 +75,9 @@ class BasicAuth(Auth):
         request.headers["Authorization"] = self._auth_header
         yield request
 
+    @staticmethod
     def _build_auth_header(
-        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+        username: typing.Union[str, bytes], password: typing.Union[str, bytes]
     ) -> str:
         userpass = b":".join((to_bytes(username), to_bytes(password)))
         token = b64encode(userpass).decode()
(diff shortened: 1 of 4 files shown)


tomchristie:
I'm a bit lukewarm towards doing this throughout. I get that we *could*, and that the methods would be more tightly scoped if we did, but also it's not terrible that Python code often uses standard methods as a name-spacing, even when they are not necessarily also state changing.

Having said that, I've not taken a proper review over it, so it's possible that it makes good sense in places.

felix-hilden:
I see! Lukewarm's no good, is it 😄 I have not considered that one could want an unused self, but that's probably my inexperience, as I don't quite understand what it means to use standard methods as a name-spacing. Nevertheless, this is here for you to accept or reject. I'd be happy to make any changes to include only a subset in suitable places if you find it appropriate!

florimondmanca:
I guess my take on this is that all these methods are `@staticmethod`-convertible mostly _by accident_. If we did need to access class state we'd need to convert them back again to standard methods. "Practicality beats purity", I guess. So while I appreciate you proposing these changes, I'm also a bit "_meh_" on going forward with these. 😅

felix-hilden:
Fair enough!

PR #1163, opened 2020-08-11
Map rfc3986 exceptions

### LINKED ISSUE #1158: Map `rfc3986.exceptions.InvalidAuthority` to `httpx.InvalidURL`
We're wrapping our underlying `h11`/`h2` library exception classes, but we're currently missing an `rfc3986` case...

We should reintroduce the `InvalidURL` exception for URLs that really are improperly formed and *cannot* be parsed (not our previous case of "Invalid because it's not http/https").

The only case I can see that we need to map is `rfc3986.exceptions.InvalidAuthority`, although perhaps I'm missing some other cases that `rfc3986` can raise? We ought to wrap our single `rfc3986.iri_reference(...)` usage, and map the `rfc3986` exception to our own exception class.

```python
>>> httpx.URL('https://😇/')
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/rfc3986/iri.py", line 125, in idna_encoder
    std3_rules=True)
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/idna/core.py", line 360, in encode
    s = alabel(label)
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/idna/core.py", line 281, in alabel
    check_label(label)
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/idna/core.py", line 261, in check_label
    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
idna.core.InvalidCodepoint: Codepoint U+1F607 at position 3 of 'ðð😇' not allowed

  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/httpx/_models.py", line 69, in __init__
    self._uri_reference = rfc3986.api.iri_reference(url).encode()
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/rfc3986/iri.py", line 133, in encode
    for part in self.host.split(".")])
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/rfc3986/iri.py", line 133, in <listcomp>
    for part in self.host.split(".")])
  File "/Users/tomchristie/Temp/venv/lib/python3.7/site-packages/rfc3986/iri.py", line 127, in idna_encoder
    raise exceptions.InvalidAuthority(self.authority)
rfc3986.exceptions.InvalidAuthority: The authority (ðð😇) is not valid.
```

So, let's change from...

```python
self._uri_reference = rfc3986.iri_reference(url).encode()
```

To...

```python
try:
    self._uri_reference = rfc3986.iri_reference(url).encode()
except rfc3986.exceptions.InvalidAuthority as exc:
    raise httpx.InvalidURL(message=str(exc)) from None
```

We *don't* want this to subclass `RequestException`, because it *isn't* necessarily. The user might be using a base `httpx.URL()`.

We then possibly also want to wrap the `url = URL(location)` usage in `_redirect_url` to wrap `httpx.InvalidURL` and instead raise `httpx.RemoteProtocolError("Invalid URL in location header: {message}")`, since if a parse error occurs at that point, then it *is* a request error. (We *do* have an associated request at this point, and it is due to a remote server error, rather than invalid user input.)

I'll label this as "refactor" since it's such a minor edge bit of API.

j178:
Closes #1158

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -347,7 +347,7 @@ class BaseClient:
         except InvalidURL as exc:
             raise RemoteProtocolError(
                 f"Invalid URL in location header: {exc}.", request=request
-            )
+            ) from None
 
         # Handle malformed 'Location' headers that are "absolute" form, have no host.
         # See: https://github.com/encode/httpx/issues/771
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -21,6 +21,8 @@ from ._exceptions import (
     HTTPCORE_EXC_MAP,
     RequestBodyUnavailable,
     TooManyRedirects,
+    InvalidURL,
+    RemoteProtocolError,
     map_exceptions,
 )
 from ._models import URL, Cookies, Headers, QueryParams, Request, Response
@@ -340,7 +342,10 @@ class BaseClient:
         """
         location = response.headers["Location"]
 
-        url = URL(location)
+        try:
+            url = URL(location)
+        except InvalidURL as exc:
+            raise RemoteProtocolError(f"Invalid URL in location header: {exc}.", request=request)
 
         # Handle malformed 'Location' headers that are "absolute" form, have no host.
         # See: https://github.com/encode/httpx/issues/771
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -19,10 +19,10 @@ from ._config import (
 from ._content_streams import ContentStream
 from ._exceptions import (
     HTTPCORE_EXC_MAP,
-    RequestBodyUnavailable,
-    TooManyRedirects,
     InvalidURL,
     RemoteProtocolError,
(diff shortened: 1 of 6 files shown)


httpx/_client.py
```diff
@@ -340,7 +342,12 @@ def _redirect_url(self, request: Request, response: Response) -> URL:
         """
         location = response.headers["Location"]
 
-        url = URL(location)
+        try:
+            url = URL(location)
+        except InvalidURL as exc:
+            raise RemoteProtocolError(
+                f"Invalid URL in location header: {exc}.", request=request
+            )
```
tomchristie, line 350: Let's use `from None` here - we don't need to expose the extra traceback info.

PR #1166, opened 2020-08-11
Include invalid url exception in docs

tomchristie:
Minor fix here.
Realised we didn't include this along with #1163


docs/exceptions.md
```diff
@@ -76,6 +76,7 @@ except httpx.HTTPStatusError as exc:
         * TooManyRedirects
         * RequestBodyUnavailable
     * HTTPStatusError
+* IvalidURL
```
florimondmanca, line 79:
```suggestion
* InvalidURL
```

tomchristie:
> There's an extra commit that belongs to #1165 here

😬

PR #1177, opened 2020-08-14
Make `name` in httpx.__init__ private

j178:
Prompted by https://github.com/encode/httpcore/pull/156/commits/35ecd2274faf1fdd4a6218bc07839d96b9791d7b

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -112,6 +112,6 @@ __all__ = [
 
 
 _locals = locals()
-for name in __all__:
-    if not name.startswith("__"):
-        setattr(_locals[name], "__module__", "httpx")  # noqa
+for _name in __all__:
+    if not _name.startswith("__"):
+        setattr(_locals[_name], "__module__", "httpx")  # noqa


jcugat:
Would be worth bringing [this test](https://github.com/encode/httpcore/pull/156/files#diff-b742918391873180aa8283b7dd4885d2) to httpx, since it's the one that surfaced the issue. Also, I'd use private variables `_*` instead of using `del`.

j178:
@jcugat Thanks for reminding me! I added a test and reorder members in `__all__`.


> Also, I'd use private variables `_*` instead of using `del`.

I think these are just temporary variables, they will not be used anywhere, I prefer `del`ing them from the namespace.

j178:
Updated.

jcugat:
@tomchristie The test was added because one of the imports was not present inside `__all__`, and it's easy to forget to keep them synced. I don't understand what extra maintenance work that test is adding.

tomchristie:
You're right sorry. I was taking a brief look on a phone screen, and it seemed a bit superfluous  But it's actually reasonable enough.

PR #1182, opened 2020-08-15
Drop urllib3 in favor of public gist

### LINKED ISSUE #1143: Move URLLib3Transport out of core and into a stand-alone package.
We should remove `URLLib3Transport` from `httpx`, and make it a stand-alone package, with a simple README documenting how to install and use it, with `transport=...`.

In fact, we *could* even just drop it from core, and *aim* at delivering a third party package but *not* treat having to do so as a 1.0 release blocker. (I mean, is anyone *actually* using it at the moment?)

florimondmanca:
Fixes #1143 

This PR removes the `URLLib3Transport` from HTTPX, and moves it to a public gist that we're now pointing to in the "Custom transports" docs.

Also adds a small additional section to the Requests compatibility guide, so that people know this option exists if they're migrating from Requests to HTTPX.


tomchristie:
Fantastic! Let's hold off on pulling in API changes until we're ready to roll a major release. We *could* consider maintaining a release branch?

tomchristie:
The only other tiny tweak I can think of here in the gist would be changing from this:

```python
        except (urllib3.exceptions.SSLError, socket.error) as exc:
            tb = exc.__traceback__
            raise httpcore.NetworkError(exc).with_traceback(tb) from None
```

To this:

```python
        except (urllib3.exceptions.SSLError, socket.error) as exc:
            raise httpcore.NetworkError(exc) from exc
```

Which I think gets us the traceback behaviour we actually.

https://gist.github.com/florimondmanca/d56764d78d748eb9f73165da388e546e#file-urllib3_transport-py-L101

tomchristie (approved):
Okay, it's got a `do-not-merge` label on, which makes sense, but otherwise yup this is *ace*.
So, approved.

Do we think we ought to be maintaining a version branch so that we can pull this in, and then maintain the version branch until we're ready to go?

tomchristie:
In the gist, let's just use `raise httpcore.NetworkError(exc) from exc` styles, instead of `.with_traceback(tb)`, in these two cases:

* https://gist.github.com/florimondmanca/d56764d78d748eb9f73165da388e546e#file-urllib3_transport-py-L18-L19
* https://gist.github.com/florimondmanca/d56764d78d748eb9f73165da388e546e#file-urllib3_transport-py-L100-L101

Once that's tweaked let's get this merged!

docs/compatibility.md
```diff
@@ -83,3 +83,11 @@ Besides, `httpx.Request()` does not support the `auth`, `timeout`, `allow_redire
...
 
 If you need to mock HTTPX the same way that test utilities like `responses` and `requests-mock` does for `requests`, see [RESPX](https://github.com/lundberg/respx).
+
+## Networking layer
+
+`requests` defers most of its HTTP networking code to the excellent [`urllib3` library](https://urllib3.readthedocs.io/en/latest/).
+
+On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as its core HTTP networking layer, which is a different project than `urllib3`.
+
+If you're transitioning from Requests to HTTPX, you might want to consider using HTTPX with an [`urllib3` custom transport](/advanced#urllib3-transport) at first, so that `urllib3` is retained as the underlying HTTP networking implementation, allowing you to deal with any differences in behavior as a second phase.
```
tomchristie, line 93: I don't think we should be recommending this.

I know it might have been mentioned at some point in the project's past, but it's unlikely to be a sensible route for many folks to follow at this point.

tomchristie:
I'd suggest we also include a link in the "third party packages" docs.

florimondmanca:
>  let's just use raise httpcore.NetworkError(exc) from exc styles, instead of .with_traceback(tb), in these two cases

✅

> I'd suggest we also include a link in the "third party packages" docs.

✅

Also dropped the recommendation to use `urllib3-transport` while migrating — I reckon that's not something even I might not actually do if I were to migrate a Requests-based project to use HTTPX.

docs/third-party-packages.md
```diff
@@ -23,3 +23,13 @@ An asynchronous GitHub API library. Includes [HTTPX support](https://gidgethub.r
 [GitHub](https://github.com/lundberg/respx) - [Documentation](https://lundberg.github.io/respx/)
 
 A utility for mocking out the Python HTTPX library.
+
+## Gists
+
+<!-- NOTE: this list is in alphabetical order. -->
+
+### urllib3-transport
```
florimondmanca, line 31: Not too keen to put this under "Plugins", since it's not really an installable package, but then also not keen on the existing "Plugins" section — perhaps we ought to just list everything as h2-level headers. We can address this as a follow-up.

tomchristie:
Fantastic. 🌟

PR #1183, opened 2020-08-16
Fix #1146 Set default allow_redirects to true

### LINKED ISSUE #1146: `allow_redirects` with `httpx.head(...)`
We've currently got a different `allow_redirects` default on HEAD requests to all other request methods.

https://github.com/encode/httpx/blob/a25d924bb910e85ca597c7b4ed819c6e9e17aecd/httpx/_api.py#L217

This is in line with behaviour in `requests`, but I'm not convinced it's something we ought to do. It'd seem to me that we'd do better just to stay consistent rather than have this odd exclusion. I get why it *could* be a reasonable choice, but I think it'd *probably* be better to just stay consistent, and allow the user to switch `allow_redirects=False` if that's their intent.

Flagging this to be addressed pre-1.0.

dongskyler:
# Fix #1146

This PR sets default `allow_redirects` to `True`. It was `False` before this PR.

This will align our behavior with most other http request libraries, except for [Requests](https://requests.readthedocs.io/en/v0.8.4/api/#requests.request).

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -214,7 +214,7 @@ def head(
     headers: HeaderTypes = None,
     cookies: CookieTypes = None,
     auth: AuthTypes = None,
-    allow_redirects: bool = True,  # Note: Aligns with usual default (except for Requests).
+    allow_redirects: bool = True,
     cert: CertTypes = None,
     verify: VerifyTypes = True,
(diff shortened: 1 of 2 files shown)


dongskyler:
Sounds great.

PR #1187, opened 2020-08-16
Docs: Use pycon for Python console code blocks

hugovk:
When code examples use the `>>>` IDLE shell, use `pycon` (short for `py`thon `con`sole) syntax instead of `python`, to fix/improve syntax highlighting.

For example:

# python

```python
>>> import httpx
>>> async with httpx.AsyncClient() as client:
>>>     r = await client.get('https://www.example.org/')
>>> r
<Response [200 OK]>
```

# pycon

```pycon
>>> import httpx
>>> async with httpx.AsyncClient() as client:
>>>     r = await client.get('https://www.example.org/')
>>> r
<Response [200 OK]>
```

Docs: https://pygments.org/docs/lexers/#lexers-for-python-and-related-languages

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ release, so that you're able to properly review [API changes between package upd
 
 Let's get started...
 
-```python
+```pycon
 >>> import httpx
 >>> r = httpx.get('https://www.example.org/')
 >>> r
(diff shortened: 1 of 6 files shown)


hugovk:
Yeah, it's useful, and not to be confused with the the *Py*thon *Con*ference :)

PR #1190, opened 2020-08-17
Map httpcore exceptions for `Response` read methods

### LINKED ISSUE #1188: httpcore.ReadTimeout not mapped to httpx.ReadTimeout
I'm running the following to perform requests and catch `httpx.RequestError` exceptions:
```
async def async_get_url(client, url, retry=True):
    async with semaphore.get():
        try:
            response = await client.get(url)
            response.raise_for_status()
            result = handle_reponse(reponse)
        except httpx.RequestError as exc:
            if retry:
                print(f"Retrying {url} ...")
                result = await async_get_url(
                    client, url, retry=False
                )
            else:
                result = ...
            result = ...
        return result
```

At some point it failed with a `httpcore._exceptions.ReadTimeout` exception:
```
  File "/home/user/main.py", line 351, in <module>
    asyncio.run(run())
  File "/usr/lib64/python3.8/asyncio/runners.py", line 43, in run
  File "/usr/lib64/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/user/main.py", line 317, in run
    await f
  File "/usr/lib64/python3.8/asyncio/tasks.py", line 608, in _wait_for_one
    return f.result()  # May raise f.exception().
  File "/home/user/main.py", line 173, in get_updates
    item_api_response = await f
  File "/usr/lib64/python3.8/asyncio/tasks.py", line 608, in _wait_for_one
    return f.result()  # May raise f.exception().
  File "/home/user/main.py", line 53, in async_get_url
    response = await client.get(url)
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_client.py", line 1292, in get
    return await self.request(
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_client.py", line 1143, in request
    response = await self.send(
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_client.py", line 1167, in send
    await response.aread()
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_models.py", line 958, in aread
    self._content = b"".join([part async for part in self.aiter_bytes()])
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_models.py", line 958, in <listcomp>
    self._content = b"".join([part async for part in self.aiter_bytes()])
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_models.py", line 969, in aiter_bytes
    async for chunk in self.aiter_raw():
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpx/_models.py", line 1002, in aiter_raw
    async for part in self._raw_stream:
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_async/connection_pool.py", line 50, in __aiter__
    async for chunk in self.stream:
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_bytestreams.py", line 72, in __aiter__
    async for chunk in self._aiterator:
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_async/http11.py", line 143, in _receive_response_data
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_async/http11.py", line 160, in _receive_event
    data = await self.socket.read(self.READ_NUM_BYTES, timeout)
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_backends/asyncio.py", line 139, in read
    return data
  File "/usr/lib64/python3.8/contextlib.py", line 131, in __exit__
  File "/home/user/.pyenv/versions/venv/lib64/python3.8/site-packages/httpcore/_exceptions.py", line 12, in map_exceptions
    raise to_exc(exc) from None
httpcore._exceptions.ReadTimeout
```
According to: https://github.com/encode/httpx/blob/8c7729e42c9ebf860e75820d5325741b85d64970/httpx/_exceptions.py#L360-L363 
The `httpcore.ReadTimeout` exception should have been mapped to `httpx.ReadTimeout` and therefore catched appropriately with `except httpx.RequestError as exc`, but that doesn't happen.

### Environment

- OS: Linux
- Python version: 3.8.5
- HTTPX version: 0.14.0
- HTTPCore version: 0.10.0
- Async environment: asyncio
- HTTP proxy: no
- Custom certificates: no

j178:
Fix #1188
Add `map_exceptions` to `Response.iter_raw` and `Response.aiter_raw`, making sure all `httpcore` exceptions are mapped.
And add a slow_stream response test to make sure `ReadTimeout` ocurred while the client is reading the response body, not in the request time.

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -33,7 +33,7 @@ def test_httpcore_exception_mapping(server) -> None:
     with pytest.raises(httpx.ConnectError):
         httpx.get("http://doesnotexist")
 
-    # Make sure Response.iter_raw() exceptinos are mapped
+    # Make sure streaming methods also map exceptions.
     url = server.url.copy_with(path="/slow_stream_response")
     timeout = httpx.Timeout(None, read=0.1)
     with httpx.stream("GET", url, timeout=timeout) as stream:
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -24,7 +24,7 @@ def test_httpcore_all_exceptions_mapped() -> None:
         pytest.fail(f"Unmapped httpcore exceptions: {not_mapped}")
 
 
-def test_httpcore_exception_mapping() -> None:
+def test_httpcore_exception_mapping(server) -> None:
     """
     HTTPCore exception mapping works as expected.
     """
@@ -32,6 +32,13 @@ def test_httpcore_exception_mapping() -> None:
     # Make sure we don't just map to `NetworkError`.
     with pytest.raises(httpx.ConnectError):
         httpx.get("http://doesnotexist")
+
+    # Make sure Response.iter_raw() exceptinos are mapped
+    url = server.url.copy_with(path="/slow_stream")
+    timeout = httpx.Timeout(None, read=1e-6)
+    with pytest.raises(httpx.ReadTimeout):
+        with httpx.stream("GET", url, timeout=timeout) as stream:
+            stream.read()
 
     # Make sure it also works with custom transports.
     class MockTransport(httpcore.SyncHTTPTransport):
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -34,10 +34,10 @@ def test_httpcore_exception_mapping(server) -> None:
         httpx.get("http://doesnotexist")
 
     # Make sure Response.iter_raw() exceptinos are mapped
-    url = server.url.copy_with(path="/slow_stream")
-    timeout = httpx.Timeout(None, read=1e-6)
-    with pytest.raises(httpx.ReadTimeout):
-        with httpx.stream("GET", url, timeout=timeout) as stream:
+    url = server.url.copy_with(path="/slow_stream_response")
+    timeout = httpx.Timeout(None, read=0.1)
+    with httpx.stream("GET", url, timeout=timeout) as stream:
+        with pytest.raises(httpx.ReadTimeout):
             stream.read()
 
     # Make sure it also works with custom transports.
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -24,6 +24,7 @@ from ._decoders import (
     TextDecoder,
 )
 from ._exceptions import (
+    HTTPCORE_EXC_MAP,
     CookieConflict,
     HTTPStatusError,
     InvalidURL,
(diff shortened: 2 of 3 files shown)


tests/test_exceptions.py
```diff
@@ -33,6 +33,13 @@ def test_httpcore_exception_mapping() -> None:
     with pytest.raises(httpx.ConnectError):
         httpx.get("http://doesnotexist")
 
+    # Make sure Response.iter_raw() exceptinos are mapped
```
florimondmanca, line 36:
```suggestion
    # Make sure streaming methods also map exceptions.
```

florimondmanca (commented):
Interesting…!

The main reason I'm not immediately "OK let's go" on this is that `ReadError` & co only have a `request=...` parameter, but in this case we _do_ have a `response` available that we could attach to the exception. Perhaps no need to sweat it since we have the `response` object available in the scope anyway when doing local exception handling, but just wanted to leave this heads up.

```python
with httpx.stream(...) as response:
    try:
        for line in response.iter_lines():
            ...
    except httpx.ReadError as exc:
        print("Stream broken", response)  # No need for `exc.response` anyway?
```

tomchristie:
@florimondmanca Well noted. There's already other exception classes that can fall into that category, which I'm totally okay with, because it's consistent & simple. As you say, the response will be available in the local scope anyways.

tomchristie (approved):

tomchristie:
There's also an alternate implementation where we wrap around the stream that we pass to the `Response` instance, here...

https://github.com/encode/httpx/blob/d10b7cdc51a332d65edec774c2386213b0ee1a01/httpx/_client.py#L795-L801

Eg. something along these lines...

```python
ResponseStream(ContentStream):
    def __int__(self, request, stream):
        self._request = request
        self._transport_stream = stream

    def __iter__(self):
        with map_exceptions(HTTPCORE_EXC_MAP, request=self._request)
            for chunk in self._transport_stream:
                yield chunk

    def close(self):
         self._transport_stream.close()
```

Then...

```python
response = Response(
   ...
   stream=ResponseStream(request, stream)
)
```

I don't see any reason for us to prefer that here & now, but mentioning it in case it's a useful reference point for anything in the future.

Issue #871, opened 2020-03-23
Add warning about unclosed AsyncClient


kamilglod:
It would be nice to have an info about some unclosed `AsyncClient()` instance. Here is an example how it is implemented in aiohttp: https://github.com/aio-libs/aiohttp/blob/693ed0cce0c67646f8df18c93e448d10782c2811/aiohttp/client.py#L268-L286

florimondmanca:
Hi @kamilglod 

Can you share what the UX for this in aiohttp looks like? (eg maybe a snippet of an interactive interpreter session…)

I'm just wondering if this would actually provide value, or just add noise when shutting down an interpreter (which should force-close any dangling resources anyway, no?).

kamilglod:
@florimondmanca closing all connections sounds fine but question is whether it is technically doable as `close()` is an async function and we cannot `await` it inside `__dell__()`. Additionally I think that such an information is very helpful in tracking buggy code with unclosed resources
Here is an example output from aiohttp: 
![Screenshot from 2020-03-25 12-40-08](https://user-images.githubusercontent.com/8168459/77532796-e470c680-6e95-11ea-88ab-bb5564c4c122.png)

florimondmanca:
Thanks, the REPL snippet is useful.

An implementation similar to aiohttp is technically possible — ie log a warning but don’t attempt to close the client (notably because, as you said, it’s an async operation).

One bit to keep in mind is that this would need to be async library agnostic - so the check needs to be deferred to the async backends at some point - so this doesn’t seem like a trivial addition. But if you’d like to give it a shot feel free to. :)

Pradhvan:
@florimondmanca If anyone is not working on it. I would like to give it a try. 😄

Pradhvan:
I was setting up the project locally and in the step of installing dependencies. Should it explicitly mention that `scripts/install` will automatically create a virtual environment?

> You can now install the project and its dependencies using:
   $ cd httpx
   $ scripts/install

Or is it just me that went ahead creating a virtual env and running the command `scripts/install` hoping it would install dependencies in the environment I created.

cdeler:
@Pradhvan do you mind if I take this?

cdeler:
@florimondmanca shall we add the similar check for `httpx.Client`?

Pradhvan:
@cdeler yes 😄

PR #1191, opened 2020-08-17
Warn a ResourceWarning when AsyncClient hasn't been closed (#871)

cdeler:
I added a warning in the `__del__` function inside the `AsyncClient` (I'm not sure if we can do more)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1,5 +1,6 @@
 import functools
 import typing
+import warnings
 from types import TracebackType
 
 import httpcore
@@ -1160,6 +1161,7 @@ class AsyncClient(BaseClient):
             for key, proxy in proxy_map.items()
         }
         self._proxies = dict(sorted(self._proxies.items()))
+        self._is_closed = False
 
     def _init_transport(
         self,
@@ -1627,6 +1629,7 @@ class AsyncClient(BaseClient):
         for proxy in self._proxies.values():
             if proxy is not None:
                 await proxy.aclose()
+        self._is_closed = True
 
     async def __aenter__(self) -> "AsyncClient":
         return self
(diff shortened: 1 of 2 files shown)


cdeler:
Reviewing the PR please check this build https://github.com/encode/httpx/pull/1191/checks?check_run_id=997511344

I don't know why it failed, after that I just touched one of the files, force-pushed my branch and the test went back to the normal.

cdeler:
I introduces a new piece of changes in https://github.com/encode/httpx/pull/1191/commits/1fdab27a9e6a6e99e8a3579aa787c42a0a2007ae

As soon as `aclose()` has set `_is_closed = True`, we cannot send any requests or call `aclose` again.

felix-hilden:
Now this PR is doing two things: warning on an unclosed `AsyncClient` and raising errors on making requests with a closed client. As Tom pointed out in Gitter, it would be nice to separate them. Additionally, I think that the call restrictions ought to apply to the synchronous `Client` as well.

A small design consideration for the other PR, but I'll mention it already because you started implementation: a simple check and error could be more easily implemented and maintained as a decorator.
```python
from typing import Callable
from functools import wraps

def check_not_closed(method: Callable):   # or a better name :D
    @wraps(method)
    def wrapper(self, *args, **kwargs):
        if self.is_closed:
            raise RuntimeError('Cannot use a closed client!')
        return method(self, *args, **kwargs)
    return wrapper

class Client:
    @check_not_closed
    def send(self, request):
        ...
```
Or if the message should change, a parameterised decorator. It's a pretty small thing to extract, but in my humble opinion, could be worth it.

cdeler:
> Now this PR is doing two things: warning on an unclosed AsyncClient and raising errors on making requests with a closed client. As Tom pointed out in Gitter, it would be nice to separate them
Sure, I was out of keyboard, now I am to do that. I'm closing the current PR

> A small design consideration for the other PR, but I'll mention it already because you started implementation: a simple check and error could be more easily implemented and maintained as a decorator.
Java side of my brain want to introduce new interface a-la "closable"

cdeler:
Close to split the PR into 2 parts

PR #1196, opened 2020-08-19
Add connect retries

### LINKED ISSUE #1141: Retries
Our yardstick for 1.0 has generally been to achieve feature parity with requests, plus the following additional support...

* Sync + Async support.
* HTTP/2 support.
* Strict timeout defaults.
* Type annotations throughtout.
* Streaming multipart uploads.

Something we've so far omitted is `retires=...`, which are *not* included in either the requests QuickStart guide, or in the Advanced Usage section. However `requests` *does* have some retry support, can be enabled by mounting a custom adapter... https://requests.readthedocs.io/en/master/api/#requests.adapters.HTTPAdapter

Screenshot 2020-08-07 at 15 34 27

I'd suggest that if we *do* want to add retry support we should start off by doing so in a limited fashion, and *only* provide a simply `retries=<int>` argument on the client, which matches the same "only retry on connection failures" behaviour that `requests` defaults to for integer retry arguments.

That doesn't *necessarily* preclude that we could consider more complex retry behaviour at some point in the future, but I'm prefer we push back on that for as long as possible.

Having more dials for configuration is something we should *generally* avoid where possible. Also, we've gone to a great deal of effort over nicely spec'ing our Transport API, and it's perfectly feasible for developers to build against that to deal with any more complex behaviours that they'd like to see.

With all that in mind, I'd suggest something we *might* consider would be...

```python
class Client(..., retries: int=0):
    ...
```

With the implementation handled inside our `_send_single_request` method. Something like...

```python
transport = self._transport_for_url(request.url)
retries = self.retries

while True:
    with map_exceptions(HTTPCORE_EXC_MAP, request=request):
        try:
            (
                http_version,
                status_code,
                reason_phrase,
                headers,
                stream,
            ) = transport.request(
                request.method.encode(),
                request.url.raw,
                headers=request.headers.raw,
                stream=request.stream,
                timeout=timeout.as_dict(),
            )
       except (httpcore.ConnectError, httpcore.ConnectTimeout):
            if retries <= 0:
               raise
            retries -= 1

response = Response(
    status_code,
    http_version=http_version.decode("ascii"),
    headers=headers,
    stream=stream,  # type: ignore
    request=request,
)

self.cookies.extract_cookies(response)

status = f"{response.status_code} {response.reason_phrase}"
response_line = f"{response.http_version} {status}"
logger.debug(f'HTTP Request: {request.method} {request.url} "{response_line}"')

return response
```

I'm still not 100% sure that we want this, but perhaps this is a decent low-impact feature.
Any thoughts?

florimondmanca:
Closes #1141

Adds off-by-default "retry for connect errors" behavior, used as `httpx.Client(..., retries=<int>)`.

* Makes sure that only `ConnectError`/`ConnectTimeout` are retried on, and that we only retry on idempotent HTTP verbs (i.e. exclude those that by definition should not be retried on because it could duplicate state changes on the server side).
* `httpx.request(<method>, connect_retries=<int>)` is not included, since I'm not sure we want this knob in the high-level API.
* The `httpx.Client(retries=<int>)` API is a good start to any future enhancements to retry functionality, such as an `httpx.Retry` class for encapsulating retries configuration, or providing a custom retries API, while preserving "`retries=<int>` retries on connect errors" for the most basic use cases.

Rendered docs preview:

Screenshot 2020-09-01 at 23 11 54


tomchristie:
Fab. 👍✨

Thoughts...

* I *guess* we should probably use the inverse form instead of `if request.method in ("POST", "PATCH")`, eg. `if request.method in IDEMPOTENT_METHODS` with `IDEMPOTENT_METHODS = ["GET", "HEAD", "PUT", "DELETE", "OPTIONS", "TRACE"]`, so that requests with nonstandard methods are not treated as idempotent.
* Wondering if we ought to be keeping the API as narrow as possible, and initially only supporting `retries: int = 0`? Rather than exposing `httpx.Retries()` and `backoff_factor`. Pushing back on the API surface area as long as possible is generally a good idea.
* It'd be good to review `requests`/`urllib3` behaviour here, and just double check that we're matching up with requests' behaviour when installing an adapter with `retries=<int>`. Any working through of that would be great.

tomchristie:
The [`requests` docs give](https://requests.readthedocs.io/en/latest/api/#requests.adapters.HTTPAdapter)...

> max_retries – The maximum number of retries each connection should attempt. **Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server.** By default, Requests does not retry failed connections. If you need granular control over the conditions under which we retry a request, import urllib3’s Retry class and pass that instead.

However, it looks like if `retries != 0` then `Retry.from_int(...)` is used...

https://github.com/psf/requests/blob/48237afd9d064c639b9c2bcea7e75ab7b717b181/requests/adapters.py#L116-L119

Which I *think* means you'll end up with `read=None` on the class (Not the same as `read=False`), and that read retries will be used... https://github.com/urllib3/urllib3/blob/f0c43419e45db17a3c5f287be80aa0d14e2f58c6/src/urllib3/util/retry.py#L243

So I *think* the actual behaviour that `requests` uses isn't quite as documented, and is actually...

* `ConnectError` and `ConnectTimeout` always retry if `retries=<positive int>`.
* `ReadError` retries if the method is idempotent and `retries=<positive int>`.

?

florimondmanca:
Based on https://github.com/encode/httpx/issues/1141#issuecomment-679994648, I updated this PR to solely focus on adding support for `httpx.Client(connect_retries=<int>)`.

httpx/_client.py
```diff
@@ -66,6 +74,7 @@ def __init__(
         cookies: CookieTypes = None,
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
+        connect_retries: int = DEFAULT_CONNECT_RETRIES,
```
tomchristie, line 77: Perhaps `retries: int = DEFAULT_RETRIES` is okay.
That gives us more scope to later keep the same API, but extend it to also support more granular retry behaviours? (But with a plain `retries=<int>` always just using basic connect retry behaviour)

tomchristie (dismissed):
This is really nicely done.

I *think* this probably strikes the right balance for us at the moment.
One comment for us to think about around the parameter naming.

Plus I guess we'd want the docs in before we merge it?

But yeah I think this is pretty fab. 👍👍👍

florimondmanca:
@tomchristie Changed the `connect_retries` naming, and added some docs. :-) Leaving this open for a bit in case you or anyone else has any comments on the docs, or remaining comments on the implementation.

httpx/_utils.py
```diff
@@ -529,3 +533,20 @@ def __eq__(self, other: typing.Any) -> bool:
...
+
+async def sleep(seconds: float) -> None:
+    library = sniffio.current_async_library()
+    if library == "trio":
+        import trio
+
+        await trio.sleep(seconds)
+    else:
+        assert library == "asyncio"
+        await asyncio.sleep(seconds)
```
tomchristie, line 546: I guess we may as well also include curio here, in anticipation of the incoming support in `httpcore` - https://github.com/encode/httpcore/pull/168

We can always just `pragma: nocover` it for the moment, right?

httpx/_client.py
```diff
@@ -793,6 +807,22 @@ def _send_handling_auth(
...
 
+    def _send_handling_retries(self, request: Request, timeout: Timeout) -> Response:
+        retries_left = self.retries
+        delays = exponential_backoff(factor=RETRIES_BACKOFF_FACTOR)
+
+        while True:
+            try:
+                return self._send_single_request(request, timeout)
+            except (ConnectError, ConnectTimeout):
+                if request.method not in IDEMPOTENT_METHODS:
```
tomchristie, line 818: I don't think we need to care about idempotency for connect retries.
If we've not established a connection to the server, then we're okay to re-issue the request *whatever* the method.

Eg. see connect retries vs. read retries behaviour in urllib3 https://github.com/urllib3/urllib3/blob/f0c43419e45db17a3c5f287be80aa0d14e2f58c6/src/urllib3/util/retry.py#L415-L427

florimondmanca:
@tomchristie Addressed latest feedback — we should be good to go on this now? 😄

tests/client/test_retries.py
```diff
@@ -0,0 +1,191 @@
...
+
+@pytest.mark.parametrize(
+    "factor, expected",
+    [
+        (0.1, [0, 0.1, 0.2, 0.4, 0.8]),
+        (0.2, [0, 0.2, 0.4, 0.8, 1.6]),
+        (0.5, [0, 0.5, 1.0, 2.0, 4.0]),
+    ],
+)
+def test_exponential_backoff(factor: float, expected: List[int]) -> None:
```
tomchristie, line 186: Since this is the only test case that requires a private import, perhaps we could put this one in `tests/test_utils.py`?

Working towards making sure we're only using `import httpx` in our test cases, but the utils cases are an exception to that.

tomchristie:
Okay, so I was leaving this on pause for a bit to try to figure out if connection retries really make sense on the client on at the transport layer, and I think we really *do* want this on the transport layer, wrapping as close to the connection opening as possible.

The reason here is that if we pull connection retries out to the client layer, then we can end up in a gnarly situation when there's high contention on the connection pool. With connection retries out at the client layer an incoming request with a retry will:

* Wait to acquire the max connections semaphore.
* Attempt a connection and fail.
* Release the max connections semaphore.
* Wait to acquire the max connections semaphore.
* Attempt a connection and fail.
* Release the max connections semaphore.

We'd really prefer a behaviour where once the semaphore is acquired, the connection will perform any connect+retry within that scope, so...

* Wait to acquire the max connections semaphore.
* Attempt a connection and fail.
* Attempt a connection and fail.
* Release the max connections semaphore.

This means much less thrashing when there's high contention.
Also supposing we have a timeout configuration like... `httpx.Timeout(5.0, pool=60.0)`, and `retries=2`, then the maximum possible timeout is a more expected 60+5+5+5, rather than a possibly surprising 60+60+60+5+5+5.

Anyways, upshot of this is that I *think?* I'm keen on us pushing the implementation details of this into the transport layer. Possibly by passing `retries` as part of the `__init__` configuration, and handling the retry behaviour within the `Connection._open_socket` method?... https://github.com/encode/httpcore/blob/92d3e9bbd3c442c6035860eea9e15a2249e0cad8/httpcore/_async/connection.py#L99

florimondmanca:
@tomchristie The motivations wrt timeouts and pooling are strong, indeed. Thoughts on my reservations related to testing here? https://github.com/encode/httpx/issues/1141#issuecomment-683329538 On HTTPCore we can't just "swap out a transport", so I'm not sure how we'd test this w/o monkeypatch since it would be right into the socket opening logic…

tomchristie:
> Thoughts on my reservations related to testing here? #1141 (comment) On HTTPCore we can't just "swap out a transport", so I'm not sure how we'd test this w/o monkeypatch since it would be right into the socket opening logic…

Ah really great point yup.

This is kinda the point at which we need to be digging into exposing `backend=<BackendClass>`, and having tests that run against mock backends, so that we can do stuff like testing connection retries, and all sorts of other in-detail bits.

florimondmanca:
Opened https://github.com/encode/httpcore/pull/221 against HTTPCore, closing this one for now!

PR #1197, opened 2020-08-19
Issue warning on unclosed `AsyncClient`.

### LINKED ISSUE #871: Add warning about unclosed AsyncClient

cdeler:
As a part of https://github.com/encode/httpx/issues/871 it has been decided to prevent `httpx.Client` and `httpx.AsyncClient` from sending requests after being closed. 

Closes #871

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1062,10 +1062,13 @@ class Client(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        self._transport.__exit__(exc_type, exc_value, traceback)
-        for proxy in self._proxies.values():
-            if proxy is not None:
-                proxy.__exit__(exc_type, exc_value, traceback)
+        if not self.is_closed:
+            self._is_closed = True
+
+            self._transport.__exit__(exc_type, exc_value, traceback)
+            for proxy in self._proxies.values():
+                if proxy is not None:
+                    proxy.__exit__(exc_type, exc_value, traceback)
 
     def __del__(self) -> None:
         self.close()
@@ -1682,10 +1685,12 @@ class AsyncClient(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        await self._transport.__aexit__(exc_type, exc_value, traceback)
-        for proxy in self._proxies.values():
-            if proxy is not None:
-                await proxy.__aexit__(exc_type, exc_value, traceback)
+        if not self.is_closed:
+            self._is_closed = True
+            await self._transport.__aexit__(exc_type, exc_value, traceback)
+            for proxy in self._proxies.values():
+                if proxy is not None:
+                    await proxy.__aexit__(exc_type, exc_value, traceback)
 
     def __del__(self) -> None:
         if not self.is_closed:
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1691,12 +1691,8 @@ class AsyncClient(BaseClient):
         if not self.is_closed:
             warnings.warn(
                 f"Unclosed {self!r}. "
-                "Please call\n"
-                "\t>>> await async_client.aclose()\n"
-                "or use it as a context manager\n"
-                "\t>>> async with httpx.AsyncClient() as client:\n"
-                "\t>>>     ...",
-                ResourceWarning,
+                "See https://www.python-httpx.org/async/#opening-and-closing-clients "
+                "for details."
             )
 
 
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1,5 +1,6 @@
 import functools
 import typing
+import warnings
 from types import TracebackType
 
 import httpcore
@@ -79,6 +80,14 @@ class BaseClient:
         self.max_redirects = max_redirects
         self._trust_env = trust_env
         self._netrc = NetRCInfo()
+        self._is_closed = True
+
+    @property
+    def is_closed(self) -> bool:
+        """
+        Check if the client being closed
+        """
+        return self._is_closed
 
     @property
     def trust_env(self) -> bool:
@@ -696,6 +705,8 @@ class Client(BaseClient):
 
         [0]: /advanced/#request-instances
         """
+        self._is_closed = False
+
         timeout = self.timeout if isinstance(timeout, UnsetType) else Timeout(timeout)
 
         auth = self._build_request_auth(request, auth)
@@ -1029,16 +1040,20 @@ class Client(BaseClient):
         """
         Close transport and proxies.
         """
-        self._transport.close()
-        for proxy in self._proxies.values():
-            if proxy is not None:
-                proxy.close()
+        if not self.is_closed:
+            self._is_closed = True
+
+            self._transport.close()
+            for proxy in self._proxies.values():
+                if proxy is not None:
+                    proxy.close()
 
     def __enter__(self) -> "Client":
         self._transport.__enter__()
         for proxy in self._proxies.values():
             if proxy is not None:
                 proxy.__enter__()
+        self._is_closed = False
         return self
 
     def __exit__(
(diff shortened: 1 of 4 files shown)


httpx/_client.py
```diff
@@ -56,6 +56,18 @@
 KEEPALIVE_EXPIRY = 5.0
 
 
+def check_not_closed(method: typing.Callable) -> typing.Callable:
```
tomchristie, line 59: I'd prefer for us *not* to use a decorator style.
I'll use inline comments below to walk through the alternative...

httpx/_client.py
```diff
@@ -619,6 +639,7 @@ def _transport_for_url(self, url: URL) -> httpcore.SyncHTTPTransport:
 
         return self._transport
 
+    @check_not_closed
```
tomchristie, line 642: I guess we should instead put the check around `def send()` rather than `def request()`.

tomchristie: Rather than the decorator, I'd prefer the plainer style of starting the method with...

```python
if self._is_closed:
   raise RuntimeError("Cannot send requests on a closed client instance.")
```

httpx/_client.py
```diff
@@ -1015,6 +1036,7 @@ def delete(
             timeout=timeout,
         )
 
+    @check_not_closed
```
tomchristie, line 1039: Could we instead just let things pass silently if the client is already closed, and `.close()` is called again?
So...

```python
if not self._is_closed:
    self._transport.close()
    for proxy in self._proxies.values():
        if proxy is not None:
            proxy.close()

File: httpx/_client.py [hunk 4]
```diff
@@ -1628,6 +1654,8 @@ async def aclose(self) -> None:
             if proxy is not None:
                 await proxy.aclose()
 
+        self._is_closed = True
```
tomchristie, line 1657: We should put `self._is_closed = True` as the very first line, rather than the very last line, since that'd handle potential race conditions more cleanly.

tomchristie (commented):
Thanks! ✨😀

Seems simple on the surface, but there's a few bits & pieces to work through.

In particular I'd prefer we don't use the decorator style, and prefer a plain approach instead.
Direct, simple, easier to reason through, neater tracebacks.
I've included some comments to help work through what I think could be tweaked.

httpx/_client.py
cdeler: I've removed the decorator. Simple is better than complex.

httpx/_client.py
cdeler: > I guess we should instead put the check around def send() rather than def request().

Done

> Rather than the decorator, I'd prefer the plainer style of starting the method with...

Done, but I used `self.is_closed` in the condition

httpx/_client.py
cdeler: @tomchristie It a bit confuses me, since there should be another ways to avoid race conditions, e.g. locks.

cdeler: I saw several same places where race condition _is possible_ 

I can try to write a test, which strictly cause a race condition here

cdeler:
@tomchristie 

I wrote the test, which shows the race condition. I can fix it (but I'd like to do it in separate PR since it involves things such as `threading.Lock()`.

Here is the test:
```python3
class SlowTestTransport(httpcore.SyncHTTPTransport):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.closing_counter = 0

    def request(
        self,
        method: bytes,
        url: URL,
        headers: Headers = None,
        stream: SyncByteStream = None,
        timeout: TimeoutDict = None,
    ) -> Tuple[bytes, int, bytes, List[Tuple[bytes, bytes]], SyncByteStream]:
        return b"", 0, b"", [(b"", b"")], ByteStream(b"Hello")

    def close(self) -> None:
        self.closing_counter += 1


class SlowClient(httpx.Client):
    @property
    def is_closed(self) -> bool:
        result = super().is_closed

        time.sleep(0.5)

        return result


def test_that_client_handles_race_condition_while_being_closed():
    slow_transport = SlowTestTransport()
    client = SlowClient(transport=slow_transport)

    futures = []
    with ThreadPoolExecutor(max_workers=2) as executor:
        futures.append(executor.submit(client.close))
        futures.append(executor.submit(client.close))

    for _ in concurrent.futures.as_completed(futures):
        pass

    assert slow_transport.closing_counter == 1
```

tomchristie:
Let's not get into that level of testing at the moment. *Yes* you can potentially do silly things if you're concurrently closing clients *and* issuing requests at the same time, but we don't necessarily need to protect users from themselves at quite that level of granularity.

So, working through this has actually made it more clear to me exactly what behaviour we *do* want here...

* Clients should *start* in `._is_closed = True`
* Calling into `.send()` or calling into either `__enter__`/`__aenter__` should set `._is_closed = False`
* Calling into `.close()`/`.aclose()` or calling into either `__exit__`/`__aexit__` should set `._is_closed = True`
* The `Client.__del__` method should check `._is_closed` and call `.close()` if needed.
* The `AsyncClient.__del__` method should check `._is_closed` and warn if needed.
* Making a request with a closed client is *fine*. It just implicitly opens the client.

That's because we don't actually want simply *importing* this...

```python
client = httpx.AsyncClient()
```

To raise a warning. Which if we treat the client as open simply by instantiating it, we would.

The client is *only* opened once you're either using...

* `async with client:` - Enter the context on an existing instance.
* `async with httpx.AsyncClient() as client:` - Enter the context, creating an instance.
* `await client.request()` - Implicitly opening the client, by issuing a request. Not recommended for async, where we advise using the client in a context manager, but perfectly common with `httpx.Client()`.

cdeler:
> So, working through this has actually made it more clear to me exactly what behaviour we do want here...
> ...

Nice, then I'm going to implement that (in this PR)

cdeler:
@tomchristie I've done with the changes you mentioned.

> Calling into .close()/.aclose() or calling into either __exit__/__aexit__ should set ._is_closed = True

I prevent the clients from being double-closed. Am I right, or I should remove this check?

tomchristie:
So I tried this locally, and it took me a while to figure out that I wasn't seeing anything because `ResourceWarning` is ignored by default. I'm not sure what we think about that?

We *could* issue it as a `UserWarning` instead, perhaps?
Or a custom subclass?
Or perhaps it's okay as a mostly silent `ResourceWarning` and we're okay with that?

cdeler:
@tomchristie yes, I didn't pay attention to the fact, that `ResourceWarning` is silent by default. May be the `RuntimeWarning` (it's verbose)?

> exception RuntimeWarning¶
Base class for warnings about dubious runtime behavior.


(c) python language reference ([link](https://docs.python.org/3/library/exceptions.html#RuntimeWarning))

tomchristie:
@cdeler - Thanks for all your fantastic work on this.

I think it probably *is* worth issuing a PR just dealing with the "close async clients properly in our test cases" first, since we can pull that one into the `master` branch, wheras this one we need to keep holding onto until we're ready to issue a new major release. Doing it that way around would help us avoid possibly messy merge conflicts, and would give us two independently reviewable PRs.

cdeler:
> I think it probably _is_ worth issuing a PR just dealing with the "close async clients properly in our test cases" first

Yes, sure, I'm to create this "close async clients properly in our test cases" PR today

cdeler:
@tomchristie 

According to [your comment](https://github.com/encode/httpx/pull/1219#issuecomment-680814430), I rebased this branch to the `master` after https://github.com/encode/httpx/pull/1219 .

The only thing which is not done is 
> Tackle any cleanup in test cases, and always prefer just using `httpx.Client` unless there's a good reason.

tomchristie (approved):
Great stuff, let's get this in now! 👍

PR #1198, opened 2020-08-19
Add `proxies` parameter to top-level API functions

### LINKED ISSUE #1195: Allow `proxies` in `httpx.request()` high-level API
Spun up from #1194 

The high-level API functions such as `httpx.get()` and `httpx.post()` don't allow passing `proxies=...` yet (only possible on `Client`), but Requests does support this: https://requests.readthedocs.io/en/master/user/advanced/#proxies

We should add support for this argument:

```pycon
>>> import httpx
>>> proxies = "http://localhost:8899"
>>> r = httpx.get("https://example.com", proxies=proxies)
```

We may need to update some documentation (for example, the [HTTP Proxying](https://www.python-httpx.org/advanced/#http-proxying) docs *may* need to clarify that `proxies` does not _require_ using clients), but we can treat that as a follow-up if it's unclear what should be updated exactly.

j178:
Closes #1195 

Should I add tests to cover these parameters in top-level API functions?

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -262,9 +262,10 @@ client = httpx.Client(trust_env=False)
 
 ## HTTP Proxying
 
-HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization.
+HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization or top-level api `httpx.get(..., proxies=...)`.
 
 _Note: SOCKS proxies are not supported yet._
+_Note: `proxies` parameter passed to `Client.request` is not supported yet._
 
 <div align="center">
     <img src="https://upload.wikimedia.org/wikipedia/commons/thumb/2/27/Open_proxy_h2g2bob.svg/480px-Open_proxy_h2g2bob.svg.png"/>
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -265,7 +265,6 @@ client = httpx.Client(trust_env=False)
 HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization or top-level API functions like `httpx.get(..., proxies=...)`.
 
 _Note: SOCKS proxies are not supported yet._
-_Note: `proxies` parameter passed to `Client.request` is not supported yet._
 
 <div align="center">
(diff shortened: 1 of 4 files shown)


httpx/_api.py
```diff
@@ -56,6 +58,8 @@ def request(
     request.
     * **auth** - *(optional)* An authentication class to use when sending the
     request.
+    * **proxies** - *(optional)* A dictionary mapping HTTP protocols to proxy
```
florimondmanca, line 61: Perhaps "mapping proxy keys to proxy URLs"? Proxy routing supports more than just routing by HTTP protocol.

docs/advanced.md
```diff
@@ -262,9 +262,10 @@ client = httpx.Client(trust_env=False)
 
 ## HTTP Proxying
 
-HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization.
+HTTPX supports setting up [HTTP proxies](https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers) via the `proxies` parameter to be passed on client initialization or top-level API functions like `httpx.get(..., proxies=...)`.
 
 _Note: SOCKS proxies are not supported yet._
+_Note: `proxies` parameter passed to `Client.request` is not supported yet._
```
florimondmanca, line 268: I don't think we need this note, especially since we most likely won't support this at all in the foreseeable future (it's a consequence of our tighter transport API).

florimondmanca (commented):
Nice, thanks!

> tests

Sure! Perhaps a single test against `httpx.request` is enough here, but up to you!

j178:
I find it a little difficult to test whether these parameters passed to top-level API functions are working as expected, do you have any ideas?

florimondmanca:
I see! Indeed I went through our proxy tests here and they're mostly about the proxy routing functionality - no existing test actually integrates with HTTPCore and a live proxy. Well no sweat then, I think it's alright (we certainly don't have tests for all options in the top level API anyway).

docs/compatibility.md
```diff
@@ -42,6 +42,8 @@ This is different to the `requests` usage of `proxies={"http": ..., "https": ...
 
 This change is for better consistency with more complex mappings, that might also include domain names, such as `proxies={"all://": ..., "all://www.example.com": None}` which maps all requests onto a proxy, except for requests to "www.example.com" which have an explicit exclusion.
 
+Also noted that `requests.Session.request(...)` allows a `proxies=...` parameter, whereas `httpx.Client.request(...)` does not.
```
florimondmanca, line 45: Typo: "also note"

florimondmanca (approved):
💯

Found a tiny typo

docs/compatibility.md
j178: Thanks, updated.

PR #1200, opened 2020-08-20
Add stream docstring

### LINKED ISSUE #1154: Add API docs for streaming
- [x] There are no similar issues or pull requests for this yet.
- [x] I discussed this idea in #1152 and feedback is positive.

There is no documentation in either the top level `stream` or `Client.stream`, and they are not mentioned in generated documentation. At least a short description of functionality and parameters (pointing to `httpx.request` if applicable) should be added.

j178:
Closes #1154

Add a docstring for `stream()`, and update `docs/api.md`.

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -115,9 +115,14 @@ def stream(
     trust_env: bool = True,
 ) -> StreamContextManager:
     """
-    Streaming response content.
+    Alternative to `httpx.request()` that streams the response body
+    instead of loading it into memory at once.
 
     **Parameters**: See `httpx.request`.
+
+    See also: [Streaming Responses][0]
+
+    [0]: /quickstart#streaming-responses
     """
     client = Client(cert=cert, verify=verify, trust_env=trust_env)
     request = Request(
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -114,6 +114,11 @@ def stream(
     cert: CertTypes = None,
     trust_env: bool = True,
 ) -> StreamContextManager:
+    """
+    Streaming response content.
+
+    **Parameters**: See `httpx.request`.
+    """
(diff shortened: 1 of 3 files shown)


httpx/_api.py
```diff
@@ -114,6 +114,11 @@ def stream(
     cert: CertTypes = None,
     trust_env: bool = True,
 ) -> StreamContextManager:
+    """
+    Streaming response content.
```
florimondmanca, line 118: What do we think of some more elaborate copy like this...?

```
Alternative to `httpx.request()` that streams the response body instead of loading it into memory at once.

**Parameters**: See `httpx.request`.

See also: [Streaming Responses][0]

[0]: /quickstart#streaming-responses
```

Also I think we could add this docstring onto `Client.stream` and `AsyncClient.stream` as well. :-)

httpx/_api.py
```diff
@@ -226,9 +231,7 @@ def head(
     **Parameters**: See `httpx.request`.
 
     Note that the `data`, `files`, and `json` parameters are not available on
-    this function, as `HEAD` requests should not include a request body. The
-    `HEAD` method also differs from the other cases in that `allow_redirects`
-    defaults to `False`.
+    this function, as `HEAD` requests should not include a request body.
```
florimondmanca, line 234: Should this be submitted as a separate cleanup PR?

florimondmanca (commented):
Thanks!

httpx/_api.py
j178: Sure 💯

httpx/_api.py
j178: Nice 👍 
But there are no `Client.stream` and `AsyncClient.stream` methods 😄

tomchristie (approved):
Lovely, yup!

PR #1201, opened 2020-08-20
Cleanup docstring leftover for #1183

j178:
Remove description about `head(..., allow_redirects=False)` (left by #1183)


tomchristie (approved):
Good catch, yup!

PR #1212, opened 2020-08-21
Add an HTTPX cli. 😇

tomchristie:
Given just how wonderful the `click` and `rich` packages are these days, we could do some pretty wonderful stuff with an `httpx` command line tool, pretty easily.

Here's a pull request that demonstrates the basics, along with a couple of examples.

Installation is `pip install httpx[cli]`.

Then we get:

```shell
$ httpx https://www.example.org
```

Screenshot 2020-08-21 at 16 17 25

```shell
$ httpx https://httpbin.org/post -m POST -j '{"hello": "world"}'
```

Screenshot 2020-08-21 at 16 17 06

Obviously there's more too it that this, in particular you'd want...

* Controls for changing what output we display (eg. response headers, request line, redirects etc...)
* Nice handling of how we display the possible error types.

And a few bits of other functionality, such as...

* `--stream` - Disable formatting, and just display the content as it arrives.
* `--upload` - Set upload content directly from a file.
* `--download` - Download directly to a file. Default to displaying a progress bar if we're on a terminal display.
* `--raise-for-status` - Return a non-zero error code, and display as an error, rather than the response itself.
* `--max-redirects`, `--http2` - Client-level controls that are not available in `httpx.request`, but that we might want in the cli.
* Authentication controls. Haven't quite figured out how they'd look, since there's a few different options.

But some things that I think are nice about this are:

* Just how much value we'd be getting for relatively minimal effort.
* Having a really clean mapping between the CLI options and the `httpx` API, which I think could really help users get to grips with both of them.

Normally I might think of this kind of thing as scope-creep, but if we can make it tie in really nicely, then perhaps it's just a really nice way to make `httpx` an all-round more valuable tool?

Obvs. https://httpie.org/ exists and is *fantastic*, but I can still see the value in us presenting an alternative here, right?


tomchristie:
Lookin nice... ⚡️

CJLhe08ZzO

StephenBrown2:
Rivaling httpie in no time!

tomchristie:
Some TODO items...

- [ ] Support `--upload [FILE]` for uploading content from file.
- [ ] Support `--stream` for streaming output to console as soon as available. (Won't use syntax highlighting)
- [ ] Support `--session [FILE]` for persistent cookie sessions.
- [ ] Support `--verify [FILE]`.
- [ ] Support `--cert [FILE]` for client signing.
- [ ] Support `--auth-scheme [basic|digest]`.
- [ ] Transfer speed with `rich.progress` doesn't look correct to me. Moves around too much. Value can revise downwards once the download completes.
- [ ] `--verbose` should also display any additional auth, redirect requests. Requires `auth`/`redirect` hooks.
- [ ] Syntax highlighting should preserve console background colour.

httpx/_cli/cli.py
```diff
@@ -0,0 +1,310 @@
...
+    format_response_headers,
+    get_download_filename,
+    get_lexer_for_response,
+)
+
+
+def print_request_headers(request: httpx.Request) -> None:
+    console = rich.console.Console()
+    http_text = format_request_headers(request)
+    syntax = rich.syntax.Syntax(http_text, "http", theme="ansi_dark", word_wrap=True)
```
StephenBrown2, line 23: Any particular reason for going with `theme="ansi_dark"` instead of `background_color="default"`?

tomchristie: Either the `ansi_light` and `ansi_dark` themes render okay on both light and dark console backgrounds...

Screenshot 2020-09-22 at 09 19 56

Screenshot 2020-09-22 at 09 20 02

Using `background_color="default"` is really nice with `monokai` on dark consoles, but terrible on light backgrounds...

Screenshot 2020-09-22 at 09 18 19

Screenshot 2020-09-22 at 09 18 05

tomchristie:
So, perhaps the route to go down here is to publish this under a separate `httpx-cli` package?
We could *still* include it prominently in the `httpx` docs, but it'd just be packaged separately.

The benefit there would be that we're not tied to bumping up `httpx` versions when we add new stuff to the client, and our 1.0 release doesn't have a whole bunch of extra considerations for us to make wrt. the cli interface.

httpx/_cli/cli.py
StephenBrown2: Gotcha. I have a dark terminal, so I preferred the default theme. Perhaps adding a `--theme={default,dark,light}` option could be an thing?

tomchristie:
Closing this, in favour of a little beta over here: https://github.com/encode/httpx-cli

PR #1215, opened 2020-08-24
Event hooks

### LINKED ISSUE #790: Support for Event Hooks

tomchristie:
Closes #790, based on the design outlined in https://github.com/encode/httpx/issues/790#issuecomment-679068648

I've narrowed this down to *only* `request` and `response` initially. I think that `auth` and `redirect` also make sense, but let's treat that separately. (I want to be careful, since it could be a bit fiddly making sure that raised exceptions within those hooks don't leave responses open.)

Also worth noting that we're not exposing the `EventHooks` data structure as public API. Event hooks are always set using a regular dictionary, but have smart behaviour when the property is updated, so that they *always* return lists of installed handlers, but can be set *either* from a list *or* from a single callable.

For example:

```python
client = httpx.Client()
assert dict(client.event_hooks) == {'request': [], 'response': []}

client.event_hooks['request'] = some_callable
assert dict(client.event_hooks) == {'request': [some_callable], 'response': []}

client.event_hooks['request'].append(some_other_callable)
assert dict(client.event_hooks) == {'request': [some_callable, some_other_callable], 'response': []}

del client.event_hooks['request']
assert dict(client.event_hooks) == {'request': [], 'response': []}
```

Rendered docs:

Screenshot 2020-08-24 at 15 28 19

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -35,6 +35,7 @@ def request(
     verify: VerifyTypes = True,
     cert: CertTypes = None,
     trust_env: bool = True,
+    event_hooks: dict = None,
 ) -> Response:
     """
     Sends an HTTP request.
@@ -84,7 +85,12 @@ def request(
     ```
     """
     with Client(
-        proxies=proxies, cert=cert, verify=verify, timeout=timeout, trust_env=trust_env,
+        proxies=proxies,
+        cert=cert,
+        verify=verify,
+        timeout=timeout,
+        trust_env=trust_env,
+        event_hooks=event_hooks,
     ) as client:
         return client.request(
             method=method,
@@ -117,6 +123,7 @@ def stream(
     verify: VerifyTypes = True,
     cert: CertTypes = None,
     trust_env: bool = True,
+    event_hooks: dict = None,
 ) -> StreamContextManager:
     """
     Alternative to `httpx.request()` that streams the response body
@@ -128,7 +135,13 @@ def stream(
 
     [0]: /quickstart#streaming-responses
     """
-    client = Client(proxies=proxies, cert=cert, verify=verify, trust_env=trust_env)
+    client = Client(
+        proxies=proxies,
+        cert=cert,
+        verify=verify,
+        trust_env=trust_env,
+        event_hooks=event_hooks,
+    )
     request = Request(
         method=method,
         url=url,
@@ -162,6 +175,7 @@ def get(
     verify: VerifyTypes = True,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
     trust_env: bool = True,
+    event_hooks: dict = None,
 ) -> Response:
     """
     Sends a `GET` request.
@@ -184,6 +198,7 @@ def get(
         verify=verify,
         timeout=timeout,
         trust_env=trust_env,
+        event_hooks=event_hooks,
     )
 
 
@@ -200,6 +215,7 @@ def options(
     verify: VerifyTypes = True,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
     trust_env: bool = True,
+    event_hooks: dict = None,
 ) -> Response:
     """
     Sends an `OPTIONS` request.
@@ -222,6 +238,7 @@ def options(
         verify=verify,
         timeout=timeout,
         trust_env=trust_env,
+        event_hooks=event_hooks,
     )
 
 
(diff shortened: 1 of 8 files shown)


httpx/_client.py
```diff
@@ -710,6 +723,13 @@ def send(
             finally:
                 response.close()
 
+        try:
+            for hook in self._event_hooks["response"]:
+                hook(response)
+        except Exception:
+            response.close()
+            raise
```
j178, line 731:
```suggestion
        finally:
            response.close()
```

httpx/_client.py
```diff
@@ -1312,6 +1338,13 @@ async def send(
             finally:
                 await response.aclose()
 
+        try:
+            for hook in self._event_hooks["response"]:
+                await hook(response)
+        except Exception:
+            await response.aclose()
+            raise
+
```
j178, line 1347:
```suggestion
        finally:
            await response.aclose()
```

j178 (commented):
💯 
Just some minor suggestions 😄

httpx/_api.py
```diff
@@ -35,6 +35,7 @@ def request(
     verify: VerifyTypes = True,
     cert: CertTypes = None,
     trust_env: bool = True,
+    event_hooks: dict = None,
```
florimondmanca, line 38: I think I'd like us to prefer a stricter typing here, and a more straightforward style that requires using a list of callables:

```python
client = httpx.Client(event_hooks={"request": [on_request], "response": [on_response])
```

It would make it clearer that `client.event_hooks` is really just always a dict of lists, rather than "a dict of sometimes a callable, sometimes a list of callables, depending on what you've set".

Hence this suggestion (which would need to be applied across `_client.py` as well):

```suggestion
    event_hooks: Dict[str, List[Callable]] = None,
```

httpx/_config.py
```diff
@@ -410,6 +411,48 @@ def __repr__(self) -> str:
         )
 
 
+class EventHooks(MutableMapping):
```
florimondmanca, line 414: I think if we apply my suggestion of using a plain `Dict[str, List[Callable]]` always, we can mostly drop this custom data structure, correct?

I'm also not sure we _need_ to enforce that the only possible keys be `request` and `response`. Currently we do this with a bit of a surprising behavior ("don't actually store items for unknown keys"), while I think making it so that `event_hooks` behaves just like a normal dict would be beneficial.

So perhaps we could replace this with a simple factory helper…

```python
def create_event_hooks(initial: Dict[str, List[Callable]] = None) -> Dict[str, List[Callable]]):
    event_hooks = {"request": [], "response": []}
    event_hooks.update(initial or {})
    return event_hooks
```

tests/client/test_async_client.py
```diff
@@ -166,3 +166,59 @@ async def test_100_continue(server):
 
     assert response.status_code == 200
     assert response.content == data
+
+
+@pytest.mark.usefixtures("async_environment")
```
florimondmanca, line 171: What do we think of putting these tests (and those for the sync client) in a dedicated `tests/client/test_event_hooks.py` file?

tests/client/test_properties.py
```diff
@@ -1,36 +1,36 @@
-from httpx import URL, AsyncClient, Cookies, Headers, Timeout
+from httpx import URL, Client, Cookies, Headers, Timeout
 
 
 def test_client_base_url():
-    client = AsyncClient()
+    client = Client()
```
florimondmanca, line 5: I think these changes belong to a separate PR? Or most likely, this PR isn't up to date with `master`. :-)

tests/client/test_properties.py
```diff
@@ -49,3 +49,12 @@ def test_client_timeout():
...
     assert client.timeout.write == expected_timeout
     assert client.timeout.pool == expected_timeout
+
+
+def test_client_event_hooks():
+    def on_request(request):
+        pass  # pragma: nocover
+
+    client = Client()
+    client.event_hooks = {"request": on_request}  # type: ignore
```
florimondmanca, line 59: Another reason why I'm advocating for a straightforward "dict of lists of callables" is that we wouldn't need any `# type: ignore`…

```suggestion
    client.event_hooks = {"request": [on_request]}
```

tests/test_event_hooks.py
```diff
@@ -0,0 +1,81 @@
+from httpx._config import EventHooks
+
+
+def callable():
+    pass  # pragma: nocover
+
+
+def test_event_hooks_init():
```
florimondmanca, line 8: These tests would be simplified if we adopted a simple `create_event_hooks()` helper that returns a plain dict.

PR #1217, opened 2020-08-25
Add support for sync-specific or async-specific auth flows

### LINKED ISSUE #1176: Supporting async auth flows.
Hello,

I try to make asynchronous calls to the Microsoft Graph API using `httpx.AsyncClient`. For authentication I use the `azure-identity` package that provides several client classes to easily get an access token:

```python
from azure.identity.aio import EnvironmentCredential

async with EnvironmentCredential() as credential:
    token = await credential.get_token(*scopes)
```

I'd like then to subclass `httpx.Auth` and implement the `auth_flow` method

```python
class AzureAuth(Auth):

    async def auth_flow(self, request):
        token = await self._credential.get_token(*self._scopes) # Get the access token or retrieve it from cache
        request.headers["Authorization"] = "Bearer " + token.token
        yield request
```

so that I can do things like

```python
async with AsyncClient(base_url="https://graph.microsoft.com", auth=AzureAuth(...)) as client:
    response = await client.get("/groups")
```

The problem is that `httpx.Auth.auth_flow` is **synchronous**. How can I make it asynchronous ?

florimondmanca:
Closes #1176, prompted by https://github.com/encode/httpx/issues/1176#issuecomment-680074866

Allow authors or auth classes to provide a sync-specific and/or async-specific implementation of the auth flow, so that they can do I/O, synchronization or concurrency without blocking the event loop:

```python
class Auth:
    def auth_flow(self, request) -> Generator[Request, Response, None]:
        yield request  # Current API

    def sync_auth_flow(self, request: Request) -> Generator[Request, Response, None]:
        ...  # Defer to `.auth_flow()` by default

    async def async_auth_flow(self, request: Request) -> AsyncGenerator[Request, Response]:
        ...  # Defer to `.auth_flow()`
```

Doesn't affect the user-side auth API, nor existing auth implementations.

Rendered docs:

capture


tomchristie:
Great! A bit more discussion about this over at https://github.com/encode/httpx/issues/1176#issuecomment-680763226

tomchristie:
Okay, I'm getting warm fuzzies about this. 😻

Suggestion that I'm curious to get your feedback on...

What do we think to also:

* Adding a `sync_auth_flow` mirroring `async_auth_flow`
* Bringing in the request/response body loading logic into the Auth class itself.

So, something like this...

```python
class Auth:
    """
    Base class for all authentication schemes.
    To implement a custom authentication scheme, subclass `Auth` and override
    the `.auth_flow()` method.
    If the authentication scheme does I/O, such as disk access or network calls, or uses
    synchronization primitives such as locks, you should override `.async_auth_flow()`
    to provide an async-friendly implementation that will be used by the `AsyncClient`.
    Usage of sync I/O within an async codebase would block the event loop, and could
    cause performance issues.
    """

    requires_request_body = False
    requires_response_body = False

    def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
        """
        Execute the authentication flow.
        To dispatch a request, `yield` it:
        ```
        yield request
        ```
        The client will `.send()` the response back into the flow generator. You can
        access it like so:
        ```
        response = yield request
        ```
        A `return` (or reaching the end of the generator) will result in the
        client returning the last response obtained from the server.
        You can dispatch as many requests as is necessary.
        """
        raise NotImplementedError("Override 'auth_flow' to implement a custom auth class.")
        yield request  # pragma: nocover 

    def sync_auth_flow(
        self, request: Request
    ) -> typing.AsyncGenerator[Request, Response]:
        """
        Execute the authentication flow asynchronously.
        By default, this defers to `.auth_flow()`. You should override this method
        when the authentication scheme does I/O, such as disk access or network calls,
        or uses concurrency primitives such as locks.
        """
        if self.requires_request_body:
            request.read()

        flow = self.auth_flow(request)
        request = next(flow)

        while True:
            response = yield request
            if self.requires_response_body:
                response.read()

            try:
                request = flow.send(response)
            except StopIteration:
                break

    async def async_auth_flow(
        self, request: Request
    ) -> typing.AsyncGenerator[Request, Response]:
        """
        Execute the authentication flow asynchronously.
        By default, this defers to `.auth_flow()`. You should override this method
        when the authentication scheme does I/O, such as disk access or network calls,
        or uses concurrency primitives such as locks.
        """
        if self.requires_request_body:
            await request.aread()

        flow = self.auth_flow(request)
        request = next(flow)

        while True:
            response = yield request
            if self.requires_response_body:
                await response.aread()

            try:
                request = flow.send(response)
            except StopIteration:
                break
```

Here's why that might be a nice thing to do...

* Most users will override `auth_flow`, that'll support sync+async. Works for any cases where the only I/O you need is sending auth requests and perhaps reading request/response bodies.
* For sync I/O users can override `sync_auth_flow`. The class will error if you attempt to use it with an async client.
* For async I/O users can override `async_auth_flow. The class will error if you attempt to use it with a sync client.
* Or users can override `sync_auth_flow`+`async_auth_flow`.

Also...

* The request/response body logic is more tightly encapsulated, rather than being spread partly across the auth class, and partly in the client.
* Testing the auth class becomes a bit easier, as you don't need to think about handling the request/response body logic in the test case.

So I think testing looks something like this?...

(Okay, we can't actually create `Response` instances quite like I've done in this example just yet, but I've got some work to talk about planned here, so just roll with it for now...)

#### Test for Basic Auth

```python
auth = httpx.BasicAuth()
request = httpx.Request("GET", "https://www.example.com")

# The initial request should include a basic auth header.
flow = auth.sync_auth_flow(request)
request = next(flow)
assert request.headers['Authorization'].startswith('Basic')

# No other requests are made.
response = httpx.Response(text='Hello, world!', status_code=200)
with pytest.raises(StopIteration):
    flow.send(response)
```


#### Test for Digest Auth with a 200 response

```python
auth = httpx.DigestAuth()
request = httpx.Request("GET", "https://www.example.com")

# The initial request should not include an auth header.
flow = auth.sync_auth_flow(request)
request = next(flow)
assert 'Authorization' not in request.headers

# If a 200 response is returned, then no other requests are made.
response = httpx.Response(text='Hello, world!', status_code=200)
with pytest.raises(StopIteration):
    flow.send(response)
```

#### Test for Digest Auth with a 401 response

```python
auth = httpx.DigestAuth()
request = httpx.Request("GET", "https://www.example.com")

# The initial request should not include an auth header.
flow = auth.sync_auth_flow(request)
request = next(flow)
assert 'Authorization' not in request.headers

# If a 401 response is returned, then a digest auth request is made.
headers = {'WWW-Authenticate': 'Digest realm="...", qop="...", nonce="...", opaque="..."'}
response = httpx.Response(text='Auth required', status_code=401, headers=headers)
request = flow.send(response)
assert request.headers['Authorization'].startswith('Digest')

# No other requests are made.
response = httpx.Response(text='Hello, world!', status_code=200)
with pytest.raises(StopIteration):
    flow.send(response)
```

Testing in an async context looks the same except calling `.async_auth_flow()`, `anext`, and `asend`.

What do we think?

tests/client/test_auth.py
```diff
@@ -641,3 +666,23 @@ def test_sync_auth_reads_response_body() -> None:
...
+async def test_sync_async_auth() -> None:
+    """
+    Test that we can use a different auth flow implementation in the async case, to
+    support cases that require performing I/O or using concurrency primitives (such
+    as checking a disk-based cache or fetching a token from a remote auth server).
+    """
+    url = "https://example.org/"
+    auth = SyncOrAsyncAuth()
+
+    client = AsyncClient(transport=AsyncMockTransport())
```
tomchristie, line 681: Let's use `async with AsyncClient(...) as client` here, since we're starting to get strict about that now.

May as well also use the same for the sync case too, just for consistency within the test case.

tests/client/test_auth.py
```diff
@@ -184,6 +186,29 @@ def auth_flow(self, request: Request) -> typing.Generator[Request, Response, Non
...
+    """
+    A mock authentication scheme that uses a different implementation for the
+    sync and async cases.
+    """
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._async_lock = asyncio.Lock()
+
+    def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
```
tomchristie, line 199: This ought to be `sync_auth_flow`, really.

httpx/_client.py
```diff
@@ -760,15 +760,12 @@ def _send_handling_auth(
         auth: Auth,
         timeout: Timeout,
     ) -> Response:
-        if auth.requires_request_body:
-            request.read()
+        auth_flow = auth.sync_auth_flow(request)
+        request = auth_flow.send(None)  # type: ignore
```
tomchristie, line 764: Wondering which of `auth_flow.send(None)` vs `next(auth_flow)` we should prefer for this? I don't really mind either way, but I guess whichever we stick with we should use consistently across:

* The docs.
* The client implementation.
* The auth implementation.

florimondmanca: I'll stick to our current usage of `next()` / `.__anext__()` for the purpose of this PR. Given that this works more nicely with type checking it's a pretty okay approach.

florimondmanca:
Added docs that were missing to make this PR complete — ready for a new round of reviews!

httpx/_auth.py
```diff
@@ -46,6 +51,56 @@ def auth_flow(self, request: Request) -> typing.Generator[Request, Response, Non
         """
```
tomchristie, line 51: Should we raise a `NotImplementedError()` here by default?

tomchristie: Okay, I can see we're using `Auth()` in no-auth cases at the moment, so no that wouldn't work trivially.
Might be something to think about there, but let's treat it separately if we do.

tomchristie:
Added in auth unit tests for good measure.
We might(?) even want to doc up unit testing auth classes at some point, but can treat that separately if we do want to.

Anyways, this one is *fantastic* - very pleased with it! ✨🍰✨

tomchristie (approved):

Issue #1145, opened 2020-08-07
Context Managed Transports


tomchristie:
We need to ensure that the Client `__enter__`/`__exit__` and `__aenter__`/`__aexit__` methods also call into the transport dunder methods, for all installed transports/proxies.

A good proof of having resolved the issue here would be streaming responses in the `ASGITransport`, which require nursery functionality in order to work properly. See notes here... https://github.com/encode/httpx/pull/998#issuecomment-653519294

florimondmanca:
Hey,

Agreed we want clients to call into context manager dunder methods wherever appropriate.

While looking over #998 I also think we have currently a problem re: `.aclose()` vs `__aexit__(*exc_info)` in clients (and perhaps transports?).

Right now the client calls `.aclose()` in its `__aexit__()` method, which calls into the `.aclose()` of transports, but this is not correct. If an exception occurs, then we want it to propagate in the `__aexit__()` of transports. Right now it won't, because we're calling `.aclose()` on transports which does nothing. It's a bit hard to explain in words but I might submit a PR to demonstrate this point…

Also, as I think I've said already, I _really_ don't think this issue applies much to `ASGITransport`. The nursery object we need for `ASGITransport` is request-level, not client-level (since you want the app to stop when the request is finished), so whether clients call into `__aenter__()`/`__aexit__()` is not really relevant. (However the problem I described above _will_ pose a problem on trio since it'll detect that the nursery is exited but the parent context is not propagating exceptions properly.)

PR #1218, opened 2020-08-26
Context managed transports

tomchristie:
Closes https://github.com/encode/httpx/issues/1145

Note that once #977 is addressed we'll be able to test the proxies case more elegantly.

We're not being clever here about "ensure other transports still close if an exceptions occurs during `__exit__` in one transport". (Or the other fiddly case, of ensuring the `__exit__` is called on any already `__enter__`'d transports, if an exception occurs during `__enter__`.)

I *think* we should probably treat that as a follow up, and it's *possible* that we might only want to do so once #977 is addressed.

I think thats probably an okay approach to take since it's not any *worse* than our existing behaviour, which anyway doesn't handle "ensure other transports still close if an exceptions occurs during `close` in one transport", and besides exceptions during transport `__enter__`/`__exit__` are pretty much unrecoverable issues, and it's not really any worse if they happen to potentially leave other transports hanging open.

(Actually having talked through it I'm not event convinced that we *should* bother handling that case at all, but either way let's talk it through as a follow-up)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1047,10 +1047,10 @@ class Client(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        self._transport.__exit__()
+        self._transport.__exit__(exc_type, exc_value, traceback)
         for proxy in self._proxies.values():
             if proxy is not None:
-                proxy.__exit__()
+                proxy.__exit__(exc_type, exc_value, traceback)
 
 
 class AsyncClient(BaseClient):
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1035,6 +1035,10 @@ class Client(BaseClient):
                 proxy.close()
 
     def __enter__(self) -> "Client":
+        self._transport.__enter__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                proxy.__enter__()
         return self
 
     def __exit__(
@@ -1043,7 +1047,10 @@ class Client(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        self.close()
+        self._transport.__exit__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                proxy.__exit__()
 
 
 class AsyncClient(BaseClient):
@@ -1639,6 +1646,10 @@ class AsyncClient(BaseClient):
                 await proxy.aclose()
 
     async def __aenter__(self) -> "AsyncClient":
+        await self._transport.__aenter__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                await proxy.__aenter__()
         return self
 
     async def __aexit__(
@@ -1647,7 +1658,10 @@ class AsyncClient(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        await self.aclose()
+        await self._transport.__aexit__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                await proxy.__aexit__()
 
 
 class StreamContextManager:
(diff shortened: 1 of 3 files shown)


httpx/_client.py
```diff
@@ -1043,7 +1047,10 @@ def __exit__(
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        self.close()
+        self._transport.__exit__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                proxy.__exit__()
```
florimondmanca, line 1053: I believe we need to send the exception information down, like this:

```suggestion
        self._transport.__exit__(exc_type, exc_value, traceback)
        for proxy in self._proxies.values():
            if proxy is not None:
                proxy.__exit__(exc_type, exc_value, traceback)
```

(I'm actually surprised the code works without passing these - HTTPCore transports allow `= None` but I'm not _sure_ this is part of the context manager protocol strictly?)

httpx/_client.py
```diff
@@ -1647,7 +1658,10 @@ async def __aexit__(
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        await self.aclose()
+        await self._transport.__aexit__()
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                await proxy.__aexit__()
```
florimondmanca, line 1664:
```suggestion
        await self._transport.__aexit__(exc_type, exc_value, traceback)
        for proxy in self._proxies.values():
            if proxy is not None:
                await proxy.__aexit__(exc_type, exc_value, traceback)
```

tests/client/test_async_client.py
```diff
@@ -166,3 +167,38 @@ async def test_100_continue(server):
...
+
+        async def aclose(self):
+            self.events.append("transport.aclose")
+
+        async def __aenter__(self):
+            await super().__aenter__()
+            self.events.append("transport.__aenter__")
+
+        async def __aexit__(self):
+            await super().__aexit__()
```
florimondmanca, line 186:
```suggestion
        async def __aexit__(self, *args):
            await super().__aexit__(*args)
```

tests/client/test_client.py
```diff
@@ -208,3 +209,37 @@ def test_pool_limits_deprecated():
...
+
+        def close(self):
+            self.events.append("transport.close")
+
+        def __enter__(self):
+            super().__enter__()
+            self.events.append("transport.__enter__")
+
+        def __exit__(self):
+            super().__exit__()
```
florimondmanca, line 227:
```suggestion
        def __exit__(self, *args):
            super().__exit__(*args)
```

tests/client/test_async_client.py
```diff
@@ -166,3 +167,38 @@ async def test_100_continue(server):
...
+    # also neccessarily include proxies.
+    transport = Transport()
+    async with httpx.AsyncClient(transport=transport, proxies="http://www.example.com"):
+        pass
+
+    assert transport.events == [
+        "transport.__aenter__",
+        "transport.aclose",
+        "transport.__aexit__",
+    ]
```
florimondmanca, line 204: I'm curious why `transport.aclose` is included at all here? Do we call `.aclose()` anywhere before `Client.__aexit__()`?

Edit: hmm, probably because the base `.__aexit__()` implementation calls into `.aclose()`?

tomchristie: Yes.

The base implementation is "Just call `.close()`/`.aclose()`".

One of the allowable things for a concrete transport to do is "Just implement `close`/`aclose`", which is in fact what our existing implementations do.

That gives us both the simple case, while still allowing *some* implementations to do more complex stuff with `__aenter__`, `__aexit__`.

tomchristie:
@florimondmanca Your reviews are *always* on-point. Wonderful, thank you. ✨🔥✨

florimondmanca (approved):
:100:

tests/client/test_async_client.py
tomchristie: Actually maybe worth a comment there against the test case.

PR #1219, opened 2020-08-26
Closing AsyncClient in all tests (#871)

cdeler:
As we discussed in https://github.com/encode/httpx/pull/1197#issuecomment-680770927 , it was decided to move the test refactoring out from https://github.com/encode/httpx/pull/1197

In this PR I added closing to all unclosed `AsyncClient` in the tests

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -5,12 +5,11 @@ import typing
 import httpcore
 import pytest
 
+import httpx
 from httpx import (
     URL,
-    AsyncClient,
     Auth,
     BasicAuth,
-    Client,
     DigestAuth,
     ProtocolError,
     Request,
@@ -189,8 +188,8 @@ async def test_basic_auth() -> None:
     url = "https://example.org/"
     auth = ("tomchristie", "password123")
 
-    client = AsyncClient(transport=AsyncMockTransport())
-    response = await client.get(url, auth=auth)
+    async with httpx.AsyncClient(transport=AsyncMockTransport()) as client:
+        response = await client.get(url, auth=auth)
 
     assert response.status_code == 200
     assert response.json() == {"auth": "Basic dG9tY2hyaXN0aWU6cGFzc3dvcmQxMjM="}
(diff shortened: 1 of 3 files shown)


tomchristie:
Great!

At the moment you're using `async with` in some cases here, and `.aclose()` in others.

I'm wondering if we should instead just always be using `async with`?
Maybe it doesn't really mater either way? What do other @encode/maintainers think?

cdeler:
@tomchristie 
Some tests, for example https://github.com/encode/httpx/blob/cfda19ed8c17e56afc8444d624049705cd8aa8c4/tests/client/test_auth.py#L313-L326 

are long. In such cases I was adding `await client.aclose()` to prevent adding new indentation level

florimondmanca:
> I'm wondering if we should instead just always be using async with?

> In such cases I was adding await client.aclose() to prevent adding new indentation level (but I can change all test to make them use identical AsyncClient closing method, e.g. async with)

Yeah, imo, we should probably stick to `async with` everywhere, even if it increases the diff size here a bit (we can ease review by enabling "ignore whitespace" in the GitHub review UI).

tomchristie:
Yup, fair call. I've also noticed that there seem to be a bunch of places in the test cases where we're using `AsyncClient`, but that I think we should probably *just* be using `Client` instead.

For example https://github.com/encode/httpx/blob/master/tests/client/test_headers.py where it looks to me like it'd be a whole bunch neater if it was just testing against a plain client.

Wondering if we should do the following in order?

* Tackle any cleanup in test cases, and always prefer just using `httpx.Client` unless there's a good reason.
* Make sure we're using context management in any remaining places where we *are* still using `httpx.AsyncClient`.
* Update #1197 against the `master` branch at that point.

(Also @cdeler thanks for being so patient, there's been a whole bunch of extra stuff we've had to work through in order to approach this as carefully and throughly as this.)

tomchristie:
See #1222.

cdeler:
I'm to work on this ticket in 2 days

tomchristie:
Sure thing - don't put yourself under pressure, or feel that you need to let us know when you will or won't be working on stuff. Thanks again for all your work on this. 😄

cdeler:
@tomchristie 

I did what you asked:
1. for now all `AsyncClient` usage is gone by `async with` block (except the`test_async_proxy_close`, where it's done in finally block)
1. I changed all `AsyncClient` references to `httpx.AsyncClient`

tomchristie:
Great! It looks like there's still some work we could be doing on preferring the use of `httpx.Client` over `httpx.AsyncClient` wherever possible in the tests, but we can probably just treat that separately to the rest of this now.

I'm happy enough for this to go in whenever. Good work. 👍

tomchristie (approved):

cdeler:
> Great! It looks like there's still some work we could be doing on preferring the use of httpx.Client over httpx.AsyncClient wherever possible in the tests, but we can probably just treat that separately to the rest of this now.

let me do that, I really missed that from the above messages:
> Tackle any cleanup in test cases, and always prefer just using httpx.Client unless there's a good reason.

PR #1221, opened 2020-08-26
Add data to client delete

CircleOnCircles:
according to

https://github.com/encode/httpx/pull/900
https://github.com/encode/httpx/pull/410

the api delete should accept the data param.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+pythonenv*/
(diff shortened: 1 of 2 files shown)


CircleOnCircles:
i'm not sure if there is a need for test case.

tomchristie:
The HTTP specs are that DELETE requests should not include a body...

https://tools.ietf.org/html/draft-ietf-httpbis-semantics-05#section-7.3.5

> A payload within a DELETE request message has no defined semantics;
> sending a payload body on a DELETE request might cause some existing
> implementations to reject the request.

If you *really* want to issue a DELETE request with a body anyway, then you can use `httpx.request('DELETE', data=...)`.

See also https://github.com/encode/httpx/pull/900#issuecomment-612049818

PR #1225, opened 2020-08-27
chore: add config for pointing to Gitter chat

Midnighter:
This configuration enables an entry in the issue template picker that points to the Gitter channel instead.


Midnighter:
black had an update last night and now annoyingly formats code differently. I don't think it makes sense to handle that in this PR.

tomchristie:
Great. How about we keep that *changes* this introduces a bit minimal, so instead of this...

0ED6C464-8EEB-4897-9C07-C39EFB69116E

We could instead drop the “Question” issue template, and instead replace it with a “Question” external link, with exactly the same text. Simply “Ask a question”.

Also let’s keep the “Open a blank issue” as it is for now.

Midnighter:
Okay, shall I do the same for https://github.com/encode/uvicorn/pull/768 then?

tomchristie:
I reckon so, yup.

tomchristie (approved):
Great, let's give it a go.

tomchristie:
Fantastic, I think the new flow looks great!

PR #1229, opened 2020-08-27
Use and pin black 20

florimondmanca:
Quick fix to un-break CI.

A nice follow-up would be to pin all deps and rely on Dependabot, similar to https://github.com/encode/httpcore/pull/162.


euri10 (approved):
haha you're smarter than uvicorn guys :duck:  and didnt fell into the trailing comma multi-line auto-reformat  !!

Issue #1227, opened 2020-08-27
Supporting `Response(content=..., text=..., html=..., json=...)`


tomchristie:
We’d like to support instantiating `Response` instances directly.

One really good example of why we might want to do that is for unit testing custom auth classes - https://github.com/encode/httpx/pull/1217#issuecomment-681825882

Currently `Response` instances are only ever created by the client itself, and...

* Have a mandatory `request=...` parameter.
* Have a mandatory `status_code` parameter.

The first things we should do here are:

* Make the `request` parameter become optional. Have a `.request` property that returns a `Request` instance or raises a RuntimeError if the instance is not set. Have a property setter for the `.request` property.
* Make the status code be enforced as kwarg-only, and make it optional, with a default of 200.

Once we’ve done those things we should add support for setting the content of a response, using `content=..., text=..., html=..., json=...` parameters, which mirror the request parameters of `data=..., files=..., json=...`, but differ because HTTP responses use different content types to HTTP requests.

The `encode(...)` method in `content_streams.py` would become `encode_request_body(data=..., files=..., json=..., boundary=...)`, and would be mirrored by a `encode_response_body(content=..., text=..., html=..., json=...)` method, which would use the following content streams:

* `content` - Support either bytes or byte iterator/aiterators. Setting `Content-Length` or `Transfer-Encoding: chunked`
* `text` - Set `Content-Type: text/plain; charset=“utf-8” `, and a `Content-Length` header.
* `html` - Set `Content-Type: text/html; charset=“utf-8”` and a `Content-Length` header.
* `json` - Set `Content-Type: application/json` and a `Content-Length` header.

We can tie in an `encoding=...` argument too, but let’s talk that through as a follow up.

What this would then all give us is the ability to create responses like so...

```python
response = httpx.Response(json={“hello”: “world”})
response = httpx.Response(html=“<body><h1>hello, world</h1></body>”})
response = httpx.Response(text=“Not found”, status_code=404)
response = httpx.Response(content=b’...’, headers=...)
```

Somewhat refs #1091

tomchristie:
If any contributors want to take a go at this I'd suggest approaching it in parts...

* Issue a PR for making the `request=...` parameter not mandatory.
* Issue a PR making `status_code` a keyword-only argument, and defaulting it to 200.
* Issue a PR with the response body encoding handling (more complex than the other two.)

Pradhvan:
[...]

PR #1232, opened 2020-08-28
Make status_code keyword-only

j178:
Make `status_code` a keyword-only argument, and defaulting it to 200. 

First part of #1227 .

### DIFF
--- a/tests/models/test_cookies.py
+++ b/tests/models/test_cookies.py
@@ -69,7 +69,7 @@ def test_multiple_set_cookie():
         ),
     ]
     request = httpx.Request("GET", "https://www.example.org")
-    response = httpx.Response(200, request=request, headers=headers)
+    response = httpx.Response(status_code=200, request=request, headers=headers)
 
     cookies = httpx.Cookies(jar)
     cookies.extract_cookies(response)
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -803,7 +803,7 @@ class Client(BaseClient):
                 timeout=timeout.as_dict(),
             )
         response = Response(
-            status_code,
+            status_code=status_code,
             http_version=http_version.decode("ascii"),
             headers=headers,
             stream=stream,  # type: ignore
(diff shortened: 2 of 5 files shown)


tests/models/test_cookies.py
```diff
@@ -69,7 +69,7 @@ def test_multiple_set_cookie():
         ),
     ]
     request = httpx.Request("GET", "https://www.example.org")
-    response = httpx.Response(200, request=request, headers=headers)
+    response = httpx.Response(status_code=200, request=request, headers=headers)
```
florimondmanca, line 72: Do we want to keep these `status_code=200` in the sale of "explicit over implicit", or should we rely on the new "200 by default"?

florimondmanca (commented):
Almost all good, just a bit question :)

tomchristie:
Having reviewed the change footprint on this I'm thinking we *might* want to hold fire on this part? Looking at the test cases that change it's not so obviously a UX improvement. Once we've made `request=` optional, we could reassess?

tests/models/test_cookies.py
j178: I think explicitly specifying `status_code=200` is better.

florimondmanca (commented):
Thanks for rebasing this against master! Reviewing what this PR has become, I'm not sure this change brings much value in terms of API usability. Eg `httpx.Response(200)` reads quite well as "HTTP 200", while `httpx.Response(status_code=200)` feels a bit _forced_? What do you think?

tomchristie:
I'm going to suggest that we close this one off.

On reassessing, I *don't* think we want it to be keyword-only, and #1227 doesn't actually imply that we should do this. There's nothing wrong with this...

```python
response = httpx.Response(200, json={"hello": "world"})
```

It's *possible* that at some point we might decide to include a *default* value, so that we also support the following...

```python
response = httpx.Response(json={"hello": "world"})
```

But it's not obvious if we do or don't want that, and there's no reason for us to require it for #1227, so let's just pass on it for now.

j178:
Fair enough, I am totally OK with or without this change. Going to close this.

PR #1238, opened 2020-08-30
Make the response's request parameter optional

tbascoul:
Make the response's request parameter optional

Part of #1227

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -689,7 +689,7 @@ class Response:
         self,
         status_code: int,
         *,
-        request: Request,
+        request: Request = None,
         http_version: str = None,
         headers: HeaderTypes = None,
         stream: ContentStream = None,
@@ -700,7 +700,8 @@ class Response:
         self.http_version = http_version
         self.headers = Headers(headers)
 
-        self.request = request
+        self._request: typing.Optional[Request] = request
+
         self.call_next: typing.Optional[typing.Callable] = None
 
         self.history = [] if history is None else list(history)
@@ -725,6 +726,19 @@ class Response:
                 "has been read or closed."
             )
         return self._elapsed
+
+    @property
+    def request(self) -> Request:
+        """
+        Returns the request instance associated to the current response.
+        """
+        if self._request is None:
+            raise RuntimeError("'.request' may only be accessed if initialized")
+        return self._request
+
+    @request.setter
+    def request(self, value: Request) -> None:
+        self._request = value
 
     @property
     def reason_phrase(self) -> str:
@@ -811,7 +825,7 @@ class Response:
                 value = value.strip().lower()
                 try:
                     decoder_cls = SUPPORTED_DECODERS[value]
-                    decoders.append(decoder_cls(request=self.request))
+                    decoders.append(decoder_cls(request=self._request))
                 except KeyError:
                     continue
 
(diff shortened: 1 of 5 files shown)


tomchristie:
Okay, so it turns out this issue is actually more awkawrd than realised at first, because although we *do* want `request=...` top be optional on a `Response()` instance, we *don't* want it to be optional on `RequestError`/`HTTPError` exceptions, where we *know* it *will* always be set.

Here's what I think we're going to need to do here...

* The signatures on our exception classes should not change.
* The decoders should *not* raise `DecodeError`, or accept a `request` instance on `__init__`. Instead any decoding errors should just raise a plain old `ValueError` exception.
* The `Response` instance should wrap any decoding codeblocks instead a `try ... except ValueError as exc`. *If* `response._request` is set then the exception should be re-raised using `DecodeError(message=str(exc), request=self.request) from exc`, but if `response._request` is *not* set, then the plain `ValueError` exception should just be re-raised, using `raise exc`.

That'll allow us to have...

* An optional `request=...` parameter on the `Response.__init__()`.
* Without changing the exception signatures.
* Properly raising `httpx.DecodeError()` on exceptions that occur during a request flow, where we *will* have an associated `request` instance.
* For unit-test style response instances, that do not have a `request=...` instance set, we'll get plain `ValueError` exceptions for any decoding errors.

You're welcome to take on these follow ups yourself, or else I'd also be happy to take a look at them.

tbascoul:
Thanks for the detailed guidance.
Still some work to do on the testing part.

httpx/_models.py
```diff
@@ -843,10 +858,14 @@ def raise_for_status(self) -> None:
 
         if codes.is_client_error(self.status_code):
             message = message.format(self, error_type="Client Error")
-            raise HTTPStatusError(message, request=self.request, response=self)
+            if self._request is None:
```
tbascoul, line 861: I defaulted to raise ValueError on missing request but it seems a bit off.

httpx/_models.py
```diff
@@ -843,10 +858,14 @@ def raise_for_status(self) -> None:
...
+                raise ValueError(message)
+            raise HTTPStatusError(message, request=self._request, response=self)
```
tbascoul, line 863: Raising a ValueError inside raise_for_status seems a bit off ?

httpx/_models.py
```diff
@@ -726,6 +728,19 @@ def elapsed(self) -> datetime.timedelta:
             )
         return self._elapsed
 
+    @property
+    def request(self) -> Request:
+        """
+        Returns the request instance associated to the current response.
+        """
+        if self._request is None:
+            raise RuntimeError("'.request' may only be accessed if initialized")
```
tomchristie, line 737:
```suggestion
            raise RuntimeError("The 'request' instance has not been set on this response.")
```

httpx/_models.py
tomchristie: I'd suggest we start the method with something like...

```python
if self._request is None:
    raise RuntimeError("Cannot call `raise_for_status` as the request instance has not been set on this response.")
```

tomchristie (approved):
Great, yup! Thanks so much for your time on this!

I've added in a `_wrap_decode_errors` context manager, to remove some of the replication, and get the test coverage back up to 100%

PR #1243, opened 2020-09-01
Drop Response(..., request=...) style in test cases.

tomchristie:
Now that `request=...` is optional on response instances, we can drop it's usage in test cases, except in any cases where it is actually required by the test.

### DIFF
--- a/tests/models/test_responses.py
+++ b/tests/models/test_responses.py
@@ -36,19 +36,19 @@ def test_response():
 
 
 def test_raise_for_status():
-    REQUEST = httpx.Request("GET", "https://example.org")
+    request = httpx.Request("GET", "https://example.org")
 
     # 2xx status codes are not an error.
-    response = httpx.Response(200, request=REQUEST)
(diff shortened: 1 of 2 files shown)


tests/models/test_responses.py
```diff
@@ -22,17 +20,24 @@ async def async_streaming_body():
...
     assert response.text == "Hello, world!"
-    assert response.request is REQUEST
+    assert response.request.method == "GET"
+    assert response.request.url == "https://example.org"
     assert response.elapsed >= datetime.timedelta(0)
     assert not response.is_error
 
 
 def test_raise_for_status():
+    REQUEST = httpx.Request("GET", "https://example.org")
```
florimondmanca, line 39: I guess let's use the same "use lowercase if not using globals" convention here, even if it results in a slightly larger diff since we have to change the rest of this test case?
```suggestion
    request = httpx.Request("GET", "https://example.org")
```

florimondmanca (approved):
💯

Just a nit, then good to go!

PR #1246, opened 2020-09-02
Event hooks

### LINKED ISSUE #790: Support for Event Hooks

tomchristie:
Based on https://github.com/encode/httpx/pull/1215 but addressing review comments from @florimondmanca.

Minor API change here is that hooks *must* now be set as a list of callables, and do not accept the shorthand "set a single callable" style.

Closes #790

Valid examples:

```python
client = httpx.Client()
assert dict(client.event_hooks) == {'request': [], 'response': []}

client.event_hooks['request'] = [some_callable]
assert dict(client.event_hooks) == {'request': [some_callable], 'response': []}

client.event_hooks['request'].append(some_other_callable)
assert dict(client.event_hooks) == {'request': [some_callable, some_other_callable], 'response': []}

client.event_hooks['request'] = []
assert dict(client.event_hooks) == {'request': [], 'response': []}
```

There's probably marginally more scope for accidentally breaking things than in #1215 (Eg. `del client.event_hooks['request']` would end up causing a failure, as would `client.event_hooks['request'] = some_callable`) but on the flip side it's a bit simpler without the `EventHooks` mapping structure.

In either case I'm happy enough with this take, since if we *did* ever later decide to add the extra structure in #1215, then we could still choose to do so, since it would be backwards compatible with this slightly more minimal approach.

Rendered docs...

Screenshot 2020-09-02 at 10 45 10


httpx/_client.py
```diff
@@ -765,6 +794,10 @@ def _send_handling_auth(
 
         auth_flow = auth.auth_flow(request)
         request = next(auth_flow)
+
+        for hook in self._event_hooks["request"]:
+            hook(request)
```
florimondmanca, line 799: Getting my thoughts down here — are we sure this is the correct/best place to call `request` hooks? AFAICS Requests does not provide a `request` hook so we're in green field here.

There are a few options that I can see…

1. In `.send()` or at the beginning of `_send_handling_redirects()` - Hide redirect requests, auth requests, retries requests.
1. At the beginning of `._send_handling_auth()` - Hide auth requests, retries requests.
1. At the beginning of `._send_single_request()` - Hide retries requests.

As I understand it, currently this PR suggests option 2/.

But interestingly the `response` hook is located at level 1/. So in case of a redirect, we'd have `request` hooks called twice or more, but `response` hooks called once.

Is this appropriate, or do we want `request` being called only once (for the initial request) (which corresponds to option 1/)?

Another data point from the monitoring use case perspective — I can see that the Datadog/Requests tracer integration [wraps `requests.Session.send()`](https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/contrib/requests/connection.py), which _I think_ corresponds to case 1/ above.

Perhaps there's a case to be made about having `request`, `redirect_request`, `auth_request`, and `single_request` hooks, or perhaps just `request` and `single_request`, or…? And then, should we match them up with corresponding `response_*` hooks?

I guess for simplicity we'd really want option 1/, i.e. call `request` hooks before calling `._send_handling_redirects()`, and call `response` hooks on the received resolved response?

docs/advanced.md
```diff
@@ -221,6 +221,57 @@ with httpx.Client(headers=headers) as client:
...
+* `response` - Called once the response has been returned. Passed the `response` instance.
+
+These allow you to install client-wide functionality such as logging and monitoring.
+
+```python
+def log_request(request):
+    print(request)
+
+def log_response(response):
+    print(response)
```
florimondmanca, line 241: A note: I initially thought that we might want to pass the `request` as an argument to `response` hooks, but actually the request would already be available as `response.request`. We might want to make this "tip" explicit in the docs?

```suggestion
def log_response(response):
    print(response, "in response to", response.request)
```

tests/test_utils.py
```diff
@@ -237,18 +237,66 @@ def test_not_same_origin():
...
+        (
+            "all://",
+            "https://example.com:123",
+            True,
+        ),
+        (
+            "",
+            "https://example.com:123",
+            True,
+        ),
```
florimondmanca, line 299: Unrelated formatting changes, it seems? :-)

tests/client/test_event_hooks.py
```diff
@@ -0,0 +1,113 @@
+import pytest
+
+import httpx
+
+
+def test_event_hooks(server):
```
florimondmanca, line 6: It seems we're not testing for "request hooks are called on redirect requests", which makes me think the decision to put request hooks at the top of `._send_handling_auth()` was not entirely deliberate? :-)

Any case, whichever we decide I guess we'd need to add explicit tests for what we expect to happen as requests and responses go through the send-redirects-auth-(retries soon) stack?

tomchristie:
Here's the behaviour we'll want for event hooks...

* `request` - Called exactly once, with `callback(request)`. We want the fully prepared request at this point, including any auth headers.
* `auth` - Usually not called, except for multiple auth flows, where it's called with `callback(response, request)`.
* `redirect` - Called for redirects, with `callback(response, request)`.
* `response` - Called exactly once, with `callback(response)`.

That'd give us a nice symmetry where we always end up with the same number of request/response arguments across the full set of callbacks.

tomchristie:
Prompting me to think that we've kinda got our `send_handling_auth` and `send_handling_redirects` around the wrong way, really.

httpx/_client.py
tomchristie: Resolved via #1267

tomchristie:
@florimondmanca 🙏 😁

florimondmanca:
@tomchristie Im generally okay with these changes and approch, assuming we're clear enough that we will want support for additional hooks than requests and responses.

That said, before committing to this API I'd sure like your opinion on how this interacts with telemetry support? Especially the "per-request context" aspects.

florimondmanca (approved):
Code looks good! 👍👍

PR #1244, opened 2020-09-01
Add "Early Hints" and "Too Early" to Status Codes

StephenBrown2:
To align with Python 3.9: https://docs.python.org/3.9/whatsnew/3.9.html#http

> HTTP status codes `103 EARLY_HINTS`, `418 IM_A_TEAPOT` and `425 TOO_EARLY` are added to http.HTTPStatus. (Contributed by Dong-hee Na in bpo-39509 and Ross Rhodes in bpo-39507.)

Source: https://github.com/python/cpython/blob/da52be47690da0d9f78d0dce9ee5c3e4dbef49e1/Lib/http/__init__.py

PR #1240, opened 2020-08-31
Handle multiple auth headers correctly

tomchristie:
Closes #1234 by correctly treating WWW-Authenticate headers as possibly occurring multiple times, and only dealing with “Digest” headers, and just ignoring other types of authentications.

PR #1247, opened 2020-09-02
Version 0.14.3

tomchristie:
I think it's worth us rolling a minor release now, and then after that we can start to merge in some of the outstanding feature PRs into the `master` branch, eg...

#1197 - Currently waiting to merge, because it'll require a proper version bump.
#1182 - Currently waiting to merge, because it'll require a proper version bump.
#1217 - Major feature
#1196 - Major feature
#1246 - Major feature

---

### Added

* `http.Response()` may now be instantiated without a `request=...` parameter. Useful for some unit testing cases. (Pull #1238)
* Add `103 Early Hints` and `425 Too Early` status codes. (Pull #1244)

### Fixed

* `DigestAuth` now handles responses that include multiple 'WWW-Authenticate' headers. (Pull #1240)
* Call into transport `__enter__`/`__exit__` or `__aenter__`/`__aexit__` when client is used in a context manager style. (Pull #1218)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.14.3 (September 2nd, 2020)
+
+### Added
+
+* `http.Response()` may now be instantiated without a `request=...` parameter. Useful for some unit testing cases. (Pull #1238)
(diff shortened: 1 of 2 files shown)


florimondmanca:
🤔 Do we want a `0.15.0` release instead? (You said "minor release", but this is a patch bump.) I guess we don't _actually_ want to "sneak in" these new features in a patch release and break semver expectations so that we get as much possible feedback as we can because people would upgrade without knowing if they're currently pinned to `0.14.*`, do we? 😄

florimondmanca (approved):

PR #1248, opened 2020-09-02
Header refinements

### LINKED ISSUE #1226: Refine headers behaviour for `Request()` and `client.build_request()`
Currently `Request()` auto-generates a bunch of headers on `__init__`. Having thought about it for a bit I can see some refinements we could make here to:

* Give users a bit more control over the default headers, such as allowing them to remove the `User-Agent` header.
* Make `Request()` more generally useful, so that it only auto sets mandatory headers.

Rather than set the following headings on `Request`, we should instead include them on `client.headers`...

* `User-Agent: python-httpx/1.2.3`
* `Accept: */*`
* `Accept-Encoding: gzip, deflate`
* `Connection: keep-alive`

When passing `Client(headers={...})` or setting `client.headers = {}` we should give precedence to explicitly set headers, but should *include* these defaults if they're not otherwise set.

```python
# Create a client, with a custom Accept header.
# Other automatic headers are still included.
client = httpx.Client(headers={'Accept': 'application/json, */*;q=0.8'}
```

Or...

```python
client = httpx.Client()

# Set a custom Accept header.
# Other automatic headers are still included.
client.headers = {'Accept': 'application/json, */*;q=0.8'}
```

If a user wants to *actually* remove one of these, they *can* do so by explicitly deleting it, eg...

```python
# Create a client, without a User-Agent header.
client = httpx.Client()
del client.headers['User-Agent']
```

The recommended way of creating a request instance will continue to be `client.build_request(...)`, but we'll want to highlight that more clearly, as instantiating a plain `Request()` instance will then give you a more basic set of behaviour by only auto-setting mandatory headers, and leaving everything else up to the user...

* Set a `Host` header, if the request URL is an absolute URL.
* Requests that mandate a body will always include either a `Content-Length` or `Transfer-Encoding` header.

Both of those cases will defer to any explicitly included header if one exists, and only auto set if those mandatory headers are omitted.

Also `.prepare_headers()` should become a private method.

This all ends up being beneficial, because users still get all the regular convenience of `client.build_request()`, but there's also a plainer style for creating `Request()` instances with complete control over the headers which doesn't make any assumptions about auto-including a set of default headers.

(Somewhat refs #1091)

tomchristie:
Closes #1226

Creating a `Request` instance directly has neater more minimal header behaviour.
Only includes mandatory headers.

```python
request = httpx.Request("POST", "http://example.org", json={"test": 123})

assert request.headers == httpx.Headers({
    "Host": "example.org",
    "Content-Type": "application/json",
    "Content-Length": "13",
})
```

Instead the default headers are on the client instance...

```python
>>> client = httpx.Client()
>>> print(client.headers)
Headers({'accept': '*/*', 'accept-encoding': 'gzip, deflate, br', 'connection': 'keep-alive', 'user-agent': 'python-httpx/0.14.2'})
```

Can include additional headers...

```python
>>> client = httpx.Client(headers={'Custom-Header': 'Custom'})
>>> print(client.headers)
Headers({'accept': '*/*', 'accept-encoding': 'gzip, deflate, br', 'connection': 'keep-alive', 'user-agent': 'python-httpx/0.14.2', 'custom-header': 'Custom'})
```

Or override existing headers...

```python
>>> client = httpx.Client(headers={'User-Agent': 'Custom'})
>>> print(client.headers)
Headers({'accept': '*/*', 'accept-encoding': 'gzip, deflate, br', 'connection': 'keep-alive', 'user-agent': 'Custom'})
```

Defaults may be removed, but must be done as an explicit `del`...

```python
>>> client = httpx.Client()
>>> del client.headers['User-Agent']
>>> print(client.headers)
Headers({'accept': '*/*', 'accept-encoding': 'gzip, deflate, br', 'connection': 'keep-alive'})
```

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -4,6 +4,7 @@ from types import TracebackType
 
 import httpcore
 
+from .__version__ import __version__
 from ._auth import Auth, BasicAuth, FunctionAuth
 from ._config import (
     DEFAULT_LIMITS,
@@ -17,6 +18,7 @@ from ._config import (
     create_ssl_context,
 )
 from ._content_streams import ContentStream
+from ._decoders import SUPPORTED_DECODERS
 from ._exceptions import (
     HTTPCORE_EXC_MAP,
     InvalidURL,
@@ -54,6 +56,10 @@ from ._utils import (
 logger = get_logger(__name__)
 
 KEEPALIVE_EXPIRY = 5.0
+USER_AGENT = f"python-httpx/{__version__}"
+ACCEPT_ENCODING = ", ".join(
+    [key for key in SUPPORTED_DECODERS.keys() if key != "identity"]
+)
 
 
 class BaseClient:
@@ -73,7 +79,7 @@ class BaseClient:
 
         self._auth = self._build_auth(auth)
         self._params = QueryParams(params)
-        self._headers = Headers(headers)
+        self.headers = Headers(headers)
         self._cookies = Cookies(cookies)
         self._timeout = Timeout(timeout)
         self.max_redirects = max_redirects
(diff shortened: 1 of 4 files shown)


florimondmanca (approved):
Very nice and sweet indeed!

PR #1249, opened 2020-09-02
Drop `request.timer` attribute.

tomchristie:
Minor bit of refactoring. We don't really want a `request.timer` attribute in our public API, exposing an internal `ElapsedTimer` instance.

Instead support a `Response(..., stream=..., get_elapsed_time=...)` parameter, which allows passing in a callable that is used to set `response.elapsed` when the response is closed. We don't particularly expect our users to need to use this, but it's a little neater and more tightly scoped than having an accessible timer on the *request* instance.

Also allows us to use a backend-native timer, appropriate to whichever context we're running in.


tests/models/test_responses.py
```diff
@@ -31,7 +30,6 @@ def test_response():
     assert response.text == "Hello, world!"
     assert response.request.method == "GET"
     assert response.request.url == "https://example.org"
-    assert response.elapsed >= datetime.timedelta(0)
```
florimondmanca, line 34: Haven't checked, do we still have some tests for the `response.elapsed` property?

florimondmanca (approved):
Nice bit of uncoupling. Left a sanity check question but this looks good!

tests/models/test_responses.py
tomchristie: Yup `test_client.py` and `test_async_client.py` check the property, and `test_responses.py` has a test to ensure that it's not accessible until the response has been closed.

florimondmanca (dismissed):

tomchristie:
Before we merge this, do we think we've got the right name?
Some options might be...

* `get_elapsed_time=...` *Currently using, in this PR*
* `elapsed_timer=...`
* `elapsed_callable=...`
* `elapsed_time=...`
* `elapsed_func=...`

Anything else? Which of those is preferable?

florimondmanca:
> `elapsed_func=...`

Yup, I think I prefer this one better. (True that `.get_elapsed_time` looks like a method rather than an externally-passed argument.)

PR #1250, opened 2020-09-02
Add support for `httpx.Response(content=..., text=..., html=..., json=...)`

### LINKED ISSUE #1227: Supporting `Response(content=..., text=..., html=..., json=...)`

tomchristie:
Closes #1227

Adds support for eg...

* `httpx.Response(200, content=b'Hello, world')`
* `httpx.Response(200, text='Hello, world')`
* `httpx.Response(200, html='Hello, world')`
* `httpx.Response(200, json={'Hello': 'world'})`

Plus streaming cases...

* `httpx.Response(200, content=<byte iterator>)`
* `httpx.Response(200, content=<byte aiterator>)`

There are some test cases we'd be able to smarten up after this, but I'll leave that for a follow up.

### DIFF
--- a/httpx/_content_streams.py
+++ b/httpx/_content_streams.py
@@ -72,11 +72,8 @@ class IteratorStream(ContentStream):
     Request content encoded as plain bytes, using an byte iterator.
     """
 
-    def __init__(
-        self, iterator: typing.Iterator[bytes], close_func: typing.Callable = None
-    ) -> None:
+    def __init__(self, iterator: typing.Iterator[bytes]) -> None:
         self.iterator = iterator
-        self.close_func = close_func
         self.is_stream_consumed = False
 
     def can_replay(self) -> bool:
@@ -95,21 +92,14 @@ class IteratorStream(ContentStream):
     def __aiter__(self) -> typing.AsyncIterator[bytes]:
         raise RuntimeError("Attempted to call a async iterator on an sync stream.")
 
-    def close(self) -> None:
-        if self.close_func is not None:
-            self.close_func()
-
 
 class AsyncIteratorStream(ContentStream):
     """
     Request content encoded as plain bytes, using an async byte iterator.
     """
 
-    def __init__(
-        self, aiterator: typing.AsyncIterator[bytes], close_func: typing.Callable = None
-    ) -> None:
+    def __init__(self, aiterator: typing.AsyncIterator[bytes]) -> None:
         self.aiterator = aiterator
-        self.close_func = close_func
         self.is_stream_consumed = False
 
     def can_replay(self) -> bool:
@@ -127,10 +117,6 @@ class AsyncIteratorStream(ContentStream):
         self.is_stream_consumed = True
         async for part in self.aiterator:
             yield part
-
-    async def aclose(self) -> None:
-        if self.close_func is not None:
-            await self.close_func()
 
 
 class JSONStream(ContentStream):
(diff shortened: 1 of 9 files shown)


httpx/_content_streams.py
```diff
@@ -95,21 +92,14 @@ def __iter__(self) -> typing.Iterator[bytes]:
     def __aiter__(self) -> typing.AsyncIterator[bytes]:
         raise RuntimeError("Attempted to call a async iterator on an sync stream.")
 
-    def close(self) -> None:
-        if self.close_func is not None:
-            self.close_func()
```
tomchristie, line 100: Turns out we don't *really* need `close_func` on `IteratorStream`/`AsyncIteratorStream` anymore, which falls out from updating test cases to use `Response(content=<bytes iterator>)`

httpx/_transports/urllib3.py
```diff
@@ -15,6 +15,21 @@
...
+
+    def __iter__(self) -> Iterator[bytes]:
+        try:
+            for chunk in self._conn.stream(4096, decode_content=False):
+                yield chunk
+        except socket.error as exc:
+            raise httpcore.NetworkError(exc) from exc
+
+    def close(self) -> None:
+        self._conn.release_conn()
```
tomchristie, line 30: Changes in this module are redundant given #1182, but I've updated this all the same since this PR removes `close_func` from our internal `IteratorStream`.

tomchristie:
If we need to narrow down the scope of this for review purposes, I'd be happy to think about how best to do that.

tomchristie:
Putting this in draft to instead tackle the intent here in a more targeted manner.
See https://github.com/encode/httpx/pull/1265 for a more slimmed-down approach.

tomchristie:
Closing in favour of #1297

PR #1254, opened 2020-09-03
Set message attribute in all exceptions

### LINKED ISSUE #1252: Docs incorrectly reference `HTTPError.message` attribute
### Checklist

### Describe the bug

The documentation indicates that message field can be used in the HTTPError:
https://github.com/encode/httpx/blob/master/httpx/_exceptions.py#L54
```
    try:
        response = httpx.get("https://www.example.com")
        response.raise_for_status()
        print(f"HTTP Exception for {exc.request.url} - {exc.message}")
```
But there is not such field:
```
AttributeError: 'HTTPStatusError' object has no attribute 'message'
```

### To reproduce

Execute the example from the doc

### Expected behavior

Print the string without raising any exceptions

### Actual behavior

AttributeError is raised

### Possible fixes

1. Update the documentation to use `str(exc)` instead of `exc.message`
or
2. Set `self.message` field in HTTPError

tsotnikov:
Fixes #1252

### DIFF
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -239,6 +239,7 @@ class InvalidURL(Exception):
 
     def __init__(self, message: str) -> None:
         super().__init__(message)
+        self.message = message
 
 
 class NotRedirectResponse(Exception):
@@ -251,6 +252,7 @@ class NotRedirectResponse(Exception):
 
     def __init__(self, message: str) -> None:
         super().__init__(message)
+        self.message = message
 
 
 class CookieConflict(Exception):
@@ -262,6 +264,7 @@ class CookieConflict(Exception):
 
     def __init__(self, message: str) -> None:
         super().__init__(message)
+        self.message = message
 
 
 # Stream exceptions...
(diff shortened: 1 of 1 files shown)


magiskboy (approved):

florimondmanca (commented):
Heya!

I think I'm in line with @tomchristie's comment here - https://github.com/encode/httpx/issues/1252#issuecomment-686761772

Given that we don't expose an `exc.message` attribute in any other exception class, we should probably just update the docstring here to say:

```python
print(f"HTTP Exception for {exc.request.url} - {exc}")
```

tsotnikov:
From @tomchristie comment I understood that we prefer to set and expose `exc.message` as message is mandatory in HTTPError.

florimondmanca:
You're right, I misread his initial recommendation. Well in that case I'll have to disagree, since given that we don't expose a `.message` attribute anywhere else, I do think just updating the docs is the best low-touch approach. :-)

tsotnikov:
Ok, so we have two alternatives or add "message" attribute to all base exception or fix the doc.

Feel free to choose between this PR and #1255  and close whichever you discard

tomchristie:
Thanks all, neat resolution to this one. 👍

PR #1255, opened 2020-09-04
Fix HTTPError message attribute doc

### LINKED ISSUE #1252: Docs incorrectly reference `HTTPError.message` attribute

tsotnikov:
Fixes #1252 

Choose this PR or #1254

### DIFF
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -55,7 +55,7 @@ class HTTPError(Exception):
         response = httpx.get("https://www.example.com")
         response.raise_for_status()
     except httpx.HTTPError as exc:
-        print(f"HTTP Exception for {exc.request.url} - {exc.message}")
+        print(f"HTTP Exception for {exc.request.url} - {exc}")
     ```
     """
 


StephenBrown2 (approved):

tomchristie:
Good comment from @StephenBrown2 on this. Also makes sense not to expose `.message` as in the general case `str(exc)` can sometimes provide more information in addition to the raw message.

Eg suppose we added a `history=...` parameter to `TooManyRedirects`, then printing `exc.message` would lose that info, while `str(exc)` could display it as extra context alongside the message.

Issue #404, opened 2019-09-27
Environment Variables docs should justify each usage case.


tomchristie:
Our [environment variable documentation](https://www.encode.io/httpx/environment_variables/) should include some justification or context for each case, by doing one of the following for each envvar:

* Make it clear that an environment variable is specific to HTTPX. (Probably only relevant to anything prefixed by `HTTPX_`)
* Make it clear that an environment variable is used as a convention by various tooling, with at least one relevant hyperlink.

Eg.

`SSLKEYLOGFILE` could give some more context about usage - "The keylog file is designed for debugging purposes only." doesn't make it clear how or when to use it, or which debug tooling it's usful with.

`SSL_CERT_FILE` and `SSL_CERT_DIR` can include a link to some docs about their convention. Perhaps give an example of other tooling that supports them? 

`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` can include a link to some docs about their convention. Perhaps give an example of other tooling that supports them?

anshulxyz:
Hi @sethmlarson and @tomchristie I'd like to work on this

mwoss:
May I take over it?

tomchristie:
We'll happily take any contributions here yup.

mwoss:
Ok, I will take care of it tommorow or on Friday.

cwkang1998:
Is this task done? Does it still need help working on?

tomchristie:
I don't think there's anything worthwhile doing here, nope.

PR #1257, opened 2020-09-04
Edited documentation about proxy-envs usage (#404)

cdeler:
In #404 I found that proxies parameters might be commented better

I added NO_PROXY to the docs, also I added a couple of examples (which describes setting proxy using envs)

Also I added two links to cURL and requests, which use the same vars to configure proxy


florimondmanca:
Hello @cdeler! For traceability and reviewing purposes, could you add a PR description to clarify what this PR intends to do and what the motivation/context is? Makes it easier to know if anything is missing or could be done differently. Thanks!

cdeler:
Thank you @florimondmanca 

I updated the pr description, sorry that I missed that

docs/environment_variables.md
```diff
@@ -139,18 +139,31 @@ Example:
 SSL_CERT_DIR=/path/to/ca-certs/ python -c "import httpx; httpx.get('https://example.com')"
 ```
 
-`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`
+`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY`
```
florimondmanca, line 142: We might want to consider using a separate section for `NO_PROXY` (?), since it actually accepts a comma-separated list of URLs, and those URLs denote hosts, rather than proxies.

We have shared information about proxies in the existing section for now, but it's perfectly possible to move that as an introduction to a general "Proxies" section in this docs page.

So we can consider a structure like this…

```markdown
# Environment Variables

…

## Proxies

The environment variables documented below are used as a convention by various HTTP tooling, including:

* [cURL](https://github.com/curl/curl/blob/master/docs/MANUAL.md#environment-variables)
* [requests](https://github.com/psf/requests/blob/master/docs/user/advanced.rst#proxies)

For more information on using proxies in HTTPX, see [HTTP Proxying](/advanced/#http-proxying).

### `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`

Valid values: A URL to a proxy

...

### `NO_PROXY`

Valid values: a comma-separated list of hostnames
...
```

What do you think?

docs/environment_variables.md
```diff
@@ -139,18 +139,31 @@ Example:
...
+python -c "import httpx; httpx.get('https://www.python-httpx.org')"
+
+# This request will be sent directly, as we set `trus_env=False`
+python -c "import httpx; httpx.get('http://example.com', trust_env=False)"
 ```
 
 For more information on using proxies in HTTPX, see [HTTP Proxying](/advanced/#http-proxying).
+
+The other tools compatible with these variables convention: [cURL](https://github.com/curl/curl/blob/master/docs/MANUAL.md#environment-variables),
+[requests](https://github.com/psf/requests/blob/master/docs/user/advanced.rst#proxies).
```
florimondmanca, line 169: I think the main idea is there (namely "these env vars are conventional"), but leaving a copy improvement suggestion below:

```suggestion
These environment variables are used as a convention by various HTTP tooling, including:

* [cURL](https://github.com/curl/curl/blob/master/docs/MANUAL.md#environment-variables)
* [requests](https://github.com/psf/requests/blob/master/docs/user/advanced.rst#proxies)
```

docs/environment_variables.md
```diff
@@ -139,18 +139,31 @@ Example:
...
 
 ```bash
-export HTTP_PROXY=http://127.0.0.1:3080
+export HTTP_PROXY=http://my-external-proxy.com:1234
+export NO_PROXY=http://127.0.0.1,python-httpx.org
 
 # This request will be sent through the proxy
 python -c "import httpx; httpx.get('http://example.com')"
+
+# This request will be sent directly, bypassing the proxy
```
florimondmanca, line 158:
```suggestion
# These requests will be sent directly, bypassing the proxy
```

florimondmanca (commented):
Great start! Left a few comments :-)

docs/environment_variables.md
florimondmanca: By the way — I pushed https://github.com/encode/httpx/pull/1260 that we should review and merge before-hand so we can use the `###` style of headers on this page, like we do everywhere else. Happy to take reviews!

cdeler: Changed the doc according your comment.

Thank you for the advice about the splitting the paragraph into two

florimondmanca (approved):
Superb, thanks!

docs/environment_variables.md
```diff
@@ -134,17 +134,47 @@ Example:
...
+`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` set the proxy to be used for `http`, `https`, or all requests respectively.
 
 ```bash
-export HTTP_PROXY=http://127.0.0.1:3080
+export HTTP_PROXY=http://my-external-proxy.com:1234
 
 # This request will be sent through the proxy
 python -c "import httpx; httpx.get('http://example.com')"
+
+# This request will be sent directly, as we set `trus_env=False`
```
florimondmanca, line 158: Small typo :-)

```suggestion
# This request will be sent directly, as we set `trust_env=False`
```

PR #1259, opened 2020-09-04
Type error on invalid url type

### LINKED ISSUE #1258: Confusing error messages when passing in the wrong URL type
https://github.com/encode/httpx/blob/642aabdac093cf9798f4881cbdd8b39bd3398bb5/httpx/_models.py#L74

Say, for example, we're using a Starlette API which features a URL type: https://github.com/encode/starlette/blob/c566fc6c819f0d565f8cff432351fe009e83d866/starlette/datastructures.py#L14

And you might mistakenly think you could pass that URL type in and use it, but of course you'd be wrong because:
```
class URL:
    def __init__(self, url: URLTypes = "", params: QueryParamTypes = None) -> None:
        if isinstance(url, str):
            ...
        else:
            self._uri_reference = url._uri_reference
```

It is not a string, but also does not have a `._uri_reference` field.

The resulting error we got, which was hard for us to debug was:
```
'URL' object has no attribute '_uri_reference'
  ...
  File <redacted>, line 27, in <redacted>
    async with AsyncClient(base_url=settings.service_baselink) as client:
  File "<redacted>/lib/python3.7/site-packages/httpx/_client.py", line 1125, in __init__
    trust_env=trust_env,
  File "<redacted>/lib/python3.7/site-packages/httpx/_client.py", line 72, in __init__
    self._base_url = self._enforce_trailing_slash(URL(base_url))
  File "<redacted>/lib/python3.7/site-packages/httpx/_models.py", line 74, in __init__
    self._uri_reference = url._uri_reference
AttributeError: 'URL' object has no attribute '_uri_reference'
```

I would suggest that this block be changed to have a further check on the appropriate expected types and then raise a type error rather than an attribute error here.

wozniakty:
Closes #1258 

Note: first contribution, so let me know if I'm missing anything

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -71,8 +71,10 @@ class URL:
                 # We don't want to normalize relative URLs, since doing so
                 # removes any leading `../` portion.
                 self._uri_reference = self._uri_reference.normalize()
-        else:
+        elif isinstance(url, URL):
             self._uri_reference = url._uri_reference
+        else:
+            raise TypeError("Invalid type for url.  Expected str or httpx.URL")
 
         # Add any query parameters, merging with any in the URL if needed.
         if params:
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -74,7 +74,9 @@ class URL:
         elif isinstance(url, URL):
             self._uri_reference = url._uri_reference
         else:
-            raise TypeError("Invalid type for url.  Expected str or httpx.URL")
+            raise TypeError(
+                f"Invalid type for url.  Expected str or httpx.URL, got {type(url)}"
+            )
 
         # Add any query parameters, merging with any in the URL if needed.
         if params:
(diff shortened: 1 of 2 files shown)


httpx/_models.py
```diff
@@ -71,8 +71,10 @@ def __init__(self, url: URLTypes = "", params: QueryParamTypes = None) -> None:
                 # We don't want to normalize relative URLs, since doing so
                 # removes any leading `../` portion.
                 self._uri_reference = self._uri_reference.normalize()
-        else:
+        elif isinstance(url, URL):
             self._uri_reference = url._uri_reference
+        else:
+            raise TypeError("Invalid type for url.  Expected str or httpx.URL")
```
florimondmanca, line 77: (Nit) Consider showing the received type of `url`

```suggestion
            raise TypeError(f"Expected url to be an str or httpx.URL, got {type(url)}")
```

florimondmanca (approved):
Neat little contribution — definitely see the point in preventing users from using other types named `URL` from related libraries.

A nit, but approving in advance…

httpx/_models.py
wozniakty: @florimondmanca great suggestion, thanks 👍

florimondmanca (approved):
Thank you!

PR #1260, opened 2020-09-04
Use shebang-style headers in environment variables docs

florimondmanca:
Tiny docs refactor PR to change the "underline" headers to more usual "shebang" style for headers in our "Environment Variables" docs page (like we do everywhere lese).

Came to me while reviewing https://github.com/encode/httpx/pull/1257


cdeler (approved):
Thank you, it will be helping to organize more complex heading structure (what you proposed in #1257)

Issue #1261, opened 2020-09-04
Incompatibility: param that is None is still included 


louwers:
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Your docs state 

> HTTPX aims to be compatible with the requests API wherever possible.

so I am reporting this as a bug. When `None` is one of the parameter values (provided with `params`), it is still included. While this is not the case for requests.

### To reproduce

```
>>> import requests
>>> requests.get("https://httpbin.org/get", params={"foo": "bar", "baz": None}).json()["url"]
'https://httpbin.org/get?foo=bar'
```

versus

```
>>> import httpx
>>> import asyncio
>>>
>>> async def get_httpbin():
>>>     async with httpx.AsyncClient() as client:
>>>         r = await client.get("https://httpbin.org/get", params={"foo": "bar", "baz": None})
>>>         print(r.json()["url"])
>>> asyncio.run(get_httpbin())
https://httpbin.org/get?foo=bar&baz=
```

### Expected behavior

A request to `https://httpbin.org/get?foo=bar` is made.

### Actual behavior

A request to `https://httpbin.org/get?foo=bar&baz=` is made.

florimondmanca:
Felt like this had been discussed before, and surely, it has:

This is essentially a duplicate of #154

At the time was closed by #174 in the name of adhering to strict typing, as well as "Requests isn't doing the best thing there" — hence why we're differing in API here.

As a note: `None` values _could_ have been reserved to mean "include a query param without `=` (such as `{"q": None}` -> `q`, instead of `q=` for `{"q": ""}`), a use case that Requests would want to address but can't due to the existing behavior you're asking about here - see psf/requests#2651. Adding this was deemed as an unecessary feature though (for such esoteric cases it's probably okay to just build the query string manually).

So, closing as a duplicate — thanks!

PR #1262, opened 2020-09-04
Update compatibility.md: mention differing query parameter handling

louwers:
#1261


docs/compatibility.md
```diff
@@ -1,6 +1,6 @@
 # Requests Compatibility Guide
 
-HTTPX aims to be compatible with the `requests` API wherever possible.
+HTTPX aims to be compatible with the `requests` API where we feel it makes sense.
```
florimondmanca, line 3: I know that probably wasn't the intent, but for some reason this phrasing sounds a bit pedantic — like "we" consider that Requests' API was nonsensical in places (when really it's just a result of pioneering the field and having to live on with early decisions, which is understandable).

I also thought of replacing with "wherever sensible", but that seems to carry the same possible misunderstanding.

So I'd be keen to keep the current wording…

```suggestion
HTTPX aims to be compatible with the `requests` API where possible.
```

florimondmanca (commented):
Ah, great point updating our docs to mention the differing behavior we discussed in #1261, thanks!

docs/compatibility.md
```diff
@@ -89,3 +89,7 @@ If you need to mock HTTPX the same way that test utilities like `responses` and
 `requests` defers most of its HTTP networking code to the excellent [`urllib3` library](https://urllib3.readthedocs.io/en/latest/).
 
 On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as its core HTTP networking layer, which is a different project than `urllib3`.
+
+## Query Parameters
+
+`requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). HTTPX will include a query parameter without a value instead.
```
tomchristie, line 95: I'd probably(?) rather we were raising a TypeError here, since our annotations don't include None as a valid value for items in params. It's only by accident of implementation that it happens to currently end up being treated as equivalent to `""`.

florimondmanca: Makes sense — still, we can add these docs, probably just tweaking them like this:

```suggestion
`requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). This is not supported by HTTPX.
```

docs/compatibility.md
louwers, line 3: How about

```suggestion
HTTPX aims to be broadly compatible with the `requests` API.
```

I feel like the current wording creates the impression that HTTPX tries to be compatible unless that's not possible at all.

tomchristie: "Broadly compatible" is spot-on, yup.

tomchristie (approved):

florimondmanca (approved):

Issue #1253, opened 2020-09-03
Exposing the content stream API


tomchristie:
While we've pretty much got all of our public/private API properly demarcated now, there's one aspect where that's still not quite true, which is the content stream API. We currently expose `ContentStream` parameters or properties in the following places...

* `Request(..., stream=<ContentStream>)` - Exposed in order to copy + modify a request eg. on redirect or digest auth.
* `Response(..., stream=<ContentStream>)` - Exposed in order to pass from the transport layer when constructing the response.
* `request.stream` - Exposed in order to pass to the transport layer when sending the request, or to copy + modify a request eg. on redirect or digest auth.

We're also a bit wooly about if we *actually* mean an `httpcore.AsyncByteStream`/`httpcore.SyncByteStream` or an `httpx.ContentStream` in places where that's exposed. Furthermore, we actually *would* like to expose content streams for some advanced functionality, like being able to support streaming the content of a file, while providing replay support.

Eg...

```python
def upload_file(name):
    with open(name, "rb") as input_file:
        while True:
            chunk = input_file.read(4096)
            if not chunk:
                break
            yield chunk

# This will fail on some kinds of redirect, or on digest auth
httpx.post(data=upload_file("example.zip"))
```

But if we exposed the content stream API then we could support something like...

```python
class UploadFile(httpx.ContentStream):
    def __init__(self, name):
        self.name = name

    def __iter__(self):
        with open(self.name, "rb") as input_file:
            while True:
                chunk = input_file.read(4096)
                if not chunk:
                    break
                yield chunk

# This can replay, if needed
httpx.post(data=UploadFile("example.zip"))
```

Because we're a bit fuzzy around the `ContentStream` vs. `httpcore.AsyncByteStream`/`httpcore.SyncByteStream` I think we also need an API that's more clear there, and doesn't have this odd artifact of `.can_replay()` and `.get_headers()` methods, that are used in some contexts, but not in others.

Here's what I think we need to do here.

* Change the `httpcore` API, so that we *only* have `httpcore.ByteStream`, which is an abstract base class with for a unified sync+async byte stream. Ie. it exposes `__iter__`, `__aiter__`, `close`, `aclose` methods.
[...]

PR #1265, opened 2020-09-07
Support `Response(content=<bytes iterator>)`

tomchristie:
Refs #1253

Support `Response(content=<bytes iterator>)` and `Response(content=<bytes aiterator>)`.

Related follow ups will be...

* Adding support for `Response(content=<byte stream>)`
* Adding support for `Request(content=<bytes|bytes iterator|bytes aiterator|byte stream)`
* Deprecating `Response(stream=...)` and `Request(stream=...)`

A good motivation when considering this pull request in isolation, is to look over the test cases that were previously importing from the private API - `from httpx._content_streams import AsyncIteratorStream` and `from httpx._content_streams import IteratorStream`

For instance we've got cases that currently look like this...

```python
# Bit grungy, since `IteratorStream` is a private implementation detail inside `httpx`
stream = IteratorStream(iterator=streaming_body())
response = httpx.Response(200, stream=stream)
```

After this pull request they get slimmed down to a more natural looking API...

```python
response = httpx.Response(200, content=streaming_body())
```


tomchristie:
I'm going to merge this one in, since I don't think it's all that controversial, and it's blocking some other work from progressing at the moment.

I *think* it makes sense to use the merge-without-a-review admin privileges occasionally, so long as it's treated carefully like this (eg. controversial PRs, that have had a few days breathing room), since practically I've got a lot more time available to this right now than the rest of the maintenance team, so we're getting a bit bottle necked.

(But happy to talk about any of that if needed.)

httpx/_content_streams.py
```diff
@@ -402,3 +388,18 @@ def encode(
...
+
+
+def encode_response(content: ResponseContent = None) -> ContentStream:
+    if content is None:
+        return ByteStream(b"")
+    elif isinstance(content, bytes):
+        return ByteStream(body=content)
+    elif hasattr(content, "__aiter__"):
+        content = typing.cast(typing.AsyncIterator[bytes], content)
+        return AsyncIteratorStream(aiterator=content)
```
JayH5, line 400: Maybe you were already aware, but a thing I just learned from reviewing encode/starlette#1041: `collections.abc.AsyncIterator`/`collections.abc.Iterator` (which the `typing` equivalents derive from) have [a subclasshook](https://github.com/python/cpython/blob/v3.8.5/Lib/_collections_abc.py#L185-L189) which I think means you could just do:
```python
    elif isinstance(content, typing.AsyncIterator):
          return AsyncIteratorStream(aiterator=content)
```

Also _technically_ you may be casting Iterables to _Iterators_ here by just checking for the presence of `__aiter__`, though it seems unlikely to cause problems.

JayH5: @tomchristie

PR #1266, opened 2020-09-07
Seperate `content=...` and `data=...` parameters

### LINKED ISSUE #1210: Separate `httpx.post(content=...)` from `httpx(data=...)`?
Currently we're following the `requests` API, with `data=...` representing one of:

* `str`/`bytes`/`byte iterator`/`byte aiterator` - Content to send in the body of the request.
* `dict` - Data to send in either a URLEncoded or Multipart request.

These are really two very different types of cases. One is for submitting HTML form data, the other is for uploading raw byte content. It's possible that we should consider if we want to diverge from `requests` here, in order to provide a more clear, consistent interface?

For example...

```python
# Send a POST request, with "abc" as the request body.
httpx.post(content=b'abc')

# Send a POST request, with URLEncoded data.
httpx.post(data={"field": "value"})

# Send a POST request, with MultiPart encoded data.
httpx.post(data={"field": "value"}, files=...)
```

Would be interested to see what others think of this?...

(Obviously we'd want a gradual deprecation if we *did* decide to take this on.)

tomchristie:
Closes #1210 

Adds support for eg. `httpx.post(..., content=<bytes|byte iterator|byte aiterator>)` for uploading raw content.
This is distinct from `httpx.post(..., data=<dict>)` for sending form data. 

See also #1265.


tomchristie:
Particularly now that we have `Response(content=...)` I'd like us to press on with this one.

PR #1267, opened 2020-09-07
Swap auth/redirects ordering

tomchristie:
Internal refactoring to swap the order in which we handle auth/redirects.

Needed because `request` event hooks ought to be triggered after any initial authentication is applied to the request, but should not re-occur after redirects.

The flow also just generally makes more sense to be now:

Create the request -> Apply any authentication -> Follow any redirects -> Send the request to the network

For the sake of reviews I haven't re-ordered the methods, but I'll issue a follow-up PR to this which will swap them around so that the code ordering follows the execution flow.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -725,8 +725,12 @@ class Client(BaseClient):
 
         auth = self._build_request_auth(request, auth)
 
-        response = self._send_handling_redirects(
-            request, auth=auth, timeout=timeout, allow_redirects=allow_redirects
+        response = self._send_handling_auth(
+            request,
+            auth=auth,
+            timeout=timeout,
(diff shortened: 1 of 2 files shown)


tomchristie:
Actually this is more that just a "refactor" - it's actually a bug fix.

For example... currently we strip any `Authorization` header on cross-domain redirects, so that if we get redirected off-site, we don't blindly trust the new origin with our auth credentials.

However, we're not doing the same behaviour if `auth=(<user>, <pass>)` or `auth=httpx.BasicAuth(...)` is used. This PR now includes a test case that is failing against master, but is not failing here.

florimondmanca (approved):
Yes, that makes total sense after thinking it through!

Issue #1208, opened 2020-08-21
Monitoring download progress.


tomchristie:
For clients that would like to display something like a progress bar for downloads, there's an issue in that while you're streaming data from `iter_content()`, you *can't* properly know how far you have progressed through the download because we'll often be performing gzip decompression. The `Content-Length` header gives the length of the raw data. (And we really *don't* want to be streaming from `iter_raw()` because we really *do* want the uncompressed bytes, not the bytes on the wire.)

See also for example, this question related to `requests`... https://stackoverflow.com/questions/21778885/python3-progress-bar-and-download-with-gzip

It might make sense to provide an API for monitoring the number of bytes that have been consumed by the underlying `iter_raw` calls, for example something like... `response.num_bytes_downloaded`.

```python
with open("download.txt", "wb") as download_file:
    with httpx.stream('http://www.example.org') as response:
        content_length = response.headers.get("Content-Length")
        for chunk in response.iter_content():
            download_file.write(chunk)
            if content_length is not None:
                percent_complete = 100.0 * (response.num_bytes_downloaded / content_length)
                print(f"{percent_complete:.2f}%")
```

Perhaps we ought to provide a `response.percent_complete` that wraps this up nicely, returning either a 0.0 <= float <= 100.0 or `None` if the `Content-Length` is unknown, so...

```python
with open("download.txt", "wb") as download_file:
    with httpx.stream('http://www.example.org') as response:
        for chunk in response.iter_content():
            download_file.write(chunk)
            if response.percent_complete is not None:
                print(f"{response.percent_complete:.2f}%")
```

florimondmanca:
Regardless of the API we decide on, one way to challenge it and make sure it's right enough for most use cases is to see if it would integrate well with [tqdm](https://github.com/tqdm/tqdm)...

And, well, the way I see it there isn't really a need for us to add anything, since `tqdm` supports a manual [`.update()`](https://tqdm.github.io/docs/tqdm/#update)-based usage that was specifically designed for this kind of "read from a known-length stream" use cases.

For example:

```python
import httpx
from tqdm import tqdm

with open("download.txt", "wb") as download_file:
    data = b"*" * 1000000
    with httpx.stream("POST", "https://httpbin.org/anything", data=data) as response:
        content_length = response.headers["Content-Length"]
        with tqdm(total=int(content_length)) as progress:
            for chunk in response.iter_bytes():
                download_file.write(chunk)
                progress.update(len(chunk))
```

Output:

```console
[...]

PR #1268, opened 2020-09-08
Add progress to streaming download

cdeler:
Having found #1208, I tried to track last downloaded raw chunk of the response.

The question is:
1. should we also track decoded response chunks? I don't see any reason since decoded stream length might be vary from the raw stream length, and it has no bearing with downloading process (but may be I missed something)

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -236,10 +236,7 @@ from tqdm import tqdm
 with tempfile.NamedTemporaryFile() as download_file:
     url = "https://speed.hetzner.de/100MB.bin"
     with httpx.stream("GET", url) as response:
-        if "Content-Length" in response.headers:
-            total = int(response.headers["Content-Length"])
-        else:
-            total = None
+        total = int(response.headers["Content-Length"])
 
         with tqdm(total=total, unit_scale=True, unit_divisor=1024, unit="B") as progress:
             num_bytes_downloaded = response.num_bytes_downloaded
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -234,13 +234,20 @@ import httpx
 from tqdm import tqdm
 
 with tempfile.NamedTemporaryFile() as download_file:
-    data = b"@" * 1000000
-    with httpx.stream("POST", "https://httpbin.org/anything", data=data) as response:
-        content_length = int(response.headers["Content-Length"])
-        with tqdm(total=content_length) as progress:
+    url = "https://speed.hetzner.de/100MB.bin"
+    with httpx.stream("GET", url) as response:
+        if "Content-Length" in response.headers:
+            total = int(response.headers["Content-Length"])
+        else:
+            total = None
+
+        with tqdm(total=total, unit_scale=True, unit_divisor=1024, unit="B") as progress:
+            num_bytes_downloaded = response.num_bytes_downloaded
             for chunk in response.iter_bytes():
                 download_file.write(chunk)
-                progress.update(response.last_raw_chunk_size)
+                progress.update(response.num_bytes_downloaded - num_bytes_downloaded)
+                num_bytes_downloaded = response.num_bytes_downloaded
+        print(f"The total download size is {response.num_bytes_downloaded} bytes")
 ```
 
 ## .netrc Support
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -223,7 +223,7 @@ with httpx.Client(headers=headers) as client:
 
 ## Monitoring download progress
 
-If you need to monitor download progress of large responses, you can use stream and the `response.last_raw_chunk_size` property.
+If you need to monitor download progress of large responses, you can use stream and the `response.num_bytes_downloaded` property.
 
 For example, showing a progress bar using the [`tqdm`](https://github.com/tqdm/tqdm) library while a response is being downloaded could be done like this…
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -223,7 +223,7 @@ with httpx.Client(headers=headers) as client:
 
 ## Monitoring download progress
 
-If you need to monitor download progress of large responses, you can use stream and the `response.num_bytes_downloaded` property.
+If you need to monitor download progress of large responses, you can use response streaming and inspect the `response.num_bytes_downloaded` property.
 
 For example, showing a progress bar using the [`tqdm`](https://github.com/tqdm/tqdm) library while a response is being downloaded could be done like this…
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
     ...
 ```
 
+## Monitoring download progress
+
+If you need to monitor download progress, you can stream with using `response.last_raw_chunk_size` property.
+
+For example, you can build a nice progress bar using the `tqdm` library:
+
+```python
+import tempfile
+
+import httpx
+from tqdm import tqdm
+
+with tempfile.NamedTemporaryFile() as download_file:
+    data = b"@" * 1000000
+    with httpx.stream("POST", "https://httpbin.org/anything", data=data) as response:
+        content_length = int(response.headers["Content-Length"])
+        with tqdm(total=content_length) as progress:
+            for chunk in response.iter_bytes():
+                download_file.write(chunk)
+                progress.update(response.last_raw_chunk_size)
+```
+
 ## .netrc Support
 
 HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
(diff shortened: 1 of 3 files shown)


docs/advanced.md
```diff
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
     ...
 ```
 
+## Monitoring download progress
+
+If you need to monitor download progress, you can stream with using `response.last_raw_chunk_size` property.
```
florimondmanca, line 226: (Nit) Copy improvement:

```suggestion
If you need to monitor download progress of large responses, you can use stream and the `response.last_raw_chunk_size` property.
```

docs/advanced.md
```diff
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
...
+
+For example, you can build a nice progress bar using the `tqdm` library:
```
florimondmanca, line 228: (Nit) Copy improvement.

(Super nit) Not sure we need the "nice" adjective so I replaced it with a more neutral phrasing below :)

```suggestion
For example, showing a progress bar using the [`tqdm`](https://github.com/tqdm/tqdm) library while a response is being downloaded could be done like this…
```

florimondmanca (commented):
Still need to ponder on this a bit, but very neat implementation (code, docs and tests)!

florimondmanca:
> should we also track decoded response chunks?

Don't think so, since that's directly available as `len(chunk)` already.

tomchristie:
Great, thanks!

I think this has helped me. understand why I'm a bit wary of `response.last_raw_chunk_size` as the interface here, rather than `response.num_bytes_downloaded` (Or something similarly named.)

It seems to me that it's more likely to lead to brittle code. Also it's not super clear what the behaviour is expected to be in different cases. Eg...

* I don't think it'd work as expected if a user happened to be using `iter_lines()`.
* It's not clear how it'd be expected to interact with #1271

In contrast if we used `response.num_bytes_downloaded` then there's just less to get wrong.
Also examples like this work, where `.last_raw_chunk_size` wouldn't...

```python
response = httpx.get(...)
response.num_bytes_downloaded  # Determine the final download size.
```

It is true that `num_bytes_downloaded` happens to be *marginally* more work in the `tqdm` example, but it's not much more to deal with.

```python
num_bytes_downloaded = 0
with httpx.stream("...") as response:
    total = ...
    with tqdm(total=...) as progress:
        for chunk in response.iter_bytes():
            download_file.write(chunk)
            progress.update(response.num_bytes_downloaded - num_bytes_downloaded)
            num_bytes_downloaded = response.num_bytes_downloaded
```

What do we think?

docs/advanced.md
```diff
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
...
+
+```python
+import tempfile
+
+import httpx
+from tqdm import tqdm
+
+with tempfile.NamedTemporaryFile() as download_file:
+    data = b"@" * 1000000
+    with httpx.stream("POST", "https://httpbin.org/anything", data=data) as response:
```
tomchristie, line 238: I think the example given is a bit misleading, since it's the *download* that we're tracking the progress of here, not the upload. Perhaps just a GET of a largish file would be a little more clear?...

```python
url = "http://ipv4.download.thinkbroadband.com/100MB.zip"
with httpx.stream("GET", url) as response:
    ...
```

docs/advanced.md
```diff
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
...
+        content_length = int(response.headers["Content-Length"])
```
tomchristie, line 239: I *guess* we ought to also cover the case where `Content-Length` isn't present.

Perhaps something like:

```python
if "Content-Length" in response.headers:
    total = int(response.headers["Content-Length"])
else:
    total = None
```

docs/advanced.md
```diff
@@ -221,6 +221,28 @@ with httpx.Client(headers=headers) as client:
...
+        with tqdm(total=content_length) as progress:
```
tomchristie, line 240: I'm interested what arguments we need to display here to have tqdm display this using B, KB, MB, GB etc...

From https://github.com/tqdm/tqdm#extra-cli-options is sounds like the API might be this?...

```python
with tqdm(total=total, unit_scale=True, unit_divisor=1024, unit='B') as progress:
    ...
```

But it'd need testing to be sure.

tomchristie (commented):
Lovely bit of functionality to work towards yup!
Really liking the docs example here.
Possible that we ought to be going for a total downloaded bytes API rather than a chunk size API?

docs/advanced.md
cdeler: Good point,

I changed the examples (but I used the test file from this page https://speed.hetzner.de/)

docs/advanced.md
cdeler: I didn't think about that, but you are right

docs/advanced.md
cdeler: I checked this example, it looks fine, something like that:
```
100%|██████████████| 100M/100M [01:54<00:00, 918kB/s]
```

cdeler:
@tomchristie 

I changed `last_raw_chunk_size` to `num_bytes_downloaded`.

For now we can state that there is a correlation between the amount of data (lets say `N`) drained from `iter_raw(chunk_size=123)` and `num_bytes_downloaded`. I.e. until the end of the stream we can assert that `N <= response.num_bytes_downloaded`. And well, we cannot say the same about `iter_bytes(chunk_size=123)` due to compression.

docs/advanced.md
```diff
@@ -221,6 +221,35 @@ with httpx.Client(headers=headers) as client:
     ...
 ```
 
+## Monitoring download progress
+
+If you need to monitor download progress of large responses, you can use stream and the `response.last_raw_chunk_size` property.
```
florimondmanca, line 226:
```suggestion
If you need to monitor download progress of large responses, you can use stream and the `response.num_bytes_downloaded` property.
```

docs/advanced.md
```diff
@@ -221,6 +221,35 @@ with httpx.Client(headers=headers) as client:
...
+import httpx
+from tqdm import tqdm
+
+with tempfile.NamedTemporaryFile() as download_file:
+    url = "https://speed.hetzner.de/100MB.bin"
+    with httpx.stream("GET", url) as response:
+        if "Content-Length" in response.headers:
+            total = int(response.headers["Content-Length"])
+        else:
+            total = None
```
florimondmanca, line 242: I know about @tomchristie's comment here… https://github.com/encode/httpx/pull/1268#discussion_r485710581

But for this particular endpoint we know that we'll get a `Content-Length` in response, so should we simplify this to…?

```suggestion
        total = int(response.headers["Content-Length"])
```

I guess the motivation is to show a general-purpose example, but at the same time it's a bit confusing since _for any particular example_ there's only one branch we'll ever go through.

docs/advanced.md
tomchristie, line 226: Needs updating 😀

florimondmanca (commented):
Looking good with `num_bytes_downloaded` too, yup :+1:

docs/advanced.md
cdeler: I don't know.

On one hand it's good idea to show that `tqdm` accepts `total=None`. On other hand an extra `if` statement can mislead someone (moreover, it's the other library, not `httpx`).

I'd like to make an example easier, as you proposed.

@tomchristie , what do you think?

tomchristie: I'm happy with either...

Keeping it simple:

```python
total = int(response.headers["Content-Length"])
```

Or just being a little more careful...

```python
content_length = response.headers.get("Content-Length")
total = None if content_length is None else int(content_length)
```

docs/advanced.md
```diff
@@ -221,6 +221,32 @@ with httpx.Client(headers=headers) as client:
     ...
 ```
 
+## Monitoring download progress
+
+If you need to monitor download progress of large responses, you can use stream and the `response.num_bytes_downloaded` property.
```
tomchristie, line 226:
```suggestion
If you need to monitor download progress of large responses, you can use response streaming and inspect the `response.num_bytes_downloaded` property.
```

tomchristie (approved):

docs/advanced.md
```diff
@@ -221,6 +221,32 @@ with httpx.Client(headers=headers) as client:
     ...
 ```
 
+## Monitoring download progress
+
+If you need to monitor download progress of large responses, you can use response streaming and inspect the `response.num_bytes_downloaded` property.
+
```
tomchristie, line 227:
```suggestion

This interface is required for properly determining download progress, because the total number of bytes returned by `response.content` or `response.iter_content()` will not always correspond with the raw content length of the response if HTTP response compression is being used.

```

tomchristie:
Have made a couple of minor docs tweaks.
Looking great, thank you!

florimondmanca (approved):

PR #1269, opened 2020-09-08
Drop chardet

### LINKED ISSUE #1018: Replace chardet with plain UTF-8 detection

tomchristie:
Closes #1018

* Drops `chardet` for charset auto-detection.
* Drops `response.apparent_encoding`.
* `response.iter_text()` no longer buffers.
* Responses with `Content-Type: text/...` and no explicit `charset` no longer default to `iso-8859-1`, since the RFC there is considered obsoleted behaviour.

Instead simplifies our auto-detection approach, so that...

If an encoding is explicitly specified, then we use that. Otherwise our strategy is to attempt UTF-8, and fallback to Windows 1252.

Note that UTF-8 is a strict superset of ascii, and Windows 1252 is a superset of the non-control characters in iso-8859-1, so we essentially end up supporting any of ascii, utf-8, iso-8859-1, cp1252.

Given that UTF-8 is now by *far* the most widely used encoding, this should be a pretty robust strategy for cases where a charset has not been explicitly included.

Useful stats on the prevalence of different charsets in the wild...

* https://w3techs.com/technologies/overview/character_encoding
* https://w3techs.com/technologies/history_overview/character_encoding

The HTML5 spec also has some useful guidelines, suggesting defaults of either UTF-8 or Windows 1252 in most cases...

* https://dev.w3.org/html5/spec-LC/Overview.html

Users can override this behaviour if required with an explicit `response.encoding = ...`.

**I do also have some thoughts about exposing a `text_decoder=...` interface to allow overriding this behaviour, but I think we should probably treat that as a follow-up PR.**


tomchristie:
One little edge case to consider here is what behaviour we expect for streaming text on responses that do not include any explicit charset. There are three different things we could choose to do here...

* Decide which of utf-8 / cp1252 to use on the first decoder pass, and stick with that. (Potentially more robust, but also in edge cases could potentially result in inconsistent behaviour depending on how much data is received in the first pass.)
* Always use utf-8 in streaming cases. (More consistent, potentially less robust.)
* Buffer up some minimum amount of data before selecting a decoder. (Most robust, but introduces buffering artefacts.)

Currently this PR is using the first of those approaches.

httpx/_decoders.py
```diff
@@ -167,12 +167,13 @@ def __init__(self, encoding: typing.Optional[str] = None):
...
+            attempt_utf_8 = codecs.getincrementaldecoder("utf-8")(errors="strict")
             try:
-                return self.decoder.decode(data)
+                attempt_utf_8.decode(data)
             except UnicodeDecodeError:
-                self.decoder = codecs.getincrementaldecoder("cp1251")(errors="ignore")
-                return self.decoder.decode(data)
+                self.decoder = codecs.getincrementaldecoder("cp1252")(errors="replace")
+            else:
+                self.decoder = codecs.getincrementaldecoder("utf-8")(errors="replace")
```
StephenBrown2, line 202: Why `errors="replace"` here if it passed with `errors="strict"` to get here?

tomchristie: We need `strict` to raise an error if it doesn't appear to decode as UTF-8, but once we've made the decision we use `errors="replace"` for the most robust behaviour possible. So eg. if we've got a streaming response that initially appears to be UTF-8, but later has some non-UTF-8 bytes, then we're not raising a hard error on accessing `.text`.

(We'd like it to have a failure mode that is as graceful as possible.)

docs/quickstart.md
```diff
@@ -65,14 +65,27 @@ HTTPX will automatically handle decoding the response content into Unicode text.
...
-You can inspect what encoding has been used to decode the response.
+You can inspect what encoding will be used to decode the response.
 
 ```pycon
 >>> r.encoding
 'UTF-8'
 ```
 
-If you need to override the standard behavior and explicitly set the encoding to
+In some cases the response may not an explicit encoding, in which case HTTPX
```
florimondmanca, line 75:
```suggestion
In some cases the response may not contain an explicit encoding, in which case HTTPX
```

tomchristie:
Alrighty then, let's press on with this! 👍

PR #1270, opened 2020-09-09
Fixed test_multiple_set_cookie

cdeler:
I found that one of the tests started to fail

It was `test_multiple_set_cookie`, the reason was the cookie expiration (on 2020-09-08). I updated expiration date to 2029


tomchristie:
Awesome, thanks!
Was wondering what was going on there.

tomchristie (approved):

tomchristie:
Bumped it to 2099.
Because why not?

j178 (approved):

Issue #394, opened 2019-09-26
Allow setting `chunk_size` for `Response.iter_bytes()` etc...


sethmlarson:
Requests allowed setting chunk_size within `.iter_content()` which is currently not an option for our alternatives `.stream()` and `.stream_text()`.

For `.stream_text()` we should go the extra step and fix the issue that users sometimes run into when using this feature and use chunk-size for measuring the decoded text, not the raw bytes.

tomchristie:
I seem to remeber this playing into some primitives around streaming bytes vs text that we never ended up digging into?

tomchristie:
A good first pass onto this would be to change the decoder interface slightly, so that instead of eg. yeilding a byte chunk, they yield a *list* of byte chunks.

On the first refactoring pass, we don't need to *actually* change the internal implmentation much - the decoders can just always yield a list with a single item.

We'd *then* be able to add a `chunk_size` argument to the decoders, which would return 0, 1, or many properly-sized chunks on each yield.

florimondmanca:
Updated the title to reflect the current `Response.aiter_*` API :-) (see #610).

b0g3r:
How could I help with this issue?

florimondmanca:
Hi @b0g3r! I think this is still something we’d like to have, and given discussions in https://github.com/python-gitlab/python-gitlab/pull/1036 it seems like some folks would like to see it too. :)

Ways to move forward would be:

- Propose an API for this
- Investigate implementation details (ie how are we going to split chunks: buffering, other?)
- Draft a PR :)
-

b0g3r:
Do I understand correctly that we will need to forward chunk_size here?
https://github.com/encode/httpx/blob/a82adcc933345c6b8cb1623b031eb85723e7665b/httpx/_dispatch/urllib3.py#L112-L115

florimondmanca:
@b0g3r Careful that we're in a sort of transition state w.r.t. urllib3 usage due to #804 (we'll soon use our own sync implementation, though keeping urllib3 as an option). Due to this I wouldn't advise relying on any existing urllib3 functionality — also because we'd need to provide chunk size on the async layer too.

I think we want to look at controlling the chunk size directly from `response.iter_bytes()`/`response.aiter_bytes()`, instead…

tomchristie:
[...]

PR #1271, opened 2020-09-09
Added chunk_size to Response.iter_bytes() and Response.aiter_bytes() (#394)

cdeler:
Working on #394 , I added the `chunk_size` argument  to `Response.iter_bytes()` and `Response.aiter_bytes()` wrapping these iterators by `drain_by_chunks()` and `async_drain_by_chunks()` (I didn't use decorators to make debugging easier). The same way might be applied to  `(a)iter_text`

But I still don't know what the expected behaviour of `(a)iter_line` and `(a)iter_raw` (so therefore I omit them in the PR)

For now the code might look like:
```python
>>> import httpx
>>> response = httpx.Response(200, content=b"1234567890")
>>> for it in response.iter_bytes(3):
>>>    print(it.decode())
123
456
789
0
>>> # it's done for requests compatibility: 
>>> # https://github.com/encode/httpx/issues/394#issuecomment-598820965
>>> for it in response.iter_bytes(None):
>>>     print(it.decode())
1234567890
```

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -47,6 +47,8 @@ from ._types import (
     URLTypes,
 )
 from ._utils import (
+    async_drain_by_chunks,
+    drain_by_chunks,
     flatten_queryparams,
     guess_json_utf,
     is_known_encoding,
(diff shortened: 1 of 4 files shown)


tomchristie:
This one's a bit fiddly, so I'm going to defer on reviewing it for a little bit.

tomchristie:
So, I'd been planning for us to take an approach more along these lines... #1277

That adds a decoder class,`ByteChunker` which has the same kind of interface as all the other decoders,
and which doesn't involve having separate sync + async implementations.

I've not quite finished that work off, since it still needs:

* Using `ByteChunker` within `iter_bytes`.
* Adding a corresponding `TextChunker` class.
* Using `TextChunker` within `iter_text`.

cdeler:
@tomchristie 
It's better to close this PR?

tomchristie:
@cdeler - I think so, yup - thanks for having taken a crack at it.

PR #1272, opened 2020-09-10
Progress examples

tomchristie:
What do we think about this?...

* Include a progress bar using `rich`, in addition to the existing one using `tqdm`.
* Include GIF images demonstrating the running examples.

Rendered docs look like this...

xhb8nB5zRt


florimondmanca (approved):
Awesome sauce 😍

tomchristie:
Incidentally it looked to me like the download speed displayed by `rich` was much more jumpy than I'd expected. It *might?* be worth looking into if there's any buglette hidden in there, to do how they're calculating it based on the update samples.

A quick pass looked as if they're averaging it out over the last 30 seconds, but I would have expected a smoother rate of change from what I actually saw.

But perhaps I've got my wires crossed - didn't get a chance to dig much.

Issue #1275, opened 2020-09-10
Tightening public API on models


tomchristie:
In our lead up to 0.14 we went though the public API of `Client` in a huge amount of detail to make sure we had everything exactly right, and were only exposing *exactly* what we needed to the public API.

Eg see issues #997, #1065, #1066

To a lesser extent we still need to make sure to do the same on everything in `_models.py` before we can hit 1.0.

Here are a few items from having taken a glance over...

* `Request.prepare` should be private.
* `Response` should not expose a public `.decoder` property. Particularly given that we don't expose the `Decoder` interface anywhere. Let's replace it with a private `._get_content_decoder()` method.
* Review Headers and multi-header handling.
* URL.path should be unquoted. Review str vs. bytes and quoting throughout.

cdeler:
@tomchristie 

By
> URL.path should be unquoted. Review str vs. bytes and quoting throughout.

did you mean `urllib.parse.unquote `/`urllib.parse.unquote_plus`

tomchristie:
What we really want here is...

* `.path` - `str` *URL escaped.*
* `.query` - `bytes` *No escaping applied.*
* `.raw_path` - `bytes` *No escaping applied. Includes both path and query portion.*

And also related...

* `.userinfo` - `bytes` *No escaping applied.*
* `.username`- `str` *URL escaped.* (We've already got this)
* `.password`- `str` *URL escaped.* (We've already got this)

There's probably also some thinking about if we'd like `.host` to expose the *IDNA decoded* host, and add a `.raw_host` for the byte-wise IDNA encoded hostname.

cdeler:
@tomchristie 

I'm a bit confused with one small detail, I'm sorry for the question, but

> .query - bytes No escaping applied.

for now it's a `str`, moreover it's stored as a `str`

In the example your provided with, `value` is also a `str`:
```python3
        if value is None or isinstance(value, (str, bytes)):
            value = value.decode("ascii") if isinstance(value, bytes) else value
            items = parse_qsl(value)
```

The question is: may be it's more convenient to treat `URL.query` as a `str`, but give an alternative: `raw_query` (like `raw_path` and `path`)?

tomchristie:
To clear up one part of this... we should *rename* `.full_path` to `.raw_path` (which should return bytes)

tomchristie:
[...]

PR #1276, opened 2020-09-10
Minor decoder refactoring.

tomchristie:
* Drop `response.decoder` property, since we don't actually want to expose the decode API. (Refs #1275)
* `Decoder` -> `ContentDecoder`. (Since it's more obviously distinct from `TextDecoder`, `LineDecoder`, rather than looking like a base class for all of them, which it isn't.)


florimondmanca (approved):
👍

Issue #1091, opened 2020-07-27
Server-side requests/responses


tomchristie:
So, not something that we need to support as a primary use-case, but something we ought to just think about a little before 1.0, is making sure that our `Request` and `Response` models can be used on the server side as well.

For instance...

* It'd be beneficial if libraries like `starlette` were able to lean heavily on `httpx`.
* It'd make for easier custom transport implementations, if users were able to eg. instantiate a request instance, given the incoming raw request info. (Since that's a bit easier to worth with that the raw byte-wise API.)

A couple of API/implementation notes to go along with this...

* `URL()` should accept the raw bytes URL 4-tuple format that the `httpcore` API uses as one of it's options.
* `Request.prepare(...)` shouldn't be called automatically on `__init__`. The `Client` should call it explicitly in `build_request()`. This ensures that we'd be able to instantiate a `Request` instance from the raw HTTP core `method`/`url`/`headers/`stream` information, without having additional headers automatically generated on the model.
* We might want `Request.raw`, `Response.raw`, `Headers.raw` properties, to match up with the existing `URL.raw` property.

florimondmanca:
Hey hey,

Interesting stuff, but what would be some possible use cases for this?

I'm thinking à use case of a Starlette server side app where we proxy requests from ASGI through HTTPX and responses back again. Is this discussion related to making this kind of things easier?

tomchristie:
So heres an example of something you might use this for...

```python
class MockTransport():
    """
    A transport class that simply always returns a single given response.
    """
    def __init__(self, response: Response):
        self.response = response

    def request(self, ...):
        return self.response.raw
```

jcugat:
Could this be used to build the httpx-equivalent of [responses](https://github.com/getsentry/responses) / [aioresponses](https://github.com/pnuckowski/aioresponses)?

florimondmanca:
@jcugat There's already [RESPX](https://github.com/lundberg/respx) by @lundberg which I think was intended to be such an equivalent (although I believe the API has parted ways a bit). :-) But yes this `MockTransport` example totally rings a bell and might be a way to make RESPX even simpler in design and implementation…

lundberg:
True @florimondmanca , the API has parted somewhat, it's hard to keep up with the speed `httpx` evolves and improves, awesome work!

[...]

PR #1281, opened 2020-09-11
Refactor tests to use `MockTransport(<handler_function>)`

tomchristie:
Refactors our test cases so that we're able to test against simple request/response handler functions.

```python
import httpx
from tests.utils import MockTransport

def app(request: httpx.Request) -> httpx.Response:
    ...

def test_something():
    client = httpx.Client(transport=MockTransport(app))
    response = client.get(...)
```

This is actually kind of a big deal. 🤗

Everything is still passing through our really tightly defined transport API, but we get the benefits of being able to work with proper request/response instances, rather than having to build out test behaviours by scratching around with bytewise variables, raw header lists, and `httpcore.SyncIteratorStream`/`httpcore.AsyncIteratorStream` instances.

Still need to clean up `client/test_auth.py` using this, but have covered all the other client test modules.

The implementation of `MockTransport` in the `tests/utils.py` module also gives us a *really* good concrete case to work against for working through some of the more nitty gritty details of the request/response APIs.

Really we'd like for the implementation to be as simple as...

```python
class MockTransport(httpcore.SyncHTTPTransport):
    def __init__(self, handler: Callable) -> None:
        self.handler = handler

    def request(
        self,
        method: bytes,
        stream: httpcore.SyncByteStream = None,
        timeout: Mapping[str, Optional[float]] = None,
    ) -> Tuple[bytes, int, bytes, List[Tuple[bytes, bytes]], httpcore.SyncByteStream]:
        request = httpx.Request(method=method, url=url, headers=headers, content=stream)
        request.read()
        response = self.handler(request)
        return (
            response.http_version.encode("ascii"),
            response.status_code,
            response.reason_phrase.encode("ascii"),
            response.headers.raw,
            response.stream
        )
```

(In fact in addition to that, we'd probably also like to end up with an API that allows `MockTransport` be usable with *either* sync *or* async clients, rather than having to have separate `MockTransport` and `AsyncMockTransport` classes, but let's save that conversation for another day.)

The existing implementation here gives us something to work against, while tackling little parts of that incrementally. Eg. adding support for passing raw URL tuples of `(<bytes>, <bytes>, <optional int>, <bytes>)` to `httpx.URL(...)`, and also some of the details around #1253

Related to https://github.com/encode/httpx/issues/1091


tests/utils.py
```diff
@@ -18,3 +22,90 @@ def override_log_level(log_level: str):
...
+        path = raw_path.decode("ascii")
+
+        request_headers = httpx.Headers(headers)
+        data = (
+            (item for item in stream)
+            if stream
+            and (
+                "Content-Length" in request_headers
+                or "Transfer-Encoding" in request_headers
+            )
```
tomchristie, line 52: It's a bit fiddly to explain why we're having to do this at the moment, but it's related to some of the grungy bits that need ironing out in #1253.

I'll resolve it as part of a follow-up.

florimondmanca (approved):
Very nice indeed! :100: :+1:

PR #1280, opened 2020-09-11
Made Request.prepare private (i.e. renamed it to _prepare)

cdeler:
I created a PR regarding to https://github.com/encode/httpx/issues/1275

I renamed `Request.prepare` to `_prepare` making this method private

PR #1283, opened 2020-09-11
Silence PIE786 in some cases

cdeler:
I've got a problem in my pr: [build link](https://github.com/encode/httpx/pull/1280/checks?check_run_id=1102709981)

The problem (PIE786) caused by some too broad exception handlers (as flake thinks). I suppressed the error

I think that error caused by `flake8-pie` [released today](https://pypi.org/project/flake8-pie/#history) :-)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -763,7 +763,7 @@ class Client(BaseClient):
                 next_request = auth_flow.send(response)
             except StopIteration:
                 return response
-            except BaseException as exc:
+            except Exception as exc:
                 response.close()
                 raise exc from None
             else:
@@ -1386,7 +1386,7 @@ class AsyncClient(BaseClient):
                 next_request = await auth_flow.asend(response)
             except StopAsyncIteration:
                 return response
-            except BaseException as exc:
+            except Exception as exc:
                 await response.aclose()
                 raise exc from None
             else:
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -763,7 +763,7 @@ class Client(BaseClient):
                 next_request = auth_flow.send(response)
             except StopIteration:
                 return response
-            except Exception as exc:
+            except Exception as exc:  # noqa: PIE786
                 response.close()
                 raise exc from None
             else:
@@ -1386,7 +1386,7 @@ class AsyncClient(BaseClient):
                 next_request = await auth_flow.asend(response)
             except StopAsyncIteration:
                 return response
-            except Exception as exc:
+            except Exception as exc:  # noqa: PIE786
                 await response.aclose()
                 raise exc from None
             else:
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [flake8]
-ignore = W503, E203, B305, PIE786
+ignore = W503, E203, B305
 max-line-length = 88
 
 [mypy]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [flake8]
-ignore = W503, E203, B305
+ignore = W503, E203, B305, PIE786
 max-line-length = 88
 
 [mypy]
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -340,7 +340,7 @@ def map_exceptions(
 ) -> typing.Iterator[None]:
     try:
         yield
-    except Exception as exc:
+    except Exception as exc:  # noqa: PIE786
         mapped_exc = None
 
         for from_exc, to_exc in mapping.items():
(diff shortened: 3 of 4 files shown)


tomchristie:
Is it possible for us to disable that case globally?

cdeler:
@tomchristie 

It's possible changing `setup.cfg`:
```
[flake8]
ignore = W503, E203, B305, PIE786
max-line-length = 88
```

If it's acceptable in your code style -- I can suppress  PIE786, but I'd keep it on, since usually it's a bad practice to catch e.g. `BaseException`

StephenBrown2:
Interestingly, PEP-8 has this to say about exceptions: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

> Derive exceptions from `Exception` rather than `BaseException`. Direct inheritance from `BaseException` is reserved for exceptions where catching them is almost always the wrong thing to do.

and:

> When catching exceptions, mention specific exceptions whenever possible instead of using a bare `except:` clause:
> 
> ```
> try:
>     import platform_specific_module
> except ImportError:
>     platform_specific_module = None
> ```
> 
> A bare `except:` clause will catch `SystemExit` and `KeyboardInterrupt` exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:).
> 
> A good rule of thumb is to limit use of bare 'except' clauses to two cases:
> 
>     If the exception handler will be printing out or logging the traceback; at least the user will be aware that an error has occurred.
>     If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise. try...finally can be a better way to handle this case.

So I don't think it's an unreasonable lint in general, and I'm surprised that `BaseException` is being caught rather than `Exception` in some cases here. Of course it makes sense to catch all exceptions because we're mapping them, but I wouldn't want to disable the check entirely.

cdeler:
@StephenBrown2 I think there we catch `BaseException` to close everything in every cases including `SystemExit`

But yes, I'd replace it `BaseException` by `Exception` in the source and our test suite passes

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = W503, E203, B305
+ignore = W503, E203, B305, PIE786
```
StephenBrown2, line 2: I'm still opposed to ignoring this outright. Ignoring it per line is my preferred style, though if we really wanted to, we could add it to [per-file-ignores](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores):

```suggestion
ignore = W503, E203, B305
per-file-ignores =
    httpx/_client.py,httpx/_exceptions.py:PIE786
```

StephenBrown2:
@tomchristie What are your thoughts on keeping the per-line disables?

setup.cfg
cdeler: I do agree with you. May be it's better to suppress the error per line back?

StephenBrown2: That's my thought, but I'd like the other maintainers to chip in first.

cdeler:
@StephenBrown2 

I found https://github.com/encode/httpcore/pull/183 and (from my point of view) it looks better, so I disabled flake8 by-line

httpx/_client.py
```diff
@@ -763,7 +763,7 @@ def _send_handling_auth(
                 next_request = auth_flow.send(response)
             except StopIteration:
                 return response
-            except BaseException as exc:
+            except Exception as exc:  # noqa: PIE786
```
florimondmanca, line 766: Should we stick to catching `BaseException` here and in the other change in this file?

cdeler: I was thinking about that.

What the benefits of `BaseException`?

Catching `BaseException` there, we also handle `KeyboardInterrupt`, `GeneratorExit` and `SystemExit` and close the response in these cases. I don't think it's a proper behaviour (but may be I miss something)

florimondmanca: I'm okay with either, but since it would be a behavioral let's try and submit that as a separate PR, so that this only focuses on silencing any PIE786 errors?

cdeler: Sure, wait a second

tomchristie:
Let's start just by pinning our version of `flake8-pie`... https://github.com/encode/httpx/pull/1286

We should then consider the `map_exceptions` and the auth iteration exceptions as *two separate cases* and decide what we want to do with each of them.

tomchristie:
I’d suggest the same approach as https://github.com/encode/httpx/pull/1286 - start by pinning `flake8-pie`, then we can consider any code changes / linting exceptions after that.

cdeler:
Closing as already solved

PR #1288, opened 2020-09-14
Refactor test_auth.py to use `MockTransport` class.

tomchristie:
Follow up to https://github.com/encode/httpx/pull/1281 - refactors `test_auth.py` to use the nice `MockTransport` class.

Possible later follow ups here...

* We should switch over to a plain `import httpx`. (`test_auth.py` uses from `httpx import ...` for a bunch of stuff, where we're mostly now sticking with a style of `import httpx` in other test cases.)
* There's a bunch of test cases that are currently async, but could probably be plain sync cases, which we should prefer where possible.

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -10,7 +10,6 @@ import os
 import threading
 import typing
 
-import httpcore
 import pytest
 
 import httpx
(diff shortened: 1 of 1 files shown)


florimondmanca (approved):
Sweet!

PR #1293, opened 2020-09-17
Requests from transport API

tomchristie:
Related to #1091

Improving our support for instantiating a request instance, from the transport API.

* `method` may now be either `str` or `bytes`.
* `url` now supports passing a raw URL. `(bytes, bytes, Optional[int], bytes)`.

Working towards being able to do something like...

```python
class CustomTransport:
    def request(self, method, url, headers, stream, timeout):
        request = httpx.Request(method, url, headers, stream)
        ...
        response = httpx.Response(...)
        return response.raw
```

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -688,7 +688,7 @@ class Response:
         status_code: int,
         *,
         request: Request = None,
-        http_version: str = "HTTP/1.1",
+        http_version: str = None,
         headers: HeaderTypes = None,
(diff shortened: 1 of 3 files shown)


johtso:
Is the `raw` property on the response still waiting to be done? Would be useful for the transport tests that I'm fiddling with. Maybe request should expose a similar property that matches the input to the transport `.request()`?

PR #1295, opened 2020-09-17
Drop ContentStream

### LINKED ISSUE #1253: Exposing the content stream API

tomchristie:
Drop `ContentStream` completely. Closes #1253

This is a horrendously grungy bit of refactoring to have had to work through, but it's really important.

We no longer have `ContentStream` anywhere internally. Instead any stream interfaces are always typed as plain `Union[Iterable[bytes], AsyncIterable[bytes]]`.

* `request.stream` is no longer a `ContentStream` instance. Now it is `Union[Iterable[bytes], AsyncIterable[bytes]]`.
* `response._raw_stream` is no longer a `ContentStream` instance. Now it's promoted to a public `response.stream` interface and is `Union[Iterable[bytes], AsyncIterable[bytes]]`.

This PR is *not a lot of fun* at all, but it does get the public API where it needs to be, and it also preps. the ground for us switching up the httpcore API to also not need a `SyncByteStream`/`AsyncByteStream` abstraction.

We can instead move to using `yield` for the closing management (https://github.com/encode/httpcore/issues/145), and plain `Iterable[bytes]`/`AsyncIterable[bytes]` annotations for the byte streams.

It's also worth looking at what the `MockTransport` implementation looks like after this PR, since it helps demonstrate how it helps clean things up...

```python
class MockTransport(httpcore.SyncHTTPTransport):
    def __init__(self, handler: Callable) -> None:
        self.handler = handler

    def request(
        self,
        method: bytes,
        stream: httpcore.SyncByteStream = None,
        timeout: Mapping[str, Optional[float]] = None,
    ) -> Tuple[bytes, int, bytes, List[Tuple[bytes, bytes]], httpcore.SyncByteStream]:
        request = httpx.Request(
            method=method,
            url=url,
            headers=headers,
            stream=stream,
        )
        request.read()
        response = self.handler(request)
        return (
            (response.http_version or "HTTP/1.1").encode("ascii"),
            response.status_code,
            response.reason_phrase.encode("ascii"),
            response.headers.raw,
            response.stream,
        )
```

This PR also sorts out some fiddly set wrt. when to automatically include headers on requests or responses.

For example `Response(content=...)` was not currently setting any automatically populated `Content-Length` header, although it *really* should. We weren't doing so because we *don't* want responses that are instantiated from the returned info from the Transport API to have auto-headers generated. This PR sorts that out with `stream=...` being used for passing raw streams, and not implying any auto-headers, in contrast to `Response(content=...)` which is the more user-facing API for creating responses for mock tests or for other server-side contexts. This unblocks #1250.

I don't expect this PR to be actionable reviewable, but we've got our type checking in place all the way through, plus 100% coverage. I'm not expecting any issues, but if there *are* any unexpected niggles we'd just have to deal with them.


florimondmanca (approved):
Went and skimmed through the changes, and nothing super controversial or popping out of place! Great one. Looking forward to seeing what this allows us to do next :) I really like how requests and responses now behave more and more like pure model classes, with support for both high level APIs and lower level transport arguments and attributes...

tomchristie:
🤗 Fantastic. Let's crack on then.

PR #1296, opened 2020-09-18
Refactor content streams

tomchristie:
Bit of housekeeping since merging #1295

* `_content_streams.py` -> `_content.py` - Now mostly plain function implementations.
* Added `_multipart.py` to better separate out the multipart handling.


tomchristie:
Getting this one straight in, since it's a nice simple internal-only refactoring, that very clearly cleans things up.

PR #1299, opened 2020-09-18
Neater `Host` header logic on redirects

tomchristie:
Instead of re-populating the `Host` header in the client on redirects away from the origin, we instead just remove it.

This is neater since the header will be re-populated by the `Request` instance, so we've only got one place where we need logic around populating it correctly.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -464,7 +464,10 @@ class BaseClient:
             # Strip Authorization headers when responses are redirected away from
             # the origin.
             headers.pop("Authorization", None)
-            headers["Host"] = url.authority
+
+            # Remove the Host header, so that a new one with be auto-populated on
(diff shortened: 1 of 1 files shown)


httpx/_client.py
```diff
@@ -464,7 +464,10 @@ def _redirect_headers(self, request: Request, url: URL, method: str) -> Headers:
             # Strip Authorization headers when responses are redirected away from
             # the origin.
             headers.pop("Authorization", None)
-            headers["Host"] = url.authority
+
+            # Remove the Host header, so that a new one with be auto-populated on
```
florimondmanca, line 468: Typo here :-) "new one s/with/will/g be"

tomchristie: 😬 Ya got me

PR #1284, opened 2020-09-11
Refactoring of models api

cdeler:
Having working on https://github.com/encode/httpx/issues/1275 in https://github.com/encode/httpx/pull/1280, I've got a reasonable comment (from @StephenBrown2) that the PR could be split into several one to reduce problems with review.

In this PR I:
1. made `Request.prepare` private
1. Changed `QueryParamTypes` so that `QueryParams` accepts `bytes`

Issue #1274, opened 2020-09-10
Finalising the Transport API for 1.0.


tomchristie:
There's a few bits of thinking around the Transport API that we'd like to be completely set on before we roll with a 1.0 release.

* We should consider splitting the sync/async interface as `.request`/`.arequest`. This would allow transport implementations that support both sync+async usage, which would be a really nice usability for some use-cases.
* There's a very strong case to be made for `.request` being a context manager. See https://github.com/encode/httpcore/issues/145 - Needs some very careful investigation.
* We need to do a bit of thinking around making custom transports easier to use. Eg from a POV of not having to specify defaults verbosely as soon as dropping down to that level. The docs at https://www.python-httpx.org/advanced/#custom-transports are a good example of where we could be making some usability improvements.
* We ought to consider exposing something like `httpx.HTTPTransport(...)` rather than forcing users to drop into `httpcore` once they're configuring transports directly.
* The work on exposing the content stream API is also somewhat related... See #1253

To give a few examples of the effect of some of these decisions...

Expose low-level configuration options on the transport but not the client, without having gnarly extra configuration needed...

```python
client = httpx.Client(transport=httpx.HTTPTransport(local_address="..."))
```

Mount a customised transport for one specific domain, without having gnarly extra configuration needed...

```python
client = httpx.Client(..., mounts={"all://www.example.com": httpx.HTTPTransport(verify=False, retries=1)})
```

Transport API using `.request` and `.arequest`...

```python
# A custom transport class, that always redirects to https URLs, and supports both sync and async API,
# meaning it can be used with either kind of client.
client = httpx.Client(..., mounts={"http": HTTPSRedirect()})
client = httpx.AsyncClient(..., mounts={"http": HTTPSRedirect()})
```

tomchristie:
I've been doing some thinking around this, and I *do* think we'll want to re-jig the Transport API slightly.
Here's where I'm at...

**1.** I think we *really* want `.request()` and `.arequest()`. Being able to have classes such as a single `httpx.MockTransport(handler)` that can work with either case is a really good argument in favour of this. Additionally I can imagine midleware-like functionality such as `HTTPSOnlyTransport`, `HSTSTransport` that'd benefit from being able to work with both async and sync clients. Switching to `.request`/`.arequest` is also a really easy one for us to do.

The next set is a bit more complex to talk through. So...

[...]

PR #1301, opened 2020-09-21
Version 0.15.0

tomchristie:
Pushing this up so that we can figure out what else we want to block a 0.15 release on.

At the moment my thoughts would be...

- [x] Change the Transport API to `.request`, `.arequest`, since it'll otherwise end up blocking other work.
- [x] Possibly also switching the transport API to `request(method, url, headers, stream, ext) -> status_code, headers, stream, ext`. See https://github.com/encode/httpx/issues/1274#issuecomment-694884553 - Might need discussion, but it's an easy switch to make, and better to make it sooner rather than later.
- [ ] Possibly get in the CLI. Because it's lovely, and might be a nice driver of adoption. https://github.com/encode/httpx/pull/1212
- [x] Go through an escalate any deprecations that we've got in the codebase into outright removals.

I *think* other stuff can wait for nice smaller increments in the 0.15.x series. The only remaining API *changes* expected at that point are the possible transport API -> transport API as a context manager switch, and some very minor review over all the model classes. (Note that `URL` is now complete in this regard.)

What do we think, and am I missing anything critical here?

### Added

* Added support for event hooks. (Pull #1246)
* Added support for authentication flows which require either sync or async I/O. (Pull #1217)
* Added support for monitoring download progress with `response.num_bytes_downloaded`. (Pull #1268)
* Added `Request(content=...)` for byte content, instead of overloading `Request(data=...)` (Pull #1266)
* Added support for all URL components as parameter names when using `url.copy_with(...)`. (Pull #1285)
* Neater split between automatically populated headers on `Request` instances, vs default `client.headers`. (Pull #1248)
* Unclosed `AsyncClient` instances will now raise warnings if garbage collected. (Pull #1197)
* Support `Response(content=..., text=..., html=..., json=...)` for creating usable response instances in code. (Pull #1265, #1297)
* Support instantiating requests from the low-level transport API. (Pull #1293)
* Raise errors on invalid URL types. (Pull #1259)

### Changed

* Cleaned up expected behaviour for URL escaping. `url.path` is now URL escaped. (Pull #1285)
* Cleaned up expected behaviour for bytes vs str in URL components. `url.userinfo` and `url.query` are not URL escaped, and so return bytes. (Pull #1285)
* Drop `url.authority` property in favour of `url.netloc`, since "authority" was semantically incorrect. (Pull #1285)
* Drop `url.full_path` property in favour of `url.raw_path`, for better consistency with other parts of the API. (Pull #1285)
* No longer use the `chardet` library for auto-detecting charsets, instead defaulting to a simpler approach when no charset is specified. (#1269)

### Fixed

* Swapped ordering of redirects and authentication flow. (Pull #1267)
* `.netrc` lookups should use host, not host+port. (Pull #1298)

### Removed

* The `URLLib3Transport` class no longer exists. We've published it instead as an example of [a custom transport class](https://gist.github.com/florimondmanca/d56764d78d748eb9f73165da388e546e). (Pull #1182)
* Drop `request.timer` attribute, which was being used internally to set `response.elapsed`. (Pull #1249)
* Drop `response.decoder` attribute, which was being used internally. (Pull #1276)
* `Request.prepare()` is now a private method. (Pull #1284)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * Drop `request.timer` attribute, which was being used internally to set `response.elapsed`. (Pull #1249)
 * Drop `response.decoder` attribute, which was being used internally. (Pull #1276)
 * `Request.prepare()` is now a private method. (Pull #1284)
+* The `Headers.getlist()` method had previously been deprecated in favour of `Headers.get_list()`. It is now fully removed.
+* The `QueryParams.getlist()` method had previously been deprecated in favour of `QueryParams.get_list()`. It is now fully removed.
+* The `URL.is_ssl` property had previously been deprecated in favour of `URL.scheme == "https"`. It is now fully removed.
+* The `httpx.PoolLimits` class had previously been deprecated in favour of `httpx.Limits`. It is now fully removed.
+* The `max_keepalive` setting had previously been deprecated in favour of the more explicit `max_keepalive_connections`. It is now fully removed.
+* The verbose `httpx.Timeout(5.0, connect_timeout=60.0)` style had previously been deprecated in favour of `httpx.Timeout(5.0, connect=60.0)`. It is now fully removed.
+* Support for instantiating a timeout config missing some defaults, such as `httpx.Timeout(connect=60.0)`, had previously been deprecated in favour of enforcing a more explicit style, such as `httpx.Timeout(5.0, connect=60.0)`. This is now strictly enforced.
 
 ## 0.14.3 (September 2nd, 2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,41 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.15.0
+
+### Added
+
+* Added support for event hooks. (Pull #1246)
+* Added support for authentication flows which require either sync or async I/O. (Pull #1217)
+* Added support for moniotoring download progress with `response.num_bytes_downloaded`. (Pull #1268)
+* Added `Request(content=...)` for byte content, instead of overloading `Request(data=...)` (Pull #1266)
+* Added support for all URL components as parameter names when using `url.copy_with(...)`. (Pull #1285)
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.15.0
+
+### Added
+
+* Added support for event hooks. (Pull #1246)
+* Added support for authentication flows which require either sync or async I/O. (Pull #1217)
+* Added support for moniotoring download progress with `response.num_bytes_downloaded`. (Pull #1268)
```
JayH5, line 13:
```suggestion
* Added support for monitoring download progress with `response.num_bytes_downloaded`. (Pull #1268)
```

florimondmanca:
Sounds good to me, great one?

tomchristie:
Right, let's keep the CLI separate. No good reason for us to bog ourselves down in a whole extra layer of API considerations (eg. which flags do we support on the command line) when we're so very close to a 1.0 release now.

I'm going to push on again, and get this released.

Where does this leave us after this?...

Our [1.0 milestone is here](https://github.com/encode/httpx/milestone/1).

* There's one significantly complex bug that needs resolving in `httpcore` around connection pooling. https://github.com/encode/httpx/issues/1171 - It'll be a decent chunk of work, and it's right in the guts of things, so I want to deal with this in a dedicated `httpcore` minor point release, that should *only* include that bugfix and nothing else. Eg. 0.11.1, or wherever we're at.
* Review any other bug-labelled tickets.
* Mount API, `httpx.HTTPTransport`, `httpx.MockTransport` - #977, #1302, #1303 are all in the same area. I figure we'll have a minor httpx release that's dedicated just to adding those, and sprucing up our docs on mounting transports etc.
* `chunk_size` #394 - We've got an outstanding PR for this, but I'd like to pull it in, in it's own minor point release. Since *all* response.text/response.content runs through that code I'd like it to be really nicely isolated from any other changes, just in case we needed to very quickly identify any unexpected problems.
* Tightening public API on models. #1275 I think this *might* just be `Cookies` that's left to review here now.
* Preserving HTTP Header casing. Blocked on https://github.com/python-hyper/h11/pull/104
* Retries #1141 
* Transport API as a context manager. Our last API change still to be made. #998 and #1274

We'll then also want to go through the API docs, and make sure that all the public API is being properly exposed in the docs.

PR #1309, opened 2020-09-23
update arg name to `max_keepalive_connections`

emlazzarin:
The old name `max_keepalive` was removed recently, so this updates the docs.


tomchristie (approved):
Fantastic, thank you!

PR #1320, opened 2020-09-24
Tweak transport closing

### LINKED ISSUE #1317: Client.close() does not close transports if unused
Currently, `Client.close()` closes its transports if and only if the client is deemed to be "open" (`not self._is_closed`).

When the client is instantiated, it is not considered open, rather only when something is sent, or when it is `__enter__`ed.

Proxies and transports are initialized when the client is instantiated.

This means that if a client is not used as a context manager, and no requests are sent with it, proxies and transport are not closed when `Client.close()` is called.

This obviously doesn't cause any problems with the standard transports, but a custom transport may be in use that sets up some other resources.

Could we not just `.close()` the transports regardless of whether the client has been used?

tomchristie:
Closes #1317

* Always close transports if `client.close()` is called, even if no requests have been made.
* Treat client instances as "open" as soon as they're instantiated, even if no requests have been made.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1119,13 +1119,12 @@ class Client(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        if not self.is_closed:
-            self._is_closed = True
+        self._is_closed = True
 
-            self._transport.__exit__(exc_type, exc_value, traceback)
-            for proxy in self._proxies.values():
-                if proxy is not None:
-                    proxy.__exit__(exc_type, exc_value, traceback)
+        self._transport.__exit__(exc_type, exc_value, traceback)
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                proxy.__exit__(exc_type, exc_value, traceback)
 
     def __del__(self) -> None:
         self.close()
@@ -1764,12 +1763,11 @@ class AsyncClient(BaseClient):
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
-        if not self.is_closed:
-            self._is_closed = True
-            await self._transport.__aexit__(exc_type, exc_value, traceback)
-            for proxy in self._proxies.values():
-                if proxy is not None:
-                    await proxy.__aexit__(exc_type, exc_value, traceback)
+        self._is_closed = True
+        await self._transport.__aexit__(exc_type, exc_value, traceback)
+        for proxy in self._proxies.values():
+            if proxy is not None:
+                await proxy.__aexit__(exc_type, exc_value, traceback)
 
     def __del__(self) -> None:
         if not self.is_closed:
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -95,7 +95,7 @@ class BaseClient:
         }
         self._trust_env = trust_env
         self._netrc = NetRCInfo()
-        self._is_closed = True
+        self._is_closed = False
 
     @property
     def is_closed(self) -> bool:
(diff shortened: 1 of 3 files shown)


florimondmanca (approved):
Exactly what I had in mind :)

httpx/_client.py
```diff
@@ -1097,20 +1095,18 @@ def close(self) -> None:
         """
         Close transport and proxies.
         """
-        if not self.is_closed:
```
johtso, line 1100: So we're safe to assume that `.close()` can be called on transports multiple times without problem? I guess anything else would be a badly written transport..

tomchristie: I think that's reasonable yeah. Also, we're in line with requests here.

I guess we *could* tweak things in a future version bump so that calling `.close()` on an already closed client would result in a loud explicit error. (Just as we'd also like to tighten things up so that calling `.request()` on a closed client should result in an error.)

johtso:
I think we might also want to expand the transport closing test?

```python
class TestTransportClosing:
    def __init__(self):
        # Note that we're including 'proxies' here to *also* run through the
        # proxy context management, although we can't easily test that at the
        # moment, since we can't add proxies as transport instances.
        #
        # Once we have a more generalised Mount API we'll be able to remove this
        # in favour of ensuring all mounts are context managed, which will
        # also necessarily include proxies.

        self.transport = Transport()
        self.client = httpx.Client(transport=self.transport, proxies="http://www.example.com")

    def test_with_context_manager(self):
        assert self.transport.events == []

        with self.client:
            pass

        assert self.transport.events == [
            "transport.__enter__",
            "transport.close",
            "transport.__exit__",
        ]

    def test_without_context_manager(self):
        assert self.transport.events == []

        self.client.close()

        assert self.transport.events == [
            "transport.close"
        ]
```

johtso:
Also, we now get a lot of unclosed client UserWarnings when running the tests.

tomchristie:
> Also, we now get a lot of unclosed client UserWarnings when running the tests.

Okay, right, which kinda gets to the heart of the two different options here.

We can either:

* Suggest that transport instances *shouldn't* hold resources on init, but should either do so on `__enter__` or if opened implicitly the first time a `.request()` is made.
* Treat clients as open as soon as they're instantiated, and if they've not been explicitly closed, then close them on `__del__`.

We're in a slightly different position with `Client` and `AsyncClient` because `Client.__init__` might perform I/O and allocate resources, and `Client.__del__` is always able to release them, whereas `AsyncClient.__init__` won't ever perform I/O, and `AsyncClient.__del__` is unable to perform I/O to release resources.

tomchristie:
> Also, we now get a lot of unclosed client UserWarnings when running the tests.

So, I've made a slight tweak here, *for AsyncClient only*...

`__del__` should only raise a warning, if the client is unclosed *and* has performed I/O. (A request or `__aenter__`)

(We don't care about this for `Client` because `__del__` never needs to raise a warning, since it can always just call `.close` on an unclosed client)

Not sure what we think about this.

Incidentally wrt. CacheControl if you're expecting to expand it into sync+async support then you *will* want to only initialise the resources either on `__enter__` or on the first time `request` is called, because async can't make I/O in the `__init__` call. (because it's always a sync method)

johtso:
Was thinking exactly along those lines, makes sense to me.

> Incidentally wrt. CacheControl if you're expecting to expand it into sync+async support then you _will_ want to only initialise the resources either on `__enter__` or on the first time `request` is called, because async can't make I/O in the `__init__` call. (because it's always a sync method)

Ah yes, look forward to exploring the a/sync rabbit hole!

florimondmanca:
@tomchristie Heads up — we've got an interesting real-world case that we could test this new logic against in #1332.

tomchristie:
Superseeded by #1346

PR #1321, opened 2020-09-24
Remove seed-isort-config

johtso:
As of isort 5 seed-isort-config is no longer needed.

We can get rid of some explicit config too as it's now more clever.


tomchristie (commented):
Lovely, thanks!

tomchristie (approved):

PR #1326, opened 2020-09-25
Support header comparisons with dict or list.

### LINKED ISSUE #1325: Support header comparisons with dict / list.
We should support direct header comparisons with dicts, or with lists of two-tuples.
This would be nice because it would help folks clean up test cases, eg...

```python
response = httpx.get(...)
assert response.headers == {"Content-Length": "13"}
```

The fix will be something like changing `Headers.__eq__` from this...

```python
def __eq__(self, other: typing.Any) -> bool:
    if not isinstance(other, Headers):
        return False
    return sorted(self._list) == sorted(other._list)
```

To this...

```python
def __eq__(self, other: typing.Any) -> bool:
    try:
        other_headers = Headers(other)
    except ValueError:
        return False
    return sorted(self._list) == sorted(other_headers._list)
```

A pull request resolving this should:

* Make the fix in `Headers.__eq__`.
* Update all our tests to use the simpler comparison style wherever possible.

This would be a nice contributor friendly issue, so if anyone's keen then feel free to jump in.

musale:
Allows comparison of headers with dict or list:

```
response = httpx.get(...)
assert response.headers == {"Content-Length": "13"} # True
assert response.headers == [("Content-Length", "13")] # True
```

Closes #1325

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/mus/Projects/httpx/venv/bin/python3"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "/Users/mus/Projects/httpx/venv/bin/python3"
-}
--- a/tests/models/test_headers.py
+++ b/tests/models/test_headers.py
@@ -23,7 +23,8 @@ def test_headers():
     assert dict(h) == {"a": "123, 456", "b": "789"}
     assert repr(h) == "Headers([('a', '123'), ('a', '456'), ('b', '789')])"
     assert h == httpx.Headers([("a", "123"), ("b", "789"), ("a", "456")])
-    assert h != [("a", "123"), ("A", "456"), ("b", "789")]
+    assert h == [("a", "123"), ("A", "456"), ("b", "789")]
+    assert h == {"a": "123", "A": "456", "b": "789"}
 
     h = httpx.Headers({"a": "123", "b": "789"})
     assert h["A"] == "123"
--- a/tests/models/test_headers.py
+++ b/tests/models/test_headers.py
@@ -22,7 +22,7 @@ def test_headers():
     assert list(h) == ["a", "b"]
     assert dict(h) == {"a": "123, 456", "b": "789"}
     assert repr(h) == "Headers([('a', '123'), ('a', '456'), ('b', '789')])"
-    assert h == httpx.Headers([("a", "123"), ("b", "789"), ("a", "456")])
+    assert h == [("a", "123"), ("b", "789"), ("a", "456")]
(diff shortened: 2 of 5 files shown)


tests/models/test_headers.py
```diff
@@ -23,7 +23,8 @@ def test_headers():
     assert dict(h) == {"a": "123, 456", "b": "789"}
     assert repr(h) == "Headers([('a', '123'), ('a', '456'), ('b', '789')])"
     assert h == httpx.Headers([("a", "123"), ("b", "789"), ("a", "456")])
-    assert h != [("a", "123"), ("A", "456"), ("b", "789")]
+    assert h == [("a", "123"), ("A", "456"), ("b", "789")]
+    assert h == {"a": "123", "A": "456", "b": "789"}
```
tomchristie, line 27: To keep coverage at 100% we'll also need a test against a no-headers-like item.

Eg...

```python
assert h != "a: 123\nA: 456\nb: 789"
```

tomchristie (commented):
Great, thanks!

There's also several other places in the test cases that could be fixed up too...

```shell
$ grep "== httpx.Headers" -r tests -n
tests/models/test_requests.py:29:    assert request.headers == httpx.Headers(
tests/models/test_requests.py:39:    assert request.headers == httpx.Headers(
tests/models/test_requests.py:52:    assert request.headers == httpx.Headers(
tests/models/test_requests.py:76:    assert request.headers == httpx.Headers(
tests/models/test_responses.py:47:    assert response.headers == httpx.Headers(
tests/models/test_responses.py:60:    assert response.headers == httpx.Headers(
tests/models/test_responses.py:74:    assert response.headers == httpx.Headers(
tests/models/test_responses.py:88:    assert response.headers == httpx.Headers(
tests/models/test_responses.py:745:    assert response.headers == httpx.Headers({"Transfer-Encoding": "chunked"})
tests/models/test_responses.py:754:    assert response.headers == httpx.Headers({"Content-Length": "8"})
tests/models/test_headers.py:25:    assert h == httpx.Headers([("a", "123"), ("b", "789"), ("a", "456")])```

**Comment by musale** [comment]
Great, let me fix up.

File: .vscode/settings.json [hunk 2]
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/mus/Projects/httpx/venv/bin/python3"
+}
```
tomchristie, line 3: We don't really want this file. You might want to setup a global gitignore to avoid including config that's specific to your local dev environment.

musale: Slipped me. Fixing.

tomchristie:
Fantastic stuff! 💚
We're good to go on this, right?

musale:
I see my master is behind. Let me ff.

musale:
Oh you did it. perfect!

tomchristie (approved):
Ace!

PR #1327, opened 2020-09-25
Version 0.15.4

tomchristie:
### Added

* Support direct comparisons between `Headers` and dicts or lists of two-tuples. Eg. `assert response.headers == {"Content-Length": 24}` (Pull #1326)

### Fixed

* Fix automatic `.read()` when `Response` instances are created with `content=<str>` (Pull #1324)


florimondmanca:
_One release a day keeps the bugs away_ 🎉 Love this workflow too!

PR #1330, opened 2020-09-28
Allow passing an ext to a Request

johtso:
This is nice if we want to create a `Request` server side (or in a `Transport`).

See #1091

Also, do we actually want to update `Request.ext` rather than making a copy?


tomchristie:
For now I'm going to close this off as stale.
I expect we *will* end up wanting this in due course, but I'd prefer to wait until we've got an absolutely clear framing use-case first.

PR #1334, opened 2020-10-01
Add `response.next_request`

tomchristie:
Currently when using `allow_redirects=False`, we have interfaces for `response.next()` and `response.anext()`,
which can be used to issue the next redirect request.

The intent of these is to mirror the interface from `requests`, which also has a `response.next`.

*However* It turns out that I've made an error in what I thought the interface we were mirroring does. When using `requests` with `allow_redirects=False`, the `response.next` *isn't* a callable. It's an optional request instance, that can be used with `client.send(...)` to send the next response.

Since one of our desired aims is for full parity with requests, we *ought* to be covering that API, and *not* the callable `.next` interface that we're currently exposing.

```python
client = httpx.Client()
request = client.build_request("GET", ...)
while request is not None:
    response = client.send(request, allow_redirects=False)
    request = response.next_request
```

This pull request adds `response.next_request` as a more clearly named version of the requests API.

In the next version bump I think we should follow up on this PR by removing `response.next()` and `response.anext()` as unnecessary bits of API.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -31,7 +31,7 @@
 
 ::: httpx.delete
     :docstring:
-    
+
 ::: httpx.stream
     :docstring:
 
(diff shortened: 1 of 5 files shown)


jcugat (approved):
I love how this feature it's just a couple lines of code, brilliant!

PR #1336, opened 2020-10-01
Allow covariants of __enter__, __aenter__, and @classmethod

Congee:
The problem we currently have is the return type of classes such as
Client does not allow covariants when subclassing or context manager.
In other words:

```python
class Base:
    def __enter__(self) -> Base:  # XXX
        return self

class Derived(Base):
    ...

with Derived() as derived:
   # The type of derived is Base but not Derived. It is WRONG
    ...
```

There are three approaches to improve type annotations.
1. Just do not type-annotate and let the type checker infer
   `return self`.
2. Use a generic type with a covariant bound
   `_AsyncClient = TypeVar('_AsyncClient', bound=AsyncClient)`
3. Use a generic type `T = TypeVar('T')` or `Self = TypeVar('Self')`

They have pros and cons.
1. It just works and is not friendly to developers as there is no type
   annotation at the first sight. A developer has to reveal its type via
   a type checker. Aslo, documentation tools that rely on type
   annotations lack the type. I haven't found any python docuementation
   tools that rely on type inference to infer `return self`. There are
   some tools simply check annotations.

2. This approach is correct and has a nice covariant bound that adds
   type safety. It is also nice to documentation tools and _somewhat_
   friendly to developers. Type checkers, pyright that I use, always
   shows the the bounded type '_AsyncClient' rather than the subtype.
   Aslo, it requires more key strokes. Not good, not good.

   It is used by `BaseException.with_traceback`
   See https://github.com/python/typeshed/pull/4298/files

3. This approach always type checks, and I believe it _will_ be the
   official solution in the future. Fun fact, Rust has a Self type
   keyword. It is slightly unfriendly to documentation, but is simple to
   implement and easy to understand for developers. Most importantly,
   type checkers love it.

   See https://github.com/python/mypy/issues/1212

But, we can have 2 and 3 combined:

```python
_Base = typing.TypeVar('_Base', bound=Base)

class Base:
   def __enter__(self: _Base) -> _Base:
      return self

class Derive(Base): ...

with Derived() as derived:
   ...  # type of derived is Derived and it's a subtype of Base
```

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1811,7 +1811,7 @@ class StreamContextManager:
         self.timeout = timeout
         self.close_client = close_client
 
-    def __enter__(self: _StreamContextManager) -> _StreamContextManager:
+    def __enter__(self) -> "Response":
         assert isinstance(self.client, Client)
         self.response = self.client.send(
             request=self.request,
@@ -1833,7 +1833,7 @@ class StreamContextManager:
         if self.close_client:
             self.client.close()
 
-    async def __aenter__(self: _StreamContextManager) -> _StreamContextManager:
+    async def __aenter__(self) -> "Response":
         assert isinstance(self.client, AsyncClient)
         self.response = await self.client.send(
             request=self.request,
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -56,6 +56,11 @@ from ._utils import (
     warn_deprecated,
 )
 
+# The type annotation for @classmethod and context managers here follows PEP 484
+# https://www.python.org/dev/peps/pep-0484/#annotating-instance-and-class-methods
+T = typing.TypeVar("T", covariant=True)
+
+
 logger = get_logger(__name__)
 
 KEEPALIVE_EXPIRY = 5.0
@@ -491,14 +496,6 @@ class BaseClient:
             return None
 
         return request.stream
-
-
-# The type annotation approach here follows PEP 484. A custom name `_Client` is
-# chosen over the convention `T` is because some documentation tools display
-# type annotations vertatimly. `_Client` indicates more practical meaning than
-# a simple `T`
-# https://www.python.org/dev/peps/pep-0484/#annotating-instance-and-class-methods
-_Client = typing.TypeVar("_Client", bound="Client")
 
 
 class Client(BaseClient):
@@ -1114,7 +1111,7 @@ class Client(BaseClient):
                 if proxy is not None:
                     proxy.close()
 
-    def __enter__(self: _Client) -> _Client:
+    def __enter__(self: T) -> T:
         self._transport.__enter__()
         for proxy in self._proxies.values():
             if proxy is not None:
@@ -1138,9 +1135,6 @@ class Client(BaseClient):
 
     def __del__(self) -> None:
         self.close()
-
-
-_AsyncClient = typing.TypeVar("_AsyncClient", bound="AsyncClient")
 
 
 class AsyncClient(BaseClient):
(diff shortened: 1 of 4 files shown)


Congee:
Hi, this PR is a follow up to #1333 . This PR introduces a typing problem for `StreamContextManager`, it returns a `httpx.Response`. I am not sure how to type it. How would you like it to be typed?

httpx/_client.py
```diff
@@ -1800,7 +1811,7 @@ def __init__(
         self.timeout = timeout
         self.close_client = close_client
 
-    def __enter__(self) -> "Response":
```
tomchristie, line 1803: I don't think there's any reason to change this line.

httpx/_client.py
```diff
@@ -1822,7 +1833,7 @@ def __exit__(
         if self.close_client:
             self.client.close()
 
-    async def __aenter__(self) -> "Response":
```
tomchristie, line 1825: Or this one?

httpx/_client.py
Congee: Yeah, if nobody is going to subclass StreamContextManager

Congee: Will revert this part back

httpx/_client.py
Congee: reverted

httpx/_client.py
```diff
@@ -493,6 +493,9 @@ def _redirect_stream(
         return request.stream
 
 
+_Client = typing.TypeVar("_Client", bound="Client")
```
tomchristie, line 496: Wondering if we should have a comment here linking to https://www.python.org/dev/peps/pep-0484/#annotating-instance-and-class-methods (?)

Congee: Did, and added some reasons behind the change here 😉

johtso: What about just naming it `ClientType`?

johtso: Or maybe that would imply `Type[Client]`..

johtso: May just be bikeshedding, if so ignore me. I'm not sure if there is actually a name that neatly encapsulates the concept of an upper bounded type variable. Maybe not worth trying :)

johtso: Possible colors to paint the shed: `TClient` `AnyClient`

tomchristie: I guess(?) I'd probably prefer that we follow the PEP and stick with `T`.

Congee: So, what's the conclusion here? I can make a change

florimondmanca: I'd also personally go for `T` (or `C`, or whatevs), as `_Client` feels more like a "private base class" than a "generic bounded type variable".

Congee: Will make a change. Keep in mind that switching to `T = TypeVar('T', covariant=True)` loses the ability to use `TypeVar(.., bound=...)`. After all, `T` is a generic name used everywhere and python's typing system does not support namespace.

Congee: changes pushed

Congee: Oops, without `bound=...`, mypy complains

Congee: As I predicted, lacking of namespace causes trouble. The following works:
```python
T = typing.TypeVar("T", bound="Client")
U = typing.TypeVar("U", bound="AsyncClient")
```
What do you think?

tomchristie (approved):

florimondmanca (approved):

PR #1339, opened 2020-10-02
Drop `.next()`/`.anext()` in favour of `response.next_request`

tomchristie:
See also #1334

Our `.next()` interface was intended to provide parity with `requests`, but actually provides the wrong API.

Dropping I/O methods from the `Response` class is surely a good thing to do if we can, and I'd far rather we had a simple `request` instance stored on the model instance, than a partially bound function call.

Using `allow_redirects=False` and explicitly following redirects now looks like this:

```python
client = httpx.Client()
request = client.build_request("GET", ...)
while request is not None:
    response = client.send(request, allow_redirects=False)
    request = response.next_request
```


florimondmanca (approved):
Superb 👍

jcugat (approved):
I guess we're ok doing this breaking change pre-1.0 right?

StephenBrown2:
> I guess we're ok doing this breaking change pre-1.0 right?

Since it's meant for `requests` compatibility, I'd say yes.

PR #1340, opened 2020-10-03
[PoC] Use context manager requests

florimondmanca:
Proof-of-concept pull request, that installs HTTPCore against https://github.com/encode/httpcore/pull/206 and demonstrates what would need to change to accomodate the new API.

This shouldn't be merged of course, but now I'm more confident that we can start by expecting the transport API to return either a response, _or_ a response context manager, and optionally enter the context via an exit stack.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1,3 +1,4 @@
+from contextlib import ExitStack
 import datetime
 import functools
 import typing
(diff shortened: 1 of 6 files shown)


florimondmanca:
Closing in favor of #1341.

PR #1341, opened 2020-10-03
Add support for context manager responses

florimondmanca:
Refs https://github.com/encode/httpcore/issues/145, originally proof-of-concept'd via #1340

Prepare the migration of the transport API returning context manager responses, by adding support for _either_ a response tuple (current behavior), _or_ a context manager containing a response tuple.

Once this is in, we can go ahead and move forward on the HTTPCore side, e.g. with https://github.com/encode/httpcore/pull/206, without risking breakage on the HTTPX side (since we'll support both versions of the transport API).


tomchristie:
Okey dokes, so I've worked through a POC showing how we can deal with this, by changing the stack so that we're using context manager interfaces throughout... https://gist.github.com/tomchristie/e6e5e06b6c7b77b9c726b74d1ba0acbc

Some nice things drop out of this, because it's always clear if a response is still in an opened state or not, because you're always inside a `with ... as response:` block. And you don't ever need to wrap bits of code up in `try ... except response.close()`, which for example we were doing with event hooks.

florimondmanca:
Closing in favor of #1355 since that looks more promising than this PR. :-)

PR #1342, opened 2020-10-03
Run tests against Python 3.9 and add trove classifier

michael-k:
Python 3.9.0 final is expected on Monday, 2020-10-05, see https://www.python.org/dev/peps/pep-0596/#schedule

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,11 +14,11 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8"]
+        python-version: ["3.6", "3.7", "3.8", "3.9.0-rc.2 - 3.9"]
 
     steps:
       - uses: "actions/checkout@v2"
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Python 3.9.0 is out!
https://www.python.org/downloads/
https://docs.python.org/release/3.9.0/whatsnew/changelog.html

.github/workflows/test-suite.yml
```diff
@@ -14,11 +14,11 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8"]
+        python-version: ["3.6", "3.7", "3.8", "3.9.0-rc.2 - 3.9"]
```
michael-k, line 17: I'll update that once 3.9.0 appears in https://github.com/actions/python-versions/blob/main/versions-manifest.json

tomchristie: Great yup. For now let's get this merged anyways, and update that once possible.

tomchristie (approved):

PR #1346, opened 2020-10-06
Tighten client closed-state behaviour

### LINKED ISSUE #1317: Client.close() does not close transports if unused

### LINKED ISSUE #1332: Unintuitive behavior with using an AsyncClient after it's been aclosed
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

I'm playing around with httpx to learn how async programming works in Python, and I've noticed it's pretty easy to get into a state where I have an async task using an AsyncClient that keeps running after I've exited the `with` block that created the client, i.e., after the the `__aclose__` method of the client has been called.

Here's a simple example in asyncio:

```python
import httpx
import asyncio

async def get(client, url):
    r = await client.get(url)
    print(r.text)

async def main():
        t1 = asyncio.create_task(get(client, "https://httpbin.org/delay/1"))
        t2 = asyncio.create_task(get(client, "https://httpbin.org/delay/2"))

    await t1
    await t2

if __name__ == "__main__":
    asyncio.run(main())
```

Here's a simple example in trio (where I ran into this):

```python
import httpx
import trio

async def get(client ,url):
    x = await client.get(url)
    print(x.text)

async def main():
    async with trio.open_nursery() as nursery:
            nursery.start_soon(get, client, "https://httpbin.org/delay/1")
            nursery.start_soon(get, client, "https://httpbin.org/delay/2")

if __name__ == "__main__":
    trio.run(main)
```

These produce the same output: they run the two requests in parallel, print the output, and then display the warning
```
.../venv/lib/python3.7/site-packages/httpx/_client.py:1777: UserWarning: Unclosed <httpx.AsyncClient object at 0x7fbe37ad33c8>. See https://www.python-httpx.org/async/#opening-and-closing-clients for details.
  f"Unclosed {self!r}. "
```

It's surprising to me that the code both _works_ and prints the warning. What's actually going on is that the `__aclose__` method sets a flag indicating the client's been closed and then calls `__aclose__` on the transport, which flushes and closes any open connections, but does not actually leave the transport in an unusable state. That means that new connections can still get created. Since in both cases the tasks are scheduled for after the `with` block has exited, I think that they _deterministically_ don't start until after `__aclose__` finishes, ie., they aren't affected by tearing down any existing connections, and they go on to create new ones. Then the `send` method goes and unsets the `_is_closed` flag (which I'm confused about - I see this behavior is intentional per https://github.com/encode/httpx/pull/1197#issuecomment-676489918 but I don't follow why) and you're left with an unclosed client at GC time.

I think it would be helpful for `__aclose__` to leave the client in an unusable state, e.g., set `self._transport = None`, or have `send` fail if `_is_closed` is set, if the intention is that clients should not be reopened once closed. That would prevent users from writing code where they think they're cleaning up properly, but they're actually cleaning up too early, while requests are still in flight.

If the intention is they can be reopened, it would be helpful for the URL above to mention this - I was greatly confused because I was using exactly the method in that URL (as you can see above), it just turns out I was _unintentionally_ reopening them. (Perhaps reopening a closed client should issue a warning of its own?)

### To reproduce

See above.

### Expected behavior

I expect this to either work properly and not print a warning, or to fail, as described above.

### Actual behavior

The code works but prints a warning because it works accidentally, as described above.

### Debugging material

See above.

### Environment

macOS, Python 3.7.3, httpx 0.15.4 from pip, trio 0.17.0.
- OS: macOS
- Python version: 3.7.3
- HTTPX version: 0.15.4
- Async environment: Both, as indicated above. trio 0.17.0
- HTTP proxy: no
- Custom certificates: no

### Additional context

none

tomchristie:
Closes #1317
Closes #1332

Tighens up the client open/closed state by making it tri-state under the hood: `UNOPENED`/`OPENED`/`CLOSED`.

End result of the changes here, are that only the following styles are valid...

```python
client = httpx.Client()
client.request(...)
client.close()  # If omitted, this will be called on `__del__` with sync client, or raise a warning on `__del__` with async client.

# Cannot send requests anymore.
```

Or...

```python
    client.request(...)

# Cannot send requests anymore.
```

Because we're tri-state now we can make sure that simply importing but not using a client won't raise errors.
(Because under the hood, the client is initially in an `UNOPENED` state.)

```python
# If we don't actually use this client, then `__del__` won't raise a warning here.
client = httpx.AsyncClient()
```

* Clients may no longer used once closed. Calling `client.get(...)` and pals will raise a runtime error if attempted.
* Clients that have not sent a request should still call `.close()` on the underlying transports if closed.
* AsyncClients that have not been opened should not raise a warning if deleted.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1,4 +1,5 @@
 import datetime
+import enum
 import functools
 import typing
 import warnings
@@ -71,6 +72,12 @@ ACCEPT_ENCODING = ", ".join(
 )
 
 
+class ClientState(enum.Enum):
+    UNOPENED = 1
+    OPENED = 2
+    CLOSED = 3
+
+
 class BaseClient:
     def __init__(
         self,
@@ -101,14 +108,14 @@ class BaseClient:
         }
         self._trust_env = trust_env
         self._netrc = NetRCInfo()
-        self._is_closed = True
+        self._state = ClientState.UNOPENED
 
     @property
     def is_closed(self) -> bool:
         """
         Check if the client being closed
         """
-        return self._is_closed
+        return self._state == ClientState.CLOSED
 
     @property
     def trust_env(self) -> bool:
@@ -750,8 +757,10 @@ class Client(BaseClient):
 
         [0]: /advanced/#request-instances
         """
-        self._is_closed = False
+        if self._state == ClientState.CLOSED:
+            raise RuntimeError("Cannot send a request, as the client has been closed.")
 
+        self._state = ClientState.OPENED
         timeout = self.timeout if isinstance(timeout, UnsetType) else Timeout(timeout)
 
         auth = self._build_request_auth(request, auth)
@@ -1104,8 +1113,8 @@ class Client(BaseClient):
         """
         Close transport and proxies.
         """
-        if not self.is_closed:
-            self._is_closed = True
+        if self._state != ClientState.CLOSED:
+            self._state = ClientState.CLOSED
 
             self._transport.close()
             for proxy in self._proxies.values():
@@ -1113,11 +1122,12 @@ class Client(BaseClient):
                     proxy.close()
 
     def __enter__(self: T) -> T:
+        self._state = ClientState.OPENED
+
         self._transport.__enter__()
         for proxy in self._proxies.values():
             if proxy is not None:
                 proxy.__enter__()
-        self._is_closed = False
         return self
 
     def __exit__(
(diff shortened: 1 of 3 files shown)


florimondmanca (approved):
👍👍👍

Fab!

Issue #1311, opened 2020-09-23
ConnectError: [Errno -2] Name or service not known when URL contains IPv6


filwie:
### Describe the bug

When trying to connect to HTTP or HTTP/2 server using URL containing IPv6 address
 `httpx.ConnectError: [Errno -2] Name or service not known` exception is raised.

### To reproduce

1. Run some http server.

    In my case it is a Docker container with [IPv6 enabled](https://docs.docker.com/config/daemon/ipv6/), so for example: 
    
   ```sh
   docker run --name=serv --rm hashicorp/http-echo -text=hello
   ```
   To get IPs of the container:

   ```sh
   docker inspect serv --format '{{ .NetworkSettings.GlobalIPv6Address  }}    {{ .NetworkSettings.IPAddress }}'
   ```

2. Run below code (after adjusting URLs to match IPsextracted from `docker inspect`
```python
import httpx
import requests

url = 'http://172.17.0.2:5678'
url6 = 'http://[2001:db8:1::242:ac11:2]:5678'

print(requests.get(url))
print(requests.get(url6))

print(httpx.get(url))
print(httpx.get(url6))
```

### Expected behavior

Response is returned

```txt
<Response [200 OK]>
```

### Actual behavior

`httpx.ConnectError: [Errno -2] Name or service not known` exception is raised

### Debugging material

Output from above snippet:

```
<Response [200]>
<Response [200]>
<Response [200 OK]>
Traceback (most recent call last):
  File "/home/user/repos/lb-test/.venv/lib/python3.7/site-packages/httpx/_exceptions.py", line 339, in map_exceptions
    yield
  File "/home/user/repos/lb-test/.venv/lib/python3.7/site-packages/httpx/_client.py", line 858, in _send_single_request
    ext={"timeout": timeout.as_dict()},
  File "/home/user/repos/lb-test/.venv/lib/python3.7/site-packages/httpcore/_sync/connection_pool.py", line 195, in request
    method, url, headers=headers, stream=stream, ext=ext
  File "/home/user/repos/lb-test/.venv/lib/python3.7/site-packages/httpcore/_sync/connection.py", line 87, in request
    self.socket = self._open_socket(timeout)
  File "/home/user/repos/lb-test/.venv/lib/python3.7/site-packages/httpcore/_sync/connection.py", line 113, in _open_socket
    local_address=self.local_address,
[...]

PR #1349, opened 2020-10-07
Correctly handle ipv6 addresses as a part of URL

cdeler:
Closes https://github.com/encode/httpx/issues/1311

There was a bug with processing an ipv6 address as a part of request URL.

What has been changed:
1. I implemented the functionality from https://github.com/encode/httpx/issues/1311#issuecomment-704224592
1. Also I found the problem in `URL` constructor. If we pass a `RawURL` with ipv6 as a host, then host should be enclosed by square brackets, otherwise `rfc3986.iri_reference(...)` won't process it

### DIFF
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -289,12 +289,8 @@ def test_url_with_url_encoded_path():
     assert url.raw_path == b"/path%20to%20somewhere"
 
 
-@pytest.mark.parametrize(
-    "url_str",
-    ["http://[::ffff:192.168.0.1]:5678/"],
-)
-def test_ipv6_ulr(url_str):
-    url = httpx.URL(url_str)
+def test_ipv6_url():
+    url = httpx.URL("http://[::ffff:192.168.0.1]:5678/")
 
     assert url.host == "::ffff:192.168.0.1"
     assert url.netloc == "[::ffff:192.168.0.1]:5678"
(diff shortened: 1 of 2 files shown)


tests/models/test_url.py
```diff
@@ -287,3 +287,37 @@ def test_url_with_url_encoded_path():
...
+    url = httpx.URL("http://[::ffff:192.168.0.1]:5678/")
+
+    assert url.host == "::ffff:192.168.0.1"
+    assert url.netloc == "[::ffff:192.168.0.1]:5678"
+
+
+@pytest.mark.parametrize(
+    "url_str",
+    [
+        "http://192.168.0.1:1234",
```
cdeler, line 302: The test with this parameter doesn't make sense.
```suggestion
        "http://127.0.0.1:1234",
```

tests/models/test_url.py
```diff
@@ -287,3 +287,37 @@ def test_url_with_url_encoded_path():
...
+    assert url.host == "::ffff:192.168.0.1"
+    assert url.netloc == "[::ffff:192.168.0.1]:5678"
+
+
+@pytest.mark.parametrize(
+    "url_str",
+    [
+        "http://127.0.0.1:1234",
+        "http://example.com:1234",
+        "http://[::ffff:192.168.0.2]:1234",
```
cdeler, line 304: Change this host as the above
```suggestion
        "http://[::ffff:127.0.0.1]:1234",
```

tomchristie (approved):
Looks great yup, nice one!

PR #1364, opened 2020-10-17
Add new project to Third Party Packages

abersheeran:


docs/third-party-packages.md
```diff
@@ -40,3 +40,13 @@ A utility for record and repeat an http request.
...
+
+## Projects
+
+<!-- NOTE: this list is in alphabetical order. -->
+
+### rpc.py
+
+[Github](https://github.com/abersheeran/rpc.py) - [Documentation](https://github.com/abersheeran/rpc.py#rpcpy)
+
+A RPC framework base on ASGI/WSGI. Use HTTPX as the client of the RPC service.
```
jcugat, line 52: Typo?

```suggestion
A RPC framework based on ASGI/WSGI. Use HTTPX as the client of the RPC service.
```

jcugat (approved):
LGTM but leaving the merge decision to @tomchristie or @florimondmanca to check if we want to add this new section.

florimondmanca:
@abersheeran To be honest, I'm having mixed feelings about adding this package to the list, given how _young_ it seems compared to existing entries. At the same time I can hear arguments in favor of being more liberal and welcoming all valid entries that do showcase some kind of integration with HTTPX, so not too fussed either way, but just wanted to voice this concern.

That being said, I think we'd be better off placing this under the "Plugins" section for now. (This isn't really a plugin, but the same applies to some other entries in the existing "Plugins" section, so what this prompts us is potentially renaming that section into something else.)

florimondmanca (approved):
Thanks!

florimondmanca:
What's wrong with tests timing out — do we have a deadlock somewhere? 🤔

PR #1366, opened 2020-10-24
[Sandbox] Trigger CI to investigate build timeout

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/1364#issuecomment-715883069

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+test
+
 <p align="center">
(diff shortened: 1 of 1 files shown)


florimondmanca:
OK, it's a `uvicorn` bug. Latest is `0.12.2`, but doesn't reproduce on `0.12.1`.

PR #1372, opened 2020-10-28
expose UNSET

lepture:
Since UNSET is widely used in lots of functions' parameters. When people want to subclass and rewrite a method, or create a function to call httpx, they will need to use this UNSET to define their own parameters too.

A simple example:

```python
def request_token(client, auth=UNSET, **kwargs):
    url = 'https://example.com/token'
    return client.get(url, auth=auth, **kwargs)
```

Here are some real examples:

1. https://github.com/lepture/authlib/blob/v0.15.2/authlib/integrations/httpx_client/oauth2_client.py#L117
2. https://github.com/lepture/authlib/blob/v0.15.2/authlib/integrations/httpx_client/oauth2_client.py#L136

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -2,7 +2,7 @@ from .__version__ import __description__, __title__, __version__
 from ._api import delete, get, head, options, patch, post, put, request, stream
 from ._auth import Auth, BasicAuth, DigestAuth
 from ._client import AsyncClient, Client
-from ._config import Limits, Proxy, Timeout, create_ssl_context
+from ._config import UNSET, Limits, Proxy, Timeout, create_ssl_context
 from ._exceptions import (
     CloseError,
     ConnectError,
@@ -93,6 +93,7 @@ __all__ = [
     "TimeoutException",
     "TooManyRedirects",
     "TransportError",
+    "UNSET",
     "UnsupportedProtocol",
     "URL",
     "WriteError",


florimondmanca:
@lepture Hi,

I think it's an interesting point, and I don't think I have an answer right now. I do agree that having this specific type leak through is a bit of a smell, eventually. What would a world in which each library has its own `UNSET` type, to account for the fact that `undefined` doesn't exist in Python? 😅 

But for now, for the specific use case of the two code examples you listed, it doesn't seem like we need to solve it _just yet_. In your 2 examples, it doesn't struck me why the `auth` has to be exposed outside of `**kwargs`? There are no special operation performed on `auth`, it's just passed as `self.post(..., auth=auth, **kwargs)`.

florimondmanca:
Closing pre-emptively since I do agree on the _problem_ (it's likely that `UnsetType` shouldn't leak to public methods), but exposing `UNSET` is not the right _solution_ I believe. We can still discuss alternatives, though.

lepture:
Yes, my given examples are not proper. Here is a proper one:     https://github.com/lepture/authlib/blob/2af1b9f2c81d5cb847491e96cf762f69215fcf2d/authlib/integrations/httpx_client/oauth2_client.py#L84

There is a check if auth is defined, if not it should attach one.

lepture:
And also, I have a suggestion, if there is a breaking change, please don’t release it with a bugfix version.

florimondmanca:
Gotcha. ;-)

PR #1385, opened 2020-11-07
Turn UNSET into a synonym for ellipsis

### LINKED ISSUE #1384: Reconsider UNSET
Prompted by #1372 

Right now several of our public API functions or methods have a reference to an `UnsetType` instantiated as `UNSET` in their signature. For example:

https://github.com/encode/httpx/blob/589c6e0f2f4f3821095bfc247417aa98a5dad838/httpx/_client.py#L690-L705

Here's the full list of public API objects and methods that refer to `UnsetType`:

* `Client.request`
* `Client.stream`
* `Client.send`
* `Client.get`, `Client.post`, `Client.put`, `Client.patch`, `Client.delete`, `Client.options`, `Client.head`
* Equivalent methods on `AsyncClient`
* `Timeout(...)`

This style was introduced in preference to eg `auth=None`. The goal was to be able to distinguish between for example "`auth` was not passed" (indicating "use the default `auth` on the client"), and "`auth` was passed explicitly with `None`" (indicating that the user _really does not want auth to be used_).

But the problem is that now this `UNSET` instance and `UnsetType` are part of the public API: if people want to override methods on `Client` and keep the default behavior, then they need to import `UNSET` and default to it in the method.

```python
import httpx
from httpx._config import UNSET  # Oops

class MyClient(httpx.Client):
    def post(self, *args, auth=UNSET, **kwargs):
        ...  # Perhaps pre-process `auth`
```

It's fair to ask whether those use cases are valid, and if people can't always find a way to _not_ have to refer to `UNSET`.

But there's also a case to be made for us dropping that "sentinel value" idea entirely. For example, we could:

* Accept a `**kwargs` and pull `auth` from there.
  * Pros: no need for a special sentinel value.
  * Cons: we lose type hinting on `auth` and `timeout` (at least until typed `**kwargs` are a thing, for example: python/mypy#4441) and autocompletion.
* Use a builtin instead, for example `auth="__UNSET__"`.
  * Pros: builtin type (string) means it removes the "I need to import `UNSET`" problem.
  * Cons:
    * Still a sentinel value which users need to be aware of.
    * To type this properly we'd need `Literal["__UNSET__"]`, and `Literal` is Py3.8+ (or we need to add `typing_extensions` as a dependency).

I'm not sure what option is better, or if there are other options to consider, but I'm flagging this for 1.0 because this touches the public API.

florimondmanca:
Closes #1384, refs https://github.com/encode/httpx/issues/1384#issuecomment-723429932.

For consideration, but I quite like this :-)

This PR changes `UNSET` and `UnsetType` so they refer to `...` and `builtins.ellipsis` respectively, i.e. the ellipsis special value.

This is a first step that gets us the behavior we want (users can override by setting `auth=...`, `timeout=...`, and then comparing `if auth is ...: ...`). But there's an obvious follow-up to update our code to drop those references internally. We may want to decide to keep the `UnsetType` and `UNSET` names available in `0.17.*` but mark them as deprecated — we can decide of this as a follow-up.

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -1,3 +1,4 @@
+import builtins
 import os
 import ssl
 import typing
@@ -32,12 +33,8 @@ DEFAULT_CIPHERS = ":".join(
 
 logger = get_logger(__name__)
 
-
-class UnsetType:
-    pass  # pragma: nocover
-
-
-UNSET = UnsetType()
+UnsetType = builtins.ellipsis
+UNSET = ...
 
 
 def create_ssl_context(


florimondmanca:
… :-)

```python
>>> import builtins
>>> builtins.ellipsis
AttributeError: module 'builtins' has no attribute 'ellipsis'
```

And yet `mypy` accepts `builtins.ellipsis`, since `builtins.pyi` defines it… Hmm, so, anyway, this won't lead us anywhere (`mypy` doesn't accept `Ellipsis` (which _does_ exist at runtime), saying it's an "improper type" (?)), so, closing…

tomchristie:
Nice little exploration yup!
I'd been considering exactly the same.

PR #1388, opened 2020-11-11
fix code example, in async.md

podhmo:
TypeError is occured.

```
TypeError: hello() takes 0 positional arguments but 1 was given
```


j178 (approved):
Thanks

Issue #1302, opened 2020-09-21
Add `httpx.HTTPTransport()` and `httpx.AsyncHTTPTransport()`


tomchristie:
Raising this separately out of https://github.com/encode/httpx/issues/1274 since I think it's worth surfacing in it's own right.

PR #1398, opened 2020-11-24
Add keepalive_expiry to Limits config

### LINKED ISSUE #1395: Customize Keep-Alive timeout
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

The change https://github.com/encode/httpx/pull/627 added a 5 second timeout for Keep-Alive connections. Once that timeout is reached and no new requests are made in that timeframe then the connection in the pool is dropped.

I'm working with an endpoint that has a longer timeout before the server sends the RST/ACK on an idle connection (~2 minutes). While there should usually be a steady stream of requests being made there can be a time where nothing was sent in that 5 second window. Once the connection is dropped from the pool then the client needs to create a new connection and do any authentication that is required. This authentication process could be quite expensive as depending on the auth protocol used, multiple round trips may be required.

## Describe the solution you would like.

Potentially a way to define the timoeut in `httpx.Timeout(keep_alive=x)` or maybe under `httpx.Limits(keep_alive=x)`.

## Describe alternatives you considered

I'm no expert on the HTTP spec but I would have assumed we could just rely on the server to send the RST, ACK as a way to determine if the socket has been closed and to only remove it from the pool then.

tomchristie:
Closes #1395

We do *also* want #1302, but I think it makes sense to include `keepalive_expiry` in the `Limits` config either ways around.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -64,7 +64,6 @@ U = typing.TypeVar("U", bound="AsyncClient")
 
 logger = get_logger(__name__)
 
-KEEPALIVE_EXPIRY = 5.0
 USER_AGENT = f"python-httpx/{__version__}"
 ACCEPT_ENCODING = ", ".join(
     [key for key in SUPPORTED_DECODERS.keys() if key != "identity"]
@@ -656,7 +655,7 @@ class Client(BaseClient):
             ssl_context=ssl_context,
             max_connections=limits.max_connections,
             max_keepalive_connections=limits.max_keepalive_connections,
-            keepalive_expiry=KEEPALIVE_EXPIRY,
+            keepalive_expiry=limits.keepalive_expiry,
             http2=http2,
         )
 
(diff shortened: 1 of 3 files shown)


florimondmanca (approved):
Nice and simple enough, yes. :-)

Don't think we have it yet, but it strikes me that a "Pooling configuration" section in the advanced docs could be a good contribution, and this new option could easily be showcased there.

tomchristie:
@florimondmanca Yeah that'd be a good one. TBH I think we could really do with *completely* overhauling our docs structure. We've mirrored requests so far, but actually I could see us doing better with more sections, and going into depth in each.

URLs, Headers, Request/Response content, Cookies, Authentication, Concurrency, Proxies, SSL, HTTP/2, Connection Pooling, etc...

Having fuller sections for each would also let us fill them out in more depth with good learning material, whereas when everything follows on in the advanced section we kinda have to keep each bit relatively brief.

florimondmanca:
@tomchristie Yes, I also feel like our docs could use some overhaul and restructuring... As always, docs from trio projects like trio and h11 are great inspiration sources on what _could be_. :) Tutorials for common problems and highlighting some best practices or common gotchas, in-depth guides about various pieces of functionality, etc.

PR #1399, opened 2020-11-25
Add `HTTPTransport` and `AsyncHTTPTransport`

### LINKED ISSUE #1302: Add `httpx.HTTPTransport()` and `httpx.AsyncHTTPTransport()`

tomchristie:
Closes #1302

Two points to note here...

1. I've not added corresponding proxy cases here. We've got two options there...

* Add `httpx.ProxyTransport()` / `httpx.AsyncProxyTransport()` classes.
* Add a `proxy` argument to the existing class. Switch which `httpcore` instance is used for `._pool` depending on that argument. `httpx.HTTPTransport(proxy=...)`, `httpx.AsyncHTTPTransport(proxy=...)`. I think I quite like the way that'd have a lower footprint.

2. I've currently used open-ended `kwargs` for the advanced options. I'm kinda thinking to switch these out to named `uds`, `local_address`, etc... options instead, although `kwargs` does at least have the benefit of allowing us to add options to `httpcore` and have them be exposed without any changes in `httpx`.


docs/async.md
```diff
@@ -112,6 +112,19 @@ async def upload_bytes():
 await client.post(url, data=upload_bytes())
 ```
 
+### Explicit transport instances
+
+When instantiating a transport instance directly, you need to use `httpx.AsyncHTTPProxy`.
```
florimondmanca, line 117: I think you meant `AsyncHTTPTransport`?

httpx/_transports/default.py
```diff
@@ -0,0 +1,131 @@
...
+The following additional keyword arguments are currently supported by httpcore...
+
+* uds: str
+* local_address: str
+* retries: int
+* backend: str ("auto", "asyncio", "trio", "curio", "anyio", "sync")
+
+Example usages...
+
+# Disable HTTP/2 on a single specfic domain.
```
florimondmanca, line 13: This kind of super granular usage enabled by mounts is _really_ cool. :-)

httpx/_transports/default.py
```diff
@@ -0,0 +1,131 @@
...
+class HTTPTransport(httpcore.SyncHTTPTransport):
+    def __init__(
+        self,
+        verify: VerifyTypes = True,
+        cert: CertTypes = None,
+        http2: bool = False,
+        limits: Limits = DEFAULT_LIMITS,
+        trust_env: bool = True,
+        **kwargs: typing.Any,
+    ):
```
florimondmanca, line 49: (Nit) I think we need `-> None` in these constructors for mypy to take them into account when type checking... ?

florimondmanca (approved):
Looking good! Is it expecting that keepalive changes are also in here, though?

docs/async.md
tomchristie: 🤭 Thanks!

httpx/_transports/default.py
tomchristie: Oop. Yup!

httpx/_transports/default.py
tomchristie: Indeed!
I've added this one to the docs too, now.

florimondmanca (approved):

httpx/_transports/default.py
```diff
@@ -0,0 +1,134 @@
...
+
+class HTTPTransport(httpcore.SyncHTTPTransport):
+    def __init__(
+        self,
+        verify: VerifyTypes = True,
+        cert: CertTypes = None,
+        http2: bool = False,
+        limits: Limits = DEFAULT_LIMITS,
+        trust_env: bool = True,
+        **kwargs: typing.Any,
```
florimondmanca, line 49: Final thoughts on `**kwargs` vs explicit params? I had switched to explicit params in #1431. My idea there was avoiding the [kwargs problem](http://ivory.idyll.org/blog/on-kwargs.html). I can see the benefit of `**kwargs` for maintainability (additions in HTTPCore are directly available in HTTPX without a code change), but at the same time there's a case to be made for readability & user experience. Was curious about your opinion. :-)

tomchristie:
@florimondmanca Okay, so...

I've gone for explicit arguments throughout. (Maxim of: Of two otherwise equally valid design options, choose the more constrained) Note that it highlights the fact that `uds`, `local_address`, and `retries` aren't currently supported by the proxy cases. Perhaps some of them should or not, unsure, but we should discuss that separately either way - we're not backtracking on anything here, just making it more obviously apparent.

I've also added the `proxy=...` argument directly here. It ends up rather neat, and is the most minimal increase the API surface area, so I reckon we ought to roll with that.

florimondmanca (approved):
Okay, this is looking neat. :+1:

remingtonc:
This is an exciting addition - any idea when it will be released?

PR #1401, opened 2020-11-25
Add `httpx.MockTransport()`

### LINKED ISSUE #1303: Add `httpx.MockTransport()`
Based on https://github.com/encode/httpx/blob/8ceb34f48672f89de2458df371e2624659e2c24f/tests/utils.py#L27-L75

Usage:

```python
def handler(request: httpx.Request) -> httpx.Response:
    return httpx.Response(200, json={"hello": "world"})


client = httpx.Client(transport=httpx.MockTransport(handler))
```

Since #1297 and #1293 we're almost at the point where we can have a really nice simple `httpx.MockTransport(handler=...)` implementation, which I think would be worthwhile, given how useful we find it in our own tests, and how it'd help round out the usefulness of the Transport API to our users (and provide a point of demonstration around being able to use `Request`/`Response` models in that context.)

~~The one outstanding bit of work that we still need really is the `.request`+`.arequest` refactoring mentioned in https://github.com/encode/httpx/issues/1274~~

Also worth mentioning that the following all fit together as as a nice related set of enhancements...

* Mount API - #977
* `httpx.HTTPTransport()` - #1302
* `httpx.MockTransport()` - This PR.

So it's *possible* that we ought to aim to release them together, just to bring some extra coherence to eg. "Here's the mount API we've just added, and here's the sort of stuff you can do with it"

tomchristie:
Closes #1303

Docs as follows...

---

### Mock transports

During testing it can often be useful to be able to mock out a transport, and return pre-determined responses, rather than making actual network requests.

The `httpx.MockTransport` class accepts a handler function, which can be used to map requests onto pre-determined responses:

```python
def handler(request):
    return httpx.Response(200, json={"text": "Hello, world!"})


# Switch to a mock transport, if the TESTING environment variable is set.
if os.environ['TESTING'].upper() == "TRUE":
else:
    transport = httpx.HTTPTransport()

```

For more advanced use-cases you might want to take a look at [the third-party mocking library, RESPX](https://lundberg.github.io/respx/), or the [pytest-httpx library](https://github.com/Colin-b/pytest_httpx).


florimondmanca (approved):
Superb! Excitingly clean. :-)

lundberg:
Hooray 🎉

Latest release of [RESPX](https://lundberg.github.io/respx/) (`0.16.2`) now has optional support for *not* patching the `httpcore` transports, and passing a RESPX router handler to the `MockTransport`.

```py
import httpx
import respx


mock_router = respx.MockRouter(using=None)
route = mock_router.get("https://example.org/").mock(return_value=httpx.Response(204))

transport = httpx.MockTransport(mock_router.handler)

with httpx.Client(transport=transport) as client:
    response = client.get("https://example.org/")
    assert response.status_code == 204
    assert route.called
```

florimondmanca (approved):
:100:

simonw:
Last minute suggestion: it looks like the `handler` function currently cannot be an `async def ...` function, even for the async version of the mock transport:

https://github.com/encode/httpx/blob/181639322e2f68c00393435ef9e2359ff2691331/httpx/_transports/mock.py#L43-L56

I can imagine cases where I might want to be able to write a handler function that uses `await ...` - to access an async database layer for example.

Maybe apply something like the [await me maybe pattern](https://simonwillison.net/2020/Sep/2/await-me-maybe/) here, so users can define a regular function or an async function for their handler?

florimondmanca:
@simonw Yes, that sounds pretty sensible, thanks for noticing! :) I also think async mock views could be useful to me in some cases. I'm not sure we need to impose "async def for async client", so something like await me maybe sounds sensible...

Issue #1405, opened 2020-12-01
Cannot request resources with paths that include a mix of path separator slashes and URL-encoded slashes


jbaayen:
With httpx 0.15.0 and newer, we are getting some unexpected behaviour with URLs that include url-encoded slashes.

For example, a call to the URL "http://a.b/c/a%2Fb" ends up requesting the path "/c/a/b" from the server, instead of "/c/a%2Fb", which is what we would expect to happen.

Using recent httpx versions, we don't see any way to request resources including a mix of path separator slashes and URL-encoded slashes.

jbaayen:
My interpretation of https://tools.ietf.org/html/rfc3986, Section 2.1, is that "%2F" is a valid way of including a non-delimiting slash in a URL.

jbaayen:
I get the impression that the problem is here: https://github.com/encode/httpx/blob/master/httpx/_models.py#L237

unquote(a/b) = a/b
unquote(a%2Fb) = a/b

and such there is no way to distinguish the two, after unquoting.

florimondmanca:
Hi @jbaayen 

Can you share a snippet of code you're trying to run?

In the models.py file you linked to, you can see that there's also a `raw_path` property, and that's what gets used in `.raw`, which is used when calling the HTTP transport.

If you're just passing in a string URL then it's possible we do some pre-processing based on `.path` that we shouldn't be doing (eg applying base URLs).

jbaayen:
Hi @florimondmanca , we use the `AsyncClient` `get` and `post` methods, passing in the URL as a string.

A snippet would be this: `await client.get("http://localhost/a%2Fc")`

vdwees:
@jbaayen The issue you mention wouldn't occur in that snippet, but does occur if the client is using a base_url. This is because the url is corrupted when accessing the URL.path attr here: https://github.com/encode/httpx/blob/master/httpx/_client.py#L330

URL.path should not be decoding slashes, and I wonder why it is doing any decoding at all.

florimondmanca:
Thanks all, useful pointers. Here's a sample reproduction case…

"Echo path" server:

```python
import json


async def app(scope, receive, send):
    assert scope["type"] == "http"

    body = json.dumps({"raw_path": scope["raw_path"].decode()}).encode()

    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [[b"content-type", b"text/plain"]],
        }
    )
    await send({"type": "http.response.body", "body": body})
```

Test script:

[...]

PR #1406, opened 2020-12-01
URL.path: don't unquote slashes as this corrupts the path

vdwees:
This PR addresses issue https://github.com/encode/httpx/issues/1405

Using `unquote` in a blanket way on the path can result in corrupting the path if the path contains encoded slashes. This change has some precedent in https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote, where the default behaviour is to treat the slash as a special case.

Personally, I am not sure URL.path should be doing `unquote` at all. My preference would be to drop the `unquote` behaviour entirely. Nevertheless, with this PR at least `URL.path` does not corrupt paths.

I'd be happy to write some tests if this approach is approved.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -234,7 +234,7 @@ class URL:
         assert url.path == "/pa th"
         """
         path = self._uri_reference.path or "/"
-        return unquote(path)
+        return "%2F".join(map(unquote, path.split("%2F")))
 
     @property
     def query(self) -> bytes:


vdwees:
I think this change is ugly, and #1407 solves #1405 much better, so I am closing it. However, I still am concerned that if URL.path was accidentally used for constructing paths within httpx itself, wouldn't users be likely to make the mistake also?

tomchristie:
We're very careful and deliberate around our `URL` API.

`.path` returns a string, without the query, and with any percent-encoded characters being URL escaped. That'll give you the same behaviour you'd typically see within a web application framework.

`.raw_path` returns full raw bytes, of the entire path (including any query string component.) No url escaping. No nonsense. This is what gets used for the `TARGET` part when constructing an HTTP request.

PR #1407, opened 2020-12-01
Properly preserve encoded slashes when using `base_url`

### LINKED ISSUE #1405: Cannot request resources with paths that include a mix of path separator slashes and URL-encoded slashes

florimondmanca:
**Update**: Prompted by #1405 

An alternative to #1406 that tweaks URL merging to operate on `.raw_path`, rather than `.path`. Rationale: `.path` is URL-decoded and might swallow pre-existing `%2F` characters into `/`. On the other hand `.raw_path` doesn't do URL-decoding, so it is suitable for handling encoded slashes as well.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -120,9 +120,9 @@ class BaseClient:
         return self._trust_env
 
     def _enforce_trailing_slash(self, url: URL) -> URL:
-        if url.path.endswith("/"):
+        if url.raw_path.endswith(b"/"):
             return url
-        return url.copy_with(path=url.path + "/")
+        return url.copy_with(raw_path=url.raw_path + b"/")
 
     def _get_proxy_map(
         self, proxies: typing.Optional[ProxiesTypes], allow_env_proxies: bool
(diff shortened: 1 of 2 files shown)


tomchristie:
Great, yup this is a reasonable change.

I'm going to scrub the "Closes 1405" notation in the description, since the OP there doesn't note any usage of base_url, so although it's been a useful prompt onto this, which is a nice little tweak, I *don't* think it's relevant to what they're seeing.

florimondmanca:
@tomchristie Makes sense! Are we ✅ on this one?

tomchristie (approved):
Lovely & neat, yup!

jbaayen:
I can confirm that this PR fixes #1405 .

tomchristie:
👍 Thanks all!

jbaayen:
Thank you!

Do you know yet when a new release might be made?

PR #1426, opened 2020-12-11
Allow tuple as input of query parameters.

SarunasAzna:
In the documentation it is stated that params can be dict, string or two
tuples. This allows to used two tuples. Previously it was possible to
use only tuple inside a list.


StephenBrown2:
[The documentation](https://www.python-httpx.org/api/#client) is not wrong, it states specifically that "a **list** of two-tuples" is allowed.
> `params` - _(optional)_ Query parameters to include in request URLs, as a string, dictionary, or list of two-tuples.

While a tuple of two-tuples should be fine, I'd like to hear the other maintainers thoughts on it. If accepted, you will also need to update the documentation to include "list or tuple of two-tuples." everywhere query params are mentioned, and adjust the `QueryParamTypes` signature: https://github.com/encode/httpx/blob/354c4cac1f1796ee4d8861a4c978807398852ad5/httpx/_types.py#L34-L41 to include `Tuple[Tuple[str, PrimitiveData]],` which should fix the currently failing mypy tests.

httpx/_models.py
```diff
@@ -441,7 +441,7 @@ def __init__(self, *args: QueryParamTypes, **kwargs: typing.Any) -> None:
             items = parse_qsl(value)
         elif isinstance(value, QueryParams):
             items = value.multi_items()
-        elif isinstance(value, list):
+        elif type(value) in [list, tuple]:
```
StephenBrown2, line 444: From the [docs for `type()`](https://docs.python.org/3.8/library/functions.html#type):
> The `isinstance()` built-in function is recommended for testing the type of an object, because it takes subclasses into account.

This would be better written as:
```suggestion
        elif any(isinstance(value, t) for t in [list, tuple]):
```

florimondmanca, line 444: In theory any `Sequence` of 2-tuples would do. We _could_ consider switching to `isinstance(value, typing.Sequence)`. But hmm, I don't think there's much of a case for anything else than lists and tuples, so this sounds good enough.

SarunasAzna:
@StephenBrown2 good point about the docs! So actually I think it is important to mention how I found this thing - at the moment we are moving some API agents from `requests` to ` httpx` and the test with two tuples failed.

StephenBrown2:
Ah yes, that is an important note! We certainly want to maintain requests compatibility where possible, so go ahead and make the changes I suggested and the others can weigh in once that's done.

httpx/_models.py
StephenBrown2: I'm also in favor of using the more generic `Sequence`, makes the docs easier to change as well.

florimondmanca: Even better is this... :)

```python
if isinstance(... (tuple, list)): ...
```

SarunasAzna: Hey, sorry for the ignorance of mypy, but this error message seems to make no sense to me:

httpx/_models.py:445: error: Incompatible types in assignment (expression has type "Union[**Sequence[Tuple[str, Union[str, int, float, bool, None]]]**, str, bytes]", variable has type "**Sequence[Tuple[str, Union[str, int, float, bool, None]]]**")

Does anyone have an idea what I am doing wrong here?

SarunasAzna: As per `isinstance(value, Sequence)` - it is a bit too broad. For example - String is also isinstance of Sequence:
```
>>> isinstance('asfd', Sequence)
True
```

florimondmanca (approved):
:+1: LGTM as this is — "sequence" in the docs, `(list, tuple)` in implementation.

florimondmanca:
@SarunasAzna Thanks!

PR #1431, opened 2020-12-18
Add default transport helpers

### LINKED ISSUE #1302: Add `httpx.HTTPTransport()` and `httpx.AsyncHTTPTransport()`

florimondmanca:
Closes #1302

Alternative to #1399 

Instead of adding new classes (`HTTPTransport`, `AsyncHTTPTransport`), we focus on the "provide defaults" part by providing "factory helpers", rather than wrapper classes. I think that might be lower-maintenance, but also it's more easily testable since we have a direct access to the HTTPCore default transport object.

Also contains unit tests, and an integration test for the "disable HTTP/2 on a single domain" use case.

* ~I'm also on the fence wrt `**kwargs` versus explicit parameters. Perhaps the latter?~ — Switched to explicit params.
* With this PR, it's clearer that for proxies we should definitely prefer a `proxy=...` parameter to `create_default_transport()`, that would switch between returning a connection pool or a proxy pool. But at the same time, it's not clear to me _which use cases_ a `proxy=...` parameter would allow solving that's not already solvable through the `proxies=...` routing and the `Proxy` config class… Well, obviously there's customizing `max_keepalive_connections` and the like, but that's super-low-level, so… Leaving it out for now.


tomchristie:
Interesting. 🤔

tomchristie:
From a little bit of thinking over...

My natural inclination here would be to prefer #1399 on the basis of consistency.

* `httpx.HTTPTransport`
* `httpx.MockTransport`
* `httpx.WSGITransport`
* `httpx.ASGITransport`

florimondmanca:
@tomchristie Yeah, upon reviewing again I'm actually on the same boat. :-)

Issue #1433, opened 2020-12-21
The handshake operation timed out - On requests https requests


FlorianREGAZ:
### Checklist
Wasn't able to find help in Issues or a solution Pull requests.

### Describe the bug

Error while requesting on any site via HTTPS.
HTTP works, but I need HTTPS for what I'm planning to do.

### To reproduce

```
# User/Pass Proxy
proxy = {
    'http://': 'http://{0}:{1}@{2}:{3}'.format(splitted[2], splitted[3], splitted[0], splitted[1]),
    'https://': 'https://{0}:{1}@{2}:{3}'.format(splitted[2], splitted[3], splitted[0], splitted[1])
}
c = httpx.Client(proxies=proxy)
response = c.get("https://ipinfo.io/json").text
print(response)
```

### Actual behavior
Error: 
> httpx.ProxyError: _ssl.c:1091: The handshake operation timed out

### Debugging material

```
Traceback (most recent call last):
  File "D:\x\venv\lib\site-packages\httpcore\_sync\http_proxy.py", line 217, in _tunnel_request
    ) = proxy_connection.request(
  File "D:\x\venv\lib\site-packages\httpcore\_sync\connection.py", line 93, in request
    self.socket = self._open_socket(timeout)
  File "D:\x\venv\lib\site-packages\httpcore\_sync\connection.py", line 119, in _open_socket
    return self.backend.open_tcp_stream(
  File "D:\x\venv\lib\site-packages\httpcore\_backends\sync.py", line 143, in open_tcp_stream
    return SyncSocketStream(sock=sock)
  File "C:\Program Files\Python38\lib\contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "D:\x\venv\lib\site-packages\httpcore\_exceptions.py", line 12, in map_exceptions
    raise to_exc(exc) from None
httpcore.ConnectTimeout: _ssl.c:1091: The handshake operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\x\venv\lib\site-packages\httpx\_exceptions.py", line 326, in map_exceptions
    yield
  File "D:\x\venv\lib\site-packages\httpx\_client.py", line 861, in _send_single_request
    (status_code, headers, stream, ext) = transport.request(
  File "D:\x\venv\lib\site-packages\httpcore\_sync\http_proxy.py", line 124, in request
    return self._tunnel_request(
[...]

Issue #1428, opened 2020-12-14
Invalid request URL using AsyncClient + HTTP/2 + FORWARD_ONLY Proxy


ech0-py:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug
The httpx.AsyncClient with HTTP/2 enabled and using FORWARD_ONLY Proxy builds incorrect request URL: the proxy URL concatenates with requested URL

### To reproduce
Pre-conditions:
- proxy server is configured to work in forward mode with enabled ssl, alpn, http/2
- destination server: simple aiohttp web server that gives valid resposes on any valid requests

The client's code:
```
import httpx

proxy_https4 = httpx.Proxy(
    url=f"https://10.9.0.1:4443",  # proxy addr:port
    mode='FORWARD_ONLY')

client = httpx.AsyncClient(
    proxies=proxy_https4,
    verify=False, 
    http2=True)

r = await client.get("https://some-domain.com/handler2")
print('>>>> r.text:', r.text)
await client.aclose()  # don't want to use context manager
```

### Expected behavior
Request passes through proxy-server correctly and 200 response from destination server is given back

### Actual behavior
Request passes through proxy-server but destination server returns 400 because it cannot handle the request due to broken url

### Debugging material
stdout log:
```
TRACE [2020-12-14 10:03:30] httpx._config - load_ssl_context verify=True cert=None trust_env=True http2=False
TRACE [2020-12-14 10:03:30] httpx._config - load_verify_locations cafile=/home/myuser/venv_37_111/lib/python3.7/site-packages/certifi/cacert.pem
TRACE [2020-12-14 10:03:30] httpx._config - load_ssl_context verify=True cert=None trust_env=True http2=False
TRACE [2020-12-14 10:03:30] httpx._config - load_verify_locations cafile=/home/myuser/venv_37_111/lib/python3.7/site-packages/certifi/cacert.pem
TRACE [2020-12-14 10:03:30] httpx._config - load_ssl_context verify=False cert=None trust_env=True http2=False
TRACE [2020-12-14 10:03:30] httpx._config - load_ssl_context verify=False cert=None trust_env=True http2=False
TRACE [2020-12-14 10:03:30] httpx._config - load_ssl_context verify=True cert=None trust_env=True http2=False
[...]

Issue #1424, opened 2020-12-08
ProxyError with https proxy url


ElliottMoos:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

I get a ProxyError when setting `TUNNEL_ONLY` to true in `httpx.Proxy()` for an https proxy url.

### To reproduce

```python
import httpx

proxy = httpx.Proxy(
       f"https://{PROXY_USER}:{PROXY_PASS}@{PROXY_HOST}:{PROXY_HTTPS_PORT}",
       mode="TUNNEL_ONLY",
)

proxies = {
    "http://": proxy,
    "https://": proxy,
}

def main():
    with httpx.Client(proxies=proxies) as client:
        response = client.get('https://ifconfig.me/ip')
        print('httpx', response.text)
```

### Expected behavior

```bash
httpx <ip>
```

### Actual behavior

```bash
received exception of type <class 'httpx.ProxyError'>, url=https://ifconfig.me/ip,
msg=[SSL: PRE_MAC_LENGTH_TOO_LONG] invalid alert (_ssl.c:1091)
```

### Environment

- OS: macOS
- Python version: 3.7.7
- HTTPX version: 0.16.1
- Async environment: No
- HTTP proxy: Yes
- Custom certificates: No

### Additional context

Issue [#897](https://github.com/encode/httpx/issues/897) is a similar issue but it is considered resolved.

florimondmanca:
@ElliottMoos Hello!

First question Id have here is: does this reproduce if not setting a particular proxy mode?

The default behavior is to tunnel HTTPS requests, so given that you're requesting an HTTPS resource (ifconfig.me) I would expect the issue to remain if you just dropped the mode option and just passed:

```python
proxies="https://..."
```

If that's the case then we can look at the dance happening when using an HTTPS proxy URL.

Also, are you able to make requests through this HTTPS proxy using another client, such as cURL or Requests?

ElliottMoos:
@florimondmanca Hi! Thanks for checking in.

I can reproduce this without the `TUNNEL_ONLY` setting, which makes sense considering the functionality you described.

cURL and requests both work. Here is the output of a little test:

requests version: 2.25.0

Code:
```python
proxies = {
    "http://": f"https://{PROXY_USER}:{PROXY_PASS}@<proxy>",
    "https://": f"https://{PROXY_USER}:{PROXY_PASS}@<proxy>",
[...]

Issue #1412, opened 2020-12-03
Proxy CONNECT request timed out when I set the local agent use window10


tracyqan:
when I set the local agent use window10

![20201203102732121](https://user-images.githubusercontent.com/45054429/100955955-79ae7980-3552-11eb-89b7-ddb32bb903ab.png)

use httpx will raise httpx.ConnectTimeout:

```python
httpx.get('https://github.com/')
```

```console
Traceback (most recent call last):
  File "***", line 74, in <module>
    httpx.get('https://github.com/')
  File "E:\Anaconda3\lib\site-packages\httpx\_api.py", line 193, in get
    trust_env=trust_env,
  File "E:\Anaconda3\lib\site-packages\httpx\_api.py", line 104, in request
    allow_redirects=allow_redirects,
  File "E:\Anaconda3\lib\site-packages\httpx\_client.py", line 734, in request
    request, auth=auth, allow_redirects=allow_redirects, timeout=timeout
  File "E:\Anaconda3\lib\site-packages\httpx\_client.py", line 772, in send
    history=[],
  File "E:\Anaconda3\lib\site-packages\httpx\_client.py", line 809, in _send_handling_auth
    history=history,
  File "E:\Anaconda3\lib\site-packages\httpx\_client.py", line 837, in _send_handling_redirects
    response = self._send_single_request(request, timeout)
  File "E:\Anaconda3\lib\site-packages\httpx\_client.py", line 866, in _send_single_request
    ext={"timeout": timeout.as_dict()},
  File "E:\Anaconda3\lib\contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "E:\Anaconda3\lib\site-packages\httpx\_exceptions.py", line 343, in map_exceptions
    raise mapped_exc(message, **kwargs) from exc  # type: ignore
httpx.ConnectTimeout: _ssl.c:1029: The handshake operation timed out
```

but when I use requests , not raise an error

```python
requests.get('https://github.com/')
```

may I ask why or how I should do?

florimondmanca:
@tracyqan Hey! Just saw this question.

Is that the full traceback? I would have expected another lower-level exception from HTTPCore to be present in the stack (see `from exc` in the `map_exceptions` traceback line). Could you share that here as well? That'd give a better hint at where exactly this handshake failure occurs.

Thanks!

tracyqan:
**Thanks for your reply!
When i run this code.**

```
import httpx
import requests
from loguru import logger
[...]

PR #1435, opened 2020-12-21
Add troubleshooting guide, with initial proxies entries

florimondmanca:
Prompted by https://github.com/encode/httpx/issues/1433#issuecomment-749012975

We're getting a bunch of related / similar / duplicate questions about common proxies gotchas. This PR introduces a new "Troubleshooting" page to the docs, as well as two entries that address two questions we've been having for some time:

* "`The handshake operation timed out`" on HTTPS requests when using a proxy (#1412, #1433, and more previous issues + questions in the community chat)
* Error when making requests to an HTTPS proxy (#1424, #1428)

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -583,6 +583,10 @@ with httpx.Client(proxies=proxies) as client:
     r = client.get("http://example.com")
 ```
 
+### Troubleshooting proxies
+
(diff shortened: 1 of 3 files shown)


euri10 (approved):
This is indeed super useful !

florimondmanca:
I need to do a new pass because the HTTP/2 issues were not actually exactly related to HTTPS proxies (user was using an HTTPS url to an HTTP proxy, which we might want to emphasize in this guide again).

PR #1437, opened 2020-12-25
Add curio docs in Supported async environments

### LINKED ISSUE #1418: Curio is missing in "Supported async environments"
HTTPCore has support for `curio`, but it is not listed yet in the docs: https://www.python-httpx.org/async/#supported-async-environments

Would be very nice if we added a new sub-section for `curio`, with a similar example usage snippet.

shan7030:
Add curio docs in Supported async environments as new sub-section below AsyncIO, Trio

Fixes #1418

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -156,6 +156,27 @@ trio.run(main)
 !!! important
     The `trio` package must be installed to use the Trio backend.
 
+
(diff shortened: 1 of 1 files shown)


florimondmanca (approved):
Superb, thank you @shan7030!

PR #1453, opened 2021-01-18
Moved all the metadata into setup.cfg.

KOLANICH:
Added pyproject.toml.
Version is now fetched from git tags.


florimondmanca:
@KOLANICH Hi, thanks for your submission.

Obviously since this is touching on tooling used by maintainers and contributors, I hope you'll understand there will be some... consideration before accepting a contribution like this.

For me the key question to ask is: What practical problem does switching things around solve for the HTTPX project? You mentioned git tags. We haven't really had issues with assigning versions in the past. Is there something else?

KOLANICH:
Just replacing non-declarative metadata with declarative. Under "non-declarative" I mean "you cannot reliable parse tuat in principle, you have to really execute Turing-complete code".

>We haven't really had issues with assigning versions in the past.

You have a custom piece of code for extracting a version from your code. Though it can be replaced with a reusable well-known module `read_version`, I feel like replacing fetching it from source with populating it into source and metadata from a git tag would be more useful, since it removes some unneeded manual work and saves part of a commit hash into version, so if someone gets troubles, his bug report would be a bit more informative.

tomchristie:
I'm going to close this off.  I'm sure it's a perfectly reasonable alternative, but I don't think the motivation is sufficiently clear vs. the cost of churn.

PR #1456, opened 2021-01-23
Added missing Request __init__ parameters

simonw:


florimondmanca (approved):
Thanks :)

florimondmanca:
Oh, looks like mypy 0.800 is out since Jan 22nd. Need to fix up a few things it seems.

https://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html?m=1

PR #1458, opened 2021-01-29
fix-1457: URL's full_path -> raw_path from pull #1285 in docs as well

### LINKED ISSUE #1457: Docs contain a removed URL property and lack an existing property
URL.full_path was replaced with URL.raw_path in pull #1285 but not in docs

DavidBord:
fixes #1457


florimondmanca (approved):
Good catch, thanks!

PR #1460, opened 2021-02-06
Fix doc capitalization

dsaxton:
Tiny aesthetic nitpick that's also consistent with the rest of the file

Edit: Unrelated mypy error had to be ignored, not sure what's happening there but looks like a change in mypy


florimondmanca (approved):
Thanks :)

Yes mypy 0.800 seems to be requiring some work to get along with new changes.

PR #1462, opened 2021-02-09
Add Changelog link to PyPI page

adamchainz:
Makes it easy to find out about changes, and it gets a little "present" icon, for example https://pypi.org/project/django-linear-migrations/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
     version=get_version("httpx"),
     url="https://github.com/encode/httpx",
     project_urls={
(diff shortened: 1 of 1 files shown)


florimondmanca (approved):
Thanks, lovely.

We already include the changelog in the `long_description`, so it does show up on the PyPI page: https://pypi.org/project/httpx/

But I'm happy with this extra hint to the changelog in GitHub. :-)

adamchainz:
Thanks! I'm gradually trying to "make this a thing" so there's a bit of a standard.

Issue #1464, opened 2021-02-10
An exception in StreamContextManager.__aclose__() overrides the original


adamhooper:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

I alluded to this in #1463. That issue is about the type of exception being raised; _this_ bug is about the _source_ of the exception being raised.

In this code block:

```
async with httpx_client.stream("GET", url) as response:
    async for blob in response.aiter_raw():
        pass
        # ... Now, imagine `aiter_raw()` raises an exception...
# ... and imagine `response.aclose()` raises an exception within `__aclose__()`.
```

The exception from `aiter_raw()` is caught by `__aclose__()`; the exception from `response.aclose()` is the one I see.

### To reproduce

See #1463 for example code that causes a "Connection reset" error within the `aiter_raw()` block.

### Expected behavior

I expect to see a stack trace with `aiter_raw()` in it, because `aiter_raw()` caused an exception. (Sure, `response.aclose()` _also_ throws an exception; but as a user, do I really care?)

### Actual behavior

The stack trace is for the error in `response.aclose()`. `aiter_raw()` doesn't appear in it.

### Debugging material

See #1463

### Environment

- OS: Linux
- Python version: 3.8.7
- HTTPX version: 0.16.1
- Async environment: asyncio
- HTTP proxy: no
- Custom certificates: no

### Additional context

As a workaround, this code block will raise the correct exception:

```
request = client.build_request("GET", f"http://{host}:{port}")
response = await client.send(request, stream=True)
try:
    async for blob in response.aiter_raw():
        pass
except Exception as err:
    try:
        await response.aclose()
    finally:
        raise err
```

... this uses the inside knowledge that `aiter_raw()` calls `aclose()` to handle these three cases:

* Exception while streaming and exception from `aclose()`: ignore the exception from `aclose()` and raise the one from `aiter_raw()`.
* Exception while streaming and success from `aclose()`: raise the exception from `aiter_raw()`.
* Streaming succeeds but exception in `aiter_raw()`'s call to `aclose()`: we'll call `aclose()` again in our exception handler; it will no-op (because it's a second invocation); and we'll raise the exception from the original call to `aclose()`.

I've been using Python for decades, and I feel very uncomfortable nesting exception handlers like this.

florimondmanca:
@adamhooper I'm not sure yet I fully understood the problem here.

[...]

PR #1465, opened 2021-02-10
map_exceptions in request.aclose()

### LINKED ISSUE #1463: `response.aclose()` raises httpcore errors, not HTTPError
### Checklist

### Describe the bug

If a server resets the connection, then `response.aiter_bytes()` correctly raises an `httpx.HTTPError`.

-BUT- the error handling doesn't end there. If `response.aiter_raw()` doesn't succeed, it doesn't call `response.aclose()`. (I imagine it's the caller's responsibility?) So the caller will call `response.aclose()` ... and that will raise an `httpcore` exception. It can be caught ... but that's a big hassle and it's a hole in the abstraction layer.

```
response = await httpx_client.send(request, stream=True, timeout=timeout)  # raise HTTPErrors
try:
    async for blob in response.aiter_raw():
        pass
finally:
    await response.aclose()  # raise httpcore errors
```

```
async with httpx_client.stream("GET", url) as response:
    async for blob in response.aiter_raw():
        pass
        # but if aiter_raw() raises an exception, the aclose() will raise an httpcore error
```

### To reproduce

I'm having trouble with the server init, but this works enough of the time....:

```
import _socket
import asyncio
import io
import struct

import httpx
import pytest


@pytest.mark.asyncio
async def test_download_stream_connection_reset():
    async def respond_then_die(reader, writer):
        # Read request -- as we should
        await reader.readuntil(b"\r\n\r\n")
        # Send HTTP header and start sending data -- so the client is streaming
        writer.write(
            b"HTTP/1.1 200 OK\r\nContent-Length:1000\r\nContent-Type:text/plain\r\n\r\nstart"
        )
        await writer.drain()
        # Now force a TCP RST packet, which will cause response.aclose() to fail
        # http://deepix.github.io/2016/10/21/tcprst.html
        fd = writer._transport._sock.fileno()
        linger = struct.pack("=II", 1, 0)
        _socket.socket(fileno=fd).setsockopt(
            _socket.SOL_SOCKET, _socket.SO_LINGER, bytearray(linger)
        )
        writer.close()

    server = await asyncio.start_server(respond_then_die, "127.0.0.1")
    async with server:
        host, port = server.sockets[0].getsockname()
            async with client.stream("GET", f"http://{host}:{port}") as response:
                async for blob in response.aiter_raw():
                    pass
```

### Expected behavior

`httpx.ReadError: [Errno 104] Connection reset by peer`

### Actual behavior

`httpcore.CloseError: [Errno 104] Connection reset by peer`

### Debugging material

```
.tox/py38/lib/python3.8/site-packages/httpx/_client.py:1846: in __aexit__
    await self.response.aclose()
.tox/py38/lib/python3.8/site-packages/httpx/_models.py:1299: in aclose
    await self._on_close(self)
.tox/py38/lib/python3.8/site-packages/httpx/_client.py:1513: in on_close
    await stream.aclose()
.tox/py38/lib/python3.8/site-packages/httpcore/_async/connection_pool.py:69: in aclose
    await self.stream.aclose()
.tox/py38/lib/python3.8/site-packages/httpcore/_bytestreams.py:78: in aclose
    await self._aclose_func()
.tox/py38/lib/python3.8/site-packages/httpcore/_async/http11.py:192: in _response_closed
    await self.aclose()
.tox/py38/lib/python3.8/site-packages/httpcore/_async/http11.py:202: in aclose
    await self.socket.aclose()
.tox/py38/lib/python3.8/site-packages/httpcore/_backends/asyncio.py:198: in aclose
    await self.stream_writer.wait_closed()  # type: ignore
/usr/lib64/python3.8/contextlib.py:131: in __exit__
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

map = {<class 'OSError'>: <class 'httpcore.CloseError'>}

    @contextlib.contextmanager
    def map_exceptions(map: Dict[Type[Exception], Type[Exception]]) -> Iterator[None]:
        try:
            yield
        except Exception as exc:  # noqa: PIE786
            for from_exc, to_exc in map.items():
                if isinstance(exc, from_exc):
>                   raise to_exc(exc) from None
E                   httpcore.CloseError: [Errno 104] Connection reset by peer

.tox/py38/lib/python3.8/site-packages/httpcore/_exceptions.py:12: CloseError
------------------------------------------------- Captured log call --------------------------------------------------
ERROR    asyncio:base_events.py:1707 Exception in callback _SelectorSocketTransport._call_connection_lost(None)
handle: <Handle _SelectorSocketTransport._call_connection_lost(None)>
  File "/usr/lib64/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib64/python3.8/asyncio/selector_events.py", line 970, in _call_connection_lost
    super()._call_connection_lost(exc)
  File "/usr/lib64/python3.8/asyncio/selector_events.py", line 730, in _call_connection_lost
    self._sock.close()
  File "/usr/lib64/python3.8/socket.py", line 500, in close
    self._real_close()
  File "/usr/lib64/python3.8/socket.py", line 494, in _real_close
    _ss.close(self)
OSError: [Errno 9] Bad file descriptor
```

(That `map_exceptions()` is the one within `httpcore`, not the one that I would expect to adorn aclose() in `httpx`.)

### Environment

- OS: Linux
- Python version: Python 3.8.7
- HTTPX version: 0.16.1
- Async environment: asyncio
- HTTP proxy: no
- Custom certificates: no

### Additional context

I'm trying to perform HTTP fetches for non-tech-savvy users. I need to handle every exception to present a friendly message.

You might want to split this into _two_ errors, actually?:

1. `response.aclose()` exceptions aren't wrapped
2. `StreamContextManager.__aclose__()` raises whatever exception `response.aclose()` raised (should it instead raise the _original_ exception -- the one from `aiter_raw()`?)

adamhooper:
Now, clients can catch an exception that occurs during close.

[closes #1463]

### DIFF
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -338,17 +338,15 @@ async def test_response_aclose_map_exceptions():
     import asyncio
 
     server = await asyncio.start_server(respond_then_die, "127.0.0.1")
-    assert server.sockets is not None
-    addr = server.sockets[0].getsockname()
-    host, port = addr
-
-    serving = asyncio.ensure_future(server.serve_forever())  # py3.6
     try:
+        assert server.sockets is not None
+        addr = server.sockets[0].getsockname()
+        host, port = addr
+
         async with httpx.AsyncClient() as client:
             async with client.stream("GET", f"http://{host}:{port}") as response:
                 with pytest.raises(httpx.CloseError):
                     await response.aclose()
     finally:
-        serving.cancel()
         server.close()
         await server.wait_closed()
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -323,11 +323,13 @@ async def test_response_aclose_map_exceptions():
         # http://deepix.github.io/2016/10/21/tcprst.html
         sock = writer.transport.get_extra_info("socket")
         if os.name == "nt":
-            linger = struct.pack("=HH", 1, 0)
+            linger = struct.pack("=HH", 1, 0)  # pragma: nocover
         elif os.name == "posix":
-            linger = struct.pack("=II", 1, 0)
+            linger = struct.pack("=II", 1, 0)  # pragma: nocover
         else:
-            raise RuntimeError("Don't know how to reset connection on this OS")
+            raise RuntimeError(
+                "Don't know how to reset connection on this OS"
+            )  # pragma: nocover
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, bytearray(linger))
         writer.close()
 
@@ -336,7 +338,8 @@ async def test_response_aclose_map_exceptions():
     import asyncio
 
     server = await asyncio.start_server(respond_then_die, "127.0.0.1")
-    async with server:
+    await server.start_serving()  # py3.6
+    try:
         assert server.sockets is not None
         addr = server.sockets[0].getsockname()
         host, port = addr
@@ -344,3 +347,6 @@ async def test_response_aclose_map_exceptions():
             async with client.stream("GET", f"http://{host}:{port}") as response:
                 with pytest.raises(httpx.CloseError):
                     await response.aclose()
+    finally:
+        server.close()
+        await server.wait_closed()
--- a/tests/client/test_async_client.py
+++ b/tests/client/test_async_client.py
@@ -1,6 +1,3 @@
-import os
-import socket
-import struct
 import typing
 from datetime import timedelta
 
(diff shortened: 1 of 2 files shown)


adamhooper:
_sigh_ I'm having a hard time reproducing this error on py36. Is that strictly necessary? I feel the code fix is obviously correct and the unit test is spectacularly difficult to write.

florimondmanca:
@adamhooper Thanks!

For the unit test — I think we can leverage `transport=...`. I would suggest writing a quick custom transport that returns a `stream` whose `.aclose()` method raises an `httpcore` exception, and make sure that an `httpx` one is raised on the client side. If that makes sense?

tests/client/test_async_client.py
```diff
@@ -301,3 +301,22 @@ async def test_mounted_transport():
...
+        async def aclose(self):
+            raise httpcore.CloseError(OSError(104, "Connection reset by peer"))
+
+    def handle(request: httpx.Request) -> httpx.Response:
+        return httpx.Response(200, stream=BrokenStream())
+
+    async with httpx.AsyncClient(transport=httpx.MockTransport(handle)) as client:
+        async with client.stream("GET", "http://example.com") as response:
+            with pytest.raises(httpx.CloseError):
+                await response.aclose()
```
florimondmanca, line 322: Am I correct saying this would also reproduce for a regular `client.request()`? If so, should we update this to…?

```suggestion
        with pytest.raises(httpx.CloseError):
            await client.request("GET", "http://example.com")
    
        with pytest.raises(httpx.CloseError):
            async with client.stream("GET", "http://example.com"):
                pass  # Close on exit.
```

(Also dropped the explicit `response.aclose()` since the expected behavior is that exiting the `stream()` async context manager would do that for us, and fail with the proper exception.)

florimondmanca (commented):
Looking very neat!

Just a minor suggestion to increase test coverage slightly?

tests/client/test_async_client.py
adamhooper: You are correct. And the `await client.request()` test is a worthy addition.

You're the maintainer, so you are welcome to change to nix the `aclose()` in the test ... but first, I'll explain my motivation.

During debugging, I was surprised by #1464 -- a broken stream raises the "wrong" error on shutdown.

I imagine a reasonable `__aexit__()` might _handle_ an errno=104 error -- by doing nothing and raising nothing. So I aimed for an explicit test that avoids `__aexit__()`.

The change would be _correct_. But as a user, I find `__axit__()` behavior confusing; so I wrote a more-explicit test.

tomchristie: Yup, I like the way @adamhooper has approached the test case here, too.

tomchristie (approved):
Good catch here, yup. Happy with this one.

PR #1470, opened 2021-02-17
Explain SSL_CERT_DIR specific format

pquentin:
It's easy to believe that any .pem files there will get picked up automatically, but that's not the case. I've reused the link that the Python docs use in https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations.

Quoting a coworker: "weird thing is that we tried to use `SSL_CERT_DIR` and it didn't work, but `SSL_CERT_FILE` did".


tomchristie (approved):
Much better, thank you!

PR #1487, opened 2021-02-24
Added an .editorconfig matching the settings mandatory for the contributors of this project.

KOLANICH:


florimondmanca:
@KOLANICH Thanks for putting some time to open this PR, but I don't think we need this. I assume most editors follow most of these rules for Python code already.

If you'd like your IDE to enforce these rules feel free to edit a global `~/.gitignore` so you can have the file but exclude it from VCS.

Thank you!

PR #1491, opened 2021-03-01
Add support for response context manager interface

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/1355#issuecomment-787878698

This PR is now a plain refactor that adds support for HTTPCore returning context-managed responses, should we choose to move forward with https://github.com/encode/httpcore/pull/206.

This ends up basically being a re-issuing of #1341.

I know we discussed "let's just switch to context managers throughout the stack" (https://github.com/encode/httpx/pull/1341#issuecomment-704368692), but eventually all the changes in #1355 didn't seem to be worth the hassle to me. If all we want is to make the Transport API more resource-leak-safe, then it shouldn't change anything to HTTPX's API since we already handle resource closure properly here.

We could look at reorganizing things to use the context-managed interface more thoroughly _internally_ (to simplify a bunch of `try/except/finally` code around `close()` and `aclose()`), but all I'm saying is that it doesn't matter as far as compatibility with https://github.com/encode/httpcore/pull/206 is concerned.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -870,7 +870,7 @@ class Client(BaseClient):
             )
             try:
                 (status_code, headers, stream, ext) = raw_response
-            except ValueError:
+            except ValueError:  # pragma: no cover
                 response_ctx = cast_context_manager(raw_response)
                 (status_code, headers, stream, ext) = exit_stack.enter_context(
                     response_ctx
@@ -1522,7 +1522,7 @@ class AsyncClient(BaseClient):
             )
             try:
                 (status_code, headers, stream, ext) = await raw_response
-            except ValueError:
+            except ValueError:  # pragma: no cover
                 response_ctx = cast_async_context_manager(raw_response)
                 (
                     status_code,
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1,3 +1,4 @@
+import contextlib
 import datetime
 import enum
 import typing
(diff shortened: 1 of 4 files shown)


tomchristie:
Okay, after much hair-tearing I think it's time for us to call a close on this.

There's no other existing Python networking libraries that adopts this approach wrt. handling the network close, and while it *does* have some point to recommend it, it also results in longer tracebacks, more awkward under the hood wrangling, and ultimately a Transport API that'd be *less* obvious to most of our user base, than a standard ol' file-like-ish stream with `.close()`/`.aclose()` methods.

adriangb:
> There's no other existing Python networking libraries that adopts this approach wrt. handling the network close

@tomchristie I may be missing something, but doesn't aiohttp exclusively use context managers?

tomchristie:
> doesn't aiohttp exclusively use context managers?

I'm not sure what the API for that was at the time I originally made that note.

So...

Yes they do. *At the response level*.

But not [within the lower-level stack](https://github.com/aio-libs/aiohttp/blob/6285f614dbc816cbdec5147557505763dbc368d0/aiohttp/client.py#L326) or at their [Connector API layer](https://docs.aiohttp.org/en/stable/client_reference.html#connectors).

We *also* exclusively use context managers at the response level. Ie. You need to use a `client.stream` block if you want to hold the connection open, while other methods return a closed response.

But we don't use them at the transport API.

PR #1495, opened 2021-03-03
Incompatible with httpcore 0.12.0

Simon-Le:
The version 0.12.0 of httpcore is not compatible. When this version is installed, all the tests using `httpx.Client()` will fail with the following error

`AttributeError: 'Client' object has no attribute '_transport'`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
         "certifi",
         "sniffio",
         "rfc3986[idna2008]>=1.3,<2",
-        "httpcore==0.12.*",
+        "httpcore>=0.12.1,<0.13",
     ],
     extras_require={
         "http2": "h2==3.*",


florimondmanca:
Hey :wave:

Indeed, I'm also getting an error about `retries` which was only added in 0.12.1:

```python
$ pip install httpx httpcore==0.12.0
$ python
>>> import httpx
>>> client = httpx.Client()
  File "/Users/florimond.manca/.pyenv/versions/3.8.5/lib/python3.8/site-packages/httpx/_client.py", line 605, in __init__
    self._transport = self._init_transport(
  File "/Users/florimond.manca/.pyenv/versions/3.8.5/lib/python3.8/site-packages/httpx/_client.py", line 652, in _init_transport
    return HTTPTransport(
  File "/Users/florimond.manca/.pyenv/versions/3.8.5/lib/python3.8/site-packages/httpx/_transports/default.py", line 58, in __init__
    self._pool = httpcore.SyncConnectionPool(
TypeError: __init__() got an unexpected keyword argument 'retries'

And then exiting the interpreter shows this exception in `Client.__del__`:

```python
Exception ignored in: <function Client.__del__ at 0x1013363a0>
  File "/Users/florimond.manca/.pyenv/versions/3.8.5/lib/python3.8/site-packages/httpx/_client.py", line 1134, in __del__
    self.close()
  File "/Users/florimond.manca/.pyenv/versions/3.8.5/lib/python3.8/site-packages/httpx/_client.py", line 1106, in close
    self._transport.close()
AttributeError: 'Client' object has no attribute '_transport'
```

The reason is that `self._transport = ...` was aborted mid-way, so the `Client` object that's still in memory doesn't have a `_transport` attribute but `__del__` assumes so.

florimondmanca (approved):
Thanks!

johtso:
Argh, awkward. Is there any reasonable way to catch this kind of thing in tests? Testing with every combination of dependency versions specified in `setup.py` probably isn't viable. Maybe there's something simple we could do that would catch the majority of slip ups though? Maybe just test with the oldest and the newest version of httpcore allowed?

Colin-b:
i guess httpcore 0.12.1 should have been 0.13.0 if it was breaking compat, that's the real issue

florimondmanca:
Testing against the oldest HTTPCore version as per the HTTPX requirement sounds sensible.

I'm not sure about running this as part of every CI run, but maybe we could start with that. Only putting it as a task in the release job would mean we only catch these issues once the release is ready, ideally that would be before.

Another option is to actually work locally with the lowest HTTPCore possible. That would mean pinning HTTPCore strictly in our `requirements.txt`, such that any API changes done separately are caught because we'd need to bump that version to continue working on HTTPX. Then we could have a simple test in the test suite to verify that this number is effectively the lowest possible number for the range specified in `setup.py`. I think I'd prefer this approach. Thoughts?

PR #1514, opened 2021-03-16
Add typing to httpx._status_codes.StatusCodeCompat.__call__

hramezani:


florimondmanca:
Heya,

Thanks for this submission. Not sure what's the motivation behind it — esp since only one method out of 3 that are type-ignored was changd — but given that these methods are deprecated, I don't think we should proceed with this. Users should consider switching to `httpx.codes`, which provides full type hinting.

Thanks!

PR #1516, opened 2021-03-19
Context managed client.send()

tomchristie:
Reworking of #1355 in order to keep thinking a few things over.

* Client.send() changes from returning an "open" Response to being a context managed interface.
* AsyncClient.send() changes from returning an "open" Response to being a context managed interface.
* The internal StreamContextManager disappears — we don't need it anymore.
* Response(on_close) disappears - we don't need it anymore.
* Response.close() *does* still exist, but it's no longer an I/O operation, it's purely for state marking. (Hey, we're not going to let you attempt to read the stream anymore)

There's a bunch of stuff for us to discuss around 1.0 and these last bits of "should we be using a context managed interface for (1) the Transport API and (2) the send() method."

For me the important step forward is the core idea in #1515, that the Transport API should be an HTTPX interface, in the HTTPX package. The really nice thing there is that once we do that, the `httpcore` implementation and API doesn't strictly have to define what we're presenting to the end-user in httpx, which just makes everything easier, and allows use to consider each of these three separate issues *in isolation from one another*...

1. Should the Transport API be a context managed request/response interface?
2. Should the client.send() API (and httpx internals) be a context managed interface?
3. Should the httpcore API and internals use a context managed request/response interface?

I need to follow up on this one with a bit more docs & chat, but parking it here for now.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -2,6 +2,7 @@ import datetime
 import enum
 import typing
 import warnings
+from contextlib import contextmanager
 from types import TracebackType
 
 import httpcore
@@ -55,6 +56,12 @@ from ._utils import (
     same_origin,
     warn_deprecated,
 )
+
+try:
+    from contextlib import asynccontextmanager
+except ImportError:  # pragma: nocover
+    from async_generator import asynccontextmanager  # type: ignore
+
 
 # The type annotation for @classmethod and context managers here follows PEP 484
 # https://www.python.org/dev/peps/pep-0484/#annotating-instance-and-class-methods
@@ -232,51 +239,6 @@ class BaseClient:
     @params.setter
     def params(self, params: QueryParamTypes) -> None:
         self._params = QueryParams(params)
-
-    def stream(
-        self,
-        method: str,
-        url: URLTypes,
-        *,
-        content: RequestContent = None,
-        data: RequestData = None,
-        files: RequestFiles = None,
-        json: typing.Any = None,
-        params: QueryParamTypes = None,
-        headers: HeaderTypes = None,
-        cookies: CookieTypes = None,
-        auth: typing.Union[AuthTypes, UnsetType] = UNSET,
-        allow_redirects: bool = True,
-        timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
-    ) -> "StreamContextManager":
-        """
-        Alternative to `httpx.request()` that streams the response body
-        instead of loading it into memory at once.
-
-        **Parameters**: See `httpx.request`.
-
-        See also: [Streaming Responses][0]
-
-        [0]: /quickstart#streaming-responses
-        """
-        request = self.build_request(
-            method=method,
-            url=url,
-            content=content,
-            data=data,
-            files=files,
-            json=json,
-            params=params,
-            headers=headers,
-            cookies=cookies,
-        )
-        return StreamContextManager(
-            client=self,
-            request=request,
-            auth=auth,
-            allow_redirects=allow_redirects,
-            timeout=timeout,
-        )
 
     def build_request(
         self,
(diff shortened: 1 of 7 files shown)


tomchristie:
Potential API alternative... Keep the `send` interface as a plain ol' return a response, and have a `send_and_stream_response` for the streaming case.

So, this continues to work the same before and after...

```python
request = client.build_request("GET", "https://www.example.com")
```

But `send(stream=True)` is no longer a *thing*. The streaming case exposed an explicitly different method...

```python
request = client.build_request("GET", "https://www.example.com")
with client.send_and_stream_response(request) as response:
    ...
```

httpx/_models.py
```diff
@@ -1243,13 +1243,12 @@ def iter_raw(self, chunk_size: int = None) -> typing.Iterator[bytes]:
...
+        Mark the response as being closed. This will ensure that the response
+        body is no longer readable.
+
         Automatically called if the response body is read to completion.
         """
-        if not self.is_closed:
-            self.is_closed = True
-            if self._on_close is not None:
-                self._on_close(self)
+        self.is_closed = True
```
florimondmanca, line 1251: It seems we still have `on_close: Callable = None` as part of the Response constructor for now, right? (Probably an oversight given the PR description.)

httpx/_client.py
```diff
@@ -56,6 +57,12 @@
     warn_deprecated,
 )
 
+try:
+    from contextlib import asynccontextmanager  # type: ignore
+except ImportError:  # pragma: nocover
+    from async_generator import asynccontextmanager  # type: ignore
```
florimondmanca, line 63: Thoughts on the `_compat.py` pattern? Move this bit there and then do `from ._compat import asynccontextmanager` everywhere. Always found it ends up being worth doing eventually, and keeps compat code in a clear area.

httpx/_client.py
```diff
@@ -705,7 +667,8 @@ def request(
 
         ```python
-        response = client.send(request, ...)
+        with client.send(request, ...) as response:
+            response.read()
```
florimondmanca, line 671: I think _this_ might be worth discussing.

If we give folks `client.send()` and this returns a `response` context manager, shouldn't they expect `response` to be readily-usable, eg `response.text` returns the text right away?

Why would we want `.send()` to return a context manager, if we don't _actually_ need that interface except in the streaming case? Having to add `response.read()` each time as boilerplate seems like a smell, right?

Also, how do we help users understand what the effect of _not_ calling `response.read()` would be?

Given the pre-existing Requests model of ".request() gives me a response with the body readily available", I wonder about the implications of this.

tomchristie:
So I'm going to close this off, but a little discussion to illuminate first...

To expand on the motivation of why we might have wanted to do this:

Following in the footsteps of trio's structured concurrency constraints but wrt. resources... making sure that when *any* OS resource is opened we only do so within a `with` block, ensuring that we *don't* give developers a footgun with which to create unclosed resources.

If we wanted to follow that constraint it'd imply that we *shouldn't* expose any "return an opened response" API, which either means dropping `response = client.send(stream=True)`, either in favour of `send` always being a context block API, or in favour of a plain send method with no streaming and a `client.send_and_stream()` method.

But the downsides here are that our call stack gets wrapped in decorators all the way through. Tracebacks are harder to follow, and it's not *necessarily* true that it's easier for developers to follow the code flow.

There's also the downside, that in some contexts a framework might *not* provide the constructs that we need in order to be able to work within a context managed fashion. See eg. using starlette to proxy and stream a response, which currently *requires* a close-callback style. (Which is a pointer to a design failure there, but nonetheless, having to point devs to a "if you really need to do this, here's how to call `__(a)enter__`/`__(a)exit__` directly" section really just isn't graceful.

In any case the `.build_request()/.send()` pair is already a somewhat lower-level API, with `.request()` and `.stream()` providing the nicer "don't give devs an easy footgun" APIs.

So short story. I can't see this as a worthwhile trade off. The "resources *MUST* be context managed" approach clearly has some nice design properties, but I think it backs us into too much of a corner. (And https://github.com/encode/httpx/pull/998 helps show that we can bridge to that neatly enough for some parts if we really need to.)

PR #1518, opened 2021-03-20
Improve header byte pairs decoding in WSGITransport

laggardkernel:
According to `Headers.encoding()`, it's supported to decode header byte pairs encoded with utf-8 or latin-1 except ascii. Cause the header byte pairs received from httpcore connection may be non-ascii encoded.

On the other hand, it's also possible to send `Request` with non-ascii encoded `Headers`, 

```python
import httpx

headers = httpx.Headers({"foobar": "甲乙"}, encoding="utf-8")
url = "https://httpbin.org/anything"

r = httpx.get(url, headers=headers)
```

Non-ascii encoded `Headers` in `Request` is not a problem for `HTTPTransport`, `AsyncHTTPTransport` and `ASGITransport`. In these transports, the headers are sent as bytes. But for `WSGITransport`, we need to decode header byte pairs and put them into a `environ` dict.

In the current implementation of `WSGITransport`, we assume the header byte pairs in `Request` are only "ascii" encoded. This PR extends the header encoding support in `WSGITransport`, by piggybacking the encoding detect strategy of property `Headers.encoding()`.

### DIFF
--- a/httpx/_transports/wsgi.py
+++ b/httpx/_transports/wsgi.py
@@ -5,6 +5,8 @@ from urllib.parse import unquote
 
 import httpcore
 
+from .._models import Headers
+
 
 def _skip_leading_empty_chunks(body: typing.Iterable) -> typing.Iterable:
     body = iter(body)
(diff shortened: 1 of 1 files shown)


laggardkernel:
Updated once to separate the typo fix into another pr.

PR #1519, opened 2021-03-21
Make redirect strategy on HEAD complied with the doc

laggardkernel:
`httpx` follows the redirect strategy of `requests` that all APIs follow redirections by default **except** `HEAD`. This PR prevents `HEAD` following redirections when `allow_redirects` param is not set.

https://www.python-httpx.org/quickstart/#redirection-and-history


laggardkernel:
Sorry, did some seaching and found following redirection was enable by default on `HEAD` on purpose in GH-1183. Seems it's the doc that should be updated.

PR #1520, opened 2021-03-21
Fix redirect description about HEAD

laggardkernel:
`HEAD` is changed to follow redirects by default in GH-1183. Update the corresponding description in the doc.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -124,8 +124,11 @@ On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as 
 
 `requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). This is not supported by HTTPX.
 
-## Determining the next redirect request
+## HEAD redirection
+In `requests`, all top-level API follow redirects by default except `HEAD`.
+In consideration of consistency, we make `HEAD` follow redirects by default in HTTPX.
 
+## Determining the next redirect request
 When using `allow_redirects=False`, the `requests` library exposes an attribute `response.next`, which can be used to obtain the next redirect request.
 
 In HTTPX, this attribute is instead named `response.next_request`. For example:
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -125,10 +125,12 @@ On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as 
 `requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). This is not supported by HTTPX.
 
 ## HEAD redirection
+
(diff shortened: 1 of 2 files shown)


docs/compatibility.md
```diff
@@ -124,7 +124,9 @@ On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as
 
 `requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). This is not supported by HTTPX.
 
-## Determining the next redirect request
+## Redirection
```
florimondmanca, line 127: Should we consider a dedicated section? The next info about `response.next` is really separate and the current "Determining the next redirect request" is well-fitted for it.

Eg we could use a new section for this info, named:

```markdown
## HEAD redirection
```

florimondmanca (changes requested):
Thanks, worth adding to the docs indeed!

docs/compatibility.md
laggardkernel: @florimondmanca Done. Amended the commit and separated this part as "HEAD redirection".

docs/compatibility.md
```diff
@@ -124,8 +124,11 @@ On the other hand, HTTPX uses [HTTPCore](https://github.com/encode/httpcore) as
 
 `requests` omits `params` whose values are `None` (e.g. `requests.get(..., params={"foo": None})`). This is not supported by HTTPX.
 
-## Determining the next redirect request
+## HEAD redirection
+In `requests`, all top-level API follow redirects by default except `HEAD`.
+In consideration of consistency, we make `HEAD` follow redirects by default in HTTPX.
 
+## Determining the next redirect request
```
florimondmanca, line 131: Adding newlines to reduce diff and use a more consistent style:

```suggestion
## HEAD redirection

In `requests`, all top-level API follow redirects by default except `HEAD`.
In consideration of consistency, we make `HEAD` follow redirects by default in HTTPX.

## Determining the next redirect request

```

docs/quickstart.md
```diff
@@ -408,7 +408,11 @@ with additional API for accessing cookies by their domain or path.
 
 ## Redirection and History
 
-By default, HTTPX will follow redirects for anything except `HEAD` requests.
+By default, HTTPX will follow redirects for all HTTP methods.
+
+!!! warning
+    For anyone comes from Requests, you may be familar that `HEAD` in Requests 
+    doesn't follow redirects by default. But that is not the case in HTTPX.
```
florimondmanca, line 415: I'd suggest we leave that out? We already document it in the "Requests Compatibility Guide" now, and it would be duplicate info.

```suggestion
```

florimondmanca (approved):
Thanks @laggardkernel!

PR #1522, opened 2021-03-22
Transport API

tomchristie:
We have a little niggle in our API. Almost all the way through `httpcore` is treated as an implementation detail, that just happens to be used by the default transport implementation, `httpx.HTTPTransport`.

The one place where this isn't the case is in our "Transport API". In that case we're explicitly documenting "subclass httpcore.SyncHTTPTransport and implement it's API".

To make this clearer, here's how the "custom transports" docs look after this change...

---

### Writing custom transports

A transport instance must implement the low-level Transport API, which deals with sending a single request, and returning a response. You should either subclass `httpx.BaseTransport` to implement a transport to use with `Client`, or subclass `httpx.AsyncBaseTransport` to implement a transport to use with `AsyncClient`.

At the layer of the transport API we're simply using plain primitives. No `Request` or `Response` models, no fancy `URL` or `Header` handling. This strict point of cut-off provides a clear design separation between the HTTPX API, and the low-level network handling.

See the `handle_request` and `handle_async_request` docstrings for more details on the specifics of the Transport API.

A complete example of a custom transport implementation would be:

```python
import json
import httpx


class HelloWorldTransport(httpx.BaseTransport):
    """
    A mock transport that always returns a JSON "Hello, world!" response.
    """

    def handle_request(self, method, url, headers, stream, extensions):
        message = {"text": "Hello, world!"}
        content = json.dumps(message).encode("utf-8")
        stream = [content]
        headers = [(b"content-type", b"application/json")]
        extensions = {}
        return 200, headers, stream, extensions
```

Which we can use in the same way:

```pycon
>>> import httpx
>>> client = httpx.Client(transport=HelloWorldTransport())
>>> response = client.get("https://example.org/")
>>> response.json()
{"text": "Hello, world!"}
```

---

Here's a checklist of differences between the previous interface, for developers of custom transports.

- [x] Transports should subclass `httpx.BaseTransport` or `httpx.AsyncBaseTransport`, not the previous `httpcore` classes.
- [x] The override method is now `handle_request()`/`handle_async_request()`, not `request()`/`arequest()`. I think it's far cleaner not to have a name clash vs. `client.request()`.
- [x] The 'ext' argument is now named 'extensions'. Because.
- [x] The 'headers', 'stream', and 'extensions' arguments are no longer optional. Helps simplify implementations marginally.
- [x] Transports should raise exceptions from the `httpx.TransportError` hierarchy, instead of `httpcore` exceptions. Note that `httpcore`->`httpx` exception mapping is no longer provided by the client, but instead is handled more correctly within the `HTTPTransport()` implementation. 
- [x] Streams are now plain `typing.Iterable[byte]`/`typing.AsyncIterable[byte]`. Transports that require a close callback can use the `close`/`aclose` extension. Implementations that call *into* a transport should ensure they use a `try...finally` construct, and call any close callback within the `finally` block.
- [x] We now use `reason_phrase` instead of `reason` as an extension key, because it then properly follows the HTTP spec naming.
- [x] The `reason_phrase` and `http_version` extensions are now `bytes`, instead of `str`, for consistency with other elements in the interface, and because, when present in the response they are bytes on the wire. (Note that HTTP/2 onwards does not include version or reason information.)

Follow-ups that we're likely to want, that we can treat as seperate PRs once this is in...

- [ ] We'll want to issue an `httpcore` 0.13 version that *also* has an equivalent API update. Not strictly necessary, but it'd be neater.
- [ ] We *probably* want to introduce a convenience method `transport.send_request(method, url, ...)` that provides a nice `httpx`-level API onto transports, and returns fully fledged `Response` instances. This would be implemented once on the base classes, and would allow users to test transport instances directly. Eg. `r = t.send_request("GET", "https://www.example.com")` (Also note the naming `send_request()`/`handle_request()`, makes sense together, and avoids clashes with any methods on the client)


StephenBrown2:
Looking real nice!

> The 'ext' argument is now named 'extensions'. Because.

That bugged me too. Glad to see it.

Also in favor of the follow-ups, all seem reasonable and beneficial.

tomchristie:
I think this one is ready to go now.

@florimondmanca not sure what your thoughts on context managed interface at this layer are, but even if we *do* want to discuss that further I think we'd be way better off doing that with this as the starting point. The following core points all apply just as well regardless of if `handle_request` is a regular function call, or if it is a context manager...

* The Transport API should be an HTTPX interface.
* Exceptions raised by transports should be HTTPX exceptions. The client shouldn't have any baked-in httpcore knowledge.
* Let's use plain byte iterables for stream interfaces.
* We'd like to call this `handle_request` in preference to `request`.
* We don't need optional arguments at this interface layer.
* Let's use "extensions" not "ext"
* Let's use "reason_phrase" not "reason"
* Specify "reason_phrase" and "http_version" as bytes for preserving raw off-the-wire info.

tomchristie:
Presumably this proposed 0.18 API will be relevant to @lundberg (Because RESPX) and @johtso (Because I recall they've done some work on an HTTP Caching transport?)

florimondmanca:
@tomchristie What would be the status of HTTPCore at this point? Mostly an implementation that works well if wrapped around the HTTPX transport API?

tomchristie:
There's not really any difference in `httpcore` except that from the perspective of HTTPX users it's now entirely an implementation detail. We're not leaking any httpcore specific stuff into the HTTPX Transport API anymore.

The `handle_request` makes sense to me in for two reasons:

1. In order to properly differentiate it from the `request` method on the client.
2. So that we can introduce a convenience  `response = t.send_request("GET", "http://www.example.com")` method on the transport class. (A nice usability one for testing etc). A `handle_request`/`send_request` pair makes way more sense than a `handle_request`/`request` pair.

florimondmanca:
Sounds good. :-)

CHANGELOG.md
```diff
@@ -4,14 +4,31 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.17.1
+## Master
+
+The 0.18.x release series formalises our low-level Transport API, introducing the
+base classes `httpx.BaseTransport` and `httpx.AsyncBaseTransport`.
+
+See the "Writing custom transports" documention and the `httpx.BaseTransport.handle_request()`
```
florimondmanca, line 12:
```suggestion
See the "Writing custom transports" documentation and the `httpx.BaseTransport.handle_request()`
```

CHANGELOG.md
```diff
@@ -4,14 +4,31 @@ All notable changes to this project will be documented in this file.
...
+
+## 0.17.1 (March 15th, 2021)
 
 ### Fixed
 
 * Type annotation on `CertTypes` allows `keyfile` and `password` to be optional. (Pull #1503)
 * Fix httpcore pinned version. (Pull #1495)
 
-## 0.17.0
+## 0.17.0 (Februray 28th, 2021)
```
florimondmanca, line 31:
```suggestion
## 0.17.0 (February 28th, 2021)
```

httpx/_transports/base.py
```diff
@@ -0,0 +1,129 @@
...
+        traceback: TracebackType = None,
+    ) -> None:
+        self.close()
+
+    def handle_request(
+        self,
+        method: bytes,
+        url: typing.Tuple[bytes, bytes, typing.Optional[int], bytes],
+        headers: typing.List[typing.Tuple[bytes, bytes]],
+        stream: typing.Iterator[bytes],
```
florimondmanca, line 25: I understand this used to be an `httpcore.SyncByteStream` type, and we're now switching to standard "can be used in a `for` loop" type.

Wondering if we should be using `Iterable[bytes]` instead?

Mostly because of this…

```python
def f(stream: Iterator[bytes]) -> None:
    pass

stream = [b"hello, world"]

f(stream)  # error: Argument 1 to "f" has incompatible type "List[bytes]"; expected "Iterator[bytes]"
```

And…

```python
def f() -> typing.Iterator[bytes]:
    return [b"content"]  # error: Incompatible return value type (got "List[bytes]", expected "Iterator[bytes]"
```

So:

```suggestion
        stream: typing.Iterable[bytes],
```

httpx/_transports/base.py
```diff
@@ -0,0 +1,129 @@
...
+        traceback: TracebackType = None,
+    ) -> None:
+        await self.aclose()
+
+    async def handle_async_request(
+        self,
+        method: bytes,
+        url: typing.Tuple[bytes, bytes, typing.Optional[int], bytes],
+        headers: typing.List[typing.Tuple[bytes, bytes]],
+        stream: typing.AsyncIterator[bytes],
```
florimondmanca, line 119:
```suggestion
        stream: typing.AsyncIterable[bytes],
```

httpx/_transports/base.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def handle_request(
+        self,
+        method: bytes,
+        url: typing.Tuple[bytes, bytes, typing.Optional[int], bytes],
+        headers: typing.List[typing.Tuple[bytes, bytes]],
+        stream: typing.Iterator[bytes],
+        extensions: dict,
+    ) -> typing.Tuple[
+        int, typing.List[typing.Tuple[bytes, bytes]], typing.Iterator[bytes], dict
```
florimondmanca, line 28:
```suggestion
        int, typing.List[typing.Tuple[bytes, bytes]], typing.Iterable[bytes], dict
```

florimondmanca (commented):
Very satisfying.

Leaving some nits, and a somewhat important one about `Iterator` vs `Iterable` in the Transport API signature.

httpx/_transports/base.py
florimondmanca: (There are more of these in mock/asgi/wsgi transports.)

httpx/_transports/default.py
```diff
@@ -91,21 +163,44 @@ def __exit__(
...
+        extensions: dict,
+    ) -> typing.Tuple[
+        int, typing.List[typing.Tuple[bytes, bytes]], typing.Iterator[bytes], dict
+    ]:
+        with map_httpcore_exceptions():
+            status_code, headers, byte_stream, extensions = self._pool.request(
+                method=method,
+                url=url,
+                headers=headers,
+                stream=stream,  # type: ignore
```
florimondmanca, line 181: Do we / how do we plan to deal with this `type: ignore`? Shall we wrap in `httpcore.SyncIteratorStream` in the interim?

tomchristie: Good call. Using the concrete `httpcore.IteratorByteStream()` and `httpcore.AsyncIteratorByteStream()` classes to wrap this in the correct interface.

httpx/_transports/base.py
tomchristie: Yup. So easy to get those two confused!

docs/advanced.md
```diff
@@ -1084,23 +1092,22 @@ which transport an outgoing request should be routed via, with [the same style
...
         if port is None:
             location = b"https://%s%s" % (host, path)
         else:
             location = b"https://%s:%d%s" % (host, port, path)
-        stream = httpcore.PlainByteStream(b"")
+        stream = [b""]
         headers = [(b"location", location)]
-        ext = {"http_version": b"HTTP/1.1"}
+        extensions = {}
         return 303, headers, stream, ext
```
florimondmanca, line 1111:
```suggestion
        return 303, headers, stream, extensions
```

docs/advanced.md
```diff
@@ -1015,31 +1015,39 @@ This [public gist](https://gist.github.com/florimondmanca/d56764d78d748eb9f73165
...
+subclass `httpx.BaseTransport` to implement a transport to use with `Client`,
+or subclass `httpx.AsyncBaseTransport` to implement a transport to
+use with `AsyncClient`.
+
+At the layer of the transport API we're simply using plain primitives.
+No `Request` or `Response` models, no fancy `URL` or `Header` handling.
+This strict point of cut-off provides a clear design separation between the
+HTTPX API, and the low-level network handling.
+
+See the `handle_request` and `handle_async_request` docstrings for more details
```
florimondmanca, line 1029: Here's a nice example of why I'd be excited to get the ease of interlinking with Sphinx in HTTPX as well. :+1: https://github.com/encode/httpcore/pull/285

florimondmanca (commented):
Looking good, some last things I found!

httpx/_transports/asgi.py
```diff
@@ -68,17 +70,14 @@ def __init__(
...
         url: Tuple[bytes, bytes, Optional[int], bytes],
-        headers: List[Tuple[bytes, bytes]] = None,
-        stream: httpcore.AsyncByteStream = None,
-        ext: dict = None,
-    ) -> Tuple[int, List[Tuple[bytes, bytes]], httpcore.AsyncByteStream, dict]:
-        headers = [] if headers is None else headers
-        stream = httpcore.PlainByteStream(content=b"") if stream is None else stream
-
+        headers: List[Tuple[bytes, bytes]],
+        stream: typing.AsyncIterable[bytes],
```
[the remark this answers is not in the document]
tomchristie: Good catch. I've gone with consistency across the transport modules, switching `asgi.py` and `mock.py` to use `import typing`, in line with `base.py`, `default.py`, and `wsgi.py`.

This is the style we use in all our top level modules too, except for `_content.py` (Which we probably may as well switch for consistency), and `_types.py` (Which might be a bit unneccessarily grungy if we switched, so might want to leave as an exception.)

docs/advanced.md
tomchristie: I know, I get it. 🙂 We'll continue that convo over there.

docs/advanced.md
```diff
@@ -1015,31 +1015,39 @@ This [public gist](https://gist.github.com/florimondmanca/d56764d78d748eb9f73165
...
-should either subclass `httpcore.AsyncHTTPTransport` to implement a transport to
-use with `AsyncClient`, or subclass `httpcore.SyncHTTPTransport` to implement a
-transport to use with `Client`.
+A transport instance must implement the low-level Transport API, which deals
+with sending a single request, and returning a response. You should either
+subclass `httpx.BaseTransport` to implement a transport to use with `Client`,
+or subclass `httpx.AsyncBaseTransport` to implement a transport to
+use with `AsyncClient`.
+
+At the layer of the transport API we're simply using plain primitives.
```
florimondmanca, line 1024: (Nit) Drop "simply":

```suggestion
At the layer of the transport API we're using plain primitives.
```

florimondmanca (approved):
Looks like we're ready to roll. :)

tomchristie:
Ace, let's do this.

johtso:
Looks great! Will have a play trying to use this in [httpx-caching](https://github.com/johtso/httpx-caching) and see what the diff looks like :)

tomchristie:
So there's a close or aclose callback that may be present in the extensions...

https://github.com/encode/httpx/blob/master/httpx/_transports/base.py#L89

For your codebase it looks to me like where you're currently using `response.stream.close()` you'll instead want `close_callback = response.ext.get('close')` then `if close_callback is not None: close_callback()`.

(Or something similar)

johtso:
Thanks @tomchristie! I just figured this out after I saw errors when trying to serialize `extensions` :)

johtso:
So, that was fairly painless! https://github.com/johtso/httpx-caching/pull/4/files
I'm still using `httpcore.PlainByteStream` to wrap the response body in an async agnostic way when retrieved from the cache..

lundberg:
Now RESPX also has a PR https://github.com/lundberg/respx/pull/142 respecting this new transport API!

tomchristie:
@lundberg Fantastic! It'd be worth you (and @johtso) taking a look over https://github.com/encode/httpx/pull/1550, which I'm rather keen on.

balki:
This change silently breaks users of existing httpcore transport users by appearing to 'work' fine. 
E.g. https://github.com/romis2012/httpx-socks/blob/v0.3.1/httpx_socks/_async_transport.py#L11
`AsyncConnectionPool` happens to have `handle_async_request` method which gets called instead of the `arequest` method. 
This means the request does not go via the socks proxy but goes directly. Tor being a common usecase for socks proxy, with this change, users who thought they are connecting via tor are now exposed. 

May be add a check like

    if not isinstance(transport, httpx.AsyncBaseTransport): raise ....

romis2012:
> This change silently breaks users of existing httpcore transport users by appearing to 'work' fine.
E.g. https://github.com/romis2012/httpx-socks/blob/v0.3.1/httpx_socks/_async_transport.py#L11

This issue was fixed in httpx-socks v0.4.0

PR #1523, opened 2021-03-22
introduce http2_prior_knowledge option in httpx Client

p1-alexandrevaney:
This PR is to be used in conjuction with [this httpcore PR](https://github.com/encode/httpcore/pull/269).

It passes the `http2_prior_knowledge` option to the httpx client, which passes it to the httpcore `AsyncConnectionPool` instanciation.

I also added it to the http config model, to make sure that the alpn selected protocols only contains `"h2"` if the http2_prior_knowledge option is set to True.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -559,6 +559,7 @@ class Client(BaseClient):
         verify: VerifyTypes = True,
         cert: CertTypes = None,
         http2: bool = False,
+        http2_prior_knowledge: bool = False,
         proxies: ProxiesTypes = None,
         mounts: typing.Mapping[str, httpcore.SyncHTTPTransport] = None,
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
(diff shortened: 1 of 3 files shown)


tomchristie:
Okay, so...

Really this kind of PR could do with discussion first, although that's on us to help guide that process. (Which we're gradually starting to do.)

My initial thought on seeing this is that an API that adds `http1=True|False` might make more sense, since it'd be in line with our existing `http2=True|False`. When HTTP/1.x support is disabled on a client and HTTP/2 support is enabled, we could force HTTP/2 requests. (?)

p1-alexandrevaney:
Hello 😄,

Yes i agree that a boolean `http1=True|False` seems more in line with the existing code.
This PR is also made to be used with a [corresponding httpcore one](https://github.com/encode/httpcore/pull/269) which is heavier, could you please have a look at it so we can discuss what to change ?

PR #1524, opened 2021-03-22
Push httpcore exception wrapping out of client into transport

tomchristie:
Follow-on to #1522

Transports should now raise `httpx` exception classes, rather than `httpcore` exception classes.

* The httpcore -> httpx exception mapping now *only* occurs within the `HTTPTransport()` class, which uses the  `httpcore` library.
* The `RequestError` exceptions no longer require a request argument at the point of instantiation. The client class ensures that any `RequestError` raised by a transport will have a `request` attribute attached to it.
* Transports now use plain byte iterables, with `extensions['close']`/`extensions['aclose']` for any optional close callback.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -18,11 +18,10 @@ from ._config import (
 )
 from ._decoders import SUPPORTED_DECODERS
 from ._exceptions import (
-    HTTPCORE_EXC_MAP,
     InvalidURL,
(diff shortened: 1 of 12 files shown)


florimondmanca (approved):
The isolation of exception mapping to the default transports is pretty neat. 👍

PR #1535, opened 2021-03-25
Fix typo in docs to match `max_keepalive` default

lmmx:
[Simple docs typo PR]

Update the docs to match default value of `max_keepalive` as given in the variable `DEFAULT_LIMITS` in `_config` at https://github.com/encode/httpx/blob/master/httpx/_config.py#L359


tomchristie (approved):
Great, thanks!

PR #1539, opened 2021-03-25
Handle `data={"key": [None|int|float|bool]}` cases.

### LINKED ISSUE #1500: Form-encoded data with `None` as the value gets passed as string "None"
### Checklist

### Describe the bug

When posting form data using `requests` and the form value is `None`, nothing is passed. With `httpx`, the `None` is converted to the string `"None"`.

### To reproduce

```python
import requests
import httpx

url = 'https://httpbin.org/post'
data = dict(foo=None)

resp_requests = requests.post(url, data=data)
pprint(resp_requests.json())
# In the response:  'form': {},

resp_httpx = httpx.post(url, data=data)
pprint(resp_httpx.json())
# In the response:  'form': {'foo': 'None'},
```

### Expected behavior

1. Expected behavior to match `requests`, since "broadly" the behavior is expected to match
2. Did not expect that `"None"` would be passed, rather e.g. an empty string might make sense?

### Actual behavior

The passed form-encoded data contains the string "None"

### Debugging material

None, but see code example

### Environment

Ubuntu 20.04 docker image

### Additional context

Writing a PoC to replace `requests` with `httpx` right here.
https://github.com/SAP/cloud-pysec/blob/6e3f92fefb2e0dc4f3c99aa7ed5dc2d70e6734ba/sap/xssec/security_context.py#L511
The method `request_token_for_client` passes `None` as a default for "all scopes".

tomchristie:
Closes #1500

Allow integers, floats, booleans, and None as valid values for URLEncoded and Multipart content.

Use `"true"` and `"false"` string literal for booleans. `""` for the None case. And `str(value)` for anything else.

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -58,7 +58,7 @@ def normalize_header_value(
 
 def primitive_value_to_str(value: "PrimitiveData") -> str:
     """
-    Coerce a primitive data type into a string value for query params.
+    Coerce a primitive data type into a string value.
 
     Note that we prefer JSON-style 'true'/'false' for boolean values here.
     """
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -56,7 +56,7 @@ def normalize_header_value(
     return value.encode(encoding or "ascii")
 
 
-def str_query_param(value: "PrimitiveData") -> str:
+def primitive_value_to_str(value: "PrimitiveData") -> str:
     """
     Coerce a primitive data type into a string value for query params.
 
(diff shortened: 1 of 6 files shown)


httpx/_utils.py
```diff
@@ -56,7 +56,7 @@ def normalize_header_value(
     return value.encode(encoding or "ascii")
 
 
-def str_query_param(value: "PrimitiveData") -> str:
+def primitive_value_to_str(value: "PrimitiveData") -> str:
     """
     Coerce a primitive data type into a string value for query params.
```
florimondmanca, line 61: (Nit)

```suggestion
    Coerce a primitive data type into a string value.
```

florimondmanca (approved):
Sweet. :-)

PR #1541, opened 2021-03-26
Neater close logic

tomchristie:
- [x] Initially raised as discussion #1540

Only call into `.close()` on a client instance if it is garbage collected in the `OPENED` state.
This seems a self-evident improvement to me, since it now properly mirrors the async case.

Also:

* Extra comments on `UNOPENED`, `OPENED`, `CLOSED`
* Extra comments on `AsyncClient.__del__`

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -69,8 +69,16 @@ ACCEPT_ENCODING = ", ".join(
 
 
 class ClientState(enum.Enum):
+    # UNOPENED:
+    #   The client has been instantiated, but has not been used to send a request,
+    #   or been opened by entering the context of a `with` block.
     UNOPENED = 1
+    # OPENED:
+    #   The client has either sent a request, or is within a `with` block.
     OPENED = 2
+    # CLOSED:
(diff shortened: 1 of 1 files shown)


j178 (approved):

Hearty07:
so what is the different between UNOPENED and CLOSED? why it can't call get() again after close()?  what if I want to close current TCP connection and build another just like http.get() but not client.get()? if I keep the client's connection with server and do not close it all the way, it may cause reaching the max stream id and the server will not response any more. but I can't find any function to set client._state to UNOPENED again in the code, it means I can't call get() again after calling close(). but in fact, I can call get() again if I set client._state = ClientState.UNOPENED. so we back to the topic, what is the different between the state CLOSED and UNOPENED? can we let the state back to UNOPENED after calling close()? or can we offer a function to reconnect to the server? thanks.

httpx.RemoteProtocolError: <ConnectionTerminated error_code:ErrorCodes.NO_ERROR, last_stream_id:1999, additional_data:None>

PR #1549, opened 2021-04-01
Fix extenstions typo in AsyncHTTPTransport

lundberg:
Seems like a typo sneaked by in the `AsyncHTTPTransport` 😉 

I couldn't find a better way of testing it than asserting that the `http_version` was passed from the `server` all the way up to the `resonse.extensions`.


tomchristie:
Goodness. Thank you.
And yup, that test case makes sense.

PR #1550, opened 2021-04-01
Stream interface

tomchristie:
This pull request allows us to compare the following two alternatives...

* The HTTPX Transport API, using `stream` as a plain primitive `Iterable[bytes]`/`AsyncIterable[bytes]`, with `close`/`aclose` extensions for handling close callbacks. 
* The HTTPX Transport API, using simple `httpx.SyncByteStream`/`httpx.AsyncByteStream` interfaces for streams.

Note that the `request.stream` and `response.stream` interfaces also change as a result of this.

The pros of this approach are that the Transport API presents a more natural interface...

**Before:**

```python
with httpx.HTTPTransport() as transport:
    status_code, headers, stream, extensions = transport.handle_request(
        method=b'GET',
        url=(b'https', b'www.example.com', 443, b'/'),
        headers=[(b'Host', b'www.example.com')],
        stream=[],
        extensions={}
    )
    try:
        body = [part for part in stream]
    finally:
        if "close" in extensions:
            extensions["close"]()
```

**After:**

Standard case:

```python
with httpx.HTTPTransport() as transport:
    status_code, headers, stream, extensions = transport.handle_request(
        method=b'GET',
        url=(b'https', b'www.example.com', 443, b'/'),
        headers=[(b'Host', b'www.example.com')],
        stream=httpx.ByteStream(b""),
        extensions={}
    )
    body = stream.read()
```

Streaming case:

```python
with httpx.HTTPTransport() as transport:
    status_code, headers, stream, extensions = transport.handle_request(
        method=b'GET',
        url=(b'https', b'www.example.com', 443, b'/'),
        headers=[(b'Host', b'www.example.com')],
        stream=httpx.ByteStream(b""),
        extensions={}
    )
    try:
        body = [part for part in stream]
    finally:
        stream.close()
```

The cons of this approach are that we're adding `httpx.SyncByteStream` and `httpx.AsyncByteStream` instead of the "`handle_request` is just plain datatypes everywhere". We're also adding the `httpx.ByteStream()` class for the simple concrete bytes-only case.

Having had a bunch of time to think over the three possible cases...

* A stream interface
* A plain bytes interface, with `close`/`aclose` extensions to handle resource closing.
* A plain bytes interface, with context management to handle resource closing.

My feel is that this one probably hits the sweet spot of "nice low-level interface" while still being neatly usable and obvious from a usability perspective.

It's also lower impact if we want to update httpcore with an equivalent interface, since https://github.com/encode/httpcore/pull/296 in it's current state already matches up with this. (But gets more complex if we start adding in `close`/`aclose` extensions.)

### DIFF
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -24,6 +24,14 @@ def create_event() -> "Event":
         return asyncio.Event()
 
 
+class ASGIResponseStream(AsyncByteStream):
+    def __init__(self, body: typing.List[bytes]) -> None:
+        self._body = body
+
+    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+        yield b"".join(self._body)
+
+
 class ASGITransport(AsyncBaseTransport):
     """
     A custom AsyncTransport that handles sending requests directly to an ASGI app.
@@ -154,13 +162,6 @@ class ASGITransport(AsyncBaseTransport):
         assert response_complete.is_set()
         assert status_code is not None
         assert response_headers is not None
-
-        class ASGIResponseStream(AsyncByteStream):
-            def __init__(self, body: typing.List[bytes]) -> None:
-                self._body = body
-
-            async def __aiter__(self) -> typing.AsyncIterator[bytes]:
-                yield b"".join(self._body)
 
         stream = ASGIResponseStream(body_parts)
         extensions = {}
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -3,7 +3,7 @@ from urllib.parse import unquote
 
 import sniffio
 
-from .base import AsyncBaseTransport
+from .base import AsyncBaseTransport, AsyncByteStream
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     import asyncio
@@ -74,10 +74,10 @@ class ASGITransport(AsyncBaseTransport):
         method: bytes,
         url: typing.Tuple[bytes, bytes, typing.Optional[int], bytes],
         headers: typing.List[typing.Tuple[bytes, bytes]],
-        stream: typing.AsyncIterable[bytes],
+        stream: AsyncByteStream,
         extensions: dict,
     ) -> typing.Tuple[
-        int, typing.List[typing.Tuple[bytes, bytes]], typing.AsyncIterable[bytes], dict
+        int, typing.List[typing.Tuple[bytes, bytes]], AsyncByteStream, dict
     ]:
         # ASGI scope.
         scheme, host, port, full_path = url
@@ -155,9 +155,14 @@ class ASGITransport(AsyncBaseTransport):
         assert status_code is not None
         assert response_headers is not None
 
-        async def response_stream() -> typing.AsyncIterator[bytes]:
-            yield b"".join(body_parts)
+        class ASGIResponseStream(AsyncByteStream):
+            def __init__(self, body: typing.List[bytes]) -> None:
+                self._body = body
 
+            async def __aiter__(self) -> typing.AsyncIterator[bytes]:
+                yield b"".join(self._body)
+
+        stream = ASGIResponseStream(body_parts)
         extensions = {}
 
-        return (status_code, response_headers, response_stream(), extensions)
+        return (status_code, response_headers, stream, extensions)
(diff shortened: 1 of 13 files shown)


tomchristie:
NB.This change also allows us to trivially drop the kinda-implementation-detail `Response(..., on_close=...)` argument, in favour of just calling `stream.close()`, but I've omitted that on this pass. Because, ya'know.

httpx/_transports/asgi.py
```diff
@@ -155,9 +155,14 @@ async def send(message: dict) -> None:
         assert status_code is not None
         assert response_headers is not None
 
-        async def response_stream() -> typing.AsyncIterator[bytes]:
-            yield b"".join(body_parts)
+        class ASGIResponseStream(AsyncByteStream):
```
florimondmanca, line 158: Thoughts on moving this class definition at the module level? Doesn't seem like it encloses any data from this function call (that wouldn't be ideal anyway).

httpx/_transports/base.py
```diff
@@ -5,6 +5,63 @@
...
+            "The '__iter__' method must be implemented."
+        )  # pragma: nocover
+        yield b""  # pragma: nocover
+
+    def close(self) -> None:
+        """
+        Subclasses can override this method to release any network resources
+        after a request/response cycle is complete.
+
+        Streaming cases should call use a `try...finally` block to ensure that
```
florimondmanca, line 20:
```suggestion
        Streaming cases should use a `try...finally` block to ensure that
```

httpx/_transports/default.py
```diff
@@ -182,19 +182,23 @@ def handle_request(
                 ext=extensions,
             )
 
-        def response_stream() -> typing.Iterator[bytes]:
-            with map_httpcore_exceptions():
-                for part in byte_stream:
-                    yield part
+        class ResponseStream(SyncByteStream):
```
florimondmanca, line 185: Same here, how about moving this to the module level?

httpx/_transports/default.py
```diff
@@ -272,19 +276,23 @@ async def handle_async_request(
...
+                with map_httpcore_exceptions():
+                    async for part in self._httpcore_stream:
+                        yield part
 
-        async def aclose() -> None:
-            with map_httpcore_exceptions():
-                await byte_stream.aclose()
+            async def aclose(self) -> None:
+                with map_httpcore_exceptions():
+                    await byte_stream.aclose()
```
florimondmanca, line 290:
```suggestion
                    await self._httpcore_stream.aclose()
```

httpx/_transports/wsgi.py
```diff
@@ -111,9 +111,16 @@ def start_response(
             seen_exc_info = exc_info
 
         result = self.app(environ, start_response)
-        # This is needed because the status returned by start_response
-        # shouldn't be used until the first non-empty chunk has been served.
-        result = _skip_leading_empty_chunks(result)
+
+        class WSGIByteStream(SyncByteStream):
```
florimondmanca, line 115: To module level?

florimondmanca (commented):
So, I take this as essentially moving from `Iterable`/`AsyncIterable` which are somewhat too low-level / not super user-friendly, to well-defined HTTPX interfaces. I think I quite like that. At least no strong "no", and I agree the resulting close-handling is more intuitive as a result.

tomchristie:
@florimondmanca That's it, yup. Just a nicer more usable interface at the Transport API level.

We don't expect the vast majority of our users to be working with that directly, but it's probably nicer for cases where folks *are* really digging in.

It would also *potentially* allow us to address [Seth's comment here](https://github.com/urllib3/urllib3/issues/2073#issuecomment-730113761) at a future date...

> I'd prefer a file-like API for response body for better control of the data that is pulled.

...because we could also add support for `read(max_bytes=...)` if we chose to.

(Although there's also also reasons we *might* not really want to go that route, which I'm not going to go into, unless/until that convo actually gets raised as a prospect.)

tomchristie:
Righty, let's push on with this...

johtso:
Looks good, is this basically the same as the previous functionality, but pulled into httpx? Wasn't over the moon about the close function in `extensions`, so this seems better.

tomchristie:
It does, doesn't it?

PR #1554, opened 2021-04-04
added docs for startup/shutdown of ASGI apps

### LINKED ISSUE #1441: Document solution for startup/shutdown of ASGI apps
Prompted by https://github.com/encode/httpx/issues/350#issuecomment-752639213

As discussed in #350, HTTPX doesn't provide a solution for managing the startup/shutdown lifecycle of an ASGI app. That is left to user code and not planned to change soon.

However we could/should consider pointing users at [`asgi-lifespan`](https://pypi.org/project/asgi-lifespan/) in our docs so that we're pointing at a full solution.

https://www.python-httpx.org/async/#calling-into-python-web-apps

jkklapp:
Closes #1441

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -243,38 +243,3 @@ See [the ASGI documentation](https://asgi.readthedocs.io/en/latest/specs/www.htm
 It is not in the scope of HTTPX to trigger lifespan events of your app.
 
 However it is suggested to use `LifespanManager` from [asgi-lifespan](https://github.com/florimondmanca/asgi-lifespan#usage) in pair with `AsyncClient`.
-
-```python
-# example.py
-from asgi_lifespan import LifespanManager
-from starlette.applications import Starlette
-
-# Example lifespan-capable ASGI app. Any ASGI app that supports
-# the lifespan protocol will do, e.g. FastAPI, Quart, Responder, ...
-app = Starlette(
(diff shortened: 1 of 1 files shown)


docs/async.md
```diff
@@ -237,3 +237,44 @@ async with httpx.AsyncClient(transport=transport, base_url="http://testserver")
 ```
 
 See [the ASGI documentation](https://asgi.readthedocs.io/en/latest/specs/www.html#connection-scope) for more details on the `client` and `root_path` keys.
+
+## Startup/shutdown of ASGI apps
+
+It is not in the scope of HTTPX to trigger lifespan events of your app.
+
+However it is suggested to use `LifespanManager` from [asgi-lifespan](https://github.com/florimondmanca/asgi-lifespan#usage) in pair with `AsyncClient`.
```
florimondmanca, line 245: Given that the examples below don't include any extra asgi-lifespan/HTTPX integration hints, should we drop everything below this line I'm commenting on? So as to avoid duplicate information. People can just click on the link and learn what `asgi-lifespan` is about and how to use it. There's even an HTTPX example in the README there. :)

jkklapp: SGTM 👍

florimondmanca (approved):
Thanks!

PR #1557, opened 2021-04-05
add new variable to default unset params

### LINKED ISSUE #1384: Reconsider UNSET

jkklapp:
Closes #1384

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -10,7 +10,7 @@ from ._config import (
     DEFAULT_LIMITS,
     DEFAULT_MAX_REDIRECTS,
     DEFAULT_TIMEOUT_CONFIG,
-    UNSET,
+    USE_CLIENT_DEFAULT,
     Limits,
     Proxy,
     Timeout,
@@ -251,9 +251,9 @@ class BaseClient:
         params: QueryParamTypes = None,
         headers: HeaderTypes = None,
         cookies: CookieTypes = None,
-        auth: typing.Union[AuthTypes, UnsetType] = UNSET,
+        auth: typing.Union[AuthTypes, UnsetType] = USE_CLIENT_DEFAULT,
         allow_redirects: bool = True,
-        timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
+        timeout: typing.Union[TimeoutTypes, UnsetType] = USE_CLIENT_DEFAULT,
     ) -> "StreamContextManager":
         """
         Alternative to `httpx.request()` that streams the response body
@@ -396,7 +396,9 @@ class BaseClient:
             raise TypeError(f'Invalid "auth" argument: {auth!r}')
 
     def _build_request_auth(
-        self, request: Request, auth: typing.Union[AuthTypes, UnsetType] = UNSET
+        self,
+        request: Request,
+        auth: typing.Union[AuthTypes, UnsetType] = USE_CLIENT_DEFAULT,
     ) -> Auth:
         auth = self._auth if isinstance(auth, UnsetType) else self._build_auth(auth)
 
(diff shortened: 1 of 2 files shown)


felix-hilden:
Nice! Two thoughts based on the original issue:
- perhaps the new default should be imported all the way up to the top level for it to be considered public
- as this parameter is to replace `UNSET`, maybe the `kwargs` approach could also be introduced to the remaining cases and `UNSET` be removed altogether (assuming you'd like to deal with the whole issue at once, otherwise "closes 1384" doesn't entirely apply)

I see that beyond a couple of thumbs up, an explicit consensus was not necessarily reached in the original issue. But I agree that this is a sensible way of doing it. What do the maintainers think?

tomchristie (commented):
Okay, so the *one* remaining issue with this is the places in the `Timeout` class, where `USE_CLIENT_DEFAULT` is being used. In that context it doesn't actually make sense, because that's not what it's doing there.

For the purposes of this pull request I'd suggest we probably want to keep `UnsetType`/`UNSET`, and continue to use them for the `Timeout` cases without any change. (While also *adding* the `ClientDefaultType`/`USE_CLIENT_DEFAULT` as done here.)

httpx/_client.py
```diff
@@ -758,7 +766,7 @@ def request(
         params: QueryParamTypes = None,
         headers: HeaderTypes = None,
         cookies: CookieTypes = None,
-        auth: typing.Union[AuthTypes, UnsetType] = UNSET,
+        auth: typing.Union[AuthTypes, ClientDefaultType] = USE_CLIENT_DEFAULT,
         allow_redirects: bool = True,
         timeout: typing.Union[TimeoutTypes, UnsetType] = UNSET,
```
tomchristie, line 771: So, *these* cases in `_client.py` should be:

```python
    timeout: typing.Union[TimeoutTypes, ClientDefaultType] = USE_CLIENT_DEFAULT,
```

(But the signature of the arguments in the Timeout *class* should keep using `UNSET`)

tomchristie: Actually thinking about this, it's probably also a good idea to move `ClientDefaultType` and `USE_CLIENT_DEFAULT` *out* of `_types.py`, and into this module, to keep them close to where they're actually used.

jkklapp: > So, _these_ cases in `_client.py` should be:
> 
> ```python
>     timeout: typing.Union[TimeoutTypes, ClientDefaultType] = USE_CLIENT_DEFAULT,
> ```
> 
> (But the signature of the arguments in the Timeout _class_ should keep using `UNSET`)

If I do that I need to also change the the types in the `Timeout` class, otherwise `mypy` complains.

tomchristie: Okay, I'll take a look.

tomchristie:
Closed via #1634 - Thanks for taking a punt at this!

PR #1565, opened 2021-04-10
simplify tqdm example

balki:
Working Example: https://gist.github.com/balki/193af5b8f508ec5eefbded13b0c73634

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -293,11 +293,9 @@ with tempfile.NamedTemporaryFile() as download_file:
         total = int(response.headers["Content-Length"])
 
         with tqdm(total=total, unit_scale=True, unit_divisor=1024, unit="B") as progress:
-            num_bytes_downloaded = response.num_bytes_downloaded
             for chunk in response.iter_bytes():
                 download_file.write(chunk)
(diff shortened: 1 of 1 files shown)


florimondmanca:
Thanks, but this would be inaccurate in certain if not most cases.

Eg of the response is compressed then the sum of `chunk` would be greater than the number of bytes actually downloaded from the network.

This is why we have a dedicated attribute used here to reflect raw bytes before any kind of decoding.

balki:
Thanks! I didn't realize that. Leaving below helper if anyone needs to simplify:

```
async def get_chunks_with_progress(response: httpx.Response):
    total = int(response.headers["Content-Length"])
    num_bytes_downloaded = response.num_bytes_downloaded
    with tqdm.tqdm(total=total, unit_scale=True, unit_divisor=1024, unit='B') as progress:
        async for chunk in response.aiter_bytes():
            yield chunk
            progress.update(response.num_bytes_downloaded - num_bytes_downloaded)
            num_bytes_downloaded = response.num_bytes_downloaded
```

Example usage: https://gist.github.com/balki/193af5b8f508ec5eefbded13b0c73634

PR #1572, opened 2021-04-14
Drop 'Response(on_close=...)' from API

tomchristie:
Stop supporting `Response(on_close=...)` in the API, which really is a bit of niggly implementation detail that's being unnecessarily exposed.

I figure we can get away without any deprecation period on this one with a 0.18 bump, since it really has only every been intended as an implementation detail, rather than user-facing API.

Instead, once we've got a response instance we upgrade the stream on it to a bound stream, that closes the timer once it's elapsed.

Figure this is worth doing, since *API is king*, and we really don't want that public API close callback there if we can avoid it.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -86,6 +86,52 @@ class ClientState(enum.Enum):
     CLOSED = 3
 
 
+class BoundSyncStream(SyncByteStream):
+    """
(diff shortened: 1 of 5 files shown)


florimondmanca (approved):
Smart one about that "bound stream" idea. :-)

PR #1574, opened 2021-04-14
Deprecate per-request cookies

### LINKED ISSUE #1404: Cookies are not sent after redirect (even on same origin)
### Checklist

### Describe the bug

When a GET request with cookies specified is issued, and the server returns a redirect, the subsequent request is not supplied with the cookies.

### To reproduce

To observe the behavior, it's possible to use https://webhook.site/ to dump the request and https://bit.ly to provide a redirect. As the https://webhook.site/ URL is per session, you'll have to create both URLs yourself, but it only takes a minute. **This is also reproducible on same-origin redirects, not only different-origin as demonstrated here.** You can use a local server to check.

```python
DIRECT_URL="" # your webhook.site URL
REDIR_URL="" # your bit.ly URL

import httpx
import asyncio

async def main():
        await client.get(DIRECT_URL, cookies={ 'cookie': 'direct' })
        await client.get(REDIR_URL, cookies={ 'cookie': 'redir' })

asyncio.get_event_loop().run_until_complete(main())
```

### Expected behavior

One request should arrive with `cookie=direct` and the other with `cookie=redir`.

The correct behavior can be observed by using `requests`:

```python
DIRECT_URL="" # your webhook.site URL
REDIR_URL="" # your bit.ly URL

import requests

requests.get(DIRECT_URL, cookies={ 'cookie': 'direct' })
requests.get(REDIR_URL, cookies={ 'cookie': 'redir' })
```

### Actual behavior

One request arrives with `cookie=direct` and the other with no such cookie.

### Environment

- OS: Windows 10 (2004 update)
- Python version: 3.9.0
- HTTPX version: 0.16.1 (reproducible on both the lastest release and `master` branch)
- Async environment: reproducible on both `asyncio` / `trio`
- HTTP proxy: no
- Custom certificates: no

tomchristie:
Closes #1404

Documentation here is as follows...

---

## Cookies

If using a client instance, then cookies should always be set on the client rather than on a per-request basis.

This usage is supported...

```python
client = httpx.Client(cookies=...)
client.post(...)
```

This usage is deprecated...

```python
client = httpx.Client()
client.post(..., cookies=...)
```

We prefer enforcing a stricter API here because it provides clearer expectations around cookie persistence, particularly when redirects occur.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -37,6 +37,26 @@ HTTPX uses `utf-8` for encoding `str` request bodies. For example, when using `c
 
 For response bodies, assuming the server didn't send an explicit encoding then HTTPX will do its best to figure out an appropriate encoding. Unlike Requests which uses the `chardet` library, HTTPX relies on a plainer fallback strategy (basically attempting UTF-8, or using Windows-1252 as a fallback). This strategy should be robust enough to handle the vast majority of use cases.
 
+## Cookies
+
+If using a client instance, then cookies should always be set on the client rather than on a per-request basis.
+
+This usage is supported...
+
+```python
+client = httpx.Client(cookies=...)
+client.post(...)
+```
+
+This usage is deprecated...
+
+```python
+client = httpx.Client()
+client.post(..., cookies=...)
+```
+
+We prefer enforcing a stricter API here because it provides clearer expectations around cookie persistence, particularly when redirects occur.
+
 ## Status Codes
 
 In our documentation we prefer the uppercased versions, such as `codes.NOT_FOUND`, but also provide lower-cased versions for API compatibility with `requests`.
@@ -147,6 +167,6 @@ while request is not None:
 
 `requests` allows event hooks to mutate `Request` and `Response` objects. See [examples](https://requests.readthedocs.io/en/master/user/advanced/#event-hooks) given in the documentation for `requests`.
 
-In HTTPX, event hooks may access properties of requests and responses, but event hook callbacks cannot mutate the original request/response. 
+In HTTPX, event hooks may access properties of requests and responses, but event hook callbacks cannot mutate the original request/response.
 
 If you are looking for more control, consider checking out [Custom Transports](advanced.md#custom-transports).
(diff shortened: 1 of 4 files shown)


florimondmanca (approved):

httpx/_client.py
```diff
@@ -731,6 +731,14 @@ def request(
 
         [0]: /advanced/#merging-of-configuration
         """
+        if cookies is not None:
+            message = (
+                "Setting per-request cookies=... is being deprecated, because "
```
StephenBrown2, line 736: This would be better as
```suggestion
                "Setting per-request cookies=<...> is being deprecated, because "
```
for better language flow, I think (and similar to the `content=<...>` message in https://github.com/encode/httpx/pull/1573/files#diff-361c5daf6e5401b987414c619d003a20ea88f74fe9da50cbb2da72c337aeb290R152)

docs/compatibility.md
```diff
@@ -37,6 +37,26 @@ HTTPX uses `utf-8` for encoding `str` request bodies. For example, when using `c
...
+If using a client instance, then cookies should always be set on the client rather than on a per-request basis.
+
+This usage is supported...
+
+```python
+client = httpx.Client(cookies=...)
+client.post(...)
+```
+
+This usage is deprecated...
```
StephenBrown2, line 51: Colons rather than ellipses here perhaps?
```suggestion
This usage is supported:

```python
client = httpx.Client(cookies=...)
client.post(...)
```

This usage is deprecated:
```

florimondmanca: Ellipses are Tom's style. :-)

StephenBrown2: Good enough for me, then!

tomchristie: Oh, I don't know, I think it's probably a good suggestion. 😃

rafalkrupinski:
I would really hate loosing per-request cookies. Cookies are just a header, why the Cookies facility class should be disabled when doing redirects?

Wouldn't it be reasonable to expect request cookies to follow the redirects, either by default or with a flag?

PR #1575, opened 2021-04-15
Update `verify` parameter description

ascopes:
Added to documentation to show that `verify` can be an `ssl.SSLContext`, as this is not immediately obvious from the developer interface documentation due to the lack of type hints on the signature that is rendered.

image


StephenBrown2:
Seems to be a failure of the doc generator, as `VerifyTypes` is defined as including those: https://github.com/encode/httpx/blob/110ce8565259a3c6927352554eccd94ca5804463/httpx/_types.py#L62 and usage is documented in the Advanced section of the docs: https://www.python-httpx.org/advanced/#changing-the-verification-defaults

PR #1577, opened 2021-04-16
Drop StreamContextManager in favour of `contextlib.contextmanager`/`asynccontextmanager`

tomchristie:
Possible refactoring that is worth our consideration.

Somewhat related to #998, since we also use `contextlib.asynccontextmanager` there, so both that pull request and this one lean on the `async_generator` package for 3.6 backport support.


florimondmanca (approved):
Yes, definitely. :-)

PR #1578, opened 2021-04-16
Drop Response.call_next

hannseman:
Had a look at the attributes of `Response` while thinking about https://github.com/encode/httpx/discussions/1562. Looks like `call_next` is not used for anything.


florimondmanca (approved):
Indeed, seems like a left-over from https://github.com/encode/httpx/pull/1339 to me. The logic behind `call_next` was replaced there. No sign that we'd have been keeping `call_next` for any backward compat reasons either.

Thanks!

hannseman:
Thanks for the review!

PR #1579, opened 2021-04-16
Make Request and Response picklable

hannseman:
Refs: https://github.com/encode/httpx/discussions/1562

I took the liberty of creating a PR as I felt that there was a consensus about implementing this.

Some open questions is what state the instances should end up in after being loaded. I.e should we override `is_closed`, `is_stream_consumed` and if so to what? I also think we need to set `stream` to something to avoid unexpected `AttributeError` being raised on access. Currently set it to an empty `ByteStream` to make mypy happy. 

It might also be cleaner to set the overrides in the dict produced in `__getstate__` instead of directly in `__setstate__`.  

An alternative to the `__attrs__` pattern would be to pop invalid attributes from `__dict__` in `__getstate__` but being explicit about it feels safer.


tomchristie:
Okay, so thoughts here...

Firstly, I don't really like the `__attrs__` thing. Although it looks like a Python built-in, it turns out that's just a bit of requests internal naming that they decided on.

I think we probably want to approach this like so:

```python
class UnattachedStream(AsyncByteStream, SyncByteStream):
    """
    If a request or response is serialized using pickle, then it is no longer attached to a
    stream for I/O purposes. Any stream operations should result in `httpx.StreamClosed`.
    """

    def __iter__(self) -> Iterator[bytes]:
        raise StreamClosed()

    async def __aiter__(self) -> AsyncIterator[bytes]:
        raise StreamClosed()
```

```python
class Response:
    ...

    def __getstate__(self):
        return {
            name: value
            for name, value in self.__dict__.items()
            if name not in ['stream', 'is_closed', '_decoder']
        }

    def __setstate__(self, state):
        for name, value in state.items():
            setattr(self, name, value)
        self.is_closed = True
        self.stream = UnattachedStream()
```

```python
class Request:
    ...

    def __getstate__(self):
        return {
            name: value
            for name, value in self.__dict__.items()
            if name not in ['stream']
        }

    def __setstate__(self, state):
        for name, value in state.items():
            setattr(self, name, value)
        self.stream = UnattachedStream()
```

I'd consider https://github.com/encode/httpx/pull/1584 to be a pre-requisite.

hannseman:
@tomchristie thanks for the review. I agree that the `__attrs__` thing felt a bit awkward, much better 👍

httpx/_content.py
```diff
@@ -61,6 +61,21 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
+    attached to a stream for I/O purposes. Any stream operations should result
+    in `httpx.StreamClosed`.
+    """
+
+    def __iter__(self) -> Iterator[bytes]:
+        raise ResponseClosed()  # TODO: StreamClosed
+
+    async def __aiter__(self) -> AsyncIterator[bytes]:
+        raise ResponseClosed()  # TODO: StreamClosed
+        yield b""  # pragma: nocover
```
hannseman, line 76: We need the `yield` here to avoid `TypeError`:

```
    async def aread(self) -> bytes:
        """
        Read and return the request content.
        """
        if not hasattr(self, "_content"):
            assert isinstance(self.stream, typing.AsyncIterable)
>           self._content = b"".join([part async for part in self.stream])
E           TypeError: 'async for' received an object from __aiter__ that does not implement __anext__: coroutine
```

tests/models/test_responses.py
```diff
@@ -853,3 +854,40 @@ def content():
...
+
+@pytest.mark.asyncio
+async def test_response_async_streaming_picklable():
+    response = httpx.Response(200, content=async_streaming_body())
+    pickle_response = pickle.loads(pickle.dumps(response))
+    assert hasattr(pickle_response, "_content") is False
+    with pytest.raises(httpx.ResponseClosed):  # TODO: StreamClosed
+        await pickle_response.aread()
+    assert pickle_response.is_stream_consumed is False
+    assert pickle_response._num_bytes_downloaded == 0
```
tomchristie, line 885: Let's aim to test against public API wherever possible.

Eg. in this case let's access the property instead, so... `assert pickle_response.num_bytes_downloaded == 0`

tests/models/test_responses.py
```diff
@@ -853,3 +854,40 @@ def content():
...
+    assert pickle_response.request.url == response.request.url
+    assert pickle_response.extensions == {}
+    assert pickle_response.history == []
+
+
+@pytest.mark.asyncio
+async def test_response_async_streaming_picklable():
+    response = httpx.Response(200, content=async_streaming_body())
+    pickle_response = pickle.loads(pickle.dumps(response))
+    assert hasattr(pickle_response, "_content") is False
```
tomchristie, line 881: I guess testing against public API we probably want something like this instead...

```python
with pytest.raises(httpx.ResponseNotRead):
    response.content
```

tests/models/test_responses.py
hannseman: Makes sense! Thanks

httpx/_content.py
```diff
@@ -61,6 +61,21 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
 
+class UnattachedStream(AsyncByteStream, SyncByteStream):
+    """
+    If a request or response is serialized using pickle, then it is no longer
+    attached to a stream for I/O purposes. Any stream operations should result
+    in `httpx.StreamClosed`.
+    """
+
+    def __iter__(self) -> Iterator[bytes]:
+        raise ResponseClosed()  # TODO: StreamClosed
```
tomchristie, line 72:
```suggestion
        raise StreamClosed()
```

httpx/_content.py
```diff
@@ -61,6 +61,21 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
+
+    async def __aiter__(self) -> AsyncIterator[bytes]:
+        raise ResponseClosed()  # TODO: StreamClosed
```
tomchristie, line 75:
```suggestion
        raise StreamClosed()
```

httpx/_content.py
```diff
@@ -13,7 +13,7 @@
 )
 from urllib.parse import urlencode
 
-from ._exceptions import StreamConsumed
+from ._exceptions import ResponseClosed, StreamConsumed
```
tomchristie, line 16:
```suggestion
from ._exceptions import StreamClosed, StreamConsumed
```

tests/models/test_responses.py
```diff
@@ -853,3 +854,41 @@ def content():
...
+    assert pickle_response.history == []
+
+
+@pytest.mark.asyncio
+async def test_response_async_streaming_picklable():
+    response = httpx.Response(200, content=async_streaming_body())
+    pickle_response = pickle.loads(pickle.dumps(response))
+    with pytest.raises(httpx.ResponseNotRead):
+        pickle_response.content
+    with pytest.raises(httpx.ResponseClosed):  # TODO: StreamClosed
```
tomchristie, line 883:
```suggestion
    with pytest.raises(httpx. StreamClosed):
```

tests/models/test_requests.py
```diff
@@ -174,3 +175,54 @@ def test_url():
...
+
+def test_request_generator_content_picklable():
+    def content():
+        yield b"test 123"  # pragma: nocover
+
+    request = httpx.Request("POST", "http://example.org", content=content())
+    pickle_request = pickle.loads(pickle.dumps(request))
+    with pytest.raises(httpx.RequestNotRead):
+        pickle_request.content
+    with pytest.raises(httpx.ResponseClosed):  # TODO: StreamClosed
```
tomchristie, line 222:
```suggestion
    with pytest.raises(https. StreamClosed):
```

tests/models/test_requests.py
```diff
@@ -174,3 +175,54 @@ def test_url():
...
+async def test_request_async_streaming_content_picklable():
+    async def streaming_body(data):
+        yield data
+
+    data = streaming_body(b"test 123")
+    request = httpx.Request("POST", "http://example.org", content=data)
+    pickle_request = pickle.loads(pickle.dumps(request))
+    with pytest.raises(httpx.RequestNotRead):
+        pickle_request.content
+    with pytest.raises(httpx.ResponseClosed):  # TODO: StreamClosed
```
tomchristie, line 205:
```suggestion
    with pytest.raises(httpx. StreamClosed):
```

tests/models/test_requests.py
```diff
@@ -174,3 +175,54 @@ def test_url():
...
+
+def test_request_generator_content_picklable():
+    def content():
+        yield b"test 123"  # pragma: nocover
+
+    request = httpx.Request("POST", "http://example.org", content=content())
+    pickle_request = pickle.loads(pickle.dumps(request))
+    with pytest.raises(httpx.RequestNotRead):
+        pickle_request.content
+    with pytest.raises(https.StreamClosed):
```
tomchristie, line 222:
```suggestion
    with pytest.raises(httpx.StreamClosed):
```

tomchristie:
Right, let's go with this.

Great work @hannseman!

hannseman:
@tomchristie thanks a lot for the review and fixups! 🎉

PR #1583, opened 2021-04-19
For non-streaming cases, populate `request.content` automatically.

tomchristie:
Minor user-experience & consistency tweak.

For non-streaming cases we should pre-load `request.content`.

This leads to neater consistency between requests/responses, and ensures that the request content is readily accessible to introspection.

```pycon
>>> r = httpx.Request("POST", "https://www.example.com", content=b'...')
>>> r.content  # Previously this would raise 'httpx.RequestNotRead'
b'...'
```

```pycon
>>> r = httpx.Request("POST", "https://www.example.com", json={"example": "payload"})
>>> r.content  # Previously this would raise 'httpx.RequestNotRead'
b'{"example": "payload"}'
```

```pycon
>>> r = httpx.Request("POST", "https://www.example.com", content=[b'.' for i in range(3)])
>>> r.content
  File "/Users/tomchristie/GitHub/encode/httpx/httpx/_models.py", line 865, in content
    raise RequestNotRead()
httpx.RequestNotRead: Attempted to access streaming request content, without having called `read()`.
```

Or, eg...

**Previously**...

```pycon
>>> import httpx
>>> r = httpx.get("https://www.example.com")
>>> r.request.content
  File "/Users/tomchristie/GitHub/encode/httpx/httpx/_models.py", line 862, in content
    raise RequestNotRead()
httpx.RequestNotRead: Attempted to access request content, without having called `read()`  # Eh?
````

**Now**...

```pycon
>>> import httpx
>>> r = httpx.get("https://www.example.com")
>>> r.request.content
b''
```

### DIFF
--- a/httpx/_exceptions.py
+++ b/httpx/_exceptions.py
@@ -294,25 +294,23 @@ class StreamConsumed(StreamError):
 
 class ResponseNotRead(StreamError):
     """
-    Attempted to access response content, without having called `read()`
-    after a streaming response.
+    Attempted to access streaming response content, without having called `read()`.
     """
 
     def __init__(self) -> None:
         message = (
-            "Attempted to access response content, without having called `read()` "
-            "after a streaming response."
+            "Attempted to access streaming response content, without having called `read()`."
         )
         super().__init__(message)
 
 
 class RequestNotRead(StreamError):
     """
-    Attempted to access request content, without having called `read()`.
+    Attempted to access streaming request content, without having called `read()`.
     """
 
     def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


tomchristie:
Going to push on with this sans-review, because it'll otherwise block other bits of work that I'd like to push on with here.

lundberg:
Much nicer 👍 

Not sure, but maybe the `httpx.RequestNotRead` and `httpx.ResponseNotRead` should be renamed to align with #1584, e.g. turn both into a `httpx.StreamNotRead`?

tomchristie:
Interesting yup. Hierarchy feels nicer that way for sure.

PR #1590, opened 2021-04-22
Differentiate between 'url.host' and 'url.raw_host'

tomchristie:
Refs #1275

Throughout our URL model we're differentiating neatly between byte-wise cases and str cases.
We're always using bytes when escaping is not applied, and str when escaping *is* applied.

Eg...

```python
url = httpx.URL("https://jo%40email.com:a%20secret@example.com:1234/pa th")
assert url.username == "jo@email.com"
assert url.password == "a secret"
assert url.userinfo == b"jo%40email.com:a%20secret"
assert url.path == "/pa th"
assert url.raw_path == b"/pa%20th"
```

This pull request is a proposal for treating IDNA domain names similarly, so...

```python
url = httpx.URL("https://müller.de:80")
assert url.host == "müller.de"
assert url.raw_host == b"xn--mller-kva.de"
```

For API consistency this also necessarily results in `url.netloc` becoming a byte interface, which actually makes sense for the contexts in which it is used.

```python
url = httpx.URL("https://müller.de:80")
assert url.netloc == b"xn--mller-kva.de:80"
```

Finally we also introduce `.raw_scheme` for a byte-wise representation of the scheme, for a nice consistency so that:

```python
url = httpx.URL("https://müller.de:80/pa th")
assert url.raw == (url.raw_scheme, url.raw_host, url.port, url.raw_path)
assert url.raw == (b"https", b"xn--mller-kva.de", 80, b"/pa%20th")
```

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -68,7 +68,7 @@ class URL:
     assert url.password == "a secret"
     assert url.userinfo == b"jo%40email.com:a%20secret"
     assert url.host == "müller.de"
-    assert url.raw_host == b"xn--mller-kva.dem"
+    assert url.raw_host == b"xn--mller-kva.de"
     assert url.port == 1234
     assert url.netloc == b"xn--mller-kva.dem:1234"
     assert url.path == "/pa th"
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -70,7 +70,7 @@ class URL:
     assert url.host == "müller.de"
     assert url.raw_host == b"xn--mller-kva.de"
     assert url.port == 1234
-    assert url.netloc == b"xn--mller-kva.dem:1234"
+    assert url.netloc == b"xn--mller-kva.de:1234"
     assert url.path == "/pa th"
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Wouldn't `raw` indicate _unencoded_?

StephenBrown2 (approved):
Just a couple typos, and the previous question about what `raw` means.

tomchristie:
Raw, as in the "the raw bytes on the wire", or "the raw ingredients that make up the cake".
The raw representation of the host is the actual unaltered bytewise representation that's used to make the connection.

Or, in baking...

The raw ingredients: `\xf0\x9f\x8e\x82`
The cake: `🎂`

tomchristie:
Thanks so much for the review @StephenBrown2.
(Geez, me & my typos. 😬)

StephenBrown2:
> (Geez, me & my typos. 😬)

I think it was mainly just a copy-paste issue that got propagated, but no more `normlized`! :-p

PR #1591, opened 2021-04-23
App support

tomchristie:
This draft pull request demonstrates how we'd potentially be able to use `httpx` for server-side development.

For example:

```python
import httpx


def homepage(request):
    return httpx.Response(200, json={'hello': 'world'})

routes = [
    httpx.Route("/", endpoint=homepage),
]

app = httpx.WSGIApp(routes=routes)
```

Running the app using Gunicorn:

```shell
$ gunicorn example:app
[2021-04-23 10:23:32 +0100] [30102] [INFO] Starting gunicorn 20.1.0
[2021-04-23 10:23:32 +0100] [30102] [INFO] Listening at: http://127.0.0.1:8000 (30102)
[2021-04-23 10:23:32 +0100] [30102] [INFO] Using worker: sync
[2021-04-23 10:23:32 +0100] [30105] [INFO] Booting worker with pid: 30105
```

In the pull request I've also updated two of the test case modules to use routes and routers, rather than a single handler function.

* `tests/client/test_event_hooks.py`
* `tests/client/test_redirects.py`

I'm not pushing for us to *necessarily* go down this road, but it's an interesting starting point, and it's worth us keeping this in mind. Am glossing over a few details for now, but with relatively little effort we'd be able to support using httpx as the basis for server-side apps, with both WSGI and ASGI support.

Beware scope-creep, of course. But also, *...interesting*.

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -1,5 +1,7 @@
 from .__version__ import __description__, __title__, __version__
 from ._api import delete, get, head, options, patch, post, put, request, stream
+from ._apps.routing import Route, Router
+from ._apps.wsgi import WSGIApp
 from ._auth import Auth, BasicAuth, DigestAuth
 from ._client import AsyncClient, Client
 from ._config import Limits, Proxy, Timeout, create_ssl_context
@@ -100,6 +102,8 @@ __all__ = [
     "RequestNotRead",
     "Response",
     "ResponseNotRead",
+    "Route",
+    "Router",
     "StatusCode",
     "stream",
     "StreamClosed",
@@ -114,6 +118,7 @@ __all__ = [
     "URL",
     "WriteError",
     "WriteTimeout",
+    "WSGIApp",
     "WSGITransport",
 ]
 
--- a/httpx/_apps/routing.py
+++ b/httpx/_apps/routing.py
@@ -1,0 +1,111 @@
+import re
+import typing
+
+from httpx._exceptions import UnsupportedProtocol
+from httpx._models import Request, Response
+from httpx._transports.base import (
+    AsyncBaseTransport,
+    AsyncByteStream,
+    BaseTransport,
+    SyncByteStream,
+)
+
+PARAM_REGEX = re.compile("{([a-zA-Z_][a-zA-Z0-9_]*)}")
+
+
+class Route:
+    def __init__(self, path: str, endpoint: typing.Callable) -> None:
+        self.path = path
+        self.endpoint = endpoint
+        self.regex = self._compile_regex(path)
+
+    def _compile_regex(self, path: str) -> re.Pattern:
+        regex = "^"
+        idx = 0
+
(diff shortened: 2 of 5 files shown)


StephenBrown2:
I'm pretty "meh" to 👎🏻 on this. My personal opinion is that a client should remain a client, and not try to also be a server, but I think [aiohttp](https://docs.aiohttp.org/en/stable/web_quickstart.html) might prove me wrong, and I've seen lots of use of that, which httpx could handle with this.

Other's thoughts?

florimondmanca:
I think it's definitely _interesting_, but there's so much that existing specialized server side frameworks handle that HTTPX would need to handle too to be of any practical use, aside from toy/mocking. I don't know how we'd present something like this not to set unrealistic expectations.

Kind of balancing on this one.

At this point I'd say "interesting but probably not"?

tomchristie:
Sure thing. The important point from my POV is making sure that the *models* are suitable for being used in either context.

So, (for example) being in a position where a future version of Starlette or FastAPI (or whatever else) could be leaning on HTTPX for `Request`, `Response`, `QueryParams`, `Headers`, and `Cookies` models. I guess if we (or anyone) ever did explore building up apps on top of the models we're using we'd likely want that to be in as a separate package.

There's a few cases where there's some careful attention to detail that's been made in the design of requests/responses that helps make that possible, where it wouldn't be quite so simple with `requests`.

Also, I get the comparison to aiohttp on the surface, but that's a tightly integrated server+app framework+client, which doesn't really compare to what we've got here. In particular the strict separation of concerns between the `httpx` client, and the `httpcore` networking layer. Also the bits of app support demo'ed in this PR as being plain ol' ASGI/WSGI apps, without any kind of integrated server.

StephenBrown2:
Mocking is a really interesting and useful case, actually. It might preclude the use of respx and pytest-httpx, or really enhance/simplify their internal code.

lundberg:
> Mocking is a really interesting and useful case, actually. It might preclude the use of respx and pytest-httpx, or really enhance/simplify their internal code.

Well, RESPX is implemented as a `Router` for this exact purpose @StephenBrown2  😉.

And about mocking, it's *optional* to patch/mock out HTTPX when using it.

Using plain router without patching HTTPX:
```py
router = respx.Router()
router.get("https://example.org/", params={"foo": "bar"}).respond(204, ...)
router.post("https://example.org/", ...).respond(201, ...)
router.route(host__in=["foo.bar", "ham.spam"]).respond(...)

transport = httpx.MockTransport(router.handler)
```

So for test cases, I'd say it's important with [flexibility route patterns](https://lundberg.github.io/respx/guide/#routing-requests) and powerful [response sideeffects/endpoints](https://lundberg.github.io/respx/guide/#mocking-responses) to not clutter the test code and be repetitive.

I like the api of instatiating a router with pre-defined list of routes, and also the term/name *endpoint*...might borrow those to the RESPX api @tomchristie 😉

PR #1597, opened 2021-04-26
Escalate 0.17 deprecation warnings to becoming fully deprecated.

tomchristie:
The following API changes have been issuing deprecation warnings since 0.17.0 onwards, and are now fully deprecated...

* You should now use `httpx.codes` consistently instead of `httpx.StatusCodes`.
* Use `limits=...` instead of `pool_limits=...`.
* Use `proxies={"http://": ...}` instead of `proxies={"http": ...}` for scheme-specific mounting.


florimondmanca (approved):
Yes :-)

PR #1600, opened 2021-04-26
Immutable QueryParams

tomchristie:
As outlined in #1599.
Refs #1275.

Drop support for `QueryParams.__setitem__` and `QueryParams.update()` in favour of an immutable-only interface, in line with our `URL` design. All manipulation methods now return new QueryParams instances:

```python
q = httpx.QueryParams()

q = q.set("a", "123")
assert q == httpx.QueryParams("a=123")

q = q.add("a", "456")
assert q == httpx.QueryParams("a=123&a=456")

q = q.remove("a")
assert q == httpx.QueryParams("")

q = q.merge({"b": "789"})
assert q == httpx.QueryParams("b=789")
```

I've opted for a hard deprecation here, with `.update()` raising a RuntimeError, because immutability vs. mutability isn't something we ought to go "we're halfway there" on.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -386,7 +386,7 @@ class BaseClient:
         """
         if params or self.params:
             merged_queryparams = QueryParams(self.params)
-            merged_queryparams.update(params)
+            merged_queryparams = merged_queryparams.merge(params)
             return merged_queryparams
         return params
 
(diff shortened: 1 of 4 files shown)


tomchristie:
I'm pushing on with this because it's something I've already been considering for a long time, and I've got other stuff that it's going to block unless I just push on with it, so...

PR #1603, opened 2021-04-27
Perform port normalization for `http`, `https`, `ws`, `wss`, and `ftp` schemes.

tomchristie:
Normalize the port component of URLs, so that...

```python
assert httpx.URL("http://www.example.com") == httpx.URL("http://www.example.com:80")
print(httpx.URL("http://www.example.com:80"))  # URL('http://www.example.com')
```

Following this part of the WHATWG spec...

> If url’s port is url’s scheme’s default port, then set url’s port to null.

And using the set of default ports as defined by https://url.spec.whatwg.org/#url-miscellaneous

I figure this probably helps iron away a few possible corner cases. It more neatly matches browser behaviour, and it's in line with other aspects of normalization that we're providing for, eg...

```python
assert httpx.URL("http://WWW.EXAMPLE.COM") == httpx.URL("http://www.example.com")
```

It also has the nice property that `url.netloc` now properly encapsulates the same value that ought to be used for the value of the `Host` header.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -100,6 +100,13 @@ class URL:
       url = httpx.URL("http://xn--fiqs8s.icom.museum")
       assert url.raw_host == b"xn--fiqs8s.icom.museum"
 
+    * `url.port` is either None or an integer. URLs that include the default port for
+      "http", "https", "ws", "wss", and "ftp" schemes have their port normalized to `None`.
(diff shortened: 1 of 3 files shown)


florimondmanca (approved):
I'm happy enough with this. Thanks for the useful reference!

PR #1605, opened 2021-04-28
Update brotli support to use the brotlicffi package

tomchristie:
The `brotlipy` package has been superseded by the `brotlicffi` package.

See #1604

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ The HTTPX project relies on these excellent libraries:
   * `idna` - Internationalized domain name support.
 * `sniffio` - Async library autodetection.
 * `async_generator` - Backport support for `contextlib.asynccontextmanager`. *(Only required for Python 3.6)*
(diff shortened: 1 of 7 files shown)


j178 (approved):

GeekDuanLian:
Hey, please follow the instructions in the brotlicffi document to import:
```python
try:
    import brotlicffi as brotli
except ImportError:
    import brotli
```
Because it is difficult to install `brotlicffi` on cygwin, but `Brotli` is easy (existing software package).
So on cygwin, only can use `import brotli`.

GeekDuanLian:
I suggest that the way to install dependencies is also in accordance with the `brotlicffi` documentation:
```python
install_requires=[
    "brotli; platform_python_implementation == 'CPython'",
    "brotlicffi; platform_python_implementation != 'CPython'"
]
```

PR #1607, opened 2021-04-28
Stream tweak

tomchristie:
Minor refactoring around usage of `Request(stream=...)` and `Response(stream=...)`.

* Reorder code in `__init__` so that the default case comes first, and the raw `stream=...` case comes second.
* Creating requests using `request = Request(..., stream=...)` should create a request without *any* headers being pre-populated. Eg. creating a request within a transport would use something like `request = Request(method=..., url=..., headers=..., stream=...)` which should not auto-populate any additional headers.


johtso:
Ooo, nice! Will be happy to delete a bit of code in `httpx-caching` and just use `Request` objects internally.

PR #1618, opened 2021-04-29
Use either brotli or brotlicffi.

tomchristie:
See https://github.com/encode/httpx/pull/1605#issuecomment-829268489

* The `brotli` package is recommended for CPython.
* The `brotlicffi` package is recommended for PyPy.

Note the following:

* We have to use `Decompressor.decompress` if `brotlicffi` is installed, but `Decompressor.process` if `brotli` is installed.
* We can use `Decompressor.finish()` if `brotlicffi` is installed.
* We use the oddly-cased `brotli.error` instead of `brotli.Error` because it is supported as a synonym in both packages.

Also worth calling out the the older `brotlipy` package *might* be installed on the system. In which case it will shadow the `brotli` package namespace. Ug.


httpx/_decoders.py
```diff
@@ -99,37 +95,44 @@ class BrotliDecoder(ContentDecoder):
...
-        self.decompressor = brotlicffi.Decompressor()
+        self.decompressor = brotli.Decompressor()
         self.seen_data = False
         if hasattr(self.decompressor, "decompress"):
-            self._decompress = self.decompressor.decompress
+            # The 'brotlicffi' package.
+            self._decompress = self.decompressor.decompress  # pragma: nocover
         else:
+            # The 'brotli' package.
             self._decompress = self.decompressor.process  # pragma: nocover
```
j178, line 112: I think`brotlicffi`  also has a `process` alias for `decompress`:

https://github.com/python-hyper/brotlicffi/blob/68c0431673e6c59549079e6588b043a707262c80/src/brotlicffi/_api.py#L423

How about just use `self.decompressor.process`?

tomchristie: True. Tho I double checked and the `brotlipy` package doesn't, so what we want to do here depends on if we want to take care around the possibility that on some systems the older `brotlipy` package might have been installed. (Or not.)

If we do care about that awkward case, then we should just update the comment strings here.
If we don't case about that case, then we should just always use `Decompressor().process()`.

Perhaps we could check which of `brotli` vs `brotlipy` is installed the `_compat.py` module, and raise an error if `brotlipy` is installed and is shadowing the `brotli` namespace.

For example:

```python
    if hasattr(brotli.Decompressor, 'decompress'):
        raise RuntimeError(
            "The brotlipy library appears to be installed, and is shadowing the 'brotli' namespace. "
            "Uninstall brotlipy and use brotlicffi instead."
        )
```

j178 (approved):

PR #1621, opened 2021-05-01
Make the MockTransport example more robust/correct

vytas7:
The current [Mock transports](https://www.python-httpx.org/advanced/#mock-transports) example reads:
```
# Switch to a mock transport, if the TESTING environment variable is set.
```
However, at the time of writing this, if the `TESTING` envvar is not set, a `KeyError` would be raised instead of following the `HTTPTransport` branch.

This patch aims to make the example more robust.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -1100,7 +1100,7 @@ def handler(request):
 
 
 # Switch to a mock transport, if the TESTING environment variable is set.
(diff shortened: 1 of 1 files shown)


florimondmanca (approved):
Indeed, always best for examples to work out-of-the-box for non-enabled cases too. :-)

Thanks

PR #1622, opened 2021-05-02
raise RuntimeError depending on state

### LINKED ISSUE #1615: Stricter enforcement around client scoping.
Raised via: https://github.com/encode/httpx/issues/1478#issuecomment-828955977

There are really two sensible ways to use a client instance wrt. scoping the lifetime...

Implicitly opened clients...

```python
client = httpx.Client()
...
client.close()  # Optional in the sync case. Issues a warning if omitted in the async case.
```

Explicitly opened clients...

```python
    ...
```

However we currently don't guard against unexpected usages, which would likely indicate a user error, and could lead to unexpected behaviours.

* Using a client within a `with` scope multiple times in parallel.
* Using a client making plain implicitly opening requests, and *then* opening it within a `with` scope.
* Using a client within a `with` scope, but calling an explicit `close()`.
* Using a client within a `with` scope multiple times sequentially. We could *potentially* allow this? Possibly a footgun? Not sure?

We could guard against these cases and more strictly enforce correct client scoping.

For instance, the user-code error that caused https://github.com/encode/httpx/issues/1478#issuecomment-828955977 would be resolved if we added some `RuntimeError` cases around weird client state transitions on `__enter__`/`__exit__`/`__aenter__`/`__aexit__`.

So:

* On `__enter__`/`__aenter__` raise a `RuntimeError` if the client was not in an `UNOPENED` state.
* On `__exit__`/`__aexit__` raise a `RuntimeError` if the client was not in a `CLOSED` state.

jkklapp:
Closes #1615


jkklapp:
@tomchristie

httpx/_client.py
```diff
@@ -1196,6 +1201,11 @@ def __exit__(
         exc_value: BaseException = None,
         traceback: TracebackType = None,
     ) -> None:
+        if self._state != ClientState.OPENED:
+            raise RuntimeError(
+                "Cannot close a client, as the current state is not 'OPENED'."
```
ahopkins, line 1206: I'm not sure this is the most obvious error message either. Might be helpful to reference what the state is so that if it's already closed that would be obvious in the message.

tomchristie:
Closed via #1800

Issue #1614, opened 2021-04-29
Issue a warning or exception when a client is closed while requests remain in-flight.


tomchristie:
Raised via: https://github.com/encode/httpx/issues/1478#issuecomment-828955977

If developer erroneously closes a client while a request/response is still in-flight, then there's no warning that this has occurred, and the seen exception is as-if the server had disconnected early.

At a minimum we'd really want to *at least* issue a warning when this occurs. Ideally we also want the response to raise a clear exception that properly indicates the underlying cause of the error.

tomchristie:
The user-code error that caused this would also be resolved if we added some `RuntimeError` cases around weird client state transitions on `__enter__`/`__exit__`/`__aenter__`/`__aexit__`.

So:

* On `__enter__`/`__aenter__` raise a `RuntimeError` if the client was not in an `UNOPENED` state.
* On `__exit__`/`__aexit__` raise a `RuntimeError` if the client was not in a `CLOSED` state.

```python
    def __enter__(self: T) -> T:
        if self._state == ClientState.OPENED:
            # Most likely this indicates that a client is being incorrectly used in multiple with blocks.
            raise RuntimeError("Client instances should not be repeatedly opened.")
        if self._state == ClientState.CLOSED:
            # Strictly speaking we *could* allow client instances to be closed and re-opened
            # within different `with` blocks, but it's not a great idea as it can easily lead to
            # invalid/unexpected states.
            raise RuntimeError("Client instances should not be re-opened once closed.")

        self._state = ClientState.OPENED
        ...
```

```python
    def __exit__(self: T) -> T:
        if self._state == ClientState.CLOSED:
            # This should have already raised a `RuntimeError` in `__enter__`, because it
            # indicates that the clients had `__enter__`/`__exit__` called multiple times.
            raise RuntimeError("Client instances cannot be repeatedly closed.")
        if self._state == ClientState.UNOPENED:
            # This shouldn't ever occur.
            # Would indicate `__exit__` being called without a prior call to `__enter__` having occurred.
            raise RuntimeError("Client instances cannot be closed if they have not been opened.")

        self._state = ClientState.OPENED
        ...
```

tomchristie:
Somewhat related to #1615

tomchristie:
Interesting spin off thought here. We *could* indicate some basic information on the client `__repr__` here.

```pycon
>>> client
<httpx.Client (4 requests sent, 1 request in-flight)>
```

ahopkins:
[...]

PR #1623, opened 2021-05-03
check pending request count to warn

jkklapp:
related to #1614

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1284,14 +1284,6 @@ class AsyncClient(BaseClient):
             )
         self._mounts = dict(sorted(self._mounts.items()))
 
-    @property
-    def pending_requests(self) -> int:
-        return self._pending_requests
-
-    @pending_requests.setter
-    def pending_requests(self, p: int) -> None:
-        self._pending_requests = p
-
     def _init_transport(
         self,
         verify: VerifyTypes = True,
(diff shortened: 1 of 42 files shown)


httpx/_client.py
```diff
@@ -1330,6 +1332,14 @@ def __init__(
             )
         self._mounts = dict(sorted(self._mounts.items()))
 
+    @property
+    def pending_requests(self) -> int:
+        return self._pending_requests
+
+    @pending_requests.setter
+    def pending_requests(self, p: int) -> None:
+        self._pending_requests = p
```
tomchristie, line 1341: I don't think we want to add any `client.pending_requests` attribute to the API. I'd suggest we keep any state here private.

httpx/_client.py
```diff
@@ -1521,6 +1532,7 @@ async def send(
             for hook in self._event_hooks["response"]:
                 await hook(response)
 
+            self.pending_requests -= 1
```
tomchristie, line 1535: Really we'd want to decrement pending requests, at the point the response is closed.

(This functionality might be more fiddly than it looks at first.)

tomchristie:
Closed via https://github.com/encode/httpcore/commit/7ddb4cabecb551ce6a79f7db5aa540f19530cb2b

You'll now get a `RuntimeError` with `"The connection pool was closed while {requests_still_in_flight} HTTP requests/responses were still in-flight."` if you close the client prematurely.

PR #1631, opened 2021-05-07
".ok" property added to the Response object

GabrielBG99:
In the requests API we have the ".ok" attribute that returns `False` only if the status code if greather than or equal 400. I think it would be interesting to have this property in the response model as it is more convenient in some cases where we want to know if the request was not completed successfully but we do not want to handle the `HTTPStatusError`.


GabrielBG99:
@johtso Oh sorry. I hadn't seen that part of the documentation.
I will close the PR.

PR #1645, opened 2021-05-20
docs: slightly clarify event hooks

K900:
* Be more specific about when the hooks are called
* Explicitly mention and demonstrate that hooks are allowed to modify requests

See https://github.com/encode/httpx/discussions/1637


tomchristie (approved):
Great, yup. 👍

PR #1647, opened 2021-05-21
Fix: proxy http2 prior to knowledge support

p1-ra:
Hi, 

This PR is related to https://github.com/encode/httpx/pull/1624 and required the following PR at httpcore side:  https://github.com/encode/httpcore/pull/342
It fix the HTTP2  prior of knowledge support for HTTP proxies.
- https://github.com/encode/httpx/commit/5a34fdd47a722b987d5da29c657323051a6884b2 propagate client `http1` support to the configured proxies
- https://github.com/encode/httpx/commit/9825d69155abad1f88da7c87cc4df69ac256832e introduce new configurations on the Proxy object to override the client http capabilities: `http1_override`, `http2_override`

```
proxies = {
    # Route all traffic through a proxy by default, inherit http1,http2 capabilities from client
    "all://": "http://localhost:8030",
    # But don't use proxies for HTTPS requests to "domain.io"...
    "https://domain.io": None,
    # And use another proxy for requests to "example.com" and its subdomains, only support http1
    "all://*example.com": httpx.Proxy(url="http://localhost:8031", http1_override=True, http2_override=False),
    # And yet another proxy if HTTP is used,
    # and the "internal" subdomain on port 5550 is requested, only support http2 -- prior to knowlegdge
    "http://internal.example.com:5550": httpx.Proxy(url="http://localhost:8032", http1_override=False, http2_override=True),
}
```

BR,

- [ ] Initially raised as discussion #None

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -580,6 +580,10 @@ class Client(BaseClient):
     to authenticate the client. Either a path to an SSL certificate file, or
     two-tuple of (certificate file, key file), or a three-tuple of (certificate
     file, key file, password).
+    * **http1** - *(optional)* A boolean indicating if HTTP/1 support should be
+    enabled. Defaults to `True`.
+    * **http2** - *(optional)* A boolean indicating if HTTP/2 support should be
+    enabled. Default to `False`.
     * **proxies** - *(optional)* A dictionary mapping proxy keys to proxy
     URLs.
     * **timeout** - *(optional)* The timeout configuration to use when sending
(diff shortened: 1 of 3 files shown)


p1-ra:
@tomchristie @florimondmanca any input/feedbacks regarding this?

I've noticed that https://github.com/encode/httpcore/pull/342 has been closed without comment.

Any reasons to not be able to use HTTP/2 prior to knowledge with proxy?

Does the issue have been resolved elsewhere?

Thanks in advance

tomchristie:
Resolved via https://github.com/encode/httpx/pull/2009 and https://github.com/encode/httpcore/pull/468

PR #1669, opened 2021-06-07
Add a network pytest mark for tests that use the network

antlarr:
Sometimes it's useful to have the tests that use the network
marked so they can be skipped easily when we know the network
is not available.

This is useful for example on SUSE and openSUSE's build servers.
When building the httpx packages (actually, any package in the
distribution) the network is disabled so we can assure
reproducible builds (among other benefits). With this mark, it's
easier to skip tests that can not succeed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,7 @@ combine_as_imports = True
 addopts = -rxXs
 markers =
   copied_from(source, changes=None): mark test as copied from somewhere else, along with a description of changes made to accodomate e.g. our test setup
+  network: marks tests which require network connection
 
 [coverage:run]
 omit = venv/*
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,7 +18,7 @@ combine_as_imports = True
 addopts = -rxXs
 markers =
   copied_from(source, changes=None): mark test as copied from somewhere else, along with a description of changes made to accodomate e.g. our test setup
-  network: marks tests which require network connection
+  network: marks tests which require network connection. Used in 3rd-party build environments that have network disabled.
 
 [coverage:run]
 omit = venv/*
--- a/tests/client/test_proxies.py
+++ b/tests/client/test_proxies.py
@@ -122,6 +122,7 @@ def test_transport_for_request(url, proxies, expected):
 
 
 @pytest.mark.asyncio
+@pytest.mark.network
 async def test_async_proxy_close():
     try:
         client = httpx.AsyncClient(proxies={"https://": PROXY_URL})
@@ -130,6 +131,7 @@ async def test_async_proxy_close():
         await client.aclose()
 
 
+@pytest.mark.network
 def test_sync_proxy_close():
     try:
         client = httpx.Client(proxies={"https://": PROXY_URL})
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -18,6 +18,7 @@ combine_as_imports = True
 addopts = -rxXs
 markers =
   copied_from(source, changes=None): mark test as copied from somewhere else, along with a description of changes made to accodomate e.g. our test setup
+  network: marks tests which require network connection
```
florimondmanca, line 21: Should we embed a bit of the rationale for introducing this marker?

```suggestion
  network: marks tests which require network connection. Allows skipping such tests, e.g. in build environments with network disabled.
```

florimondmanca (commented):
I'm personally okay with this, though perhaps let's make sure this doesn't just appear as a "utility marker", since you'd be relying on it for building on SUSE?

antlarr:
> I'm personally okay with this, though perhaps let's make sure this doesn't just appear as a "utility marker", since you'd be relying on it for building on SUSE?

That's right, we'll use it for building the httpx SUSE (and openSUSE) packages

florimondmanca (approved):

tomchristie:
Seems fair. I'd been keen on us switching these cases out to non-network-requiring tests at some point, tho not a massive priority atm.

@antlarr Could you comment here showing how to disable running the network tests once we've got this marker? Does it require a particular envvar to be set in order to skip them? Where is this marker (and making use of it) documented?

Once we've got that info against this history, then I'm super happy for this one to go in, yup.

s-t-e-v-e-n-k:
https://docs.pytest.org/en/6.2.x/example/markers.html is the upstream documentation for custom markers like this.

antlarr:
> Seems fair. I'd been keen on us switching these cases out to non-network-requiring tests at some point, tho not a massive priority right now.
> 
> @antlarr Could you comment here showing how to disable running the network tests once we've got this marker? 
> Does it require a particular envvar to be set in order to skip them?
> 

Sure. We just use it in our build service by running the tests like this:

`%pytest -k 'not network'`

Where %pytest is a rpm macro that basically runs pytest with the given arguments for every python flavor (currently 3.6, 3.8 and 3.9) in order to test the rpm packages that are being built for each flavor.

You can see this at [the python-httpx spec file in openSUSE build service](https://build.opensuse.org/package/view_file/devel:languages:python/python-httpx/python-httpx.spec?expand=1) in line 84.

> Where is this marker (and making use of it) documented?

@s-t-e-v-e-n-k already linked the pytest documentation for markers. I also added a comment at https://github.com/encode/httpx/pull/1669/commits/5f7ee8e5efb1d85c39a5507598f7772522f9aae3 to document what that particular marker is used for, just in case some other test is added in the future that makes use of the network.

> Once we've got that info against this history, then I'm super happy for this to go in, yup.

Cool

florimondmanca:
@tomchristie With a small tweak to `scripts/test` (passing `"$@"`) we could also do this:

```bash
scripts/test -m "not network"
```

tomchristie:
@florimondmanca Gotcha thanks. We can *potentially* follow up with that, but either ways let's get this in...

PR #1675, opened 2021-06-11
Adding netrc environment variable into the documentation

meist0731:

please ensure you've discussed the your proposal prior to issuing a pull request.

### DIFF
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -132,6 +132,25 @@ Example:
 SSL_CERT_DIR=/path/to/ca-certs/ python -c "import httpx; httpx.get('https://example.com')"
 ```
 
+## `.NETRCFILE`
+
+Valid values: a filename
+
+If this environment variable is set and auth parameter is not defined, HTTPX will add auth information stored in
+.netrc file into request's header. If you do not provide NETRC environment, HTTPX will use default files. (~/.netrc
+, ~/_netrc)
+
+Example:
+
+```python
+# test_script.py
+import httpx
+import os
+os.environ["NETRC"] = "my_default_folder/.my_netrc"
+with httpx.Client(trust_env=False) as client:
+    r = client.get("https://example.org/")
+```
+
 ## Proxies
 
 The environment variables documented below are used as a convention by various HTTP tooling, including:
@@ -175,3 +194,5 @@ python -c "import httpx; httpx.get('http://example.com')"
 python -c "import httpx; httpx.get('http://127.0.0.1:5000/my-api')"
 python -c "import httpx; httpx.get('https://www.python-httpx.org')"
 ```
+
+
--- a/docs/environment_variables.md
+++ b/docs/environment_variables.md
@@ -136,12 +136,12 @@ SSL_CERT_DIR=/path/to/ca-certs/ python -c "import httpx; httpx.get('https://exam
 
 Valid values: a filename
 
-If this environment variable is set and auth parameter is not defined, HTTPX will add auth information stored in .netrc file into request's header. If you do not provide NETRC environment, HTTPX will use default files. (~/.netrc, ~/_netrc)
+If this environment variable is set but auth parameter is not defined, HTTPX will add auth information stored in the .netrc file into the request's header. If you do not provide NETRC environment either, HTTPX will use default files. (~/.netrc, ~/_netrc)
 
 Example:
 
 ```console
-NETRCFILE=/path/to/netrcfile/my_netrc python -c "import httpx; httpx.get('https://example.com')"
+NETRCFILE=/path/to/netrcfile/.my_netrc python -c "import httpx; httpx.get('https://example.com')"
 ```
 
 ## Proxies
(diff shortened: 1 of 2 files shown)


docs/environment_variables.md
```diff
@@ -175,3 +194,5 @@ python -c "import httpx; httpx.get('http://example.com')"
 python -c "import httpx; httpx.get('http://127.0.0.1:5000/my-api')"
 python -c "import httpx; httpx.get('https://www.python-httpx.org')"
 ```
+
+
```
tomchristie, line 198: For tidiness we could do without the trailing newlines here.

docs/environment_variables.md
```diff
@@ -132,6 +132,25 @@ Example:
 SSL_CERT_DIR=/path/to/ca-certs/ python -c "import httpx; httpx.get('https://example.com')"
 ```
 
+## `.NETRCFILE`
+
+Valid values: a filename
```
tomchristie, line 137: "A filename"

docs/environment_variables.md
```diff
@@ -132,6 +132,25 @@ Example:
...
+
+If this environment variable is set and auth parameter is not defined, HTTPX will add auth information stored in
+.netrc file into request's header. If you do not provide NETRC environment, HTTPX will use default files. (~/.netrc
+, ~/_netrc)
```
tomchristie, line 141:
```suggestion
If this environment variable is set and auth parameter is not defined, HTTPX will add auth information stored in .netrc file into request's header. If you do not provide NETRC environment, HTTPX will use default files. (~/.netrc, ~/_netrc)
```

docs/environment_variables.md
meist0731: Get it. This makes more sense. Thank you for this good suggestion.

docs/environment_variables.md
meist0731: Yep. Couldn't agree more. This seems far better than mine. Thank you for this kind suggestion and thoughtful modification. 🙏🏻

meist0731:
@tomchristie and @Vibhu-Agarwal, I've made changes to the description and the example part, but as for the 'Valid values: A filename', I changed it back to lowercase 'a', cuz other env variable descriptions all use the lowercase, if you need all be uppercase, do I need to create a new pull request to do so? Thanks in advance for checking them. :D

docs/environment_variables.md
tomchristie: Appreciating your style of interaction here, @meist0731. ✨

tomchristie:
> as for the 'Valid values: A filename', I changed it back to lowercase 'a', cuz other env variable descriptions all use the lowercase

Ah, so they do, yup. Good call. We should prefer consistency in that case. If we do want to uppercase 'em then treat it as a separate PR yup. (Not *too* fussed about it, but yeah after a colon the following ought to be uppercased right?)

docs/environment_variables.md
```diff
@@ -132,6 +132,18 @@ Example:
 SSL_CERT_DIR=/path/to/ca-certs/ python -c "import httpx; httpx.get('https://example.com')"
 ```
 
+## `.NETRCFILE`
```
j178, line 135: I think this env name should be `NETRC`, right?

https://github.com/encode/httpx/blob/3231211aa364c0406737b78a45268e08c434f94c/httpx/_utils.py#L134-L135

and per the doc:
https://www.python-httpx.org/advanced/#netrc-support

docs/environment_variables.md
```diff
@@ -132,6 +132,18 @@ Example:
...
+
+Valid values: a filename
+
+If this environment variable is set but auth parameter is not defined, HTTPX will add auth information stored in the .netrc file into the request's header. If you do not provide NETRC environment either, HTTPX will use default files. (~/.netrc, ~/_netrc)
+
+Example:
+
+```console
+NETRCFILE=/path/to/netrcfile/.my_netrc python -c "import httpx; httpx.get('https://example.com')"
```
j178, line 144:
```suggestion
NETRC=/path/to/netrcfile/.my_netrc python -c "import httpx; httpx.get('https://example.com')"
```

j178 (changes requested):

docs/environment_variables.md
meist0731: Yep! you're right, this should be 'NETRC', especially in the example part. I would like to confirm one thing that I noticed in the doc  (https://www.python-httpx.org/advanced/#netrc-support) both the right content table and central content title are written with a dot as the start, like ".netrc", so should I need to change this title to be ".NETRC" or simply "NETRC"?

meist0731: > I think this env name should be `NETRC`, right?
> 
> https://github.com/encode/httpx/blob/3231211aa364c0406737b78a45268e08c434f94c/httpx/_utils.py#L134-L135
> 
> and per the doc:
> https://www.python-httpx.org/advanced/#netrc-support

For consistency, '.NETRC' would be better👀may be.

j178: In this `Environment Variables` section, I think we should just use `NETRC`. 👍

meist0731: Gotcha! Thanks for this suggestion.

j178 (approved):
LGTM 💯

florimondmanca (approved):
Thank you :-)

PR #1684, opened 2021-06-14
Added pre-commit config

cansarigol:
- [x] Initially raised as discussion #1683

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 21.5b2
+    hooks:
+      - id: black
+        exclude: "setup.py"
+        args:
+          - "--check"
+          - "--diff"
+          - "--target-version=py36"
+  - repo: https://github.com/pycqa/isort
+    rev: 5.8.0
+    hooks:
+      - id: isort
+        args:
+          - "--check"
+          - "--diff"
+          - "--project=httpx"
+  - repo: https://github.com/PyCQA/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          - flake8-bugbear
+          - flake8-pie==0.5.*
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.902
+    hooks:
+      - id: mypy
+        exclude: setup.py
+        additional_dependencies:
+          - cryptography
+          - types-certifi
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -17,6 +17,10 @@ jobs:
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-beta.2"]
 
     steps:
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ matrix.python-version }}|${{ hashFiles('.pre-commit-config.yaml') }}
       - uses: "actions/checkout@v2"
       - uses: "actions/setup-python@v2"
         with:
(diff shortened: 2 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - id: check-symlinks
+      - id: check-toml
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+      - id: python-check-blanket-noqa
+      - id: python-use-type-annotations
```
tomchristie, line 17: Presumably, if we *didn't* include this section, then the other code formatting changes that we're seeing on this PR wouldn't be required?

If we *did* want to introduce pre-commit, then I'd strongly suggest that we start of with a pull request that introduces it, but *doesn't* make any other linting changes.

httpx/_config.py
```diff
@@ -218,10 +218,10 @@ def __init__(
...
             assert write is UNSET
             assert pool is UNSET
-            self.connect = timeout.connect  # type: typing.Optional[float]
-            self.read = timeout.read  # type: typing.Optional[float]
-            self.write = timeout.write  # type: typing.Optional[float]
-            self.pool = timeout.pool  # type: typing.Optional[float]
+            self.connect: typing.Optional[float] = timeout.connect
+            self.read: typing.Optional[float] = timeout.read
+            self.write: typing.Optional[float] = timeout.write
+            self.pool: typing.Optional[float] = timeout.pool
```
tomchristie, line 224: Regardless of any linting enforcement or not, these typing changes are *clearly* a neater style. I'd be very much in favour of an isolated PR that just deals with cleaning up any "type as comment" cases that we're currently doing.

tomchristie:
My strong tendency is to aim for the absolute minimal amount of tooling possible, throughout. Every piece of extra code we introduce into our toolchain brings extra clunk with it.

There's a couple of different aspects to this one...

* The linting checks running as a pre-commit hook.
* The additional linting checks that are being added here.

I'll address each of these in turn.

### The listing checks running as a pre-commit hook.

Not a bad plan for sure. We don't *necessarily* need extra tooling to get there.

An alternative that we could consider would be adding something like the following to `scripts/install`...

```shell
if [ -d '.git' ] ; then
    echo "scripts/check" > .git/hooks/pre-commit
fi
```

Which gets us to the same point, but with a more lightweight approach.

No *necessarily* better or worse than using this tool, but worth considering as an alternative. From my perspective that sort of slimline approach means less obfuscation.

###  The additional linting checks that are being added here.

I'd kinda want to consider each of these in turn, on their own merit. In some cases we might or might not want them. In other cases we might consider that we've got a valid code change to make but don't necessarily need a hard linting enforcement. In other cases it might be worth double-checking if the existing tools we're using are able to provide the checks.

cansarigol:
Actually I wanted to include that package because of its hooks functionality. There are pretty useful hooks in the list. in case it's not enough, custom hooks which is coded by httpx contributers can also be  included. If we don't want to leverage the hooks, as you said, we don't necessarily need extra tooling to get there.

Definitely agree with you. Additional linting checks should be considered item by item.

tomchristie:
My *personal* feeling on this? I'm not super keen.

Tooling indirection is so *easy* to add into a project, but it obscures what's going on by an extra layer.
I'd like us to very strongly favour a minimal tooling approach *wherever possible*, across the encode projects.

graingert:
> Tooling indirection is so easy to add into a project, but it obscures what's going on by an extra layer.
I'd like us to very strongly favour a minimal tooling approach wherever possible, across the encode projects.

Each bespoke shell script in the repository adds a brand new tool and increases the cognitive load when contributing to multiple packages. Using simpler and more generic tools like pre-commit actually reduces the overall number of tools I need to learn

PR #1714, opened 2021-06-25
use context.minimum_version in py3.7+ where available

graingert:
ssl.OP_NO_TLSv1
ssl.OP_NO_TLSv1_1

are deprecated in py3.7+ and using these options prevents
downstream users of httpx.create_ssl_context using
minimum_version

### DIFF
--- a/httpx/_compat.py
+++ b/httpx/_compat.py
@@ -20,6 +20,8 @@ if sys.version_info >= (3, 10) or (
     # 'SSLContext.minimum_version' from Python 3.7 onwards, however
     # this attribute is not available unless the ssl module is compiled
     # with OpenSSL 1.1.0g or newer.
+    # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
+    # https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.minimum_version
     def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
         context.minimum_version = ssl.TLSVersion.TLSv1_2
 
--- a/httpx/_compat.py
+++ b/httpx/_compat.py
@@ -13,22 +13,18 @@ except ImportError:
     from async_generator import asynccontextmanager  # type: ignore # noqa
 
 
-if sys.version_info >= (3, 10) or (
-    sys.version_info >= (3, 7) and ssl.OPENSSL_VERSION_INFO >= (1, 1, 0, 7)
-):
-    # The OP_NO_SSL* and OP_NO_TLS* become deprecated in favor of
-    # 'SSLContext.minimum_version' from Python 3.7 onwards, however
-    # this attribute is not available unless the ssl module is compiled
-    # with OpenSSL 1.1.0g or newer.
-    # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
-    # https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.minimum_version
-    def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
+def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
+    if sys.version_info >= (3, 10) or (
+        sys.version_info >= (3, 7) and ssl.OPENSSL_VERSION_INFO >= (1, 1, 0, 7)
+    ):
+        # The OP_NO_SSL* and OP_NO_TLS* become deprecated in favor of
+        # 'SSLContext.minimum_version' from Python 3.7 onwards, however
+        # this attribute is not available unless the ssl module is compiled
+        # with OpenSSL 1.1.0g or newer.
+        # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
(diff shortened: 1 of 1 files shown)


tomchristie:
> are deprecated in py3.7+

Could you link to the relevant Python docs to help with reviewing this?

Given that we've enabled warnings-as-errors on our CI, wouldn't we expect to be raising errors here if we were using a deprecated Python API?

graingert:
https://docs.python.org/3.7/library/ssl.html#ssl.OP_NO_TLSv1

> Deprecated since version 3.7: The option is deprecated since OpenSSL 1.1.0, use the new SSLContext.minimum_version and SSLContext.maximum_version instead.

httpx/_compat.py
```diff
@@ -13,13 +13,20 @@
...
-        context.options |= ssl.OP_NO_TLSv1
-        context.options |= ssl.OP_NO_TLSv1_1
+
+
+else:
+
+    def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
+        context.options |= (
+            ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3 | ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1
+        )
```
tomchristie, line 32: Okay. I'd suggest we aim to keep the change footprint on pull requests as low as possible. Not reformatting the `context.options |= ...` style here would help with that.

```python
def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
     if (
        sys.version_info >= (3, 10) or
        sys.version_info >= (3, 7) and ssl.OPENSSL_VERSION_INFO >= (1, 1, 0, 7)
     ):
         # The 'minimum_version' attribute is available from Python 3.10 onwards,
         # Or Python 3.7 onwards if the ssl module is compiled with OpenSSL 1.1.0g
         # or newer.
         # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
         # https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.minimum_version
         context.minimum_version = ssl.TLSVersion.TLSv1_2
     else:
         # If 'minimum_version' isn't available, we configure these options with
         # the older deprecated variants.
         context.options |= ssl.OP_NO_SSLv2
         context.options |= ssl.OP_NO_SSLv3
         context.options |= ssl.OP_NO_TLSv1
         context.options |= ssl.OP_NO_TLSv1_1
```

tomchristie (changes requested):

httpx/_compat.py
```diff
@@ -13,12 +13,20 @@
...
 
-def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
-    if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10) or (
+    sys.version_info >= (3, 7) and ssl.OPENSSL_VERSION_INFO >= (1, 1, 0, 7)
+):
+    # The OP_NO_SSL* and OP_NO_TLS* become deprecated in favor of
+    # 'SSLContext.minimum_version' from Python 3.7 onwards, however
+    # this attribute is not available unless the ssl module is compiled
+    # with OpenSSL 1.1.0g or newer.
```
graingert, line 22:
```suggestion
    # with OpenSSL 1.1.0g or newer.
    # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
    # https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.minimum_version
```

httpx/_compat.py
```diff
@@ -13,12 +13,22 @@
     from async_generator import asynccontextmanager  # type: ignore # noqa
 
 
-def set_minimum_tls_version_1_2(context: ssl.SSLContext) -> None:
-    if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10) or (
```
tomchristie, line 16: So there's still a little too much unnecessary change footprint here. We don't really need to switch the structure around here. Let's keep `set_minimum_tls_version_1_2` as the top level function, with the switch *inside* it.

We really just want to:

* Update the `sys.version_info >= (3, 10)` branch condition.
* Possibly neaten up the commenting.

tomchristie (changes requested):

httpx/_compat.py
```diff
@@ -14,11 +14,17 @@
...
+        # 'SSLContext.minimum_version' from Python 3.7 onwards, however
+        # this attribute is not available unless the ssl module is compiled
+        # with OpenSSL 1.1.0g or newer.
+        # https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version
+        # https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.minimum_version
         context.minimum_version = ssl.TLSVersion.TLSv1_2
     else:
-        # These become deprecated in favor of 'context.minimum_version'
-        # from Python 3.10 onwards.
         context.options |= ssl.OP_NO_SSLv2
```
graingert, line 28:
```suggestion
        # If 'minimum_version' isn't available, we configure these options with
        # the older deprecated variants.
        context.options |= ssl.OP_NO_SSLv2
```

tomchristie (approved):

tomchristie:
Fab! ✨

PR #1720, opened 2021-06-28
check sys.version_info and ssl.OPENSSL_VERSION_INFO once

graingert:


tomchristie (approved):
Yes, I guess it *is* a little neater this way around. 👍

PR #1721, opened 2021-06-28
📌 Pin development requirements

Kludex:
Idea raised on #1680:

> Should we make sure that we have more strictly pinned dependancies before pulling this in?
> 
> Eg... https://github.com/encode/httpcore/pull/184/files
> 
> _Originally posted by @tomchristie in https://github.com/encode/httpx/issues/1680#issuecomment-862320582_

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,5 +27,3 @@ trio==0.19.0
 trio-typing==0.5.0
 trustme==0.8.0
 uvicorn==0.14.0
-
-attrs==21.2.0  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,7 @@
+# We're pinning our tooling, because it's an environment we can strictly control.
+# On the other hand, we're not pinning package dependencies, because our tests
+# needs to pass with the latest version of the packages.
+# Reference: https://github.com/encode/httpx/pull/1721#discussion_r661241588
 -e .[http2,brotli]
 
 # Documentation
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,31 +1,31 @@
 -e .[http2,brotli]
 
 # Documentation
-mkdocs
-mkautodoc
-mkdocs-material
+mkdocs==1.2.1
+mkautodoc==0.1.0
+mkdocs-material==7.1.8
 
 # Packaging
-twine
-wheel
+twine==3.4.1
+wheel==0.36.2
 
 # Tests & Linting
-autoflake
+autoflake==1.4
 black==20.8b1
 coverage==5.3
-cryptography
-flake8
-flake8-bugbear
-flake8-pie==0.5.*
-isort==5.*
-mypy
-types-certifi
-pytest==6.*
-pytest-asyncio
-pytest-trio
-trio
-trio-typing
-trustme
-uvicorn
+cryptography==3.4.7
+flake8==3.9.2
+flake8-bugbear==21.4.3
+flake8-pie==0.5.0
+isort==5.9.1
+mypy==0.910
+types-certifi==0.1.4
+pytest==6.2.4
+pytest-asyncio==0.15.1
+pytest-trio==0.7.0
+trio==0.19.0
+trio-typing==0.5.0
+trustme==0.8.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,31 +1,31 @@
 -e .[http2,brotli]
```
tomchristie, line 1: So... we'd need the deps that this introduces to be pinned, right?

*(To resolve this I'd start by looking up how we did this with `httpcore`)* 🤷‍♂️

requirements.txt
```diff
@@ -1,31 +1,31 @@
...
+pytest==6.2.4
+pytest-asyncio==0.15.1
+pytest-trio==0.7.0
+trio==0.19.0
+trio-typing==0.5.0
+trustme==0.8.0
+uvicorn==0.14.0
 
-attrs>=19.3.0  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665
+attrs==21.2.0  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665
```
tomchristie, line 31: It's *possible* that we don't need this now, maybe? (Pinning the other cases might have resolved it?)

I *guess* we could explore that as a follow-up.

requirements.txt
Kludex: Hmm, it follows the same as `httpcore`: https://github.com/encode/httpcore/pull/184/files

requirements.txt
Kludex: > It's possible that we don't need this now, maybe? (Pinning the other cases might have resolved it?)

Correct.

requirements.txt
tomchristie: Okay, right. Seems like an oversight there too, then. (I think, right?)

Kludex: Hmm, is it? :thinking:
It feels weird to pin the extra requirements :thinking: I mean, the tests **needs** to pass with the latest version allowed by the version range on `setup.py`, right? :thinking:

tomchristie: Ahhhhhh.... right. Which helps tease out an interesting point.

(I'm glad we've taken this super slowly, because I don't think we'd have noticed it otherwise)

I'd kinda agree with your assessment here...

* We probably want our *tooling* to be pinned, because it's an environment we can strictly control, and it'll give more reproducible build results.
* We possibly don't want our *package dependencies* to be pinned, because they won't be for our users, and we probably may as well just *always* be installing vs. whatever's the latest.

If that *is* what we opt for with the pinning, then adding caching https://github.com/encode/httpx/pull/1680 probably *isn't* something we'd want to do. Because it'd mean we'll *sometimes* get the latest package dependancies, and sometimes not get them, depending on that state of the cache. Or it might mean that we *are* okay with caching, but we only cache the tooling deps, and we always make sure to update to the very latest dependencies.

Either way though, I think we should resolve what we do or don't want for the pinning, and get this PR resolved, because this is still a valid PR regardless of the outcome of #1680.

tomchristie (approved):
Okay, great.

This is simple, isolated, and feels like a positive step.
I'm happy with approving this now.

Suggested follow-ups:

* Let's get dependabot enabled. We really do want that once we're pinning the tooling, so that we can keep our tooling fresh. Updates once per month will be absolutely fine. If we're able to turn on auto-merge-if-checks-pass then that might be a good idea too.
* We might well want to add a little bit of commenting to our 'requirements.txt' to help explain what our approach is here, and why. (Ie. that we pin our tooling dependancies, but don't pin our packaging dependancies.)

requirements.txt
Kludex: > If that is what we opt for with the pinning, then adding caching #1680 probably isn't something we'd want to do. Because it'd mean we'll sometimes get the latest package dependancies, and sometimes not get them, depending on that state of the cache. Or it might mean that we are okay with caching, but we only cache the tooling deps, and we always make sure to update to the very latest dependencies.

Well, we can also invalidate the cache within a few days (say `7 days`). But not caching the requirements on `setup.py` also works, ofc. 

Anyway, I guess we should continue this discussion on #1680

Kludex:
@tomchristie I've added a note about our decision on the top of the file.

tomchristie:
This is good, yup!

PR #1724, opened 2021-06-29
Use utf-8-sig for decoding to handle BOM

fredrike:

### DIFF
--- a/httpx/_decoders.py
+++ b/httpx/_decoders.py
@@ -272,7 +272,7 @@ class TextDecoder:
         if self.decoder is None:
             # If this is the first decode pass then we need to determine which
             # encoding to use by attempting UTF-8 and raising any decode errors.
-            attempt_utf_8 = codecs.getincrementaldecoder("utf-8")(errors="strict")
+            attempt_utf_8 = codecs.getincrementaldecoder("utf-8-sig")(errors="strict")
             try:
                 attempt_utf_8.decode(data)
             except UnicodeDecodeError:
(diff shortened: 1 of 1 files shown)


tomchristie:
Interesting. I didn't know about the `utf-8-sig` codec.

* Is it documented somewhere?
* Are you able to point to an example URL with content that includes a leading BOM?

fredrike:
Here is a short discussion about `utf-8-sig` https://www.howtosolutions.net/2019/04/python-fixing-unexpected-utf-8-bom-error-when-loading-json-data/

Official doc: https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig

"Broken" page: https://service.karlskrona.se/FileStorageArea/Documents/bad/1.1/swimAreas.geojson and https://service.karlskrona.se/FileStorageArea/Documents/bad/swimAreas.json

tomchristie:
Checking the test run I can see that there's currently a linter failure... https://github.com/encode/httpx/pull/1724/checks?check_run_id=2942212371

You'll want to run `scripts/lint` in order to resolve this.

tomchristie:
Can you provide some example code that you're using?...

As I was reviewing this I was reminded that `r.json()` actually calls the `guess_json_utf` utility function, which accounts for BOM markers, and is more complex/awkward than just using `utf-8-sig`.

https://github.com/encode/httpx/blob/ab64f7c41fc0fbe638dd586fecf0689c847109bb/httpx/_utils.py#L103

fredrike:
> Can you provide some example code that you're using?...
> 
> As I was reviewing this I was reminded that `r.json()` actually calls the `guess_json_utf` utility function, which accounts for BOM markers, and is more complex/awkward than just using `utf-8-sig`.
> 
> https://github.com/encode/httpx/blob/ab64f7c41fc0fbe638dd586fecf0689c847109bb/httpx/_utils.py#L103

Smart, I'll write a short example but think that the json function is a better solution.

I faced this in home-assistant and opened a case there but realized that it might be better here. But usin the httpx json function is probably the best for all https://github.com/home-assistant/core/pull/52257.

PR #1736, opened 2021-07-04
enforce-upload-files-binary-type

### LINKED ISSUE #1706: Enforce that upload files must be binary, not text.
### Based on this discussion... https://github.com/encode/httpx/discussions/1704

<sup>Originally posted by **plotski** June 21, 2021</sup>
I just spent several hours debugging `LocalProtocolError("Too much data for declared Content-Length")`. I've figured out what's going on, but I don't know what the best fix would be or if this is entirely my own fault.

I was passing a `StringIO` object to a POST request, e.g.:

```python
>>> files = {'upload-file': ('foo', StringIO('cöntént of föö'))}
>>> r = httpx.post("https://httpbin.org/post", files=files)
```

The `StringIO` object ends up in `FileField.get_length()`, which calls `peek_filelike_length()`, which correctly reports the the length of the string. However, we must know the length of the string as `bytes` because that's what we need for the POST request.

The fix is very simple: Use `BytesIO` instead of `StringIO`. But the error message was extremely unhelpful.

Maybe `FileField` shouldn't accept a `StringIO` object? Or maybe it should convert to `bytes` first? Or maybe this is really obvious for anyone with more experience and there's no need to fix it?</div>

I'm fairly sure that the most sensible thing to do here is to raise an error on:

* Files opened in text mode.
* `StringIO` instances, used instead of `BytesIO` instances.

aminalaee:
Fixes #1706

Forces `FileContent` to be `bytes` or `BytesIO` instances. For `str` and `StringIO` instances a `TypeError` is raised.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -1,4 +1,5 @@
 import binascii
+import io
 import os
 import typing
 from pathlib import Path
@@ -80,6 +81,9 @@ class FileField:
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
             content_type = guess_content_type(filename)
+
+        if isinstance(fileobj, str) or isinstance(fileobj, io.StringIO):
+            raise TypeError(f"Expected bytes or bytes-like object got: {type(fileobj)}")
 
         self.filename = filename
         self.file = fileobj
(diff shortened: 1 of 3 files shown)


aminalaee:
Looks like Github workflow timed out (unrelated to PR). How can I re-run that?

tomchristie:
Fab, yes.. 👍

PR #1749, opened 2021-07-13
Fix typo in wsgi.py

adriangb:
I _think_ this should say WSGI, not ASGI


tomchristie:
I think so, too. 😀

Good spot.

PR #1757, opened 2021-07-19
Replace yield inside for loops with yield from

monosans:


tomchristie:
Thanks.

I have mixed feelings about this. One the whole I'm probably a -1...

`yield from` is a more obscure piece of syntax than the simple formulation, so it's not at all clear to me that we're helping our users by using a lesser-known area of Python, for the sake of slightly shorter code.

monosans:
> lesser-known area of Python

I agree that it is lesser-known, but at the same time it's more readable for those who do know this syntax. Those who don't know what **yield from** does can search the Internet in a minute and learn a little more about Python.

And in general, I don't think it's hard to guess what exactly this syntax does.

> On the other hand, it is perhaps neater?

Yeah, it looks way cleaner.



BTW, `yield from` is already used 4 times in the current code.

tomchristie:
> BTW, yield from is already used 4 times in the current code.

Yup, I'd had it on my own inner backlog to get around to removing those. I recall seeing them and thinking "I can guess what that does, but I don't know that syntax", and meaning to remove 'em, but never getting around to it partly because I couldn't remember exactly what it was I was looking for.

florimondmanca:
I think those existing `yield from` instances are in the `multipart` code? That's where I remember using it, I believe. :-)

graingert:
I'd be more comfortable if these changes were done with pyupgrade

graingert:
@tomchristie `yield from` also helps a lot with generator.close

graingert:
here's a PR introducing `yield from` using `pyupgrade` to ensure there's no mistakes: https://github.com/encode/httpx/pull/1762

tomchristie:
> `yield from` also helps a lot with generator.close

Could you provide an example, or a pointer to some more context on this?

graingert:
> > `yield from` also helps a lot with generator.close
> 
> Could you provide an example, or a pointer to some more context on this?

`yield from` delegates all generator operations as well as just `next`, so you can .close the top most generator and you don't need to rely on the GC eventually closing a stack of `for x in v: yield x` for you

tomchristie:
I was starting to warm to this change.

Then realised that the equivalent `async yield from ...`  doesn't exist in Python (I think?) which rather puts me off it, since it's certainly a nice property for us to try to ensure that whenever we've got both sync + async cases, that they ought to exactly mirror each other.

graingert:
the equivalent of:

```
yield from genfn()
```
is

```
async with contextlib2.aclosing(agenfn()) as agen:
    async for v in agen:
        yield v
```

tomchristie:
I actually like this, but the clincher is that there's no equivalent for async, so we'd end up without nicely mirrored equivelents between the two cases.

Thanks for taking a go at it though, it'd probably be a neater approach other than that.

PR #1759, opened 2021-07-20
Replace for loops with comprehensions

monosans:


tomchristie:
These read more neatly to me now, yup.

Thanks!

tests/models/test_responses.py
```diff
@@ -572,10 +572,7 @@ def test_iter_lines():
         200,
         content=b"Hello,\nworld!",
     )
-
-    content = []
-    for line in response.iter_lines():
-        content.append(line)
+    content = [line for line in response.iter_lines()]
```
tomchristie, line 575: We could also use `content = list(response.iter_lines())` here.

graingert: https://pypi.org/project/flake8-comprehensions/

PR #1762, opened 2021-07-21
Replace yield inside for loops with yield from with pyupgrade

graingert:

- [x] Initially raised as discussion https://github.com/encode/httpx/discussions/1764

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,38 @@
+repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.21.2
+    hooks:
+      - id: pyupgrade
+        args: ["--py36-plus"]
+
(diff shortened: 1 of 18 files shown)


tomchristie:
So, I'd really like use to use the "discussions first" approach in a meaningful way.
Perhaps we need to change some wording or so?
The intent is that we don't escalate into pull requests, until there's been an OK on it.

Anyways, with my maintainer-director hat on here, I'm -1 on switching to `pyupgrade`.
I'd like us to have a policy of minimising the amount of tooling we use wherever possible.

tomchristie:
See also https://github.com/encode/httpx/pull/1684#issuecomment-862289637

tomchristie:
Closing with same motivations as https://github.com/encode/httpx/pull/1757#issuecomment-897563839

PR #1770, opened 2021-07-26
tests/conftest.py: conditionally import typing_extensions on older Py…

thesamesam:
…thons

We don't always need typing_extensions -- it's a backport of the
'typing' module in Python 3.8+.

Signed-off-by: Sam James <sam@gentoo.org>


thesamesam:
Actually, I need to check this a bit more.

PR #1771, opened 2021-07-26
Do not require typing_extensions on Python 3.8+

mgorny:
typing.Literal is part of Python's stdlib since Python 3.8, so require
the typing_extensions backport only on older Python versions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import os
+import sys
 import threading
 import time
 import typing
@@ -14,9 +15,15 @@ from cryptography.hazmat.primitives.serialization import (
     PrivateFormat,
     load_pem_private_key,
 )
-from typing_extensions import Literal
 from uvicorn.config import Config
 from uvicorn.main import Server
+
+# `typing.Literal` exists from Python 3.8 onwards.
+# For <=3.7 we require the `typing_extensions` package for a backported version.
+if sys.version_info >= (3, 8):  # pragma: no cover
+    from typing import Literal
+else:  # pragma: no cover
+    from typing_extensions import Literal
 
 from httpx import URL
 from tests.concurrency import sleep
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,7 @@ pytest-trio
 trio
 trio-typing
 trustme
+typing_extensions; python_version < "3.8"
 uvicorn
 
 attrs>=19.3.0  # See: https://github.com/encode/httpx/pull/566#issuecomment-559862665


tomchristie:
Oh that's an interesting point, thanks!

Perhaps we could just avoid using the `Literal` type entirely?

Had a look at this just now...

Initially I tried changing the type annotation from `Literal["session"]` to `str`.

```python
SERVER_SCOPE: str = "session"
```

However that causes several type errors from mypy, such as...

```
tests/conftest.py:170: error: No overload variant of "fixture" matches argument type "str"
tests/conftest.py:170: note: Possible overload variant:
tests/conftest.py:170: note:     def fixture(fixture_function: None = ..., *, scope: Union[Union[Literal['session'], Literal['package'], Literal['module'], Literal['class'], Literal['function']], Callable[[str, Config], Union[Literal['session'], Literal['package'], Literal['module'], Literal['class'], Literal['function']]]] = ..., params: Optional[Iterable[object]] = ..., autouse: bool = ..., ids: Union[Iterable[Union[None, str, float, int, bool]], Callable[[Any], Optional[object]], None] = ..., name: Optional[str] = ...) -> FixtureFunctionMarker
tests/conftest.py:170: note:     <1 more non-matching overload not shown>
```

Instead I needed to change all the `pytest.fixture(scope=SERVER_SCOPE)` to `@pytest.fixture(scope="session")`.

I'd actually probably prefer that we just switch to using `@pytest.fixture(scope="session")` throughout, rather than having the `SERVER_SCOPE` indirection, plus a conditional dependancy on `typing_extensions`.

tests/conftest.py
```diff
@@ -14,10 +14,16 @@
...
     load_pem_private_key,
 )
-from typing_extensions import Literal
 from uvicorn.config import Config
 from uvicorn.main import Server
 
+# `typing.Literal` exists from Python 3.8 onwards.
+# For <=3.7 we require the `typing_extensions` package for a backported version.
+if sys.version_info >= (3, 8):
+    from typing import Literal
```
graingert, line 23: so the comment is also removed when pyupgrade is run:

```suggestion
if sys.version_info >= (3, 8):
    # `typing.Literal` exists from Python 3.8 onwards.
    # For <=3.7 we require the `typing_extensions` package for a backported version.
    from typing import Literal
```

graingert:
> I'd actually probably prefer that we just switch to using @pytest.fixture(scope="session") throughout, rather than having the SERVER_SCOPE indirection, plus a conditional dependancy on typing_extensions.

I tend to agree, it was confusing to see it when I was reading the code

tests/conftest.py
```diff
@@ -14,10 +15,16 @@
...
+else:
+    from typing_extensions import Literal
```
graingert, line 26:
```suggestion
if sys.version_info >= (3, 8):  # pragma: no cover
    from typing import Literal
else:  # pragma: no cover
    from typing_extensions import Literal
```

mgorny:
Ok, I've merged the proposed changes for completeness and now I'll look into the alternative approach of replacing scope.

mgorny:
Filed as #1773. Feel free to choose either variant and the close the other, or suggest me what I should do otherwise ;-).

graingert (approved):
I think this is an improvement

graingert:
now redundant due to the other PR `tests/conftest.py:26:5: F401 'typing_extensions.Literal' imported but unused`

PR #1773, opened 2021-07-26
Inline SERVER_SCOPE and remove typing_extensions requirement

mgorny:
Inline the confusing SERVER_SCOPE literal as suggested in #1771,
and effectively remove the Literal import bringing typing_extensions
requirement.


graingert (approved):
this one looks good too

tomchristie:
Nice little change, yup. 👍

PR #1781, opened 2021-08-05
Fix JSON wrong encoding tests on big endian platforms

mgorny:
Fix test_json_without_specified_encoding_*_error tests on big endian
platforms.  The tests wrongly assume that data encoded as "utf-32-be"
can not be decoded as "utf-32".  This is true on little endian platforms
but on big endian platforms "utf-32" is equivalent to "utf-32-be".
To avoid the problem, explicitly decode as "utf-32-le", as this should
trigger the expected exception independently of platform's endianness.

### DIFF
--- a/tests/models/test_responses.py
+++ b/tests/models/test_responses.py
@@ -735,7 +735,7 @@ def test_json_without_specified_encoding_decode_error():
     content = json.dumps(data).encode("utf-32-be")
     headers = {"Content-Type": "application/json"}
     # force incorrect guess from `guess_json_utf` to trigger error
-    with mock.patch("httpx._models.guess_json_utf", return_value="utf-32"):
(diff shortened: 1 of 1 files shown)


mgorny:
I'm sorry for not raising a discussion but for such trivial changes, even if you told me it's all wrong I've spent less time doing it wrong then I'd going through all the formalities.

tomchristie:
Sure thing...

Looking at the codebase I can see that this *just makes more sense now*, so I'm good with it.
Can you tell us how you bumped into this?

mgorny:
One of our ppc64 testers hit the test failure.

mgorny:
(with the Gentoo package)

thesamesam:
For reference, this was bug [802195](https://bugs.gentoo.org/802195) in Gentoo. We have a bunch of big (and little) endian platforms we test popular packages on.

PR #1790, opened 2021-08-09
Add `Client(allow_redirects=<...>)`

tomchristie:
Add a client default for `allow_redirects`, supporting the following usages...

```python
client = httpx.Client(allow_redirects=<True|False>)
```

This brings `allow_redirects` in line with the `timeout` and `auth` parameters, in that they both have per-request-arguments and client-defaults, and use `USE_CLIENT_DEFAULT` as the default request argument.

We'll probably later follow up on this with https://github.com/encode/httpx/discussions/1785 - but this step make sense in it's own right, deserves to be a pull request on it's own. Other steps can follow later.


florimondmanca (approved):
Sounds good. :-)

PR #1791, opened 2021-08-10
Add `charset_normalizer` detection.

tomchristie:
Use `charset_normalizer` to auto-detect character encodings in cases where no `Content-Type: text/...; charset=...` is included.

See https://github.com/encode/httpx/discussions/1657 and https://github.com/tomchristie/top-1000 for some evidence-led rationale behind this change.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -1361,14 +1361,13 @@ class Response:
         """
         Return the encoding, as detemined by `charset_normalizer`.
         """
-        if hasattr(self, "_content"):
-            if len(self.content) < 32:
-                # charset_normalizer will issue warnings if we run it with
-                # fewer bytes than this cutoff.
-                return None
-            match = charset_normalizer.from_bytes(self.content).best()
-            return None if match is None else match.encoding
-        return None
+        content = getattr(self, "_content", b"")
+        if len(content) < 32:
+            # charset_normalizer will issue warnings if we run it with
+            # fewer bytes than this cutoff.
+            return None
+        match = charset_normalizer.from_bytes(self.content).best()
+        return None if match is None else match.encoding
 
     def _get_content_decoder(self) -> ContentDecoder:
         """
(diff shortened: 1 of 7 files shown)


httpx/_models.py
```diff
@@ -1351,6 +1356,19 @@ def charset_encoding(self) -> typing.Optional[str]:
 
         return params["charset"].strip("'\"")
 
+    @property
+    def apparent_encoding(self) -> typing.Optional[str]:
+        """
+        Return the encoding, as detemined by `charset_normalizer`.
+        """
+        content = getattr(self, "_content", b"")
+        if len(content) < 32:
```
Ousret, line 1365: There are cases where the detection works just fine with small content. I would suggest silent the warning instead.

```
* Using the following `Qu'est ce que une étoile?`
chardet detect ISO-8859-1
cchardet detect IBM852
charset-normalizer detect utf-8

  * Using the following `Qu’est ce que une étoile?`
chardet detect utf-8
cchardet detect UTF-8
charset-normalizer detect utf-8

  * Using the following `<?xml ?><c>Financiën</c>`
chardet detect ISO-8859-1
cchardet detect ISO-8859-13
charset-normalizer detect utf-8

  * Using the following `(° ͜ʖ °), creepy face, smiley 😀`
chardet detect Windows-1254
cchardet detect UTF-8
charset-normalizer detect utf-8

  * Using the following `["Financiën", "La France"]`
chardet detect utf-8
cchardet detect ISO-8859-13
charset-normalizer detect utf-8
```

WDYT?

tomchristie: I went with that originally, and a couple of the tests with small amounts of content returned results I wasn't expecting. If `apparent_encoding` is wrong, then we'll end up decoding it with `'utf-8', errors='replace'`, whichI figure  is a pretty reasonable default for the corner case.

Ousret: Alrighty, that is reasonable. :+1: 
In another matter, you may run the detection anyway and check if the result has a SIG/BOM, that could be reasonable too. And discard it if len(content) < 32 and best_guess.bom is False.

```python
results = from_bytes(content)
best_guess = results.best()

if best_guess.bom:
    ...
```

PR #1793, opened 2021-08-10
Add sidebar ad placements through EthicalAds platform

tomchristie:
The intent of this pull request is to add sidebar ads served through [the EthicalAds platform](https://www.ethicalads.io).

This'd allow us to manage a set of sponsors that'd display in the sidebar.

I'd probably prefer us to be able to do this, rather than extending the custom app I originally built for [the REST framework sidebar ads](https://www.django-rest-framework.org).

However I've obviously not got something right with the integration, because at the moment, I'm just seeing this...

Screenshot 2021-08-10 at 16 05 43

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,0 +1,5 @@
+{% extends "base.html" %}
+
+{% block extrahead %}
+  <script async src="https://media.ethicalads.io/media/client/ethicalads.min.js"></script>
+{% endblock %}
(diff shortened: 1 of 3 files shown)


ericholscher:
@tomchristie This was a server misconfiguration. Basically it was returning no ads because your publisher didn't have any ads set to show on it. I've fixed that now, and the test ad should be showing on it.

tomchristie:
Here's how sidebar ads with this change...

---

Screenshot 2021-08-12 at 11 52 50

PR #1795, opened 2021-08-11
Drop mode argument, 'httpx.Proxy(..., mode=...)'

tomchristie:
Discussed in https://github.com/encode/httpx/discussions/1783


tomchristie:
Let's crack on with this.

Issue #838, opened 2020-03-01
HTTPX AsyncClient slower than aiohttp?


imbolc:
I just found async-client to be significantly slower comparing to aiohttp, escecially for single request:

| |aiohttp | httpx | aiohttp/httpx |
|---|---|---|---|
|single, rps |1062 | 141 | 7.5 |
| session, rps |1862|1197 | 1.5 |
| session/single | 1.8 | 8.5 | |

I tried both release and master, results are pretty the same. The code of the benchmark is here: https://gist.github.com/imbolc/15cab07811c32e7d50cc12f380f7f62f

florimondmanca:
Thanks, this is very interesting quantified insights!

I went ahead and added the aiohttp/httpx ratio to your table.

I used your script to run the benchmark on my own machine, and observe similar results:

| | aiohttp | httpx | aiohttp/httpx |
|--|--|--|--|
| single, rps | 220 | 48 | 4.6 |
| client, rps | 495 | 331 | 1.5 |

florimondmanca:
The `client/single` ratio for HTTPX is not surprising to me — we know that using a client significantly increases performance.

The `aiohttp/httpx` ratio in the `client` case isn't surprising either — I had already noted that we were slower than `aiohttp` in the past (don't think I posted the results on GitHub though).

What's more surprising to me is, as you said, the 3x higher `aiohttp/httpx` ratio in the `single` case. I interpret it as "setting up a client (or opening a single connection, or whatever) is not as efficient as it could be".

I'll run [httpxprof](https://github.com/florimondmanca/httpxprof) over your scripts and see what other insights I can come up with. :-)

florimondmanca:
Okay, so apparently one _massively useless_ thing we do is **eager TLS setup on `Client` instantiation**, _even though none of the eventually requested URLs uses HTTPS_.

If we run:

```
pip install -e git+https://github.com/florimondmanca/httpxprof.git#egg=httpxprof
httpxprof run async_single
httpxprof view async_single
```

We get this view:

![Screenshot 2020-03-01 at 13 50 35](https://user-images.githubusercontent.com/15911462/75625991-b31d1780-5bc3-11ea-9d08-e7986025c05d.png)

=> 92% (!) of the time spent instantiating the client (which itself is 62% of the total time making a single request with a client) is spent setting up SSL (`.load_ssl_context()`).

[...]

PR #1797, opened 2021-08-12
1.0 announcement / funding page

tomchristie:
Starting to flesh out a 1.0 announcement, call for sponsors, and roadmap for future development...


docs/funding.md
```diff
@@ -0,0 +1,53 @@
...
+
+* Documentation that guides users through every aspect of the HTTPX library, and how it all fits together, while being accessible to both newcomers and experienced developers.
+* Ensuring that any error messages guide users towards resolution with clarity and precision.
+* Every object within HTTPX ought to provide informative `__repr__` messages, helping developers fully understand the state of any instances they're working with.
+* Comprehensive logging that will allow users to dig into a fuller understanding of every part of the HTTP request/response cycle.
+* Exposing a low level networking API, allowing developers to drop into the library at whatever level of detail and precision is required. This will allow the development of features such as record/replay of network traffic, customisation of DNS lookups, and alternate connection pooling strategies.
+
+### A sustainable and transparent business model
+
+Throughout the project we'll be issuing weekly progress reports, to give your businesses confidence that
```
florimondmanca, line 51: End of sentence missing? :)

florimondmanca (commented):
Quick review from me. That's an exciting perspective!

docs/funding.md
tomchristie: Yeah... it's still in progress in places. 😅

tomchristie:
> That's an exciting perspective!

Certainly helps to lay out the direction, eh? Helpful for me too, since I can see it more clearly, and get priorities in order.

docs/funding.md
```diff
@@ -0,0 +1,53 @@
...
+* Fully supports all the features that the `requests` package provides.
+* Both HTTP/1.1 and HTTP/2 support.
+* Strict request timeouts enabled by default.
+* Support for the asyncio and trio concurrency environments.
+* A fully type annotated interface, with 100% code coverage.
+* Development & maintenance that's backed by a sustainable and transparent business model.
+
+HTTPX is the **only** Python HTTP client with **HTTP/1.1 and HTTP/2** support, as well as being the only client providing both **sync and async** support. We'd like to continue pushing the project forward, and we're asking for sponsors to help us in building the foundation of modern HTTP stack for Python.
+
+We've spent a long time finessing the API for the project, and we're now happy to release a 1.0 version, and provide a standard SEMVER policy on future API updates.
```
graingert, line 15: I think you should flesh out the stability policy per https://docs.twistedmatrix.com/en/latest/core/development/policy/compatibility-policy.html

and include a separate policy for type hints, and behaviour under adverse conditions such as cancellation

docs/funding.md
```diff
@@ -0,0 +1,53 @@
...
+
+### HTTPX 1.2
+
+This version will add an integrated command line client to HTTPX, allowing for developers to easily switch between debugging HTTP requests from the console, and within their own codebases. You'll be able to switch easily between the two, with a consistent interface onto both sets of tools.
+
+### HTTPX 1.3 & beyond...
+
+Resource limits for the total download time, and maximum allowed download size. Support for WebSockets which run transparently over either HTTP/1.1 or HTTP/2. Documentation for using advanced functionality such as HTTP/2 bi-directional streaming, or working with network streams established using CONNECT or Upgrade requests. Support for a URLLib3 backend, which will aid projects that want upgrade from requests but still keep the underlying networking changes to an absolute minimum.
```
graingert, line 37: websockets don't run over HTTP/1.1

graingert: and they *can't* run over HTTP/2 https://daniel.haxx.se/blog/2016/06/15/no-websockets-over-http2/

graingert: on HTTP 3 [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) should out-perform websockets

there also doesn't seem to be any interest in websockets over http3. I suspect because on HTTP 3 [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) should out-perform websockets.

The new developments are on https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3 that adds "websocket"-like transports with multi homing and datagram support.

docs/funding.md
graingert: you should link to the reports: https://www.encode.io/reports/

docs/funding.md
```diff
@@ -0,0 +1,53 @@
+# Announcing HTTPX 1.0
+
+HTTPX is a fully featured HTTP client for Python, which offers both HTTP/1.1 and HTTP/2 support. Learning from all of the functionality and ease of use that the fantastic `requests` library provides, while bringing additional features and keeping pace with HTTP as it evolves.
+
+* An API that is broadly compatible with `requests`.
+* Fully supports all the features that the `requests` package provides.
+* Both HTTP/1.1 and HTTP/2 support.
+* Strict request timeouts enabled by default.
+* Support for the asyncio and trio concurrency environments.
```
graingert, line 9: burying the lede here. are you dropping crucio?

docs/funding.md
graingert:
```suggestion
Resource limits for the total download time, and maximum allowed download size. Support for WebSockets which run transparently over either HTTP/1.1 or HTTP/2. Documentation for using advanced functionality such as HTTP/2 bi-directional streaming, or working with network streams established using CONNECT or Upgrade requests. Support for a urllib3 backend, which will aid projects that want upgrade from requests but still keep the underlying networking changes to an absolute minimum.
```

njsmith: That 2016 blog post is out of date. The websockets-over-http/2 spec is RFC 8441.

yjqiang:
Maybe something about the performance?
[yes, we're maybe 2x-3x slower than aiohttp, but getting up to speed there isn't a priority for 1.0](https://github.com/encode/httpx/issues/838#issuecomment-598919546)

tomchristie:
@yjqiang I don't think the phrasing there is properly indicative of the differences, but *yes* we ought to include DNS caching for the 1.3 release.

docs/funding.md
graingert: ah indeed it's shipping in chrome and firefox already https://www.chromestatus.com/feature/6251293127475200

tomchristie:
Some thoughts here:

* Scope out 1.3 so that it's not open-ended. Have a properly defined end-goal for the project. "This much functionality and no more planned beyond that". Doesn't mean with might not *potentially* accept some additions as we go, but having a well defined scope has got to be a good thing for the project.
* We could feasibly switch around 1.1 and 1.2 here depending on how we consider each of them as priorities.

docs/funding.md
graingert: so I'm reading this as websockets can run on their own "compatible with http 1.1" using "http Upgrade" (as opposed to tunnelled 'over http 1.1'), RFC 8441 websockets however *are* tunnelled over HTTP 2 ?

tomchristie: Well, it'd probably be a little clearer to say "negotiated transparently over HTTP/1.1 or HTTP/2".

The idea is that upgrading to a raw network read/write API would work equally well against either HTTP/1.1 or HTTP/2, without the developer seeing any difference.  Against HTTP/1.1 those read/write operations would map straight onto the TCP connection, but with HTTP/2 they'd map onto a single HTTP/2 stream.

We can achieve this without adding any extra API methods by exposing the stream as an extension.

```python
with client.get(..., headers={"Upgrade": ...}) as response:
    if response.status_code != 101:
        ...
    stream = response.extensions["stream"]
    ...
```

My feeling is that we probably want to tightly scope things, so that we *show* how HTTPX can be used at this kind of level of API, without necessarily also building a full websockets API on top of HTTPX. (We can leave that to other packages to build on top of.)

PR #1799, opened 2021-08-13
Update HTTP/2 support in `extras_require`

### LINKED ISSUE #1798: Update h2 pin?
### Discussed in https://github.com/encode/httpx/discussions/1485

<sup>Originally posted by **HarrySky** February 24, 2021</sup>
Hi, some time ago `h2` pin was updated in `httpcore`: https://github.com/encode/httpcore/pull/208

But it is still pinned to `3.*` in this package's `http2` extra:
https://github.com/encode/httpx/blob/0f280af8b170ed5cc48c12a894f71a8b5762f748/setup.py#L65

This is not an issue, as I can just add `h2>=3,<5` to my setup.py instead of using `httpx[http2]`, but maybe you want dependencies to be in sync with `httpcore`.

EDIT: Using git blame we can see that before `http2` extra - `httpcore[http2]` was used instead of `h2` dependency directly.</div>

tomchristie:
So, `httpx` itself doesn't have any specific requirements on which `h2` version should be installed.
That's actually up to `httpcore`.

This pull request updates the `extras_require` to reflect that.

*However*... I'm a bit wary that pip's dependency resolution *might* not actually get this right, so this pull request is a bit exploratory. 

Closes #1798

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ setup(
         "async_generator; python_version < '3.7'"
     ],
     extras_require={
-        "http2": "h2==3.*",
+        "http2": "httpcore[http2]",
         "brotli": "brotlicffi==1.*",
     },
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ setup(
         "async_generator; python_version < '3.7'"
     ],
     extras_require={
-        "http2": "httpcore[http2]",
+        "http2": "h2>=3,<5",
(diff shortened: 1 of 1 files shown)


tomchristie:
Okay, I don't trust that `httpcore[http2]` is going to do the right thing here. Does appear to work on the CI installs. Doesn't appear to work on my local install. May well behave differently under different versions of `pip`.

This approach is at least sufficient. If not all that graceful.

PR #1806, opened 2021-08-17
Switch event hooks to also run on redirects.

tomchristie:
Possible change in behaviour, based on discussion #1804

We'd also want to switch up the documentation to make it clear that `response.read()` needs to be called in an event hook if you want access to the body there.


tomchristie:
Resolved via #1807

PR #1807, opened 2021-08-18
Resolve Python 3.6 tests hanging.

tomchristie:
Test cases on Python 3.6 are hanging, both when running locally, and on CI.

Locally I ran...

```python
scripts/install -p python3.6
venv/bin/pytest --full-trace
```

in order to track down where the tests were hanging. Issue was when waiting for the uvicorn server to be available.

Pinning to Uvicorn 0.14.0 appears to resolve the issue.


tomchristie:
Right, Python 3.6 builds are hanging - unclear why.

PR #1808, opened 2021-08-18
Switch follow redirects default

tomchristie:
Implementation based on discussion #1785.

**Switches the auto-redirect behaviour to off by default**.

The benefit of this change would be to make it more explicit when redirects should be followed, and avoid code that unintentionally issues multiple requests, as a result of not requesting the correct URL.

We do still allow auto-redirects, but only if explicitly enabled...

```python
client.get(url, follow_redirects=True)
```

Or, by enabling a client-wide default...

```python
client = httpx.Client(follow_redirects=True)
```

Also note that we change `allow_redirects` to `follow_redirects`.

This is a **big change**, but I think it may well be one of those points where it's worth us diverging from requests.

It'd also play more neatly into the behaviour that we'd probably want by default from a command-line client, which [I'd really like to see on our roadmap](https://github.com/encode/httpx/pull/1797).


florimondmanca:
I think I like the "explicit better than implicit" vibe to this. At the same time I wonder about usability. But isn't there precedent to this anyway? Doesn't cURL ignore redirects and return 3xx by default too? I think so?

httpx/_api.py
```diff
@@ -34,7 +34,7 @@ def request(
     auth: AuthTypes = None,
     proxies: ProxiesTypes = None,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
-    allow_redirects: bool = True,
+    follow_redirects: bool = False,
```
florimondmanca, line 37: I assume we don't consider a soft deprecation (with a warning) to be worth it here, given the amount of code that would require?

florimondmanca (commented):
I'm enjoying the clearer follow_* name as well.

tomchristie:
This feels potentially nice, but we need to be *careful*.

Eg. Potentially it changes what expectations you might want on `raise_for_status()`.

```python
r = httpx.get(...)
r.raise_for_status()  # If 3xx responses *aren't* handled by default, then you might want this to raise exceptions for 3xx cases.
data = r.json()
```

PR #1809, opened 2021-08-18
Version 0.19.0

tomchristie:
So... I guess here's the lowdown.

Final set of changes, that we'd release in a 0.19.0 version, and then promote to a 1.0 in perhaps 2 weeks time. (I'm away on holiday this following week 23rd-26th, so getting these changes out a few days prior to that would seem to be a good idea.)

- [ ] ~Also requires https://github.com/encode/httpx/pull/1808~
- [ ] ~Might want to document the redirect differences from requests more prominently?~
- [x] Need to update the `__version__`.
- [x] Need to update the docs/README pin notice.

Regarding 1.0... I'm not 100% convinced that the transport API is necessarily absolutely definitively the most favoured option, but it seems silly at this point to keep blocking everything indefinitely. It's a little verbose, but it *is* a nice cut-off between the two slices of functionality.

There's a really [nice roadmap ahead](https://github.com/encode/httpx/pull/1797), and we should probably call it at this point, so that we can push on with:

* A revised `httpcore` implementation that starts to really neatly expose some lower-level connection pooling details, as well as fixing a couple of gnarly implementation issues we currently have.
* An integrated `httpx` command-line client, which would just be a fantastic piece of functionality to offer, for *very*, *very* little additional code.
* Future functionality, such as resource limits on total download size & download timeouts, websocket APIs, good support for CONNECT & Upgrade functionality, etc.etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-* Add support for `Client(allow_redirect=<bool>)`. (Pull #1790)
+* Add support for `Client(allow_redirects=<bool>)`. (Pull #1790)
 * Add automatic character set detection, when no `charset` is included in the response `Content-Type` header. (Pull #1791)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+* Event hooks are now also called for any additional redirect or auth requests/responses. (Pull #1806)
 * Strictly enforce that upload files must be opened in binary mode. (Pull #1736)
-* Strictly enforce that client instances can only be opened and closed once, and cannot be re-opened. (#1800)
+* Strictly enforce that client instances can only be opened and closed once, and cannot be re-opened. (Pull #1800)
 * Drop `mode` argument from `httpx.Proxy(..., mode=...)`. (Pull #1795)
 
 ## 0.18.2 (17th June, 2021)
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Latest
+
+### Added
+
+* Add support for `Client(allow_redirect=<bool>)`. (Pull #1790)
```
florimondmanca, line 11: Typo: Is it allow_redirect**S**?

florimondmanca (commented):
Changelog looking good 👍

CHANGELOG.md
tomchristie: It is indeed! That one got added prior to https://github.com/encode/httpx/pull/1808. 😬

tomchristie:
So... #1808 is probably way to big of a behavioural change to rush into.

I figure we may as well get this out the door now as 0.19.0, and just have a little more time to mull the last bits of *how do we really, really want the 1.0 API to look?*.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.19.0 (19th June, 2021)
```
terencehonles, line 7: @tomchristie should this be August 18th/19th (not sure your timezone)? I'm slightly surprised to see something mentiong a couple months ago, but it's been released ~9hours ago.

tomchristie: 🙃

florimondmanca: Oopsie — looks like that got merged without a change?

terencehonles: Thanks for updating, I believe the GitHub release title can also be updated here https://github.com/encode/httpx/releases/tag/0.19.0

tomchristie: Thanks @terencehonles!

PR #1821, opened 2021-08-27
Setup Windows CI

da1910:
Work in progress for #458 

Three tests were broken:
1. test_utils:test_get_ssl_cert_file assumed paths were always unix-like, fixed by splitting path using os.path
2. test_timeouts:test_write_timeout[asyncio] fails - it never throws the timeout
3. test_async_client:test_get[asyncio] fails - It reports that the request took 0 microseconds to complete

I am unsure what is causing 2 or 3, I would welcome any suggestions.


tomchristie:
Interesting thanks! Are you able to add exclusions of windows for those two test cases?
I'm not sure we'll want to double the test matrix like this - it's an awfully big hit to take.

da1910:
> Interesting thanks! Are you able to add exclusions of windows for those two test cases?
> I'm not sure we'll want to double the test matrix like this - it's an awfully big hit to take.

Certainly, I'll also fix the typing issue that's causing the failures in test_util.

da1910:
I've excluded the two failing tests, but couldn't work out how to exclude from only one async backend. This also means that the coverage checks are failing.

I can put nocover marks on the tests in question, if we're happy with the exclusions.

PR #1823, opened 2021-08-28
Add dependabot

Kludex:
This was discussed on #1721 and #1680.

This PR adds the native [GitHub Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates), and also introduces the [Automerge bot](https://github.com/fastify/github-action-merge-dependabot).

Merging this PR is **not enough** to enable both bots. You need to enable Dependabot on `Settings` -> `Security` and install the app [here](https://github.com/apps/dependabot-merge-action) for the Automerge bot.


Kludex:
In case we need time to discuss the Automerge bot, I can update this PR with only Dependabot.

tomchristie:
> In case we need time to discuss the Automerge bot, I can update this PR with only Dependabot.

@Kludex Yeah - let's do things that way around. Automerge may *well* be an obvious merge too, but I don't know yet. Always a good tactic to keep PRs as tightly targeted as possible.

PR #1828, opened 2021-08-31
BUG: wsgi.error should be TextIO, not BytesIO in WSGI transport

adriangb:
I found this from here: https://github.com/falconry/falcon/blob/e255bff9ae5a90d0cb3fe9af7c16917f18a92dc3/falcon/request.py#L1879

Based on the documentation at https://modwsgi.readthedocs.io/en/master/user-guides/debugging-techniques.html#apache-error-log-files, I believe Falcon is right and HTTPX is wrong.

This PR:
1. Adds a parameter so that we can test it and users can easily collect the logs in their tests if they want.
2. Changes the default from an inaccessible `BytesIO` (wrong type of stream, can't be accessed by user) to `sys.stderr` so that users (or pytest automatically) will capture the logs.

### DIFF
--- a/httpx/_transports/wsgi.py
+++ b/httpx/_transports/wsgi.py
@@ -89,7 +89,7 @@ class WSGITransport(BaseTransport):
             "wsgi.version": (1, 0),
             "wsgi.url_scheme": scheme.decode("ascii"),
             "wsgi.input": wsgi_input,
-            "wsgi.errors": io.BytesIO(),
+            "wsgi.errors": io.StringIO(),
             "wsgi.multithread": True,
             "wsgi.multiprocess": False,
             "wsgi.run_once": False,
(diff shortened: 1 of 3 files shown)


httpx/_transports/wsgi.py
```diff
@@ -62,11 +63,13 @@ def __init__(
         raise_app_exceptions: bool = True,
         script_name: str = "",
         remote_addr: str = "127.0.0.1",
+        log_file: typing.Optional[typing.TextIO] = None,
     ) -> None:
         self.app = app
         self.raise_app_exceptions = raise_app_exceptions
         self.script_name = script_name
         self.remote_addr = remote_addr
+        self.log_file = log_file
```
tomchristie, line 72: I think `wsgi_errors` would be more consistent here, right?
Loosely matching `script_name` and `remote_addr`.

adriangb: Happy to change it!

adriangb: Donzo

tomchristie:
Neat little, addition, yup! Thanks!

PR #1839, opened 2021-09-03
Fix conditional imports for pyright

RobertCraigie:
This PR fixes an issue I encountered due to conditional imports and type checking with pyright.

- [x] Initially raised as discussion #1829


tomchristie:
Yup. Thank you!

PR #1866, opened 2021-09-14
Fix console markup escaping issue

willmcgugan:
Escapes the download name so that it won't render console markup.

- [X] Initially raised as discussion https://github.com/encode/httpx/discussions/1865


tomchristie:
Thanks @willmcgugan!

PR #1904, opened 2021-10-20
Map UnsupportedProtocol to InvalidURL

### LINKED ISSUE #1832: Inconsistent Handling of Invalid Urls
If I call: `httpx.get('https://😇')`, an `InvalidURL` exception is raised. However, if I call: `httpx.get('https:/google.com')`, instead I get an `UnsupportedProtocol` exception, which seems inconsistent. I would expect the latter to raise an `InvalidURL` as well. For reference, requests raises a `InvalidURL` in _both_ cases.

aminalaee:
Fixes #1832 .

Updates mapping to raise InvalidURL for UnsupportedProtocol exceptions from httpcore.
Not sure if this is the best approach.

Any feedback is appreciated.


aminalaee:
@tomchristie Agreed. This will basically omit UnsupportedProtocol.

PR #1909, opened 2021-10-23
Replace `yield` inside for loop with `yield` from

yezz123:
Refactor Code Expression, and Replace `yield` inside for loop with `yield` from.

### DIFF
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -48,8 +48,7 @@ class IteratorByteStream(SyncByteStream):
             raise StreamConsumed()
 
         self._is_stream_consumed = True
-        for part in self._stream:
-            yield part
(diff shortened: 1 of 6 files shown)


tomchristie:
I'm going to close this pull request because there's too many different bits and pieces all mixed in together. There's some stuff in here that we might accept, and some stuff that we might not, but we'd need to make those assessments individually, rather than against a single PR.

yezz123:
I will work on it @tomchristie by the end of the week as a single PRs

tomchristie:
Okay. I'd suggest just start with one, and we could see from there. That way you don't have to invest too much time before getting feedback. (I mean we *might* not want to accept non-behaviour changing refractors at all)

PR #1916, opened 2021-10-31
Expand docs note for async custom handler responses

hexarobi:
Custom response handlers need to run `response.read()` before they can read the content of the response. However when using an AsyncClient this will produce an error of `RuntimeError: Attempted to call a sync iterator on an async stream.`. Took me some digging to figure out I just needed to use `response.aread()` here instead of `response.read()` so figured I would an MR with an expansion on the note for anyone else.
Thanks!


tomchristie:
Okay, sure. Thanks.

PR #1922, opened 2021-11-01
Update documentation to outline that binary content is also excluded in certain requests

symonk:
Evening, hope you are all doing well.  While having a little study through the docs and code, I noticed that in some requests the `json`, `data` and `files` arguments are well documented when they are excluded, however binary content via `content=` for the request body is not, this patch updates the documentation to outline that.

Thanks again for the time to review these changes and your time on the library in general.

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -183,8 +183,8 @@ def get(
 
     **Parameters**: See `httpx.request`.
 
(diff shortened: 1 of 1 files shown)


tomchristie (approved):
Good catch. Thank you. 🙏🏼

PR #1936, opened 2021-11-12
Allow custom headers in multipart/form-data requests

adriangb:
Compliment to https://github.com/encode/starlette/pull/1311

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -85,17 +85,14 @@ class FileField:
         # It would be good to think of better APIs for this that we could include in httpx 2.0
         # since variable length tuples (especially of 4 elements) are quite unwieldly
         if isinstance(value, tuple):
-            try:
-                filename, fileobj, content_type, headers = value  # type: ignore
-            except ValueError:
-                # 4th parameter (headers) not included
-                try:
-                    filename, fileobj, content_type = value  # type: ignore
-                except ValueError:
-                    # neither the 3rd parameter (content_type) nor the 4th (headers) was included
-                    filename, fileobj = value  # type: ignore
+            if len(value) == 2:
+                # neither the 3rd parameter (content_type) nor the 4th (headers) was included
+                filename, fileobj = value  # type: ignore
+            elif len(value) == 3:
+                filename, fileobj, content_type = value  # type: ignore
             else:
-                # corresponds to (filename, fileobj, content_type, headers)
+                # all 4 parameters included
+                filename, fileobj, content_type, headers = value  # type: ignore
                 headers = {k.title(): v for k, v in headers.items()}
         else:
             filename = Path(str(getattr(value, "name", "upload"))).name
@@ -107,7 +104,7 @@ class FileField:
         if content_type is not None and "Content-Type" not in headers:
             # note that unlike requests, we ignore the content_type
             # provided in the 3rd tuple element if it is also included in the headers
-            # requests does the opposite
+            # requests does the opposite (it overwrites the header with the 3rd tuple element)
             headers["Content-Type"] = content_type
 
         if isinstance(fileobj, (str, io.StringIO)):
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -81,20 +81,22 @@ class FileField:
         headers: typing.Dict[str, str] = {}
         content_type: typing.Optional[str] = None
 
+        # This large tuple based API largely mirror's requests' API
+        # It would be good to think of better APIs for this that we could include in httpx 2.0
+        # since variable length tuples (especially of 4 elements) are quite unwieldly
         if isinstance(value, tuple):
             try:
                 filename, fileobj, content_type, headers = value  # type: ignore
             except ValueError:
+                # 4th parameter (headers) not included
                 try:
                     filename, fileobj, content_type = value  # type: ignore
                 except ValueError:
+                    # neither the 3rd parameter (content_type) nor the 4th (headers) was included
                     filename, fileobj = value  # type: ignore
             else:
+                # corresponds to (filename, fileobj, content_type, headers)
                 headers = {k.title(): v for k, v in headers.items()}
-                if "Content-Type" in headers:
-                    raise ValueError(
-                        "Content-Type cannot be included in multipart headers"
-                    )
         else:
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
@@ -102,7 +104,10 @@ class FileField:
         if content_type is None:
             content_type = guess_content_type(filename)
 
-        if content_type is not None:
+        if content_type is not None and "Content-Type" not in headers:
+            # note that unlike requests, we ignore the content_type
+            # provided in the 3rd tuple element if it is also included in the headers
+            # requests does the opposite
             headers["Content-Type"] = content_type
 
         if isinstance(fileobj, (str, io.StringIO)):
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -97,11 +97,12 @@ class FileField:
         else:
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
-
+        
         if content_type is None:
             content_type = guess_content_type(filename)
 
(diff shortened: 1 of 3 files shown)


httpx/_multipart.py
```diff
@@ -78,23 +78,39 @@ def __init__(self, name: str, value: FileTypes) -> None:
...
+                try:
+                    filename, fileobj, content_type = value  # type: ignore
+                except ValueError:
+                    filename, fileobj = value  # type: ignore
+            else:
+                headers = {k.title(): v for k, v in headers.items()}
+                if "Content-Type" in headers:
+                    raise ValueError(
+                        "Content-Type cannot be included in multipart headers"
+                    )
```
tomchristie, line 97: Perhaps we don't need to do this check.

It's *odd* behaviour for the developer to set the `content_type` and then override it with the *actual* value provided in the custom headers. But it's not *broken*.

My preference would be that we don't do the explicit check here. In the case of conflicts I'd probably have `header` values take precedence.

I'm not absolute on this one, but slight preference.

adriangb: Makes sense. I thought it'd be a good idea to check what requests does here. It looks like it silently ignores the header in the `header`. That is:

```python
requests.post("http://httpbin.org/post", files=[("test", ("test.txt", b"data", "text/plain", {"Content-Type": "text/csv"}))])
```

Gets sent as `text/plain`.

Digging into _why_ this is the case, it seems like it's just an implementation detail in urllib3. It happens [here](https://github.com/urllib3/urllib3/blob/c58dbbb35a19877a8f367c1eaa99ca43c2148f59/src/urllib3/fields.py#L351).

I'm not sure what the right thing to do here is, but if you feel like it's best to go with no error and making `header` values take precedence, I'm happy to implement that.

tomchristie: > I'm not sure what the right thing to do here is, but if you feel like it's best to go with no error and making header values take precedence, I'm happy to implement that.

I reckon let's do that, yeah.

> Another alternative would be to have the 3rd parameter be either a string representing the content type or a headers dict. We can't really make the 3rd parameter always be a headers dict because that would be a breaking change for httpx.

I actually quite like that yes, neat idea. The big-tuples API is... not helpful really. But let's probably just go with the path of least resistance here. Perhaps one day we'll want an `httpx` 2.0, where we gradually start deprecating the various big-tuples bits of API in favour of a neater style.

adriangb: > I reckon let's do that, yeah.

👍 gonzo

> > Another alternative would be to have the 3rd parameter be either a string representing the content type or a headers dict. We can't really make the 3rd parameter always be a headers dict because that would be a breaking change for httpx.
> 
> I actually quite like that yes, neat idea. The big-tuples API is... not helpful really. But let's probably just go with the path of least resistance here. Perhaps one day we'll want an `httpx` 2.0, where we gradually start deprecating the various big-tuples bits of API in favour of a neater style.

Agreed! I added a comment in the code explaining the reasoning behind the big tuple API (inherited from requests) and how we might want to change it in the future.

httpx/_multipart.py
```diff
@@ -78,23 +78,44 @@ def __init__(self, name: str, value: FileTypes) -> None:
...
-                content_type = guess_content_type(filename)
+                # 4th parameter (headers) not included
+                try:
+                    filename, fileobj, content_type = value  # type: ignore
+                except ValueError:
+                    # neither the 3rd parameter (content_type) nor the 4th (headers) was included
+                    filename, fileobj = value  # type: ignore
+            else:
+                # corresponds to (filename, fileobj, content_type, headers)
+                headers = {k.title(): v for k, v in headers.items()}
```
tomchristie, line 99: I don't think we should `.title()` case here.

httpx/_multipart.py
```diff
@@ -78,23 +78,44 @@ def __init__(self, name: str, value: FileTypes) -> None:
...
+                filename, fileobj, content_type, headers = value  # type: ignore
             except ValueError:
-                filename, fileobj = value  # type: ignore
-                content_type = guess_content_type(filename)
+                # 4th parameter (headers) not included
+                try:
+                    filename, fileobj, content_type = value  # type: ignore
+                except ValueError:
+                    # neither the 3rd parameter (content_type) nor the 4th (headers) was included
+                    filename, fileobj = value  # type: ignore
```
tomchristie, line 96: An alternative would be...

```python
# Unpack filename, fileobj, and optionally content_type and headers.
filename, fileobj, content_type, headers, *_ = value + (None, None)
```

Not sure if the more concise version is better or worse here.

Or we could do explicitly length-based...

```python
if len(value) == 2:
    filename, fileobj = value
elif len(value) == 3:
    filename, fileobj, content_type = value
else:
    filename, fileobj, content_type, headers = value
```

Looking at it now, I think the length based version is probably neatest, since it's just very clear and obvious. Marginally nicer than the various indents that try..catch requires.

httpx/_multipart.py
tomchristie: Ah... I see the comparison case. Huh. Fiddly.

httpx/_multipart.py
```diff
@@ -78,23 +78,44 @@ def __init__(self, name: str, value: FileTypes) -> None:
...
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
+
+        if content_type is None:
             content_type = guess_content_type(filename)
 
+        if content_type is not None and "Content-Type" not in headers:
+            # note that unlike requests, we ignore the content_type
+            # provided in the 3rd tuple element if it is also included in the headers
+            # requests does the opposite
```
tomchristie, line 110: Okay maybe we should instead do it the other way. If the 4-tuple is used, just ignore the `content_type` variable. That'd be okay enough, matches `requests` more closely, and we can forget about fiddly case-based header checking.

httpx/_multipart.py
adriangb: Yep yep. I guess I just took what was there and extended it. Obviously it doesn't scale, I like matching on the length more.

httpx/_multipart.py
adriangb: requests does the opposite: it ignores the header in the 4th tuple element. so we'll still need the case-based header checking if we want to do exactly what requests does. either way, we need to know if the content type header exists in the 4th element tuple so we can either ignore the 3rd element or overwrite it with the 3rd element.

httpx/_multipart.py
adriangb: Ok changed it to a length based comparison

tomchristie (commented):
I'm happy with this pull request *except* that I would rather we don't force-change the casing on the headers. That introduces a hidden little bit of behaviour surprise that I'd rather avoid.

Obvs we *do* still want to do a case-insensitive comparison for the `Content-Type` case tho.

httpx/_multipart.py
```diff
@@ -78,23 +78,41 @@ def __init__(self, name: str, value: FileTypes) -> None:
...
+                filename, fileobj, content_type, headers = value  # type: ignore
+                headers = {k.title(): v for k, v in headers.items()}
         else:
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
+
+        if content_type is None:
             content_type = guess_content_type(filename)
 
+        if content_type is not None and "Content-Type" not in headers:
```
tomchristie, line 104: Perhaps...

```python
has_content_type_header = any(["content-type" in key.lower() for key in headers])
if content_type is not None and not has_content_type_header:
    ...
```

?

adriangb: I adapted it to `any("content-type" in key.lower() for key in headers)` (so it'll stop early).
Also removed the `{header.title() ...}` line.

httpx/_multipart.py
```diff
@@ -97,11 +97,12 @@ def __init__(self, name: str, value: FileTypes) -> None:
         else:
             filename = Path(str(getattr(value, "name", "upload"))).name
             fileobj = value
-
+
```
tomchristie:
```suggestion

```

tomchristie:
Lovely stuff. 👍

PR #1943, opened 2021-11-18
Refactor `httpx/_utils.py`

yezz123:
Based on what @tomchristie recommended to make https://github.com/encode/httpx/pull/1909 this PR into a shorter one to validate multiple things and let it more clear and ready for review.

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -457,19 +457,18 @@ class URLPattern:
         self.port = url.port
         if not url.host or url.host == "*":
             self.host_regex: typing.Optional[typing.Pattern[str]] = None
+        elif url.host.startswith("*."):
+            # *.example.com should match "www.example.com", but not "example.com"
+            domain = re.escape(url.host[2:])
+            self.host_regex = re.compile(f"^.+\\.{domain}$")
+        elif url.host.startswith("*"):
+            # *example.com should match "www.example.com" and "example.com"
+            domain = re.escape(url.host[1:])
+            self.host_regex = re.compile(f"^(.+\\.)?{domain}$")
         else:
-            if url.host.startswith("*."):
-                # *.example.com should match "www.example.com", but not "example.com"
-                domain = re.escape(url.host[2:])
-                self.host_regex = re.compile(f"^.+\\.{domain}$")
-            elif url.host.startswith("*"):
-                # *example.com should match "www.example.com" and "example.com"
-                domain = re.escape(url.host[1:])
-                self.host_regex = re.compile(f"^(.+\\.)?{domain}$")
-            else:
-                # example.com should match "example.com" but not "www.example.com"
-                domain = re.escape(url.host)
-                self.host_regex = re.compile(f"^{domain}$")
+            # example.com should match "example.com" but not "www.example.com"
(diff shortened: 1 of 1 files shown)


httpx/_utils.py
```diff
@@ -457,19 +457,18 @@ def __init__(self, pattern: str) -> None:
...
-                # *example.com should match "www.example.com" and "example.com"
-                domain = re.escape(url.host[1:])
-                self.host_regex = re.compile(f"^(.+\\.)?{domain}$")
-            else:
-                # example.com should match "example.com" but not "www.example.com"
-                domain = re.escape(url.host)
-                self.host_regex = re.compile(f"^{domain}$")
+            # example.com should match "example.com" but not "www.example.com"
+            domain = re.escape(url.host)
+            self.host_regex = re.compile(f"^{domain}$")
```
tomchristie, line 471: I like the refactoring you've done on this block, yup. I think it makes enough of a readability difference that it's worth having in.

httpx/_utils.py
```diff
@@ -120,7 +120,7 @@ def guess_json_utf(data: bytes) -> typing.Optional[str]:
         if sample[1::2] == _null2:  # 2nd and 4th are null
             return "utf-16-le"
         # Did not detect 2 valid UTF-16 ascii-range characters
-    if nullcount == 3:
+    elif nullcount == 3:
```
tomchristie, line 123: Either we should update all the cases here that could be `elif`, or none of them.
We're not really getting any value from the change, so I'd suggest we just leave it.

```suggestion
    if nullcount == 3:
```

httpx/_utils.py
```diff
@@ -23,7 +23,7 @@
     {chr(c): "%{:02X}".format(c) for c in range(0x1F + 1) if c != 0x1B}
 )
 _HTML5_FORM_ENCODING_RE = re.compile(
-    r"|".join([re.escape(c) for c in _HTML5_FORM_ENCODING_REPLACEMENTS.keys()])
+    r"|".join([re.escape(c) for c in _HTML5_FORM_ENCODING_REPLACEMENTS])
```
tomchristie, line 26: I could go either way on this one. Possibly it's more obvious with `.keys()` what the effect is. Possibly not. But, okay, sure.

tomchristie:
Okay, I've reverted one `elif` change that was just dangling on its own.
The other two changes make sense.

yezz123:
> Okay, I've reverted one `elif` change that was just dangling on its own. The other two changes make sense.

@tomchristie that's great, if there is any other things we can discuss about it and fix it 🚀

httpx/_utils.py
tomchristie, line 26:
```suggestion
    r"|".join([re.escape(c) for c in _HTML5_FORM_ENCODING_REPLACEMENTS.keys()])
```

tomchristie:
I've reverted the minor change to `_HTML5_FORM_ENCODING_RE` - it's unrelated to the `elif` block changes.

Given that I think *those* are a clear win, let's *just pull those in* as a single nicely scoped pull request.

tomchristie (approved):
Nice little improvement to the flow there, thank you!

PR #1945, opened 2021-11-18
Replaces `yield` as part of `for` loops with `yield from`

yezz123:
Based on what @tomchristie recommended to make https://github.com/encode/httpx/pull/1909 this PR into a shorter one to validate multiple things and let it more clear and ready for review.

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -341,7 +341,7 @@ class Proxy:
         return f"Basic {token}"
 
     def __repr__(self) -> str:
-        return f"Proxy(url={str(self.url)!r}, headers={dict(self.headers)!r})"
+        return f'Proxy(url={self.url!r}, headers={dict(self.headers)!r})'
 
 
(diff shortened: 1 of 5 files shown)


tomchristie:
So, I don't really want to use `yield from` within our codebase, because there's quite a lot of places where there's also an async version that mirrors the sync one, and we're not able to do `async yield from`.

I'd rather we ensure that the two-cases continue to neatly mirror each other.

yezz123:
> So, I don't really want to use `yield from` within our codebase, because there's quite a lot of places where there's also an async version that mirrors the sync one, and we're not able to do `async yield from`.
> 
> I'd rather we ensure that the two-cases continue to neatly mirror each other.

So, what you think to remove the `yield from` and let only the other changes?

tomchristie:
I'd be up for considering isolated pull requests addressing separate areas.
As it currently stands there's too many different concerns being tackled here.

PR #1946, opened 2021-11-19
readmes/README.scn.md

2601677867:

- [ yes] Initially raised as discussion #...

Hello, I suggest using readme.scn.md, because Chinese is divided into simplified Chinese and traditional Chinese. People who use traditional Chinese may not understand some words in simplified Chinese.


tomchristie:
Thanks! Let's start by getting this merged in. We can figure out how we want to do nice interlinking between the different available languages as a follow-up.

Other PRs here also very welcome.

PR #1983, opened 2021-12-27
Fix/lost chunk size

iamnotaprogrammer:
add params  chunk_size to iter_raw


iamnotaprogrammer:
#1982

httpx/_models.py
```diff
@@ -1577,7 +1577,7 @@ def iter_bytes(self, chunk_size: int = None) -> typing.Iterator[bytes]:
             decoder = self._get_content_decoder()
             chunker = ByteChunker(chunk_size=chunk_size)
```
tomchristie, line 1578: Passing `chunk_size` through to `iter_raw()` *isn't* the right behaviour, no.
We're already handling the chunk_size correctly here.

tomchristie:
There isn't a broken behaviour here - the `chunk_size` is already respected.
(Implementation-wise it's passed to the correct place, rather than being passed through to the `iter_raw()` method, which isn't what we want.)

PR #2016, opened 2022-01-06
Fix for stream uploads that subclass SyncByteStream/AsyncByteStream

tomchristie:
Closes #2013


hugovk:
In https://github.com/encode/httpx/releases/tag/0.21.3, should "Regression in 0.12.2" read "Regression in 0.21.2"?

tomchristie:
Good spot @hugovk - thanks.

PR #2019, opened 2022-01-08
Drop mention of backend selection for AsyncHTTPTransport() in docs

dlax:
There is no 'backend' parameter to AsyncHTTPTransport and it seems that
the backend is detected automatically for anyio as it is for other async
libraries.


tomchristie:
Thanks, yup. 👍

PR #2026, opened 2022-01-12
Dont perform implicit close/warning on `__del__`

### LINKED ISSUE #2024: httpx\_client.py", line 1991, in __del__ AttributeError: 'NoneType' object has no attribute 'OPENED'
The starting point for issues should usually be a discussion...

https://github.com/encode/httpx/discussions

Possible bugs may be raised as a "Potential Issue" discussion, feature requests may be raised as an "Ideas" discussion. We can then determine if the discussion needs to be escalated into an "Issue" or not.

This will help us ensure that the "Issues" list properly reflects ongoing or needed work on the project.

---

Sender Runtime Finished ...
[<Response [200 OK]>, <Response [200 OK]>, <Response [200 OK]>, <Response [200 OK]>]
Exception ignored in: <function AsyncClient.__del__ at 0x000001C2BCC04A60>
  File "C:\ANACONNA\lib\site-packages\httpx\_client.py", line 1991, in __del__
AttributeError: 'NoneType' object has no attribute 'OPENED'

__ If I Delete The  \_\_del\_\_ function in _client.py , my code can work without error, but if not ,the error above will just happend..
And my function can get The coroutine data return , but I still don know why this error happend? 
Is Python garbage Recycle effect this ? __

And The  \_\_del\_\_  code is 

```
    def __del__(self) -> None:
        # We use 'getattr' here, to manage the case where '__del__()' is called
        # on a partially initiallized instance that raised an exception during
        # the call to '__init__()'.
        if getattr(self, "_state", None) == ClientState.OPENED:  # noqa: B009
            # Unlike the sync case, we cannot silently close the client when
            # it is garbage collected, because `.aclose()` is an async operation,
            # but `__del__` is not.
            #
            # For this reason we require explicit close management for
            # `AsyncClient`, and issue a warning on unclosed clients.
            #
            # The context managed style is usually preferable, because it neatly
            # ensures proper resource cleanup:
            #
            #     async with httpx.AsyncClient() as client:
            #         ...
            #
            # However, an explicit call to `aclose()` is also sufficient:
            #
            #    client = httpx.AsyncClient()
            #    try:
            #        ...
            #    finally:
            #        await client.aclose()
            warnings.warn(
                f"Unclosed {self!r}. "
                "See https://www.python-httpx.org/async/#opening-and-closing-clients "
                "for details."
            )


```

tomchristie:
- Prompted by discussion #1951
- Closes #2024

We currently attempt to perform a bit of hand-holding for cases where a client instance has not been closed at the point it is deleted. The reasoning here was to try to give users a bit more of a guard against unintentional resource leaks.

Our behaviour here was...

* For sync clients, automatically `.close()` clients on `__del__`, if they're still open.
* For async clients, issue a warning on `__del__`, if they're still open.

As it turns out, this *just isn't reliable*. We really can't do this properly, because if `__del__` is called at `atexit` then there's all sorted of bit of stdlib that may not be available. We can't reliably call into `warnings.warn`, and it turns out we can't even reliably access the properties on an `intenum` class. Geez.

So. Let's just *not do this anymore*.

If you're leaking client instances (which as it happens, zillions of `requests` and `urllib3` using codebases currently already do.) Then okay, that's just the way it is. I'm not sure what flags you'd need to enable in python to have the system warn you about when that's the case, but we should rely on that level of warning, rather than attempting to add any ourselves.

Better that we stick with the current status-quo, than attempt to do better and actually just end up with user-hostile exceptions or undefined behaviours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.21.3 (6th January, 2022)
+
+### Fixed
+
+* Fix streaming uploads using `SyncByteStream` or `AsyncByteStream`. Regression in 0.12.2. (#2016)
+
 ## 0.21.2 (5th January, 2022)
 
 ### Fixed
--- a/httpx/__version__.py
+++ b/httpx/__version__.py
@@ -1,3 +1,3 @@
 __title__ = "httpx"
 __description__ = "A next generation HTTP client, for Python 3."
-__version__ = "0.21.2"
+__version__ = "0.21.3"
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1273,13 +1273,6 @@ class Client(BaseClient):
             if transport is not None:
                 transport.__exit__(exc_type, exc_value, traceback)
 
-    def __del__(self) -> None:
-        # We use 'getattr' here, to manage the case where '__del__()' is called
-        # on a partially initiallized instance that raised an exception during
-        # the call to '__init__()'.
-        if getattr(self, "_state", None) == ClientState.OPENED:  # noqa: B009
-            self.close()
-
 
 class AsyncClient(BaseClient):
     """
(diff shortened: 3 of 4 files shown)


samuelcolvin:
Nice explanation, thank you!

PR #2038, opened 2022-01-21
Fix Headers.update to correctly handle repeated headers

adriangb:
These two tests were failing beforehand


tomchristie (approved):
Took a little while to convince myself, but yes.
Very nice, thank you @adriangb.

tomchristie:
Retitled issue very slightly for neatness & consistency.
We may as well avoid "prefix:" kinda styles.

PR #2044, opened 2022-01-23
Do not recommend the deprecated `data=` kwarg for passing bytes

vytas7:
AFAICT `data=` is meant for sending form data, and deprecated for passing bytes per https://www.python-httpx.org/compatibility/#request-content.

I've grepped through other occurrences in `docs/` and it seems this is the only case of misuse of the deprecated `data=` interface (i.e., where replaceable with `content=`).


j178 (approved):

florimondmanca (approved):
Thanks, that does look like an old flake :-)

PR #2057, opened 2022-02-01
Allow rich-11

mgorny:
Relax the version bind on rich to allow v11.  There are no test
regressions with the new version, and CLI seems to work correctly.


tomchristie (approved):
Sure thing.

mgorny:
Thanks!

PR #2065, opened 2022-02-04
Always rewind files on multipart uploads.

tomchristie:
Raised in #2064

- [x] Failing test case.
- [x] Implement rewind fix.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -157,7 +157,7 @@ class FileField:
             yield self._data
             return
 
-        if hasattr(self.file, 'seek'):
+        if hasattr(self.file, "seek"):
             self.file.seek(0)
 
         chunk = self.file.read(self.CHUNK_SIZE)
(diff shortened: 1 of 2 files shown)


tomchristie:
Waiting on a release of `httpcore` with https://github.com/encode/httpcore/pull/502 in order to fix the test suite.
Once we've got that I'll re-run the suite to demo that the test currently fails, and then push the fix.

aminalaee (approved):

httpx/_multipart.py
```diff
@@ -158,9 +157,8 @@ def render_data(self) -> typing.Iterator[bytes]:
             yield self._data
             return
 
-        if self._consumed:  # pragma: nocover
+        if hasattr(self.file, "seek"):
```
aminalaee, line 160: This is more of a question, so we do `seek` for the first file upload too, right?

tomchristie: Yup

StarkBotsIndustries:
It works like charm now! Thanks

StarkBotsIndustries (approved):

Issue #2069, opened 2022-02-05
The code in the document does not work


imsgj:
Hi. The code in the [document](https://www.python-httpx.org/advanced/#:~:text=files%20%3D%20%7B%27upload%2Dfile%27%3A%20(None%2C%20%27text%20content%27%2C%20%27text/plain%27)%7D) does not work

```
import httpx
files = {'upload-file': (None, 'text content', 'text/plain')}
r = httpx.post("https://httpbin.org/post", files=files)
print(r.text)
```
The error is 
```
TypeError: Expected bytes or bytes-like object got: <class 'str'>
```
I found [this problem](https://github.com/encode/httpx/issues/481) has been raised but not solved

If the string is modified with b, it works fine
```
import httpx
files = {'upload-file': (None, b'text content', 'text/plain')}
r = httpx.post("https://httpbin.org/post", files=files)
print(r.text)
```
Compared to requests
```
import requests
import httpx

files = {'upload-file': (None, 'text content', 'text/plain')}
httpx.post("https://httpbin.org/post", files=files)  # error
requests.post("https://httpbin.org/post", files=files)  # works

files = {'upload-file': (None, b'text content', 'text/plain')}
httpx.post("https://httpbin.org/post", files=files)  # works
requests.post("https://httpbin.org/post", files=files)  # works
```

tomchristie:
Thanks for raising this @imsgj - you're absolutely correct.

We made a change here so that only binary content is accepted. However I'm not sure we really want to do that or not. The *necessary* case was that we only accept *files opened in binary mode*. But we *could* still accept the plain ol' `str` case if we wanted to.

We've got two options here.

* Accept that we've got sensible behaviour, and update the docs to match the expected usage.
* Change the behaviour so that plain `str` cases *are* accepted. (But files opened in text mode are not.)

jamesboehmer:
FWIW I am hopeful that we can allow non-files for two reasons:
1. It would be consistent with `requests`
2. I just ran into exactly this issue because the Cloudflare Images API recently fixed a long-standing issue, but in doing so changed to require `multipart/form-data` requests (see https://developers.cloudflare.com/images/cloudflare-images/upload-images/direct-creator-upload/).  This used to take urlencoded JSON, so my client instantly broke.  I worked around it by faking the files argument with an empty pydantic instance:

```
async with httpx.AsyncClient() as client:
    cloudflare_response = await client.post(
        url=url,
        data=data_dict,
[...]

PR #2078, opened 2022-02-10
Allow str in form files.

waterfountain1996:
Closes: #2069

Changed `FileField` to allow `str` content, but raise exception on text buffers.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -107,8 +107,12 @@ class FileField:
             # requests does the opposite (it overwrites the header with the 3rd tuple element)
             headers["Content-Type"] = content_type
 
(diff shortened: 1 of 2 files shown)


tomchristie:
Great - thanks!

I've tweaked the description to "Closes: #2069", so that the issue gets a backlink to here, and so that merging this would auto-close the issue. Also retitled the PR, dropping the issue number.

I'm... *not sure* which of the two possible resolutions suggested in #2069 we should go for.

Possibly(?) in favour of making the docs change, rather than making the behaviour change.

That would still give us the option of making the behaviour change at a later date if we think that's a good thing to do, because *adding* `str` support there isn't be a breaking API change. (But removing it is.)

Not sure.

PR #2080, opened 2022-02-10
add options to override tcp hostname and tls hostname

toppk:
this requires new extensions inside httpcore  (https://github.com/encode/httpcore/pull/507)

Here is a snipped I used to test this.  
`python -c 'import httpx; import sys;httpx.main(sys.argv[1:])'  https://neverin.org/headers --x-connect-to 54.221.78.73  --x-sni-hostname httpbin.org`

### DIFF
--- a/httpx/_main.py
+++ b/httpx/_main.py
@@ -80,6 +80,9 @@ def print_help() -> None:
     )
 
     table.add_row("--follow-redirects", "Automatically follow redirects.")
+    table.add_row("--x-connect-to", "Use alternative hostname for tcp connection.")
+    table.add_row("--x-sni-hostname", "Use alternative hostname for TLS handshake.")
+
     table.add_row("--no-verify", "Disable SSL verification.")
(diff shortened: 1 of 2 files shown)


httpx/_main.py
```diff
@@ -468,6 +492,13 @@ def main(
             verify=verify,
             http2=http2,
         ) as client:
+            extensions: typing.Dict[str, typing.Any] = {
+                "trace": functools.partial(trace, verbose=verbose)
+            }
```
tomchristie, line 497: Do we need to type annotate this?
If we do (because of our tooling) then does just `extensions: dict` also pass?

toppk: yes, it was autolabelling with Partial[None] from "trace".  just "dict" works. (pushed)

toppk:
i am working on a test case for connect_to, but a test case for sni_hostname will be more involved, since we don't create a tls server inside the test suit currently (i think).

PR #2094, opened 2022-02-24
Fix invalid max_keepalive_connections argument name

benfasoli:
The argument to `httpx.Limits` is named `max_keepalive_connections` but is referenced as `max_keepalive` in the docs. This updates the single place I found it referenced incorrectly.

Thanks for all the great work on `httpx` 🙏


tomchristie:
Ah good spot - thanks!

tomchristie (approved):

PR #2145, opened 2022-03-25
Apply PEP-621

Kludex:
This was brought on https://github.com/encode/starlette/pull/1359.

setuptools released 0.61.0 yesterday, and brought to light the PEP-621 which makes this PR possible. Most of the PR was inspired on this page: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

Notes:
- ~I'll continue later on. I need to check if everything is included on the metadata files.~

## How to test?

- `./scripts/build` to see that wheel and `tar.gz` are generated
- `python -m pip install -e .` for editable installs


pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+Changelog = "https://github.com/encode/httpx/blob/master/CHANGELOG.md"
+Documentation = "https://www.python-httpx.org"
+Source = "https://github.com/encode/httpx"
+
+[project.scripts]
+httpx = "httpx:main"
+
+[tool.setuptools.dynamic]
+version = { attr = "httpx.__version__.__version__" }
+readme = { file = ["README.md", "CHANGELOG.md"] }
```
florimondmanca, line 64: Oh, this is quite neat. :+1:

Kludex:
Oh... `setuptools` doesn't support Python 3.6 anymore.

johtso:
https://github.com/encode/httpx/discussions/2136

florimondmanca:
What do we think about this bit at the top of the setuptools docs page…?

> Note
> New in 61.0.0 (experimental)
>
> Warning
> Support for declaring [project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/) or configuring setuptools via pyproject.toml files is still experimental and might change (or be removed) in future releases.

Kludex:
> Is now a good time?

Maybe not. Let's wait for others to test this first 😅 👍 

> Also, what does this get us compared to the current setup.py configuration?

One less file (setup.py), and compliance with the mentioned PEP.

scripts/build
```diff
@@ -8,6 +8,6 @@ fi
 
 set -x
 
-${PREFIX}python setup.py sdist bdist_wheel
+${PREFIX}python -m build --sdist --wheel
```
florimondmanca, line 11: Seems like this is the default for `python -m build`?

```suggestion
${PREFIX}python -m build
```

https://pypa-build.readthedocs.io/en/latest/index.html

requirements.txt
```diff
@@ -14,6 +14,7 @@ mkdocs-material==8.1.4
 # Packaging
 twine==3.7.1
 wheel==0.37.0
+build==0.7.0
```
florimondmanca, line 17: Latest is now:

```suggestion
build==0.8.0
```

Kludex:
Thanks for the review @florimondmanca 🙇 

I've updated the PR, sync with the changes that happened on `setup.py`, and I've added the `setup.py` back.

florimondmanca (approved):
Personally happy with this. Still happy to read other people's opinion.

My last question would be -- have we checked `py.typed` is present in the resulting `sdist`? (If not, releasing would break type hints.)

Kludex:
It adds correctly the `py.typed`.

I've added the `content-type` on the `readme` and bumped twine, as I was having a warning.

pyproject.toml
```diff
@@ -0,0 +1,61 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
```
jhominal, line 2: I think we should add a minimal version of setuptools as a build requirement to ensure that a compatible version of setuptools is installed.

```suggestion
requires = ["setuptools>=61", "setuptools-scm"]
```

adriangb: Agreed

Kludex:
- Close this in favor of #2334

Issue #2139, opened 2022-03-22
Response not closed when timeout/cancel reading response stream, which cause RuntimeError: The connection pool was closed while 1 HTTP requests/responses were still in-flight. 


guyskk:
Response not closed when timeout/cancel reading response stream, which cause `RuntimeError: The connection pool was closed while 1 HTTP requests/responses were still in-flight.`

#### Reproduce:

```python
import asyncio
import httpx

async def main():
    url = "https://httpbin.org/drip?delay=0&duration=5"
    # Or use local httpbin server:
    #     docker run -ti -p 8088:80 kennethreitz/httpbin:latest
    url = "http://127.0.0.1:8088/drip?delay=0&duration=5"
    async with httpx.AsyncClient(timeout=10, trust_env=False) as client:
        try:
            coro = client.get(url)
            response = await asyncio.wait_for(coro, 3)
        except Exception as ex:
            print(type(ex), repr(ex))
        else:
            print(response)

if __name__ == "__main__":
    asyncio.run(main())
```

#### Output:

```python
<class 'asyncio.exceptions.TimeoutError'> TimeoutError()
Traceback (most recent call last):
  File "/Users/kk/dev/zhuwen/httpx/http_error2.py", line 22, in <module>
    asyncio.run(main())
  File "/Users/kk/.pyenv/versions/3.9.7/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/kk/.pyenv/versions/3.9.7/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/kk/dev/zhuwen/httpx/http_error2.py", line 18, in main
    print(response)
  File "/Users/kk/dev/zhuwen/httpx/httpx/_client.py", line 1978, in __aexit__
    await self._transport.__aexit__(exc_type, exc_value, traceback)
  File "/Users/kk/dev/zhuwen/httpx/httpx/_transports/default.py", line 332, in __aexit__
    await self._pool.__aexit__(exc_type, exc_value, traceback)
  File "/Users/kk/.pyenv/versions/3.9.7/envs/httpx/lib/python3.9/site-packages/httpcore/_async/connection_pool.py", line 326, in __aexit__
    await self.aclose()
[...]

PR #2148, opened 2022-03-28
Fix response not closed when read stream canceled

guyskk:
See https://github.com/encode/httpx/issues/2139#issue-1176551205, the PR only fix issues of httpx client.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -901,7 +901,7 @@ class Client(BaseClient):
 
             return response
 
-        except Exception as exc:
+        except BaseException as exc:
             response.close()
             raise exc
 
@@ -933,7 +933,7 @@ class Client(BaseClient):
                     request = next_request
                     history.append(response)
 
-                except Exception as exc:
+                except BaseException as exc:
                     response.close()
                     raise exc
         finally:
@@ -972,7 +972,7 @@ class Client(BaseClient):
                     response.next_request = request
                     return response
 
-            except Exception as exc:
+            except BaseException as exc:
                 response.close()
                 raise exc
 
@@ -1605,7 +1605,7 @@ class AsyncClient(BaseClient):
 
             return response
 
-        except Exception as exc:  # pragma: no cover
+        except BaseException as exc:  # pragma: no cover
             await response.aclose()
             raise exc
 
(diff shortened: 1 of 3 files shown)


httpx/_client.py
```diff
@@ -933,7 +933,7 @@ def _send_handling_auth(
                     request = next_request
                     history.append(response)
 
-                except Exception as exc:
+                except BaseException as exc:
```
vrslev, line 936: Maybe catch `Exception` and `asyncio.exceptions.CancelledError`?

guyskk: Here we need catch all exceptions to ensure response closed, just like `async with` context manager. And the exception will re-raised so it will not swallow errors.

tomchristie:
Great thank you.

Yup, I'm in favour of switching over to `BaseException` here - it makes sense to *always* close the response. Including for asyncio or trio cancelations or `SystemExit`, and `KeyboardInterrupt`.

I'm not keen on us having test cases that introduce sleeps, tho. Perhaps we could change those around somehow.

guyskk:
In Python 3.6, asyncio.wait_for will not cancel the task, so I skip the test in py36. https://stackoverflow.com/questions/69370252/asyncio-wait-for-didnt-cancel-the-task

florimondmanca:
I think we should move the `aclose()` calls to `finally` blocks instead, see my analysis here: https://github.com/encode/httpx/issues/2139#issuecomment-1080532962

guyskk:
Hi @florimondmanca Seems the response not always need close, eg stream request: https://github.com/encode/httpx/blob/master/httpx/_client.py#L904 . 
And in handle redirect cases, I'm not sure if the response should close https://github.com/encode/httpx/blob/master/httpx/_client.py#L969.

tomchristie:
Thanks @guyskk.

I've taken another pass at this in #2156, with a different approach to the test case, so that we don't need to introduce slow `sleep` behaviour into our test suite. (Anywhere we can avoid the tests using actual network I/O we should)

guyskk:
@tomchristie 👍👍 your test case is perfect!

tomchristie:
Thanks folks. Closing this in favour of #2156.

We *also* ought to address any similar cases in `httpcore`.

PR #2152, opened 2022-03-29
Making charset auto-detection strictly opt-in.

tomchristie:
From discussion https://github.com/encode/httpx/discussions/2083

Remove our auto-charset-guessing behaviour for cases where `response.text` is accessed, but the response has no explicit charset. The behaviour is too fuzzy and inconsistent to support as a default.

Instead we allow explicitly setting the default encoding. Furthermore we *also* allow 

- [x] Drop `response.apparent_encoding` in favour of an explicit `response.default_encoding`.
- [x] Type of `response.encoding` becomes `str`, not `Optional[str]`.
- [x] Support `Response(default_encoding=...)`.
- [x] Support `Client(default_encoding=...)`. 
- [ ] Support `Response(encoding_errors=...)`.
- [ ] Support `Client(encoding_errors=...)`.
- [ ] Remove `charset_normalizer` dependancy.
- [ ] Support `default_encoding="charset_normalizer"` and `default_encoding="chardet"` to allow for auto-detection behaviours.
- [ ] Documentation on text encodings.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -167,6 +167,7 @@ class BaseClient:
         event_hooks: typing.Mapping[str, typing.List[typing.Callable]] = None,
         base_url: URLTypes = "",
         trust_env: bool = True,
+        default_encoding: str = "utf-8",
     ):
         event_hooks = {} if event_hooks is None else event_hooks
 
@@ -184,6 +185,7 @@ class BaseClient:
             "response": list(event_hooks.get("response", [])),
         }
         self._trust_env = trust_env
+        self._default_encoding = default_encoding
         self._netrc = NetRCInfo()
         self._state = ClientState.UNOPENED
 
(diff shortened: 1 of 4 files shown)


tomchristie (commented):
(Self review)

I think the core "make charset auto-detection strictly opt-in" is a good plan.

I also think that adding support for `default_encoding=...` (and `encoding_errors=...`) makes sense.

However, I'm not convinced by having installable codecs for "charset_normalizer" and "chardet". It's a bit magic. Here's a couple of different alternatives, that seem simpler to me...

1. Add support for `default_encoding=...`, and allow it to either be a string, or a callable.

```python
# A client with utf-8 as the fallback encoding.
client = httpx.Client(default_encoding="utf-8")

# A client with autodetect as the fallback encoding.
def autodetect(content: bytes) -> str:
    return charset_normalizer.detect(content).get("encoding", "utf-8")

client = httpx.Client(default_encoding=autodetect)
```

2. Add support for `default_encoding=...`, and use the string "autodetect" if you want to switch to auto detection.

```python
# A client with utf-8 as the fallback encoding.
client = httpx.Client(default_encoding="utf-8")

# A client with autodetect as the fallback encoding.
client = httpx.Client(default_encoding="autodetect")
```

My preference out of these is probably (2). I prefer that it doesn't overload the `default_encoding` type into two different cases, and it's obvious to me from reading the code what the intent is.

PR #2154, opened 2022-03-30
Support Server-sent Events

### LINKED ISSUE #1278: Server Sent Events
It'd be really neat to offer built-in support for server sent events, which are a way for servers to stream events to an HTTP client.

(Similar to WebSockets, but uni-directional)

```python
with httpx.stream("GET", "https://www.example.org/sse") as response:
    for sse in response.iter_sse():
        print(sse.data, sse.event, sse.id, sse.retry)
```

We'd probably also want to include an `sse.json()` helper method.

Useful resources...

* https://www.html5rocks.com/en/tutorials/eventsource/basics/
* https://sysid.github.io/sse/
* https://pgjones.gitlab.io/quart/tutorials/broadcast_tutorial.html

Not sure if this has come up before or not, but it'd be a *really* neat feature for us to offer, for relatively little effort.

abersheeran:
close #1278

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -77,6 +77,7 @@
 * `def .iter_bytes()` - **bytes iterator**
 * `def .iter_text()` - **text iterator**
 * `def .iter_lines()` - **text iterator**
+* `def .iter_sse()` - **Iterator[ServerSentEvent]**
 * `def .close()` - **None**
 * `def .next()` - **Response**
 * `def .aread()` - **bytes**
@@ -84,6 +85,7 @@
 * `def .aiter_bytes()` - **async bytes iterator**
 * `def .aiter_text()` - **async text iterator**
 * `def .aiter_lines()` - **async text iterator**
+* `def .aiter_sse()` - **`AsyncIterator[ServerSentEvent]`**
 * `def .aclose()` - **None**
 * `def .anext()` - **Response**
 
--- a/docs/async.md
+++ b/docs/async.md
@@ -78,6 +78,7 @@ The async response streaming methods are:
 * `Response.aiter_bytes()` - For streaming the response content as bytes.
 * `Response.aiter_text()` - For streaming the response content as text.
 * `Response.aiter_lines()` - For streaming the response content as lines of text.
+* `Response.aiter_sse()` - For streaming the Server-sent events response.
 * `Response.aiter_raw()` - For streaming the raw response bytes, without applying content decoding.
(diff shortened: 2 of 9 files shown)


tomchristie:
Thanks @abersheeran...

Some thoughts:

* I've got mixed feelings about if we should be supporting this, or if we should just close off https://github.com/encode/httpx/issues/1278, as per [this comment](https://github.com/encode/httpx/issues/1278#issuecomment-1081690073) (For example... are folks *actually* using/wanting this, or are we just adding functionality for the sake of it?)
* If we do support it, should the return result be a dict or an object with attributes?
* I think the documentation for this should probably be in the "Advanced" section, rather than the quickstart.

abersheeran:
Server-sent events is a stable, simple specification whose behavior won't change for the foreseeable future, and I don't think it's too difficult to maintain directly within httpx.

> I think the documentation for this should probably be in the "Advanced" section, rather than the quickstart.

OK, I'll revise this part of the documentation after we've settled on a final conclusion.

ahopkins:
Don't get me wrong, I love SSE. I feel that in many situations when someone wants to reach for websockets, they can get by with a lighter-weight SSE implementation. With my :heart: for SSE established, and with the express intent to speak only for myself, here are my thoughts.

> are folks _actually_ using/wanting this, or are we just adding functionality for the sake of it?

This seems out of scope to me. If I want this support, I want to build it myself or use another lib on top of httpx. It is a protocol layer that lives above HTTP. I want my HTTP client to focus on HTTP not on handling message parsing for a protocol that is not within its core functionality. By the same logic, I would not want httpx to become a GraphQL client either. These are functions that live above and should implement httpx and not be a part of it.

> If we do support it, should the return result be a dict or an object with attributes?

Typed objects. They will be easier to use, and the protocol is simple enough that there are only a handful of models needed with fairly basic properties.

> I think the documentation for this should probably be in the "Advanced" section, rather than the quickstart.

Agreed

fafo:
If you need use cases a large part of my last web application is untested since we are widely using SSE ;)
BTW while trying to implement the proposed external solution I got stuck with this problem: using EventSourceResponse from see-starlette when I try to to res = await client.stream("GET", url) the call hangs. 
I didn't figure why yet but without a response it's impossible to proceed with an external solution.

abersheeran:
> @abersheeran would you allow me to make a package out of your code?

Ok.

Issue #2157, opened 2022-04-01
[Typing] - `HTTPError` doesn't have a `request` attribute


michaeloliverx:
### Discussed in https://github.com/encode/httpx/discussions/2151

<div type='discussions-op-text'>

<sup>Originally posted by **michaeloliverx** March 29, 2022</sup>
The `HTTPError` class serves as a base class for `RequestError` and `HTTPStatusError`. The docstring shows the following usage:
```python
try:
    response = httpx.get("https://www.example.com")
    response.raise_for_status()
except httpx.HTTPError as exc:
    print(f"HTTP Exception for {exc.request.url} - {exc}")
```

Mypy doesn't like this:
```
error: "HTTPError" has no attribute "request"  [attr-defined]
```
</div>

PR #2158, opened 2022-04-01
Add `request` getter/setter to `HTTPError`

michaeloliverx:
- [x] Initially raised as discussion #2151

Closes: #2157


tomchristie (approved):
Seems reasonable, yup.
Thanks!

PR #2159, opened 2022-04-01
Fix raise_for_status exception docs

### LINKED ISSUE #2126: `raise_for_status` now raises an exception for 1xx and 3xx responses
Hi!

I've just noticed that `response.raise_for_status()` now raises an exception for 1xx and 3xx requests. I don't think it was the case before (namely with version 0.18) and the documentation doesn't seem to have been updated here: https://www.python-httpx.org/exceptions/

> The HTTPStatusError class is raised by response.raise_for_status() on 4xx and 5xx responses.

### Steps to reproduce (even though I think this is an intentional behavior, the doc is just not up to date?)

```python
import httpx
url_301 = "https://stat.pagesjaunes.fr/redirect?target=ETvxvbcLWawl69YzLe_Ph4vo03E_Yj6_Jt1_EvjjbHFaCE3c9RGLEg==&v=2.0&p1=1&p2=1&p3=&p4=RESEAUSOC-SITE-LVS-GRATUIT&p5=LR&p6=1361647273529289141447273529289&p7=06486910&p8="
httpx.get(url_301, follow_redirects=False).raise_for_status()

johtso:
Closes #2126

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -14,7 +14,7 @@ except httpx.RequestError as exc:
     print(f"An error occurred while requesting {exc.request.url!r}.")
 ```
 
-The `HTTPStatusError` class is raised by `response.raise_for_status()` on 4xx and 5xx responses.
+The `HTTPStatusError` class is raised by `response.raise_for_status()` on responses which are not a 2xx success code.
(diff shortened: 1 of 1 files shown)


tomchristie (approved):
👍

PR #2165, opened 2022-04-05
Make charset auto-detection optional.

tomchristie:
- [x] Add `Response(..., default_encoding=...)`
- [ ] Add `Client(..., default_encoding=...)`
- [ ] Documentation.
- [ ] Switch from `autodetect` as the default to `utf-8` as the default. (Should we do that as part of this PR, or as a follow-up?)

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -441,6 +441,7 @@ class Response:
         request: Request = None,
         extensions: dict = None,
         history: typing.List["Response"] = None,
+        default_encoding: str = "autodetect",
     ):
         self.status_code = status_code
         self.headers = Headers(headers)
@@ -456,6 +457,8 @@ class Response:
 
         self.is_closed = False
         self.is_stream_consumed = False
+
+        self.default_encoding = default_encoding
 
         if stream is None:
             headers, stream = encode_response(content, text, html, json)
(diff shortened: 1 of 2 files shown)


tomchristie:
> I would actually prefer (1)

My feeling is that practically the `default_encoding="autodetect"` just ends up being a neater developer experience. The alternative is slightly more fiddly to setup, is more verbose and less readable as a typing, and means we need an extra module in `httpx`, with an additional function exposed.

Another tiny little wrinkle is that if we want the import to fast fail, then it'd need to be a public module within `httpx`, and imported as a module. Eg `from httpx.charset_autodetect import charset_autodetect`. That's at odds with *all* the other API in `httpx`, which is all within private modules, and always accessed as `import httpx` ... `# Use httpx.<...>`.

> With the current setup of "autodetect" an error because of a missing dependency may not bubble up until the response is read.

Valid point. It'd probably make sense here to attempt importing `charset_normalizer` at the point of `Client(default_encoding="autodetect")`, so that any error there is raised immediately.

bratao:
Just an +1 from me. Debugging httpx on a high speed crawler, the decode (httpx/_decoders.py:76) is a bottleneck. I would like the option to not decode it automatically to speedup my code.

adriangb:
That looks nice Tom. I think the example makes it easy to understand.

Is the `content` parameter to `autodetect` the entire content body, in bytes? I suppose it only gets called if the user called `.text`, so it wouldn't read the body into memory unless the user requested the entire body anyway, right?

tomchristie:
@bratao Gotcha - that's a different consideration, which I've opened a discussion for... https://github.com/encode/httpx/discussions/2220

tomchristie:
> Is the content parameter to autodetect the entire content body, in bytes?

Yes.

> so it wouldn't read the body into memory unless the user requested the entire body anyway, right?

Yes.

The one additional consideration here is what to do when the user calls `iter_text`. In that case we don't have the entire content available before we need to take a decision on which text encoding to use. Currently we don't do autodetect in this case and just fallback to utf-8 always. We probably just want to stick with that behaviour. (At least for now)

tomchristie:
Now updated based on @adriangb's feedback.

PR #2175, opened 2022-04-11
Support rfc3986 v2

Mokto:
Would be nice to support rfc3986 to that enables python 3.8,3.9,3.10 support.

https://rfc3986.readthedocs.io/en/latest/release-notes/unreleased.html

Thanks!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
         "certifi",
         "charset_normalizer",
         "sniffio",
-        "rfc3986[idna2008]>=1.3,<2",
+        "rfc3986[idna2008]>=1.3,<3",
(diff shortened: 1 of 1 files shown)


Mokto:
No idea to be honest.

I just know httpx prevents me from updating the rfc3986 version.

florimondmanca:
As you can see in the CI logs, v2 seems to bring changes that would need addressing. Mostly look like deprecations. Since we manage the dependency, perhaps we should switch to `>=2,<3` and address those deprecations.

tomchristie:
Thanks for spending your time on this.
Closing in favour of #2252. Let's crack on with dropping our `rfc3986` dependancy.

PR #2183, opened 2022-04-16
docs: Fix proxy examples

nyuszika7h:
Per https://www.python-httpx.org/compatibility/#proxy-keys, there should always be a `://` after the protocol. The given examples raise an exception when used as-is.


florimondmanca:
@nyuszika7h Hi, thanks, could you confirm this is the kind of error we'd see from using `all` instead of `all://`?

```pycon
>>> from httpx._utils import URLPattern
>>> URLPattern("all")
  File "/Users/florimond/.pyenv/versions/3.10.0/lib/python3.10/site-packages/httpx/_utils.py", line 447, in __init__
    raise ValueError(
ValueError: Proxy keys should use proper URL forms rather than plain scheme strings. Instead of "all", use "all://"
```

nyuszika7h:
Indeed, that's the error I get.

florimondmanca (approved):
Thanks @nyuszika7h :+1:

Issue #2184, opened 2022-04-21
Some URL can make httpx use URL with wrong info


lebr0nli:
After some research, I found that `httpx.Client` and `httpx.Proxy` may implicit parse wrong URL because of the improper implement of `httpx.URL().copy_with`.

And this issue may lead to some blacklist bypass.

For example:

- `httpx.Client`

```python
user_input_from_http_request = 'http:////admin-dashboard/secret'
u = httpx.URL(user_input_from_http_request)
assert u.host.lower() != 'admin-dashboard'
resp = httpx.Client(base_url=u).get('/') # SSRF to http://admin-dashboard/secret
print(resp.text) # sensitive data leak
```

- `httpx.Proxy`

```python
user_input_from_http_request = 'http://x@//internal-proxy:8082/'
u = httpx.URL(user_input_from_http_request)
assert u.host.lower() != 'internal-proxy'
# httpx.Proxy(u).url.netloc == b'internal-proxy:8082'
resp = httpx.Client(proxies=u).get('http://localhost/secret') # will request via http proxy at internal-proxy:8082

```

Main reason:

https://github.com/encode/httpx/blob/b07fe7b0745e62be5ef9bce1bee9e7d7a8878552/httpx/_urls.py#L487

`copy_with` parse `self._uri_reference.copy_with(**kwargs).unsplit()` before returning the new URL, but the new URL string return by `unsplit` may make some unintended changes on the new URL.

For example:

```python
print(httpx.URL('http://[invalid string!!!!]//localhost/test!'). _uri_reference.unsplit())
# output: http://localhost/test!
u = httpx.URL('http://[invalid string!!!!]//localhost/test!').copy_with(scheme='https') # I only want to change scheme
print(u.host)
# output: localhost
# oops, host changed!
```

So if a function is using `copy_with`, it may have the same issue as `httpx.Client` and `httpx.Proxy`, too (For example, `copy_set_param`).

I also made a patch PR for this issue by replacing `return URL(self._uri_reference.copy_with(**kwargs).unsplit())` to:

```python
        new_url = URL(self)
        new_url._uri_reference = self._uri_reference.copy_with(**kwargs)
        if new_url.is_absolute_url:
            new_url._uri_reference = new_url._uri_reference.normalize()
        return URL(new_url)
```

~By the way, I think this issue is similar to [CWE-172](https://cwe.mitre.org/data/definitions/707.html) and [CWE-20](https://cwe.mitre.org/data/definitions/20.html).~

[...]

PR #2185, opened 2022-04-21
Patch `copy_with`

lebr0nli:
Patch for #2184


adamserafini (commented):
Would it make sense to add a test for this change?

lebr0nli:
> Would it make sense to add a test for this change?

Yes, it makes sense.

I add a test for `copy_with` in the new commit to ensure `copy_with` won't change any URL property that the user didn't expect.

P.S. I think it's very hard to cover all scenarios, so I focus on preventing the possible attack vectors as far as I know.

tomchristie (approved):

PR #2202, opened 2022-05-03
Clarify customizing authentication docs

Ninzero:
This is a tiny documentation change. But this part actually confused me for sometime so maybe worth clarifying.


florimondmanca (approved):
Sure, thanks!

PR #2206, opened 2022-05-06
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule: 
+      interval: monthly


Kludex (approved):

tomchristie:
Interesting, I didn't know this was an option.

Questions:

* Are any of our existing actions out-of-sync with the latest versions?
* Is updating to later actions beneficial or not? Why?

naveensrinivasan:
> Interesting, I didn't know this was an option.
> 
> Questions:
> 
> * Are any of our existing actions out-of-sync with the latest versions?
> * Is updating to later actions beneficial or not? Why?

With this change dependabot would open PR’s for out of sync actions. 

To answer your second question when dependabot opens the PR it will show the diffs in code and release note which can help you decide whether or not to update. 

Hope this helps. 

Thanks

Kludex:
> Are any of our existing actions out-of-sync with the latest versions?

Yes. The only two we have... `checkout` and `setup-python` have a v3.

> Is updating to later actions beneficial or not? Why?

The idea here is more about having the _bumping PR_, the actual decision on accepting is still ours, as @naveensrinivasan said.

naveensrinivasan:
> > Are any of our existing actions out-of-sync with the latest versions?
> 
> Yes. The only two we have... `checkout` and `setup-python` have a v3.
> 
> > Is updating to later actions beneficial or not? Why?
> 
> The idea here is more about having the _bumping PR_, the actual decision on accepting is still ours, as @naveensrinivasan said.
> 
> EDIT: Even if it looks like, I'm not very strong about this PR... I don't think there's much benefit, but it's a good thing...

In the future, if you add any workflows this will keep it updated. I'll let you decide. 👍

PR #2097, opened 2022-02-26
Drop EOL Python 3.6

ofek:
Discussion: https://github.com/encode/httpx/discussions/2136

Also, now `uvicorn` can be bumped (was pinned to fix Python 3.6 in https://github.com/encode/httpx/pull/1807)

PR #2076, opened 2022-02-09
Suppress binary output on the command line.

waterfountain1996:
Closes #2049

Changed print_response to simply display the content length of a binary response.

`get_lexer_for_response` returns 'Text only' for a `Content-Type: text/...` response, so it will be properly displayed using `rich.syntax.Syntax`. Those responses which do not get a corresponding lexer are considered binary and their output is suppressed.

To make the tests pass I had to add another route to the test server that would return a binary response. Feel free to point it out if there's another way to do it.

PR #2074, opened 2022-02-09
Preserve Authorization header on HTTPS redirect (#1850)

waterfountain1996:
Closes #1850

Added a utility function to check if a URL is a HTTPS upgrade of another URL. It will only work if the default ports (80 and 443) are used.

PR #2214, opened 2022-05-09
Version 0.23.0

tomchristie:
Drafting up our next release.

Since we're dropping Python 3.6, this needs to be a median point bump.

One thing I'd like to discuss before we push on with this is resolving https://github.com/encode/httpx/discussions/2083 (See PR #2165). The median point bump is a good motivator to drop our `charset-normalizer` dependancy, so it seems like an efficient use of time to address it as part of this release, if possible.

I'm planning to push on with that by addressing @[adriangb](https://github.com/adriangb)'s comment, so that we have two alternative approaches for consideration.

---

Release notes...

### Changed

 * Drop support for Python 3.6. (#2097)
 * Use `utf-8` as the default character set, instead of falling back to `charset-normalizer` for auto-detection. To enable automatic character set detection, see [the documentation](https://www.python-httpx.org/advanced/#character-set-encodings-and-auto-detection). (#2165)

 ### Fixed

 * Fix `URL.copy_with` for some oddly formed URL cases. (#2185)
 * Digest authentication should use case-insensitive comparison for determining which algorithm is being used. (#2204)
 * Fix console markup escaping in command line client. (#1866)
 * When files are used in multipart upload, ensure we always seek to the start of the file. (#2065)
 * Ensure that `iter_bytes` never yields zero-length chunks. (#2068)
 * Preserve `Authorization` header for redirects that are to the same origin, but are an `http`-to-`https` upgrade. (#2074)
 * When responses have binary output, don't print the output to the console in the command line client. Use output like `<16086 bytes of binary data>` instead. (#2076)
 * Fix display of `--proxies` argument in the command line client help. (#2125)
 * Close responses when task cancellations occur during stream reading. (#2156)
 * Fix type error on accessing `.request` on `HTTPError` exceptions. (#2158)


hozkok:
This PR seems to be resolving [this](https://nvd.nist.gov/vuln/detail/CVE-2021-41945) **critical** vulnerability. Is there any plan for releasing this soon?

felipecwb (approved):

hozkok (approved):

Kurbezz (approved):

tomchristie:
Waiting on #2165 and #2228.

tomchristie:
Considering switching the settings so that I can merge without further reviews.

* Pro: Can address issues such as this release more quickly.
* Con: Side-steps enforcing that we have enough input from wider maintenance team.

florimondmanca (approved):

PR #2226, opened 2022-05-13
Remove `RequestBodyUnavailable` from module docstring

michaeloliverx:
The `RequestBodyUnavailable` class no longer exists?


adriangb (approved):
Yep looks to be the case

PR #2261, opened 2022-06-05
add more virtualenv to gitignore

kaboussi:
- remove long_description variable
- add .env and .venv to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ htmlcov/
 site/
 *.egg-info/
 venv*/
+env*/
+.venv*/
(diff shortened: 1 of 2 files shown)


tomchristie:
Hi there! Thanks for your interest in the project.

We probably wouldn't want to make this change - the `.gitignore` style we have here is consistent across a bunch of different `encode` projects, and also *only* includes files that we need to ignore, rather than attempting to cover different style that other developers/projects might use.

See the `scripts/` directory for how we manage our various install, test, etc. tasks for this project.

PR #2278, opened 2022-06-23
allow setting an explicit multipart boundary via headers

adriangb:
Closes #2235

### DIFF
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -8,6 +8,7 @@ from typing import (
     Dict,
     Iterable,
     Iterator,
+    Mapping,
     Optional,
     Tuple,
     Union,
@@ -15,7 +16,7 @@ from typing import (
 from urllib.parse import urlencode
 
 from ._exceptions import StreamClosed, StreamConsumed
-from ._multipart import MultipartStream
+from ._multipart import MultipartStream, get_multipart_boundary_from_content_type
 from ._types import (
     AsyncByteStream,
     RequestContent,
@@ -150,11 +151,21 @@ def encode_urlencoded_data(
 
 
 def encode_multipart_data(
-    data: dict, files: RequestFiles, boundary: Optional[bytes] = None
+    data: dict,
+    files: RequestFiles,
+    boundary: Optional[bytes] = None,
+    headers: Optional[Mapping[str, str]] = None,
 ) -> Tuple[Dict[str, str], MultipartStream]:
+    if headers and boundary is None and "content-type" in headers:
+        content_type = headers["content-type"]
+        if not content_type.startswith("multipart/form-data"):
+            raise ValueError(
+                f"Invalid content-type header for multipart request: {content_type}"
+            )
+        boundary = get_multipart_boundary_from_content_type(content_type)
     multipart = MultipartStream(data=data, files=files, boundary=boundary)
-    headers = multipart.get_headers()
-    return headers, multipart
+    new_headers = multipart.get_headers()
+    return new_headers, multipart
 
 
 def encode_text(text: str) -> Tuple[Dict[str, str], ByteStream]:
@@ -187,6 +198,7 @@ def encode_request(
     files: Optional[RequestFiles] = None,
     json: Optional[Any] = None,
     boundary: Optional[bytes] = None,
+    headers: Optional[Mapping[str, str]] = None,
 ) -> Tuple[Dict[str, str], Union[SyncByteStream, AsyncByteStream]]:
     """
     Handles encoding the given `content`, `data`, `files`, and `json`,
@@ -207,7 +219,7 @@ def encode_request(
     if content is not None:
         return encode_content(content)
     elif files:
-        return encode_multipart_data(data or {}, files, boundary)
+        return encode_multipart_data(data or {}, files, boundary, headers)
     elif data:
         return encode_urlencoded_data(data)
     elif json is not None:
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -156,10 +156,6 @@ def encode_multipart_data(
     content_type: Optional[str] = None,
 ) -> Tuple[Dict[str, str], MultipartStream]:
     if content_type:
-        if not content_type.startswith("multipart/form-data"):
-            raise ValueError(
-                f"Invalid content-type header for multipart request: {content_type}"
-            )
         boundary = get_multipart_boundary_from_content_type(content_type)
     multipart = MultipartStream(data=data, files=files, boundary=boundary)
     new_headers = multipart.get_headers()
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -8,7 +8,6 @@ from typing import (
     Dict,
     Iterable,
     Iterator,
-    Mapping,
     Optional,
     Tuple,
     Union,
@@ -154,10 +153,9 @@ def encode_multipart_data(
     data: dict,
     files: RequestFiles,
     boundary: Optional[bytes] = None,
-    headers: Optional[Mapping[str, str]] = None,
+    content_type: Optional[str] = None,
 ) -> Tuple[Dict[str, str], MultipartStream]:
-    if headers and boundary is None and "content-type" in headers:
-        content_type = headers["content-type"]
+    if content_type:
         if not content_type.startswith("multipart/form-data"):
             raise ValueError(
                 f"Invalid content-type header for multipart request: {content_type}"
@@ -198,7 +196,7 @@ def encode_request(
     files: Optional[RequestFiles] = None,
     json: Optional[Any] = None,
     boundary: Optional[bytes] = None,
-    headers: Optional[Mapping[str, str]] = None,
+    content_type: Optional[str] = None,
 ) -> Tuple[Dict[str, str], Union[SyncByteStream, AsyncByteStream]]:
     """
     Handles encoding the given `content`, `data`, `files`, and `json`,
@@ -219,7 +217,7 @@ def encode_request(
     if content is not None:
         return encode_content(content)
     elif files:
-        return encode_multipart_data(data or {}, files, boundary, headers)
+        return encode_multipart_data(data or {}, files, boundary, content_type)
     elif data:
         return encode_urlencoded_data(data)
     elif json is not None:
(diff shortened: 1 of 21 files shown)


tests/test_multipart.py
```diff
@@ -42,6 +42,64 @@ def test_multipart(value, output):
...
+    # bit grungy, but sufficient just for our testing purposes.
+    boundary = response.request.headers["Content-Type"].split("boundary=")[-1]
+    content_length = response.request.headers["Content-Length"]
+    pdict: dict = {
+        "boundary": boundary.encode("ascii"),
+        "CONTENT-LENGTH": content_length,
+    }
+    multipart = cgi.parse_multipart(io.BytesIO(response.content), pdict)
+
+    assert multipart["file"] == [b"<file content>"]
```
adriangb, line 100: This test is fails on main/master as well, but I think it would be important to fix in this PR. What needs to happen is we need to create the random boundary and then insert that into the existing header. The issue is that there is currently no machinery / precedent for `encode_request()` modifying headers, so it would take a bit more refactoring to get working.

httpx/_content.py
```diff
@@ -187,6 +198,7 @@ def encode_request(
     files: Optional[RequestFiles] = None,
     json: Optional[Any] = None,
     boundary: Optional[bytes] = None,
+    headers: Optional[Mapping[str, str]] = None,
```
tomchristie, line 201: How about we keep things a little tighter here by passing `content_type: str = None` instead? That restricts the information we're passing around to the one thing we're actually interested in.

httpx/_multipart.py
```diff
@@ -20,6 +20,16 @@
...
 
 
+def get_multipart_boundary_from_content_type(
+    content_type: str,
+) -> bytes:
+    if ";" in content_type:
+        for section in content_type.split(";"):
+            if section.strip().startswith("boundary="):
+                return section.strip().split("boundary=")[-1].encode("latin-1")
+    raise ValueError("Missing boundary in multipart/form-data content-type header")
```
tomchristie, line 30: Instead of raising `ValueError` we could just return `None` if the content type doesn't start with "multipart/form-data", or doesn't include a valid boundary.

(Users can currently submit requests with `Content-Type` headers that don't properly match up to the data they include, so just being lax here seems okay to me.)

httpx/_content.py
```diff
@@ -150,11 +151,21 @@ def encode_urlencoded_data(
...
 def encode_multipart_data(
-    data: dict, files: RequestFiles, boundary: Optional[bytes] = None
+    data: dict,
+    files: RequestFiles,
+    boundary: Optional[bytes] = None,
+    headers: Optional[Mapping[str, str]] = None,
 ) -> Tuple[Dict[str, str], MultipartStream]:
+    if headers and boundary is None and "content-type" in headers:
+        content_type = headers["content-type"]
+        if not content_type.startswith("multipart/form-data"):
```
tomchristie, line 161: I think we could drop this case, as with [this comment[(https://github.com/encode/httpx/pull/2278/files#r905947972)

tomchristie (commented):
Looking good!

httpx/_content.py
adriangb: Yup good call

httpx/_multipart.py
adriangb: Note that since `content-type` gets set [via `setdefault`](https://github.com/encode/httpx/blob/aad60a4f123801e7ce0e02bc49138e7f0f9ca0a5/httpx/_models.py#L362). If it already exists, it won't be overwritten. So in the case where the user provided the `content-type` header but did not provide the multipart boundary we'd be sending out a request with no multipart boundary, which completely violates the spec and no server will be able to parse.

httpx/_content.py
adriangb: done

httpx/_multipart.py
tomchristie: > So in the case where the user provided the content-type header but did not provide the multipart boundary we'd be sending out a request with no multipart boundary, which completely violates the spec and no server will be able to parse.

True. We could either hard-error in that case, or just ignore it and allow users to do that.

There is also a whole class of cases here where a user can set a `Content-Type` that doesn't match the encoding type they're using with `data=`/`files=`/`json=`. (Eg. they can set `Content-Type: application/json` on either form or multipart requests.) But that's probably okay.

adriangb:
@tomchristie this is ready for another review whenever you get a chance

httpx/_content.py
```diff
@@ -150,11 +150,16 @@ def encode_urlencoded_data(
...
 
 def encode_multipart_data(
-    data: dict, files: RequestFiles, boundary: Optional[bytes] = None
+    data: dict,
+    files: RequestFiles,
+    boundary: Optional[bytes] = None,
+    content_type: Optional[str] = None,
 ) -> Tuple[Dict[str, str], MultipartStream]:
+    if content_type:
+        boundary = get_multipart_boundary_from_content_type(content_type)
```
jhominal, line 159: Shouldn't we check that `boundary is None` before computing `boundary`? (In other words, what happens when `content_type` and `boundary` do not agree on the value to use for `boundary`?)

adriangb: We're the only ones ever making the call into this function and we never call it with a mismatch. If I add the error I would either have to (1) add a test just to check the error or (2) pragma: no cover it. I'll add a comment instead.

jhominal: Shouldn't the `get_multipart_boundary_from_content_type` call be made upstream? e.g. in httpx/_models.py?

httpx/_models.py
```diff
@@ -332,7 +332,13 @@ def __init__(
...
 
         if stream is None:
-            headers, stream = encode_request(content, data, files, json)
+            headers, stream = encode_request(
+                content=content,
+                data=data,
+                files=files,
+                json=json,
+                content_type=self.headers.get("content-type"),
+            )
```
jhominal, line 341: By the comment on the other thread, I mean, why not write:
```suggestion
            headers, stream = encode_request(
                content=content,
                data=data,
                files=files,
                json=json,
                boundary= get_multipart_boundary_from_content_type(self.headers.get("content-type")),
            )
```

That way, we do not have function signatures that have both a `boundary` and a `content_type` that could possibly have contradictory data?

jhominal: By which I mean, the `content_type` argument could be removed from `encode_request` and `encode_multipart_data`?

httpx/_content.py
adriangb: I think _should_ would be a bit strong of a word, we certainly _could_ do that. I think the main reason for having it here is that `_models.py` doesn't know about multipart request or other encoding stuff, so it would be a bit of a violation of the layering that is currently set up to put something like `if content_type.startswith("multipart/form-data")`: <do multipart specific stuff>`.

jhominal: Yes, sorry, I went a bit strongly on it.

adriangb: No need to apologize for a good suggestion 😄 
Like I said I'm open to it, but I do feel that it breaks a bit with the way abstractions are currently set up in the codebase

httpx/_models.py
adriangb: Okay I liked it once I wrote it, I think given that `encode_request` already takes a `boundary` parameter (which only makes sense in the context of multipart requests) moving the call to where you are suggesting is not a big deal

jhominal: Yes, that kind of was my thought (if `encode_request` did not have `boundary` I would not have suggested that). Thank you for accepting my suggestion.

I just wonder if you would also go also look at potentially reviewing the remaining changes in `httpx/_content.py`? I mean, these changes seem mostly vestigial to me (e.g. the `content_type` in `encode_request`, and the changes in `encode_multipart_data` are now purely of variable and formatting (and `boundary` is now a required argument)

httpx/_multipart.py
```diff
@@ -20,6 +20,18 @@
...
 
+def get_multipart_boundary_from_content_type(
+    content_type: typing.Optional[str],
+) -> typing.Optional[bytes]:
+    if not content_type or not content_type.startswith("multipart/form-data"):
+        return None
+    if ";" in content_type:
+        for section in content_type.split(";"):
+            if section.strip().startswith("boundary="):
+                return section.strip().split("boundary=")[-1].encode("latin-1")
```
jhominal, line 31: Is there any possibility that `boundary` could be quoted, with either simple or double quotes?

jhominal: [RFC 2046](https://www.rfc-editor.org/rfc/rfc2046#section-5.1.1) has examples where boundary is contained in quotes...

httpx/_models.py
adriangb: done

httpx/_multipart.py
adriangb: Great catch! Not sure how you find these things but thank you

jhominal: After reviewing the grammar for boundary ([RFC 2046 Appendix A](https://www.rfc-editor.org/rfc/rfc2046#page-43), referencing [RFC 2045 §5.1](https://www.rfc-editor.org/rfc/rfc2045#section-5.1), referencing [RFC 822 §3.3](https://www.rfc-editor.org/rfc/rfc822#section-3.3) which has been replaced by [RFC 5322 §3.2.4](https://www.rfc-editor.org/rfc/rfc5322#section-3.2.4)):

1. The only quoting character that seems to be valid is the double quote `"` (`'` is not valid for that usage);
2. However, when the double quote is in use, any backslash `\` will be part of a "quoted pair", meaning that it is semantically equivalent to the next character (`\\` => `\`, `\"` => `"`);
3. However, there is a limitation on allowed characters in a boundary - none of the characters that would require a quoted pair escape are allowed in a boundary;

In other words:
 * We need to handle only `"` for quoting;
 * We also need to decide what to do about quoted pairs:
   * Either we write a simpler implementation that does not handle them, as all legal boundaries can be written without using quoted pairs;
   * Or we go the extra mile of supporting quoted pairs;
 * What do we want to do about the fact that not all ascii characters are allowed as part of a boundary?

adriangb: I think we can go with the simpler implementation. I also don't think we need to handle non-valid characters, that'd be too much introspection

jhominal (approved):
I think that is good for me.

Thank you for your patience!

adriangb:
> Thank you for your patience!

Thank you for your thorough and timely reviews!

PR #2302, opened 2022-07-13
Test under 3.11.0-beta.3

j178:
Added a few warning filters because:
- `_models.py` and `test_multiplart.py` are relying on deprecated`cgi`
- `certifi` is using deprecated `importlib.resources.path`
- `rich` is using deprecated `sre_constants`

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
 
     steps:
       - uses: "actions/checkout@v3"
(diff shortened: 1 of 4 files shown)


Kludex:
Should we use 3.11-dev tag instead? So when beta 4 is available on python-setup we get it automatically and can act on it?

j178:
Good idea, I don't know we can use `3.11-dev` actually, updated.

florimondmanca:
It's up to us to solve the `cgi` warning so I opened #2309 for this.

tomchristie:
I've not looked into *why* but seems like the coverage check is failing on 3.11 but not other Python versions...

https://github.com/encode/httpx/actions/runs/3150283298/jobs/5122888953

j178:
@tomchristie This is a bug when `coverage` is working under 3.11: https://github.com/nedbat/coveragepy/issues/1419. It should be fixed by bumping `coverage` version.

Kludex (approved):

PR #2307, opened 2022-07-14
Fix link to "http2 explained"

florimondmanca:
The old `/content/en/` link was broken, as English content seems to be served from `/` directly now.

Also -- small nit -- converts to "http2 explained" instead of "HTTP2 Explained", as per the site title.


tomchristie (approved):
Easy review, this one. 😅

Issue #207, opened 2019-08-14
Improve README with the differentiators from other HTTP clients


sethmlarson:
See: https://twitter.com/cooperlees/status/1161466340286185472

We're doing awesome things, we should make sure users know why they should use our library. :)

fbidu:
Do you think it would be better to provide something like a comparison table or some code samples right on the readme comparing the alternatives?

sethmlarson:
Would be nice to answer the question "Why HTTPX versus A, B, or C?" better in the README. Even if it's just a small section.

tomchristie:
Copying this from https://github.com/encode/httpx/issues/1216 so that we've got some better context here...

The notable candidates here are `requests`, `urllib3`, `aiohttp`, and `asks`.

* https://requests.readthedocs.io/en/master/
* https://urllib3.readthedocs.io/en/latest/
* https://docs.aiohttp.org/en/stable/client.html
* https://github.com/theelous3/asks

Main headliners here for HTTPX are...

* Supports HTTP/1.1 &* HTTP/2
* Supports both synchronous and async I/O
* Supports alternate async backends besides asyncio (trio, and soon curio)
* Supports calling into ASGI and WSGI applications without I/O
* Designed for compatibility with requests, with a docs section outlining any points where the API differs.

(And some other stuff that's out on the horizon, too)

They're not the same project because we were approaching a whole bunch of complex areas (such as how to build an HTTP client that supports both sync + async usages) that needed a clean slate, and the freedom to iterate on the design.

However we've put a stack of effort into making sure that the low-level transport layer is really tightly spec'ed as a standalone package - `httpcore`. Something we'd be very interested in would be figuring out with other projects if there's any mileage in figuring out if their clients would be able to use `httpcore` as their transport layer. In order to:

* Bring HTTP/2 support to those projects, with less effort than would otherwise be required.
* Be able to consolidate our efforts on the core networking layer, while still being able to provide different client libraries on top of that, and not break API for existing users.

We've not really reached out to folks about that yet since we've been focusing on getting to a 1.0 release first, but it *might* be an area worth exploring.

tomchristie:
Still valid thx, @stale bot.

PR #2308, opened 2022-07-14
Rework docs structure

florimondmanca:
I didn't raise a discussion about this, but this PR also serves as a hands-on proposal. Happy to discuss separately if need be.

This PR restructures the documentation in four well-visible sections:

View of the updated sidebar navigation

<details>
<summary>Open for text version</summary>

* Usage
  * Quickstart
  * Advanced usage
* Guides
  * Async Support
  * HTTP/2 Support
  * Requests Compatibility
  * Troubleshooting
* API Reference
  * Developer Interface
  * Exceptions
  * Environment variables
* Community
  * Third-Party Packages
  * Contributing
  * Code of Conduct

</details>

This takes inspiration from the 4-part Tutorials/How-To's/Discussions/Reference theorised by Daniele Procida (see [What nobody tells you about documentation (talk)](https://www.youtube.com/watch?v=t4vKPhjcMZg)).

It doesn't _exactly_ match that structure -- e.g. I don't think we've got something resembling "discussions" yet, although stuff like #207 would be potential candidates.

But we do have different _kinds_ of documentation in HTTPX _already_. The motivation here was to group them to help people find what they're looking for more easily.

So, each section answers the following questions:

* Usage - "How do I generally use HTTPX?"
* Guides - "How do I do X with HTTPX?" (where X = Async, HTTP/2, migrating from Requests, troubleshooting problems)
* API Reference - "What components are available when coding?"
* Community - "What surrounds the HTTPX project?"

The only shuffling of content I had to do was split the "Exceptions" docs across Quickstart vs. API Reference, as it mixed general usage hints and reference material. This matches how there's a [Exceptions and errors](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions) section in the Requests quickstart documentation, and then an [Exceptions](https://requests.readthedocs.io/en/latest/api/#exceptions) section on its API Reference page.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -4,6 +4,8 @@ site_url: https://www.python-httpx.org/
 
 theme:
     name: 'material'
+    features:
+        - navigation.sections
 
 repo_name: encode/httpx
 repo_url: https://github.com/encode/httpx/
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -4,25 +4,31 @@ site_url: https://www.python-httpx.org/
 
 theme:
     name: 'material'
+    features:
+        - navigation.sections
```
tomchristie, line 8: * What change does this feature effect?
* Where would I find this documented?

tomchristie (approved):
The section changing is very pleasing on the eye. Certainly an aid to my mental navigation.

Screenshot 2022-07-18 at 10 06 30

Moving that part of the exceptions docs also made much sense to me.

One inline comment which could be addressed. Not strictly a blocker.

Nicely chosen changes, Florimond. 🙏🏼

mkdocs.yml
florimondmanca: This allows showing navigation sections as always-unfolded, instead of foldable groupings as is the default in mkdocs.

This is here: https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-sections

florimondmanca: Is this clarification enough or are there any changes to the code you'd think worthy?

j178 (approved):
👍

florimondmanca:
Let’s merge then! :-)

PR #2309, opened 2022-07-14
Replace cgi which will be deprecated in Python 3.11

florimondmanca:
Prompted by https://github.com/encode/httpx/pull/2302 which surfaced a warning in the upcoming Python 3.11 as `cgi` is being deprecated as per [PEP-594](https://peps.python.org/pep-0594/).

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -215,10 +215,7 @@ def parse_content_type_charset(content_type: str) -> typing.Optional[str]:
     # See: https://peps.python.org/pep-0594/#cgi
     msg = email.message.Message()
     msg["content-type"] = content_type
-    charset = msg.get_param("charset")
-    if charset is None:
-        return None
-    return str(charset).strip("'\"")
+    return msg.get_content_charset(failobj=None)
 
 
 SENSITIVE_HEADERS = {"authorization", "proxy-authorization"}
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -1,4 +1,5 @@
 import codecs
+import email.message
 import logging
 import mimetypes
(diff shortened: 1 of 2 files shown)


j178:
There is another use in `test_multipart.py` actually.

adriangb (approved):

httpx/_utils.py
```diff
@@ -209,6 +210,17 @@ def parse_header_links(value: str) -> typing.List[typing.Dict[str, str]]:
...
 
+def parse_content_type_charset(content_type: str) -> typing.Optional[str]:
+    # We used to use `cgi.parse_header()` here, but `cgi` became a dead battery.
+    # See: https://peps.python.org/pep-0594/#cgi
+    msg = email.message.Message()
+    msg["content-type"] = content_type
+    charset = msg.get_param("charset")
+    if charset is None:
+        return None
+    return str(charset).strip("'\"")
```
oyeyipo, line 221: does `str` call not seem redundant? given failobj (default to None) is returned in the absence of a param. If param is present, either a string or a 3-tuple. Is later a possibility?

florimondmanca: True, not sure why I used `str()` here. Also, I think we want `get_content_charset()`, which does an equivalent of `email.utils.collapse_rfc2231_value(msg.get_param("charset"))`, which is what the [get_param() docs](https://docs.python.org/3/library/email.compat32-message.html#email.message.Message.get_param) recommend to deal with this "string or 3-tuple" oddity.

florimondmanca, line 221:
```suggestion
    return msg.get_content_charset(failobj=None)
```

PR #2317, opened 2022-07-21
Relax `HeaderTypes` type alias definition

### LINKED ISSUE #2314: Relax `HeaderTypes` type alias definition
https://github.com/encode/httpx/blob/3aa441015820a41861bff9b0e78d787b1a4f2f68/httpx/_types.py#L46-L47

Could `Dict` be replaced with `Mapping` in this type definition?

This change would placate the type-checker when I do something like the following:

```python3
from collections.abc import Mapping
from typing import Final

_DEFAULT_HEADERS: Final[Mapping[str, str]] = {
    # ...
}
```

when my intent is that the module-level dictionary is intended to be immutable but I don't want to go so far as to write a custom frozen dictionary or wrap it in `types.MappingProxyType`.

If this is acceptable, I'm happy to open a PR for the change. I was about to open a patch but your pull request template text made me think twice because this is more than just fixing a typo despite the tiny change.

Thanks!

willfrey:
This replaces `Dict[..., ...]` with `Mapping[..., ...]` in the union definition for `HeaderTypes`. Closes #2314.

- [x] Initially raised as discussion #2314


willfrey:
If there's interest, I'd be happy to try to do the same for the following type definitions in separate PRs, replacing `Dict` with `Mapping` or `MutableMapping` and `List` with `Sequence`, `MutableSequence`, or `Iterable`, depending on what the actual constraints are in the implementations.

https://github.com/encode/httpx/blob/93de1980fa77f15c6b23cbaf2422c0a812caf243/httpx/_types.py#L35-L42

https://github.com/encode/httpx/blob/93de1980fa77f15c6b23cbaf2422c0a812caf243/httpx/_types.py#L52

https://github.com/encode/httpx/blob/93de1980fa77f15c6b23cbaf2422c0a812caf243/httpx/_types.py#L68

adriangb (approved):

tomchristie:
Seems reasonable, yes.

PR #2334, opened 2022-08-09
Update package metadata

ofek:
## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). As a result, the execution of `setup.py` files is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

This is a follow-up to https://github.com/encode/httpx/discussions/2333

## Summary of changes

This implements [PEP 621](https://peps.python.org/pep-0621/), obviating the need for `setup.py` and `MANIFEST.in`. The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), etc.

## Future

- Move config for tools like `pytest` out of `setup.cfg` and into `pyproject.toml`
[...]

PR #2351, opened 2022-08-26
Add fake setup.py

Kludex:
This is quite sad.

@michaeloliverx brought this to our attention on: https://github.com/encode/httpx/commit/45b7cfaad3a8987ea35fa5bf092bbdda485444fd#commitcomment-82042745.

The `Used by` section on the project's profile disappeared after we removed the `setup.py`. We can either accept this PR or revert https://github.com/encode/httpx/pull/2334, to have it back,

I'll let you folks decide @florimondmanca @tomchristie 🙏 

I'll update the Starlette and Uvicorn according to what is decided here.

Closes #2352

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,7 @@
 import sys
 
+from setuptools import setup
+
 sys.stderr.write(
     """
 ===============================
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,31 @@
+import sys
+
+from setuptools import setup
+
+sys.stderr.write(
+    """
+===============================
+Unsupported installation method
+===============================
+httpx no longer supports installation with `python setup.py install`.
+Please use `python -m pip install .` instead.
+"""
+)
+sys.exit(1)
+
+
+# The below code will never execute, however GitHub is particularly
+# picky about where it finds Python packaging metadata.
+# See: https://github.com/github/feedback/discussions/6456
+#
+# To be removed once GitHub catches up.
+
+setup(
+    name="httpx",
+    install_requires=[
+        "certifi",
+        "sniffio",
+        "rfc3986[idna2008]>=1.3,<2",
+        "httpcore>=0.15.0,<0.16.0",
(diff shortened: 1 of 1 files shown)


michaeloliverx:
This is the same solution as `urllib3` has taken. You can read a bit more about it at https://sethmlarson.dev/blog/preparing-for-the-wave-of-open-source-funding#dependency-chain-metadata.

Kludex:
Flask does this: https://github.com/pallets/flask/blob/main/setup.py#L3-L16

florimondmanca:
That’s a raging GitHub limitation. I’d be okay with this PR; happier to use newest standards than stick to oldest because GH is lagging behind. The sad part is we’d have to do this and duplicate dependency declaration every project. So I’m on the fence…

Kludex:
How we decide here?

Kludex:
I've created #2352 as an alternative to this PR.

setup.py
```diff
@@ -0,0 +1,31 @@
+import sys
+
+from setuptools import setup
```
samuelcolvin, line 3: I guess technically this import isn't required since it can never be used.

Kludex, line 3:
```suggestion
```

Kludex:
> "Flask does this" is a pretty good argument. 😅
> 
> > The sad part is we’d have to do this and duplicate dependency declaration every project. So I’m on the fence…
> 
> Indeed. I'm not wild about it either, but the pallets team have this on all their repos. Probably an okay solution for now.

Since that's the case. Shall we choose this PR?

setup.py
```diff
@@ -0,0 +1,29 @@
+import sys
```
Kludex, line 1:
```suggestion
import sys

from setuptools import setup
```

michaeloliverx (approved):

Kludex:
I'll merge this tonight if no objections are made. 🙏

PR #2352, opened 2022-08-28
Revert "Update package metadata"

Kludex:
Reverts encode/httpx#2334

- This is an alternative to #2351, and you can find the context for this PR there.

Closes #2351.


tomchristie:
Hrm. Don't know actually.
I'll reply on the other ticket.

PR #2367, opened 2022-09-12
add basic type annotation for extensions

adriangb:
Mentioned on gitter

Before:
Screen Shot 2022-09-12 at 1 16 40 PM

After:
Screen Shot 2022-09-12 at 1 16 59 PM

We can always come back later and add a more accurate type (`TypedDict` or something)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -38,6 +38,7 @@ from ._types import (
     QueryParamTypes,
     RequestContent,
     RequestData,
+    RequestExtensions,
     RequestFiles,
     SyncByteStream,
     TimeoutTypes,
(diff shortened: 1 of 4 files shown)


tomchristie:
Sure. We're not changing any public API here(*), and we're fixing up some tooling, so seems like a clear win to me.

(*) Specifically, we don't expose `RequestExtensions`/`ResponseExtensions` as publicly exported bits of our API.

tomchristie (approved):

tomchristie:
I assume that's pylance you're showing in the example?
Is that output supplied by the pyright linter?

PR #2382, opened 2022-09-30
Streaming multipart support

### LINKED ISSUE #2249: Add support for streaming multipart/form-data
(After initial discussion in https://github.com/encode/httpx/discussions/2227.)

As far as I see, there's no way to stream `multipart/form-data` uploads when the `Content-Length` is unknown, since HTTPX tries to get byte size some way or another (last resort is to read it all in memory and count bytes). Please, let me know if I overlooked something!

Use case would be to stream something whose size is unknown, like fetching and in parallel uploading a CSV export from a Postgres database, say.

HTTP/1.1 has chunked encoding for that. As you know, that streams the _message_ itself, not the resource. This encoding is supported for "direct file uploads" in HTTPX, but not for `multipart/form-data`. This is strange, since chunked encoding is independent of `Content-Type`. Indeed, if both chunking and compression are in place, you first compress, and then you chunk the compressed bytes.

In HTTP/2 chunked encoding does not exist, byte size is per DATA frame. As you know, a message is allowed to have several. So, in this case the API would support uploading with multiple DATA frames, coming from a generator or iterator, for example.

tomchristie:
Closes #2249

Something we're not currently handling here, that'd also need updating is how we handle non-rewindable cases.

If the file instance is a non-seekable `IOBase` then we ought to raise a `StreamConsumed` error if we attempt to read from the file more than once. This occurs if we attempt to upload a streaming multipart file, but then get a redirect. Everything's fine if we've got a rewindable file instance, but if we've got a non-rewindable case then we need to error out, since we can't repeat the upload at that point.


adriangb:
Wouldn't we get the redirect before we start reading files? I would have thought we'd assemble the multipart stuff into an `AsyncIterable[bytes]` that lazily does work -> if we get redirected we never start pulling from it.

justinjeffery-ipf:
Any update when this will be added?

adriangb:
Awesome thanks Tom and Michael!

PR #2396, opened 2022-10-05
Always use latest version of `pip`.

tomchristie:
Installation should start by updating `pip` to the latest version.

Resolves issue noted in https://github.com/encode/httpx/pull/2334#issuecomment-1268308195


Kludex (approved):

Kludex:
Mental note: same needs to be done on the other encode projects.

PR #2400, opened 2022-10-06
Allow str content for multipart upload files

### LINKED ISSUE #2069: The code in the document does not work

tomchristie:
Closes #2069.

Allow `str` for simple multipart uploads...

```python
httpx.post(url, files={"upload": "some content"})
```

However we (continue to) explicitly disallow `io.StringIO` and files opened in text mode, since we can't upfront determine their byte-length.

I think this is probably a pretty okay middle-ground, supporting the simple text-use-case, while enforcing bytes for the file-like cases. Not a breaking change, since we're *expanding* our supported cases here.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -122,8 +122,14 @@ class FileField:
             # requests does the opposite (it overwrites the header with the 3rd tuple element)
             headers["Content-Type"] = content_type
 
-        if isinstance(fileobj, (str, io.StringIO)):
-            raise TypeError(f"Expected bytes or bytes-like object got: {type(fileobj)}")
+        if "b" not in getattr(fileobj, "mode", "b"):
+            raise TypeError(
+                "Multipart file uploads must be opened in binary mode, not text mode."
+            )
+        if isinstance(fileobj, io.StringIO):
+            raise TypeError(
+                "Multipart file uploads require 'io.BytesIO', not 'io.StringIO'."
+            )
 
         self.filename = filename
         self.file = fileobj
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -80,7 +80,7 @@ ResponseExtensions = Dict[str, Any]
 
 RequestData = Mapping[str, Any]
 
-FileContent = Union[IO[bytes], bytes]
+FileContent = Union[IO[bytes], bytes, str]
 FileTypes = Union[
     # file (or bytes)
     FileContent,
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -339,18 +339,37 @@ def test_multipart_encode_files_allows_bytes_content() -> None:
         assert content == b"".join(stream)
 
 
-def test_multipart_encode_files_raises_exception_with_str_content() -> None:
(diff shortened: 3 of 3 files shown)


adriangb (approved):
Makes sense to me! Do we just encode it as utf-8?

tomchristie:
> Makes sense to me! Do we just encode it as utf-8?

Yup.

euri10:
> I think this is probably a pretty okay middle-ground, supporting the simple text-use-case, while enforcing bytes for the file-like cases. Not a breaking change, since we're _expanding_ our supported cases here.

I think this is a breaking change, at least it breaks my test suite with 
```
        if "b" not in getattr(fileobj, "mode", "b"):
>           raise TypeError(
                "Multipart file uploads must be opened in binary mode, not text mode."
            )
E           TypeError: Multipart file uploads must be opened in binary mode, not text mode.
```
this is the snippet that was previously working:
```
    files: List[Tuple[str, IO]] = [
        ("files", (open(Path(__file__).parent / "data_upload" / "ls" / ls_file)))
        for ls_file in to_upload
    ]
    response = await client.post(
        "/uploads/ls",
        files=files,
        headers={"Authorization": f"Bearer {token}"},
    )
```

tomchristie:
Okay. Error on my part - we shouldn't have issued that as a minor point release.
Still, we're here now.

**Make sure to open upload files in binary mode.**

We have that as a constraint since it ensures that you're uploading the *actual binary file content*, rather than decoding it into text and then re-encoding it into bytes when it goes out onto the wire. This also means we can correct determine the content length in advance.

```python
    files: List[Tuple[str, IO]] = [
        ("files", (open(Path(__file__).parent / "data_upload" / "ls" / ls_file, "rb")))  # Added "rb" mode.
        for ls_file in to_upload
    ]
    response = await client.post(
        "/uploads/ls",
        files=files,
        headers={"Authorization": f"Bearer {token}"},
    )
```

PR #2404, opened 2022-10-11
improve types for RequestData

adriangb:


httpx/_multipart.py
```diff
@@ -40,17 +42,7 @@ class DataField:
...
-        self, name: str, value: typing.Union[str, bytes, int, float, None]
-    ) -> None:
-        if not isinstance(name, str):
-            raise TypeError(
-                f"Invalid type for name. Expected str, got {type(name)}: {name!r}"
-            )
-        if value is not None and not isinstance(value, (str, bytes, int, float)):
-            raise TypeError(
-                f"Invalid type for value. Expected primitive type, got {type(value)}: {value!r}"
-            )
```
adriangb, line 53: I'm removing these checks and the associated tests. I don't think we should be doing manual runtime type checking.

httpx/_multipart.py
```diff
@@ -214,16 +206,18 @@ def __init__(
     def _iter_fields(
         self, data: RequestData, files: RequestFiles
     ) -> typing.Iterator[typing.Union[FileField, DataField]]:
-        for name, value in data.items():
-            if isinstance(value, (tuple, list)):
-                for item in value:
-                    yield DataField(name=name, value=item)
+        for name, data_value in data.items():
+            if data_value is None or isinstance(
+                data_value, (*PRIMITIVE_DATA_TYPES, bytes)
```
adriangb, line 211: DataField accepts `bytes` in addition to the other "primitive types". I'm not sure where else `PrimitiveTypes` is being used but we might want to consider expanding everywhere to accept bytes instead of just here.

httpx/_multipart.py
```diff
@@ -214,16 +206,18 @@ def __init__(
...
             else:
-                yield DataField(name=name, value=value)
+                for item in data_value:
+                    yield DataField(name=name, value=item)
 
         file_items = files.items() if isinstance(files, typing.Mapping) else files
-        for name, value in file_items:
-            yield FileField(name=name, value=value)
+        for name, file_value in file_items:
+            yield FileField(name=name, value=file_value)
```
adriangb, line 220: mypy complains about sharing the variable `value` for both data and file types

httpx/_content.py
```diff
@@ -137,12 +139,12 @@ def encode_content(
...
-    plain_data = []
+    plain_data: "List[Tuple[str, str]]" = []
     for key, value in data.items():
-        if isinstance(value, (list, tuple)):
-            plain_data.extend([(key, primitive_value_to_str(item)) for item in value])
-        else:
+        if value is None or isinstance(value, PRIMITIVE_DATA_TYPES):
             plain_data.append((key, primitive_value_to_str(value)))
+        else:
+            plain_data.extend([(key, primitive_value_to_str(item)) for item in value])
```
adriangb, line 147: Flipping these checks lets us handle _any_ sequence instead of special casing `List` and `Tuple`.

adriangb:
@tomchristie how would you feel about running pyright in CI so we can use it for informational purposes, maybe using it’s typing score as a rough metric for “is this PR improving things for users” without any sort of enforcement?

httpx/_multipart.py
```diff
@@ -214,16 +206,32 @@ def __init__(
...
+                "Invalid type for value."
+                " Expected primitive type or iterable of primitive types,"
+                f" got {type(data_value)}: {data_value!r}"
+            )
+            if data_value is None or isinstance(
+                data_value, (*PRIMITIVE_DATA_TYPES, bytes)
+            ):
+                yield DataField(name=name, value=data_value)
+            elif isinstance(data_value, typing.Sequence):
+                for item in data_value:
```
adriangb, line 225: Moving the type checks to this function lets us handle any sequence

adriangb:
@tomchristie I requested re-review since I removed the removal (so restored?) the runtime type checks. This is still touching multiple modules but it is dealing only with a single type (`RequestData`).

I think that the lack of changes to the tests (besides adding a check for a case that would have previously been a false positive) are a good indication that this is not changing any runtime behavior despite the internal refactoring.

tomchristie:
Closing as stale

PR #2407, opened 2022-10-13
Drop `.read`/`.aread` from `SyncByteStream`/`AsyncByteStream`

tomchristie:
We have a leftover bit of API on `SyncByteStream`/`AsyncByteStream` that is a hangover from when we were designing the Transport API.

They both have a convenience method for reading the stream, which was added to make testing at the low-level of the Transport API easier. However the methods are a bit kludgy because they don't actually fit the `read(max_bytes)` that Python file interfaces use.

I think we can just hard-drop these. They're undocumented, and you don't *ever* need them. The convenience methods are on the `Response` class.

### DIFF
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -52,9 +52,7 @@ class IteratorByteStream(SyncByteStream):
             raise StreamConsumed()
 
         self._is_stream_consumed = True
-        if hasattr(self._stream, "read") and not isinstance(
-            self._stream, SyncByteStream
-        ):
+        if hasattr(self._stream, "read"):
             # File-like interfaces should use 'read' directly.
             chunk = self._stream.read(self.CHUNK_SIZE)  # type: ignore
             while chunk:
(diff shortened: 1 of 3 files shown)


httpx/_content.py
```diff
@@ -52,9 +52,7 @@ def __iter__(self) -> Iterator[bytes]:
             raise StreamConsumed()
 
         self._is_stream_consumed = True
-        if hasattr(self._stream, "read") and not isinstance(
-            self._stream, SyncByteStream
-        ):
```
tomchristie, line 57: The `and not isinstance(self._stream, SyncByteStream)` is what prompted me to clean these up.

It's such an oddly confusing branch - I couldn't figure out why it existed.

*"Treat file-like objects as file-like objects, except for this one specific case"*.

No thanks.

httpx/_types.py
```diff
@@ -107,33 +107,7 @@ def close(self) -> None:
...
-        Streaming cases should use a `try...finally` block to ensure that
-        the stream `close()` method is always called.
-
-        Example:
-
-            status_code, headers, stream, extensions = transport.handle_request(...)
-            try:
-                ...
-            finally:
-                stream.close()
```
tomchristie, line 120: This is an example from the old-style Transport API.

We had `.read()` so that users could eg. `stream.read()` here.

The Transport API now just returns fully fledged responses, so this is redundant.

michaeloliverx (approved):

tomchristie:
Thanks for the review @michaeloliverx.
Anyone want to merge this?...

tomchristie:
...How should we handle that generally?
Default to "pull request owner" can merge once approved?

michaeloliverx:
> ...How should we handle that generally? Default to "pull request owner" can merge once approved?

I think this is okay, maybe if **you** are personally reviewing something you can go ahead and merge it yourself.

PR #2424, opened 2022-10-26
Drop cgi module from test_multipart

vfazio:
Minor PR to drop usage of the `cgi` module in favor of `multipart`

Also drops the deprecation warning filters that should no longer necessary


tomchristie (approved):
Great, thanks!

PR #2452, opened 2022-11-20
doc: add recommendations for use of clients as context managers

adriangb:
https://github.com/encode/httpx/discussions/2444#discussioncomment-4178881


madkinsz (approved):

docs/advanced.md
```diff
@@ -52,6 +52,21 @@ finally:
     client.close()
 ```
 
+There is some overhead to managing these connections, so you should try to persist your client as much as possible.
+For example, in an ASGI web application like Starlette or FastAPI you should create the client in the lifespan event so that it gets re-used for each request.
```
tomchristie, line 56: I'm not particularly keen on this call-out for Starlette and FastAPI.

docs/advanced.md
```diff
@@ -52,6 +52,21 @@ finally:
     client.close()
 ```
 
+There is some overhead to managing these connections, so you should try to persist your client as much as possible.
```
tomchristie, line 55: > There is some overhead to managing these connections, so you should try to persist your client as much as possible.

👍 Yup.

I wonder if we might add something like "This can be achieved either by having a single scoped client that's passed throughout wherever it's needed, or by having a single global client instance."

docs/advanced.md
```diff
@@ -52,6 +52,21 @@ finally:
...
+
+For example, if you are connecting to `http://auth.example.com` and `http://service.example.com` you would create two distinct client instances:
+
+```python
+with httpx.Client(base_url="http://auth.example.com") as auth_client:
+    with httpx.Client(base_url="http://service.example.com") as service_client:
+        resp = auth_client.get("/.well-known/jwks.json")        
+        ...
+        resp = service_client.get("/endpoint")
+```
```
tomchristie, line 68: This is an interesting point. Do we want two distinct client instances here? The connection pooling is smart enough to retain connections across each domain wherever possible.

trondhindenes:
It would be good if this PR got some love. Yesterday I (yet again) came across a code where the context-manager async client  was used in a hot loop, causing terrible performance/excessive cpu usage. IMHO it's quite important to get this doc update in (I would also maybe not recommend instantiating two clients, our experience is that a single "global" client deals well with all we've thrown at it

tomchristie:
@trondhindenes Good call. Would you like to open a PR with the wording you'd propose, and we can work through it from there?

tomchristie:
Let's work through this on #2727.

PR #2461, opened 2022-11-22
update requests compatibility docs on query and form params

falkben:
Following up on https://github.com/encode/httpx/discussions/2330#discussioncomment-4167768 I added a note about requests compatibility with form data and query params -- namely that httpx only supports dictionaries whereas requests also supports lists of tuples.

As suggested, I added this note after the Query Parameters section, however, it might also bear mentioning in the Request Content section?


tomchristie (approved):
Looks good, yup.

tomchristie:
> however, it might also bear mentioning in the Request Content section?

Not sure. We could leave it as-is and await further feedback perhaps?

falkben:
> > however, it might also bear mentioning in the Request Content section?
> 
> Not sure. We could leave it as-is and await further feedback perhaps?

Sounds good!

PR #2464, opened 2022-11-27
[docs] Remove `chunk_size` arg from `iter_lines()` & `aiter_lines()` in documentation.

symonk:
Hi, looking at the docs while wrapping the `Response` object a few minor discrepancies exist.  Both `iter_lines()` & `aiter_lines()` do not accept an optional `chunk_size` argument, just updating that.  I _think_ this is the right place for this change.

Thanks for your time and a great library.


symonk:
I don't _think_ the diff here is causing the failing tests

symonk:
I think there might also be a discrepency with `next()` & `anext()` also

tomchristie (approved):

PR #2466, opened 2022-11-29
Remove typeshed iscoroutine workaround

adriangb:
This was resolved in typeshed and newer mypy versions have inherited the fix.

pyright reports:
[after.txt](https://github.com/encode/httpx/files/10110416/after.txt)
[before.txt](https://github.com/encode/httpx/files/10110417/before.txt)


abersheeran (approved):
LGTM.

PR #2467, opened 2022-11-29
Typing: use wsgiref.types to validate types and fix issues uncovered

adriangb:
- start_response() must return a write(bytes) function, even though this is now deprecated. It's fine to be a no-op here.
- sys.exc_info() can return (None, None, None), so make sure to handle that case.

pyright reports:
[after.txt](https://github.com/encode/httpx/files/10115502/after.txt)
[before.txt](https://github.com/encode/httpx/files/10115503/before.txt)

### DIFF
--- a/httpx/_transports/wsgi.py
+++ b/httpx/_transports/wsgi.py
@@ -1,14 +1,38 @@
 import io
 import itertools
 import sys
+import types
 import typing
+
+import typing_extensions as te
 
 from .._models import Request, Response
 from .._types import SyncByteStream
 from .base import BaseTransport
 
+_T = typing.TypeVar("_T")
+_ExcInfo = typing.Tuple[typing.Type[BaseException], BaseException, types.TracebackType]
+_OptExcInfo = typing.Union[_ExcInfo, typing.Tuple[None, None, None]]
 
-def _skip_leading_empty_chunks(body: typing.Iterable) -> typing.Iterable:
+# backported wsgiref.types definitions from Python 3.11
+
+
+class StartResponse(te.Protocol):
+    def __call__(
+        self,
+        __status: str,
+        __headers: typing.List[typing.Tuple[str, str]],
+        __exc_info: typing.Optional[_OptExcInfo] = ...,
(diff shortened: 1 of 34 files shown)


adriangb:
@tomchristie this one brings type completeness up to 98% 🥳

httpx/_transports/wsgi.py
```diff
@@ -1,14 +1,38 @@
 import io
 import itertools
 import sys
+import types
 import typing
 
+import typing_extensions as te
```
tomchristie, line 7: I can't see where this is coming from in the `requirements.txt`... is this built-in or a package?

Also, can we just `import typing_extensions` rather than `import typing_extensions as te`?

tomchristie (approved):
Sure. Thanks.

adriangb:
Ups looks like https://github.com/encode/httpx/pull/2467#discussion_r1035086504 was too late. I'll still check and we can always follow up in a PR adding the dependency if needed.

adriangb:
I tested Flask and Falcon by creating an app and passing it to `WSGITransport`:

```python
from httpx import WSGITransport
from flask import Flask
from falcon import App
WSGITransport(app=Flask(__name__))
WSGITransport(app=App())
```

No errors from mypy, even in `--strict` mode. So I think the `Callable` will be just fine 😄

httpx/_transports/wsgi.py
```diff
@@ -1,14 +1,31 @@
...
+# backported wsgiref.types definitions from Python 3.11
+StartResponse = typing.Callable[
+    [str, typing.List[typing.Tuple[str, str]], typing.Optional[_OptExcInfo]],
+    typing.Callable[[bytes], object],
+]
+
+
+WSGIApplication = typing.Callable[
+    [typing.Dict[str, typing.Any], StartResponse], typing.Iterable[bytes]
+]
```
michaeloliverx, line 25: Hmm shouldn't we just import these directly from typeshed instead?
E.g.
```python
if typing.TYPE_CHECKING:
    from _typeshed import OptExcInfo
    from _typeshed.wsgi import WSGIApplication
```

PR #2468, opened 2022-11-29
Typing: always fill in generic type parameters

adriangb:
Being explicit about the parameters helps find bugs and makes the library easier to use for users.

- Tell mypy to disallow generics without parameter values
- Give all generic types parameters values


tomchristie (approved):
Mixed feelings about bits where the typing becomes super verbose, but... sure. 🤷‍♂️

Issue #2436, opened 2022-11-11
Stricter typing.


tomchristie:
Several of our users work with `httpx` using tools that have stricter type checking than we currently use.

We've had several pull requests and discussions around this, for example pull requests #2435, #2404, #2266.

I'd like us to take an incremental approach to considering stricter typing throughout the package. This will ensure that we have a series of tightly targeted, reviewable pull requests, rather than bundling too many different changes all up into one.

Here are the various flags that `mypy --strict` enables. I've marked the ones that we currently have enabled...

**Note**: Updating this list as we go.

- [x] `--warn-unused-configs`
- [x] `--disallow-any-generics`
- [x] `--disallow-subclassing-any`
- [x] `--disallow-untyped-calls`
- [x] `--disallow-untyped-defs`
- [x] `--disallow-incomplete-defs`
- [x] `--check-untyped-defs`
- [x] `--disallow-untyped-decorators`
- [x] `--no-implicit-optional`
- [x] `--warn-redundant-casts`
- [x] `--warn-unused-ignores`
- [x] `--warn-return-any`
- [x] `--no-implicit-reexport`
- [x] `--strict-equality`
- [x] `--strict-concatenate`

I'd suggest that an approach along these lines:

* Individually targeted pull requests which turn on a *single* `mypy` flag, and make the required changes to keep the tests passing.
* If we get all of these passing, then switching to enabling `mypy --strict`.
* Once we've done that we can assess if there are any remaining issues with `pyright`'s strict mode, and consider if/how we want to address those.

PR #2435 enables the `disallow_any_generics = True` flag. That looks like it could be a good starting point.

https://github.com/encode/httpx/pull/2435/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52

mjpieters:
The project is in a pretty good shape, I must say. Out of the box, you can enable these settings already:

- [x] `warn_unused_configs = True`
- [x] `disallow_subclassing_any = True`
- [x] `check_untyped_defs = True`
- [x] `disallow_untyped_decorators = True`
- [x] `warn_redundant_casts = True`
- [x] `strict_concatenate = True`

and `scripts/check` passes cleanly with no other changes required.

I'd enable `warn_unused_ignores = True` too, and remove the 10 `type: ignore` comments that that then highlights. Those comments do nothing, according to mypy. It's a basic code hygiene check, like Flake8 telling you about unused imports.

[...]

PR #2469, opened 2022-11-29
Typing: add explicit types for public fields

adriangb:
Without these, type inference would be required, the output of which can differ between type checkers. Explict fields also serves as documentation and can be a place to add explicit field descriptions in future.

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -305,6 +305,12 @@ class Headers(typing.MutableMapping[str, str]):
 
 
 class Request:
+    method: str
+    url: URL
+    headers: Headers
+    extensions: RequestExtensions
+    stream: typing.Union[SyncByteStream, AsyncByteStream]
+
     def __init__(
         self,
         method: typing.Union[str, bytes],
@@ -443,6 +449,16 @@ class Request:
 
 
 class Response:
+    status_code: int
+    headers: Headers
+    next_request: typing.Optional[Request]
+    extensions: ResponseExtensions
+    history: typing.List["Response"]
+    is_closed: bool
+    is_stream_consumed: bool
+    default_encoding: typing.Union[str, typing.Callable[[bytes], str]]
+    stream: typing.Union[SyncByteStream, AsyncByteStream]
+
     def __init__(
         self,
         status_code: int,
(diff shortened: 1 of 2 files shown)


adriangb:
Fair enough, happy to pause for now

adriangb:
I was thinking about what sorts of changes can happen and what would get caught or not.

## 1. New field added to

A new parameter is added to `__init__` that becomes a field. It's inferred correctly by mypy and we don't add it. Most users aren't impacted but these explicit fields slowly loose value as documentation.

## 2. Type of field changed in `__init__`

This would immediately be caught by mypy or any other type checker as soon as we try to assign to it.

## 3. Field deleted

 This would not be caught by any type checker that I know of and if we don't access the attribute in tests we'd be shipping a bug (we promised it exists but doesn't).

httpx/_transports/wsgi.py
```diff
@@ -80,6 +80,12 @@ class WSGITransport(BaseTransport):
     ```
     """
 
+    app: typing.Callable[..., typing.Any]
+    raise_app_exceptions: bool
+    script_name: str
+    remote_addr: str
+    wsgi_errors: typing.Optional[typing.TextIO]
```
tomchristie, line 87: We don't *really* want these to be public API.

The `HTTPTransport` class has it's implementation attributes use a private namespace, but the `WSGITransport` and `ASGITransport` classes currently don't.

httpx/_models.py
```diff
@@ -1007,6 +1023,8 @@ class Cookies(typing.MutableMapping[str, str]):
     HTTP Cookies, as a mutable mapping.
     """
 
+    jar: CookieJar
+
```
tomchristie, line 1027: Hrm. Restructuring `__init__` would probably stop this being ambiguous to `pyright`.

The `__init__` method should probably start with `self.jar = CookieJar()` before the branching cases. That'd seem easier to reason about from my perspective, as well as from `pyright`'s.

httpx/_models.py
```diff
@@ -305,6 +305,12 @@ def __repr__(self) -> str:
 
 
 class Request:
+    method: str
+    url: URL
+    headers: Headers
+    extensions: RequestExtensions
+    stream: typing.Union[SyncByteStream, AsyncByteStream]
```
tomchristie, line 312: Which of these are ambiguous to `pyright`? Some? All?

tomchristie:
If y'all are keen on having `pyright` pass cleanly, maybe we should start out with a discussion or issue similar to #2436, so we can make sure to tackle this fully.

tomchristie:
Going to close this off as stale.

PR #2471, opened 2022-11-29
use `# pragma: no cover` instead of `# pragma: nocover`

adriangb:
We have a mix of both. Although coverage [accepts both](https://coverage.readthedocs.io/en/6.3.3/changes.html?highlight=nocover#version-2-7-2007-07-21), the `no cover` version is the one [mentioned in the docs](https://coverage.readthedocs.io/en/6.3.3/excluding.html?highlight=no%20cover#excluding-code-from-coverage-py)

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -46,7 +46,7 @@ from ._urls import URL, QueryParams
 
 try:
     from ._main import main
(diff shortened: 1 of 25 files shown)


tomchristie (approved):
Okay.

PR #2472, opened 2022-11-29
Add more mypy flags to increase type checking strictness

adriangb:


tomchristie (approved):
Yup.

Updated the list in #2436 to reflect this.

PR #2474, opened 2022-11-29
Add strict_equality = True to mypy config

adriangb:


tomchristie:
Looks like we should go with #2480 here...

* `def test_wsgi_server_port(url: str, expected_server_port: str) -> None:` is a little neater.
* I like [the comment](https://github.com/encode/httpx/pull/2480/files#diff-f8941bdcbd6dbab82ab257551cc2d3720bb2f0670b94df1f61525437f9c94dfeR5) in the `test_status_code_as_int` function.

tomchristie:
Merged #2480

PR #2475, opened 2022-11-29
Typing: Enable no_implicit_reexport

mjpieters:
- Add explicit exports for names imported from _compat and _models
- Correct import of the uvicorn Server class

Part of #2436

### DIFF
--- a/httpx/_compat.py
+++ b/httpx/_compat.py
@@ -38,3 +38,6 @@ else:
         context.options |= ssl.OP_NO_SSLv3
         context.options |= ssl.OP_NO_TLSv1
         context.options |= ssl.OP_NO_TLSv1_1
+
+
+__all__ = ["brotli", "set_minimum_tls_version_1_2"]
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -7,8 +7,9 @@ from pathlib import Path
 import certifi
 
 from ._compat import set_minimum_tls_version_1_2
-from ._models import URL, Headers
+from ._models import Headers
 from ._types import CertTypes, HeaderTypes, TimeoutTypes, URLTypes, VerifyTypes
+from ._urls import URL
 from ._utils import get_ca_bundle_from_env, get_logger
 
 DEFAULT_CIPHERS = ":".join(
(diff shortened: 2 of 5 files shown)


httpx/_models.py
```diff
@@ -41,7 +41,7 @@
     ResponseExtensions,
     SyncByteStream,
 )
-from ._urls import URL
+from ._urls import URL as URL
```
tomchristie, line 44: I'm curious why `mypy` requires us to do this weird dance here.
Maybe we don't care why, and we just do it anyway?
Or?

tomchristie: Ah gotcha.

We're importing `URL` from the wrong place, once in `_config.py` and twice in `_utils.py`. Let's fix those instead of making this change here...

```diff
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -6,8 +6,9 @@ from pathlib import Path
 import certifi
 
 from ._compat import set_minimum_tls_version_1_2
-from ._models import URL, Headers
+from ._models import Headers
 from ._types import CertTypes, HeaderTypes, TimeoutTypes, URLTypes, VerifyTypes
+from ._urls import URL
 from ._utils import get_ca_bundle_from_env, get_logger
 
 DEFAULT_CIPHERS = ":".join(
```

And...

```diff
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -16,7 +16,7 @@ import sniffio
 from ._types import PrimitiveData
 
 if typing.TYPE_CHECKING:  # pragma: no cover
-    from ._models import URL
+    from ._urls import URL
 
 
 _HTML5_FORM_ENCODING_REPLACEMENTS = {'"': "%22", "\\": "\\\\"}
@@ -465,7 +465,7 @@ class URLPattern:
     """
 
     def __init__(self, pattern: str) -> None:
-        from ._models import URL
+        from ._urls import URL
 
         if pattern and ":" not in pattern:
             raise ValueError(
```

tomchristie (changes requested):
The `from ._urls import URL as URL` indicates that we've got some incorrect imports.

Let's fix those rather than re-exporting `URL`.

tomchristie (approved):

PR #2476, opened 2022-11-29
Typing: enable disallow_incomplete_defs

mjpieters:
The only places mypy reports issues is in the test suite.

Part of the work for #2436


tomchristie (approved):
These all look fine, yup.

PR #2478, opened 2022-11-30
Use wsgi types from typeshed

michaeloliverx:
Since these types already exist in typeshed I see no reason to maintain them ourselves. What is everyone thoughts?

See docs on importing typeshed types https://github.com/python/typeshed/tree/main/stdlib/_typeshed.

### DIFF
--- a/httpx/_transports/wsgi.py
+++ b/httpx/_transports/wsgi.py
@@ -1,28 +1,17 @@
 import io
 import itertools
 import sys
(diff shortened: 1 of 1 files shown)


tomchristie (approved):
Looks like a neater approach to me, yup.

michaeloliverx:
Refs https://github.com/encode/httpx/pull/2467

PR #2479, opened 2022-11-30
Typing: enable disallow_untyped_calls

mjpieters:
Only the test suite needed adjusting to add type hints.

Part of the work for #2436

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,21 +1,27 @@
 import os
+import typing
 
 from click.testing import CliRunner
 
 import httpx
+from httpx import main
+
+if typing.TYPE_CHECKING:  # pragma: no cover
+    # don't let mypy be misled by the fallback defined in httpx/__init__.py
+    from httpx._main import main  # noqa: F811
 
 
-def splitlines(output):
+def splitlines(output: str) -> typing.Iterable[str]:
     return [line.strip() for line in output.splitlines()]
 
 
-def remove_date_header(lines):
+def remove_date_header(lines: typing.Iterable[str]) -> typing.Iterable[str]:
     return [line for line in lines if not line.startswith("date:")]
 
 
 def test_help():
     runner = CliRunner()
-    result = runner.invoke(httpx.main, ["--help"])
+    result = runner.invoke(main, ["--help"])
     assert result.exit_code == 0
     assert "A next generation HTTP client." in result.output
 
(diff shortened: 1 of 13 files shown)


tests/test_main.py
```diff
@@ -1,29 +1,31 @@
 import os
+import typing
 
 from click.testing import CliRunner
 
 import httpx
+from httpx._main import main
```
tomchristie, line 7: What's prompted this change?

Can we continue to use `import httpx` and `httpx.main(...)` so that we're not importing private namespaces?

mjpieters: Mypy then sees the `def main()` fallback defined in `httpx/__init__.py`, which is decidedly not a `click.BaseCommand` or subclass, so all the calls to `httpx.main()` then need a cast. Which is unfortunate.

I can see if we can't trick mypy by using a `TYPE_CHECKING` block that does the private import and leave the test using the normal, public import.

tomchristie: > Without this, mypy sees the def main() fallback defined in httpx/__init__.py, which is decidedly not a click.BaseCommand or subclass, so all the calls to httpx.main() then need a typing.cast(click.BaseCommand, httpx.main). Which is unfortunate.

Ah, gotcha. You're referring to this...

https://github.com/encode/httpx/blob/57aa5b08a48c82a6e1a1500e131d94a31128b6f9/httpx/__init__.py#L47-L59

It looks to me like we should type the arguments on that.

```python
def main(*args: str) -> None:
    ...
```

Would that resolve the issue?

tomchristie: Oh right, no it wouldn't.

Hrm. Perhaps we can put it inside a conditional `if not TYPE_CHECKING` block?

tomchristie: Or else we just suck up the gnarly private import for now.

Perhaps that okay.

mjpieters: I've put the import into a TYPE_CHECKING guard with a comment. The test is against the public import. This feels marginally better. Whadayathink?

tests/test_main.py
```diff
@@ -1,29 +1,35 @@
...
+import typing
 
 from click.testing import CliRunner
 
 import httpx
+from httpx import main
 
+if typing.TYPE_CHECKING:  # pragma: no cover
+    # don't let mypy be misled by the fallback defined in httpx/__init__.py
+    from httpx._main import main  # noqa: F811
```
tomchristie, line 11: I like what you've done here - we've got other cases where the `TYPE_CHECKING` imports from private namespaces, so I think this is okay.

tomchristie (approved):

PR #2481, opened 2022-11-30
Add back in URL.raw with NamedTuple

### LINKED ISSUE #2462: Breaking drop of URL.raw in patch release 0.23.1
The patch release `0.23.1` included a breaking change in #2241 

The justification seems to be that "we don't use it", but `url.raw` is a public property used by other libraries including `ddtrace` 0.x versions: https://github.com/DataDog/dd-trace-py/blob/v0.61.3/ddtrace/contrib/httpx/patch.py#L41 

Can this breaking change be reverted and maybe soft-deprecated? @tomchristie @michaeloliverx

developerDemetri:
Closes #2462 with a basic test, happy to add tests for more scenarios

### DIFF
--- a/httpx/_types.py
+++ b/httpx/_types.py
@@ -16,6 +16,7 @@ from typing import (
     Iterator,
     List,
     Mapping,
+    NamedTuple,
     Optional,
     Sequence,
     Tuple,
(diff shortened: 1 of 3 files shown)


tomchristie (approved):
Yup, I'm happy with this.

We *could* also add a comment on the `.raw()` method metioning that it is due to be deprecated in a future release, but I don't mind too much either way, so wouldn't consider it a blocker.

tomchristie:
Okay, let's go with this.

We'll want a new minor point release.

PR #2463, opened 2022-11-25
reuse the digest auth state to avoid unnecessary requests

rettier:
this implements digest authentication state reuse as discussed in #1467.

Contrary to the [comment](https://github.com/encode/httpx/issues/1467#issuecomment-1280656164) i am not skipping most of the authentication flow if a previous challenge exists. Otherwise a failed authentication (401) would not result in a retry with the new server challenge (`test_digest_auth_resets_nonce_count_after_401` would fail).

PR #2510, opened 2022-12-20
Version 0.23.2

tomchristie:
### Added

* Support digest auth nonce counting to avoid multiple auth requests. (#2463)

### Fixed

* Multipart file uploads where the file length cannot be determine now use chunked transfer encoding, rather than loading the entire file into memory in order to determine the `Content-Length`. (#2382)
* Raise `TypeError` if content is passed a dict-instance. (#2495)
* Partially revert the API breaking change in 0.23.1, which removed `RawURL`. We continue to expose a `url.raw` property which is now a plain named-tuple. This API is still expected to be deprecated, but we will do so with a major version bump. (#2481)


Secrus (approved):

bersace:
:up: Any update on this topic ? (and a happy new year 🎊)

tomchristie:
@Kludex - Oh yup, missed that one.

@bersace - Thanks for the prompt - let's get this out now.

tomchristie:
Done. https://pypi.org/project/httpx/0.23.2/

PR #2511, opened 2022-12-20
Update pytest-asyncio

tomchristie:
Currently causing tests to fail


Kludex (approved):
As we already have anyio, we can make a replacement I guess?

tomchristie:
> As we already have anyio, we can make a replacement I guess?

I don't know? 🤷‍♂️

PR #2512, opened 2022-12-20
replace pytest-asyncio and pytest-trio with anyio

graingert:
- [x] Initially raised as PR comment: https://github.com/encode/httpx/pull/2511#pullrequestreview-1224512355


tests/client/test_auth.py
```diff
@@ -152,7 +152,7 @@ async def async_auth_flow(
         yield request
 
 
-@pytest.mark.asyncio
+@pytest.mark.anyio
```
graingert, line 155: is there a reason these were not marked `@pytest.mark.usefixtures("async_environment")` ?

michaeloliverx:
One less dependency gets my vote :)

Kludex:
As a note, because @tomchristie mentioned before on Starlette that was not necessary, using `pytest.mark.anyio` without defining a `anyio_backend` fixture will run every test with both `asyncio` and `trio`.

I'm just mentioning, I'm not expressing my opinion. JFYK

graingert:
This needs https://github.com/encode/starlette/pull/1932/files

setup.cfg
```diff
@@ -18,7 +18,7 @@ combine_as_imports = True
 addopts = -rxXs
 filterwarnings =
   error
-asyncio_mode = strict
+  ignore: You seem to already have a custom sys.excepthook handler installed. I'll skip installing Trio's custom handler, but this means MultiErrors will not show full tracebacks.:RuntimeWarning
```
tomchristie, line 21: What's causing this warning, and when can we expect to be able to remove this filter?

tomchristie (approved):

PR #2515, opened 2022-12-26
Support Enum query params

Kludex:
Raised by https://github.com/mlrun/mlrun/pull/2629.


httpx/_urls.py
```diff
@@ -554,7 +554,6 @@ def __init__(
 
         value = args[0] if args else kwargs
 
-        items: typing.Sequence[typing.Tuple[str, PrimitiveData]]
```
Kludex, line 557: This was unused.

tomchristie:
Should we support auto-casting enums here, or should we raise a type error if you pass anything other than a `boolean`/`int`/`float`/`None`?

Option 1.

*Explicit type error*

```python
import enum
import httpx

class Color(enum.Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

u = httpx.URL("https://example.com", params={"color": Color.RED.value})
print(u)  # https://example.com?color=1

u = httpx.URL("https://example.com", params={"color": Color.RED})
# TypeError: Invalid type for params value. Expected str, bool, int, or float, got <enum 'Color'>: <Color.RED: 1>
```

Option 2.

*Silent type coercion*

```python
import enum
import httpx

class Color(enum.Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

u = httpx.URL("https://example.com", params={"color": Color.RED.value})
print(u)  # https://example.com?color=1

u = httpx.URL("https://example.com", params={"color": Color.RED})
print(u)  # https://example.com?color=1
```

httpx/_utils.py
```diff
@@ -67,6 +68,8 @@ def primitive_value_to_str(value: "PrimitiveData") -> str:
         return "false"
     elif value is None:
         return ""
+    elif isinstance(value, enum.Enum):
+        return str(value.value)
```
tomchristie, line 72: Does this imply that the `PrimitiveData` type ought to change to include `enum.Enum`?

tomchristie:
Currently both `requests` and `httpx` will perform silent type coercion to `str` here, but neither special-cases the `enum` type, so both give an awkward behaviour...

```python
>>> r = requests.get("https://www.example.com", params={"a": Color.RED})
>>> r.request.url
'https://www.example.com/?a=Color.RED'
```

Kludex:
> Should we support auto-casting enums here, or should we raise a type error if you pass anything other than a boolean/int/float/None?

Hmm... Maybe this would be better:

```python
    if isinstance(value, enum.Enum):
        return primitive_value_to_str(value.value)
```
Then it would support all those cases. :thinking:

httpx/_utils.py
Kludex: Yeah, I miss that.

Do you think that `enum.Enum` should not be considered a primitive data, and hence not be in this function?

tomchristie: My honest answer?
My preference would be for us to raise nice tight `TypeError` cases here.

tomchristie: For a similar example...

```python
import json

json.dumps{"value": Color.RED}        # TypeError: Object of type Color is not JSON serializable
json.dumps{"value": Color.RED.value}  # '{"value": 1}'
```

Kludex: The thing is... I think the following should raise an error:
```python
class Color(enum.Enum):
    RED = 1
    GREEN = 2
    BLUE = 3
```

But not the following:

```python
class Color(int, enum.Enum):
    RED = 1
    GREEN = 2
    BLUE = 3
```
Because the `int` there means that each element is an integer, analogous to other types.

tomchristie: Hrm. That's a bit awkward...

```python
class Color(str, enum.Enum):
    # heeeeyyyyyy I'm just a normal str
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

c = Color.RED
isinstance(c, str)  # True
str(c)              # 'Color.RED'  yeah but you ain't
```

The upshot is *even if you're type-checking* the user still needs to correctly pass `Color.RED.value`, unless we special-case for `enum`.

Maybe the upshot is we should have strict type checking here *and* that we should also expand the primitive types to include `enum.Enum`. (Just because we're friendly like that?)

Kludex: > The upshot is even if you're type-checking the user still needs to correctly pass Color.RED.value, unless we special-case for enum.

If I understood correctly what you said, this is not the case. Type checkers understand that each member of a (str, Enum) is a str, and they don't raise errors.

Kludex: I don't think requests has the same behavior as httpx here, does it? Because the Testclient changed, so I assume requests can do the "right" thing on (str, Enum) - I'm on my phone.

tomchristie:
Okay, I've some reluctance on this, which I'm more clear now on how to phrase.

It's about consistency.

```python
class SiteURLs(str, enum.Enum):
    staging = "https://localhost"
    production = "https://www.example.com"

httpx.URL(SiteURLs.staging)
# Nope, not going to do what you want.
# You meant to pass `SiteURLs.staging.value`.

class ContentType(str, enum.Enum):
    json = "application/json"
    form = "application/x-www-form-urlencoded"

httpx.Headers({"Content-Type": ContentType.json})
# Nope, not going to do what you want.
# You meant to pass `ContentType.json.value`.
```

Same with, say `json.dumps({...})` and other APIs.

It's not obvious to me that we should be special casing `Enum` in QueryParams, when the less ambiguous API usage is for the user to pass the `.value` from enum instances.

On the other hand... (?)

> Because the Testclient changed

I saw something along those lines mentioned in the referenced ticket... but I wasn't clear what it meant, and I couldn't replicated with `requests`. For example:

```python
>>> r = requests.get("https://www.example.com", params={"a": Color.RED})
>>> r.request.url
'https://www.example.com/?a=Color.RED'  # Ooops you meant `Color.RED.value`. Same as with `httpx`.
```

Kludex:
@tomchristie did your thoughts regarding this changed or should I close it?

tomchristie:
It's really not obvious what behaviour you should expect when using a `StrEnum` if you *don't* access `.value`. For example...

```python
class CustomEnum(str, enum.Enum):
    A = "a"
    B = "b"

"%s" % CustomEnum.A
'CustomEnum.A'
"".join([CustomEnum.A])
'a'
```

😬

Anyways we probably just need to suck this weirdness up and deal with it...

```python
def primitive_value_to_str(value: "PrimitiveData") -> str:
    """
    Coerce a primitive data type into a string value.
    Note that we prefer JSON-style 'true'/'false' for boolean values here.
    """
    if value is True:
        return "true"
    elif value is False:
        return "false"
    elif value is None:
        return ""
    elif isinstance(value, (str, int)):
        if isinstance(value, enum.Enum):
            # typed enums have some pretty inconsistent
            # behaviour around expectations when passed directly.
            # See ...
            # We could either raise an exception here and force
            # developers to pass `.value` directly, or just coerce it.
            # We're being soft on ya here...
            return str(value.value)
        return str(value)

    # Passing other types here isn't type-correct against our API,
    # but we do currently automatically coerce them.
    # There has been some discussion about removing this in favour
    # stricter typing. See this reverted pull request for more details...
    # https://github.com/encode/httpx/pull/2523
    return str(value)
```

tomchristie:
Note...

Being careful about placing the `isinstance(value, enum.Enum)` inside the `(str, int)` check means you'll get the same behaviour as `requests` or the previous starlette test client wrt. typed vs untyped enums.

```python
    ...
    elif isinstance(value, (str, int)):
        if isinstance(value, enum.Enum):
    ...
```

Kludex:
Why only `int` and `str`?

httpx/_utils.py
```diff
@@ -67,6 +68,8 @@ def primitive_value_to_str(value: "PrimitiveData") -> str:
...
     return str(value)
```
florimondmanca, line 73: I think I'd be happier if HTTPX had an extra branch `if isinstance(value, str): return value` here, and then raised an explicit `TypeError` for all remaining cases. 

For `Enum`, we could _assume_ that people want `value.value` to be used, but why assume and potentially cause unexpected behavior when we can make users provide exactly what they want with a few keystrokes?

Is my thinking, at least

Kludex: The `if isinstance(value, str): return value` solves the issue that I want to solve here.

Should I proceed? Aren't we going to have some side effects?

Kludex: It was kind of what happened on https://github.com/encode/httpx/pull/2523/, and we had to revert.

tomchristie:
Closing as stale

PR #2517, opened 2022-12-28
Expose transport retries as `connect_retries` in client options

zanieb:
Opening this draft of previously discussed addition of an option to set connect retries on the client.

Next steps:
- [ ] Pass type checking in CI
- [ ] Agree on a testing strategy
- [ ] Pass tests in CI


madkinsz:
See https://github.com/encode/httpcore/pull/643 to resolve type errors (outside of tests) by adding `retries` to SOCKS proxy classes.

tomchristie:
Ah neat.

So...

I don't think we need test cases where what we're doing is passing through an argument, and then inspecting the private implementation details to see that it's been set. It should be enough that we have type-checking throughout. The `retries` API is provided by `httpcore`, and is tested by `httpcore`. Separation of concerns etc.etc.

I wonder if we should be avoiding too much top-level configuration API on the client. Perhaps `retries` or `connection_retries` should be in the `Limits(...)` config?

madkinsz:
> I don't think we need test cases where what we're doing is passing through an argument, and then inspecting the private implementation details to see that it's been set.

I agree this is really awkward — however, since the behavior differs if you provide your own transport I think we need a little bit of test coverage. Perhaps we should throw an error if you provide a value for this _and_ a transport.

> I wonder if we should be avoiding too much top-level configuration API on the client. Perhaps retries or connection_retries should be in the Limits(...) config?

I agree it'd be nice to keep things out of the top-level. Limits feels like a weird place though — everything there is non-zero by default. I don't feel like this setting is adjusting a reasonable upper bound (as limits imply), it's enabling behavior.

madkinsz:
My need for this is not pressing as I've just implemented wrapper around the client but I'd love to keep discussing the best way to handle this when you have time!

PR #2523, opened 2022-12-30
Raise `TypeError` on invalid query params.

### LINKED ISSUE #746: Support passing bytes keys/values in `params=...`
``` python
>>> httpx.get('https://example.org', params={b'testparam': b'testvalue'}).url
URL('https://example.org?b%27testparam%27=b%27testvalue%27')
```

I don't think this is intentional, as there is an isinstance check for bytes in the ``flatten_queryparams`` function, I'd expect anyone involved to have raised an exception instead of letting it go through like that, and that would be an undocumented divergence from requests compatibility.

I have tested with ``httpx==0.7.8`` as well as ``httpx==0.11.0``.

tomchristie:
Prompted by #2515. (Doesn't presume any particular resolution to that case.)

Closes #746 - we don't support bytes (or other unexpected types) as QueryParam values. Let's raise a clear error, rather than coercing an unexpected `str` value.

Before this change...

```python
>>> httpx.URL("http://www.example.com/", params={"query": b"abc"})
URL('http://www.example.com/?query=b%27abc%27')
```

After this change...

```python
>>> httpx.URL("http://www.example.com/", params={"query": b"abc"})
TypeError: Expected str, int, float, bool, or None. Got <class 'bytes'>.
```

### DIFF
--- a/tests/models/test_queryparams.py
+++ b/tests/models/test_queryparams.py
@@ -88,7 +88,9 @@ def test_empty_query_params():
 
 
 def test_invalid_query_params():
-    with pytest.raises(TypeError, match=r"Expected str, int, float, bool, or None\. Got 'bytes'\."):
+    with pytest.raises(
+        TypeError, match=r"Expected str, int, float, bool, or None\. Got 'bytes'\."
+    ):
         httpx.QueryParams({"a": b"bytes"})
 
 
--- a/tests/models/test_queryparams.py
+++ b/tests/models/test_queryparams.py
@@ -87,6 +87,11 @@ def test_empty_query_params():
     assert str(q) == "a="
 
 
+def test_invalid_query_params():
+    with pytest.raises(TypeError):
+        httpx.QueryParams({"a": b"bytes"})
+
+
 def test_queryparam_update_is_hard_deprecated():
     q = httpx.QueryParams("a=123")
     with pytest.raises(RuntimeError):
--- a/tests/models/test_queryparams.py
+++ b/tests/models/test_queryparams.py
@@ -88,7 +88,7 @@ def test_empty_query_params():
 
 
 def test_invalid_query_params():
-    with pytest.raises(TypeError):
+    with pytest.raises(TypeError, match=r"Expected str, int, float, bool, or None\. Got 'bytes'\."):
         httpx.QueryParams({"a": b"bytes"})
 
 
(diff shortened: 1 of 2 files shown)


tests/models/test_queryparams.py
```diff
@@ -87,6 +87,11 @@ def test_empty_query_params():
     assert str(q) == "a="
 
 
+def test_invalid_query_params():
+    with pytest.raises(TypeError):
```
madkinsz, line 91: Seems nice to ensure that the formatted message is correct
```suggestion
    with pytest.raises(TypeError, match=r"Expected str, int, float, bool, or None\. Got 'bytes'\."):
```

madkinsz (approved):

tests/models/test_queryparams.py
tomchristie: That's neat, thanks.

madkinsz: Sorry I broke your line length :)

tomchristie: 🤨

tests/models/test_queryparams.py
```diff
@@ -89,7 +89,7 @@ def test_empty_query_params():
 
 def test_invalid_query_params():
     with pytest.raises(
-        TypeError, match=r"Expected str, int, float, bool, or None\. Got 'bytes'\."
+        TypeError, match=r"Expected str, int, float, bool, or None. Got 'bytes'."
```
madkinsz, line 92: You don't need the `r` if you aren't escaping the periods — this is doing a `re.search` so technically the `.` will be a wildcard but 🤷‍♀️ it's not likely to cause you any problems. You could use `re.escape` if you really wanted to be correct.

madkinsz:
@tomchristie this broke Prefect since we were passing UUID objects — that's fine but I think we should make sure this is listed on the changelog for 0.23.2 for visibility.

tomchristie:
Ooofff. Okay, more care needed in the future.

tomchristie:
That's entirely on me - rushing things that didn't need to be rushed.

Do we want to roll this back in a 0.23.3, and leave it for the upcoming 0.24.0 update? (Or do we leave it as is?)

madkinsz:
It'd probably be safest to make this change in 0.24.0 instead since it changes the supported interface (I'll try to keep an eye out for this in future reviews), but I don't have strong feelings about rolling it back.

curtiscook:
> @tomchristie this broke Prefect since we were passing UUID objects — that's fine but I think we should make sure this is listed on the changelog for 0.23.2 for visibility.

FYI - This also breaks datetime -> str conversion

PR #2525, opened 2022-12-31
Drop private import of 'encode_request' in test_multipart

tomchristie:
Refs #2492 

This is just a test refactoring.

Use `httpx.Request(...)` to handle the multipart tests, rather than importing the `encode_request` function, which is a private implementation detail. Having tests against our public API makes more sense than digging into private imports.


michaeloliverx (approved):
Makes sense to dog food the public API.

Issue #2088, opened 2022-02-18
follow_redirects and  `.netrc` credentials


jjagielka:
`httpx` uses `.netrc` credentials for automatic login when http call is triggered, but it does not apply `.netrc` credentials when doing redirecting (`follow_redirects=True`)

Site 1 -> Site 2 -> Site 3
If the authorization is required on Site2, `follow_redirects` won't work even if `.netrc` contains an entry for that site.

`requests` does apply `.netrc` for traversed sites.

This is especially useful if  the `Site 2` above is a Single Sign On server.

_Originally posted by @jjagielka in https://github.com/encode/httpx/discussions/2087_

nkitsaini:
Hi, I'm up for fixing this. I'll be back once I've taken a good look at code to discuss potential solution.

tomchristie:
I'd suggest that a good approach to starting this off would be...

* Describe a complete, simple example.
* Write a test case in `tests/client/test_auth.py`, using `MockTransport` to mock the example.
* Review at this point, so we can be confident that we've properly described how we'd like the behaviour to change.
* Work on the code changes.

Working on the code changes themselves probably needs to come last, only once we've got this more precisely identified first.

tomchristie:
Okay. So... I think having an explicit `NetRCAuth` class helps clarify *a little bit here*, because it shows that we *have the same kind of behaviour for all our auth classes in this regard.*

Which is...

When a cross-domain redirect occurs, we remove the `Authorisation` header, and don't automatically re-apply it.

If you want a redirect to reapply the auth, then you'd need to handle that explicitly.

If someone would like to help out with documenting this, then I suppose a good start would be to write up an example in this thread which shows how to deal with redirects explicitly.

A starting point for figuring that out would be this part of the docs...

https://www.python-httpx.org/quickstart/#redirection-and-history

You'd want to use `request.has_redirect_location`, `request.next_request` and `Client.send(request, auth=...)`.

tomchristie:
Given that we now have an explicit `NetRCAuth` class, and that our cross-domain-redirects behaviour is the same for *all* our auth classes, I think we can close this one off.

PR #2535, opened 2023-01-02
Add `NetRCAuth()` class.

### LINKED ISSUE #2532: NetRCInfo.netrc_info() doesn't work when run in a systemd service with DynamicUser=true
### Discussed in https://github.com/encode/httpx/discussions/2526

<sup>Originally posted by **djmattyg007** January  1, 2023</sup>
The function in question, `netrc_info()`, iterates over three paths. Two of these paths refer to the home directory of the current process' user, by prefixing the paths with `~/`. These paths are run through `pathlib.Path.expanduser()`, which raises an exception if the home directory can't be determined.

When the code is running in a systemd service with `DynamicUser=true`, there is no home directory. This means `expanduser()` will always raise an exception.

Fortunately, there's a potential override: set the `NETRC` environment variable to something. As long as it points to a file that exists, and doesn't need to know the current user's home directory, this is an acceptable workaround.

Unfortunately, there's an `is_file()` check performed on the `Path` object before attempting to use it. This means I can't just set `NETRC=/dev/null` as a simple way of working around this problem, even though `netrc.netrc()` will happily accept `/dev/null`.

Why does any of this matter? Because I'm not using `httpx` directly. I'm using the library `python-telegram-bot`, and have no control over how it uses `httpx` internally.

To summarise, there are two problems:

- The call to `pathlib.Path.expanduser()` can raise an exception that isn't currently being caught
- Setting `NETRC=/dev/null` doesn't work as expected, despite it being a perfectly suitable file to use in this context

I'm happy to raise an issue for this, and also to implement a PR to resolve this.</div>

tomchristie:
Closes #2532

Prompted by https://github.com/encode/httpx/issues/2532#issuecomment-1368907498

We currently automatically handle [`netrc` authentication](https://docs.python.org/3/library/netrc.html), which is baked directly into the `Client`, and which is enabled unless the developer sets `trust_env=False`.

* Browsers, which do *not* use `netrc` authentication.
* The `curl` command line client, which *only* uses `netrc` authentication if explicitly enabled.

The suggest here is that we should make a behavioural change, and *no longer* bake `netrc` authentication directly into the client, in favor of providing an explicit `NetRCAuth()` class.

Some usage examples...

```python
# Use default netrc file
auth = httpx.NetRCAuth()
client = httpx.Client(auth=auth)

# Use explicit netrc file
auth = httpx.NetRCAuth(file="/path/to/netrc")
client = httpx.Client(auth=auth)

# Optional environ override, fallback to default netrc file otherwise
auth = httpx.NetRCAuth(file=os.environ.get('NETRC'))
client = httpx.Client(auth=auth)

# Mandatory environ override
auth = httpx.NetRCAuth(file=os.environ['NETRC'])
client = httpx.Client(auth=auth)

# Optional netrc auth, depending on if the "~/.netrc" file exists
try:
    auth = httpx.NetRCAuth()
except FileNotFound:
    auth = None
client = httpx.Client(auth=auth)
```

This would be a behavioral change and need to be part of a 0.24.0 release, but I think it's obviously more consistent and neater than our existing "netrc is a special case" behaviour.

*(I had also assumed that this change would resolve issue #2088, and added a test case for this, which showed me that my assumption there was incorrect. Failing test case for that now dropped in commits https://github.com/encode/httpx/pull/2535/commits/ebcf77ae4959f0e67c7fce070e7b79ab5b5bad2a and https://github.com/encode/httpx/pull/2535/commits/02120cccecf905a0f142b36a31cebf68974e9528.)*

**TODO:**

- [x] Drop baked-in client `netrc` auth class.
- [x] Implement `NetRCAuth` class.
- [x] Tests.
- [x] Update documentation.

Documentation link, for easier review... https://github.com/encode/httpx/blob/02120cccecf905a0f142b36a31cebf68974e9528/docs/advanced.md#netrc-support

### DIFF
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -1,4 +1,5 @@
 import hashlib
+import netrc
 import os
 import re
 import time
@@ -132,6 +133,34 @@ class BasicAuth(Auth):
     def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
         request.headers["Authorization"] = self._auth_header
         yield request
+
+    def _build_auth_header(
+        self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
+    ) -> str:
+        userpass = b":".join((to_bytes(username), to_bytes(password)))
+        token = b64encode(userpass).decode()
+        return f"Basic {token}"
+
+
+class NetRCAuth(Auth):
+    """
+    Use a 'netrc' file to lookup basic auth credentials based on the url host.
+    """
+
+    def __init__(self, file: typing.Optional[str]):
+        self._netrc_info = netrc.netrc(file)
+
+    def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
+        auth_info = self._netrc_info.authenticators(request.url.host)
+        if auth_info is None or auth_info[2] is None:
+            # The netrc file did not have authentication credentials for this host.
+            yield request
+        else:
+            # Build a basic auth header with credentials from the netrc file.
+            request.headers["Authorization"] = self._build_auth_header(
+                username=auth_info[0], password=auth_info[2]
+            )
+            yield request
 
     def _build_auth_header(
         self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -152,7 +152,7 @@ class NetRCAuth(Auth):
 
     def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
         auth_info = self._netrc_info.authenticators(request.url.host)
-        if auth_info is None or auth_info[2] is None:
+        if auth_info is None or not auth_info[2]:
             # The netrc file did not have authentication credentials for this host.
             yield request
         else:
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -428,42 +428,47 @@ with tempfile.NamedTemporaryFile() as download_file:
 
 ## .netrc Support
 
-HTTPX supports .netrc file. In `trust_env=True` cases, if auth parameter is
-not defined, HTTPX tries to add auth into request's header from .netrc file.
+HTTPX can be configured to use [a `.netrc` config file](https://everything.curl.dev/usingcurl/netrc) for authentication.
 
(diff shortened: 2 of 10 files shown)


tomchristie:
What do we think @encode/maintainers?

httpx/_auth.py
```diff
@@ -141,6 +142,34 @@ def _build_auth_header(
...
+    """
+    Use a 'netrc' file to lookup basic auth credentials based on the url host.
+    """
+
+    def __init__(self, file: typing.Optional[str]):
+        self._netrc_info = netrc.netrc(file)
+
+    def auth_flow(self, request: Request) -> typing.Generator[Request, Response, None]:
+        auth_info = self._netrc_info.authenticators(request.url.host)
+        if auth_info is None or auth_info[2] is None:
```
cdeler, line 155: I'm sorry, may I ask you why you are checking the password for `None`?

cdeler: I'm asking since previously you were checking 
```python
if credentials is not None:
    ...
```

and python documentation says ([link](https://docs.python.org/3/library/netrc.html#netrc.netrc.authenticators)) that 
> If the netrc file did not contain an entry for the given host, return the tuple associated with the ‘default’ entry. If neither matching host nor default entry is available, return None.

tomchristie: Thanks @cdeler.

The simple version of "why" is because I was blindly following our previous implementation details here.

I've had a look into it, and I don't think the password can be `None` here, but I think it can be the empty string. It looks to me like Python 3.11+ will allow the empty string for a missing password field in the `netrc` file, and the previous versions will raise a parse error for missing passwords.

I think the robust thing to do here would be...

```python
if auth_info is None or not auth_info[2]:
```

tomchristie:
Great review, thanks @cdeler - I've updated the PR to be more robust to the empty password case.

cdeler:
LGTM, but have 1 small nitpick:

Does it make sense to add a test + test fixture for the case when password is empty (it will add a test for the newly added check)?

cdeler (approved):

tomchristie:
> Does it make sense to add a test + test fixture for the case when password is empty (it will add a test for the newly added check)?

Done.

PR #2539, opened 2023-01-03
Revert "Raise `TypeError` on invalid query params. (#2523)"

madkinsz:
This reverts commit 4cbf13ece2e584b45b935df0a0c670e1863c4569 / PR #2523

See discussion at https://github.com/encode/httpx/discussions/2538 and in #2523 — this change is potentially breaking and should be considered for a future release instead (i.e. 0.24.0).

PR #2540, opened 2023-01-04
Version 0.23.3

tomchristie:
~~Depends on #2539~~

We've now had two recent cases of tidying up edges of the API *in minor point releases* which have led to breakages in user codebases. Both of these have been due to my lead. Calling these out so that we can tighten up our release process, and for me in particular to slow down on hitting go on releases.

Release draft is here... https://github.com/encode/httpx/releases/tag/untagged-9adfb8466a6a6537698a

---

### Fixed

* Version 0.23.2 accidentally included stricter type checking on query parameters. This shouldn've have been included in a minor version bump, and is now reverted. (#2523, #2539)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.23.3 (4th Jan, 2023)
+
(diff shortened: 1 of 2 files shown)


tomchristie:
Ready to go.

michaeloliverx (approved):

tomchristie:
Thanks @michaeloliverx.

PR #2541, opened 2023-01-04
Add date type to accepted param type

turgaysozen:
If there is **datetime** parameter in **httpx** request, it gets crashed with following error.
`TypeError: Expected str, int, float, bool, or None. Got 'datetime'.`

Adding **date** will be enough to fix this problem, actually converting **datetime** object to string is also helps.

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -10,6 +10,7 @@ import time
 import typing
 from pathlib import Path
 from urllib.request import getproxies
+from datetime import date
 
 import sniffio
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks.

We're rolling back that change. See #2539 and #2540.
Hopefully we'll have a fixed release on PyPI soon.

tomchristie:
Now released... https://pypi.org/project/httpx/0.23.3/

Thanks for your time and apologies for the breakage.

turgaysozen:
Thanks @tomchristie

PR #2571, opened 2023-02-05
Drop private imports from test_exceptions.py

florimondmanca:
More towards #2492


tests/test_exceptions.py
```diff
@@ -16,66 +14,39 @@ def test_httpcore_all_exceptions_mapped() -> None:
...
-        "httpcore.ConnectionPool.handle_request",
-        return_value=httpcore.Response(
-            200, headers=[], content=TimeoutStream(), extensions={}
-        ),
-    ):
-        with pytest.raises(httpx.ReadTimeout):
-            httpx.get(server.url)
 
-
-def test_httpx_exceptions_exposed() -> None:
```
florimondmanca, line 63: The purpose of this test was to verify that we don't leave any exception class defined in `httpx/_exceptions.py` out of `httpx/__init__.py`.

As such, it's impossible to perform without reaching to `httpx._exceptions`.

But I also thought it wasn't very useful after all, so I am dropping it in this PR.

br3ndonland (approved):
Thanks @florimondmanca. I agree with the suggestion in #2492 and appreciate you taking care of this chore.

I'll try to help out with reviewing some of the other related PRs in the coming days.

PR #2572, opened 2023-02-05
Drop private imports from test_urlparse.py

florimondmanca:
More towards #2492

We still access `._uri_reference`, because the goal of this module is to test our vendored `urlparse()` implementation, but at least we do so from `httpx.URL()`.


httpx/_urlparse.py
```diff
@@ -195,18 +195,6 @@ def urlparse(url: str = "", **kwargs: typing.Optional[str]) -> ParseResult:
...
-            "scheme",
-            "authority",
-            "path",
-            "query",
-            "fragment",
-            "userinfo",
-            "host",
-            "port",
-        ):
-            raise TypeError(f"'{key}' is an invalid keyword argument for urlparse()")
```
florimondmanca, line 208: We were already doing this check in `URL.__init__()`, so in the wild it would indeed be redundant as we use `httpx.URL()`, so I dropped this check.

tomchristie (commented):
Can we drop the `._uri_reference` access on these cases, and instead inspect the public properties on the URL instance?

For example...

```python
url = httpx.URL("https://example.com/")
assert url.host == "example.com"
```

tomchristie (approved):

PR #2574, opened 2023-02-05
Drop private imports from test_main.py

florimondmanca:
More towards #2492 

Looks like the issue with mypy isn't relevant anymore (due to newer mypy version?).


tomchristie (approved):
> Looks like the issue with mypy isn't relevant anymore (due to newer mypy version?).

👍

PR #2547, opened 2023-01-05
Use standard logging style

tomchristie:
Closes #2042

This proposal drops some of our customised logging handling to replace it with a more standard setup.

In particular, we would...

* Drop our custom `TRACE` level logging. Instead use `INFO` for logging requests, and `DEBUG` for logging the SSL config.
* Drop our custom auto-logging setup, previously enabled via `HTTPX_LOG_LEVEL`.

This would then give us a more standard setup.

For example, logging at INFO level...

```python
import logging
import httpx

logging.basicConfig(
    format="%(levelname)s [%(asctime)s] %(name)s - %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
    level=logging.INFO
)

httpx.get('https://www.example.com')
```

Output:

```
INFO [2023-01-05 15:25:26] httpx - HTTP Request: GET https://www.example.com "HTTP/1.1 200 OK"
```

Or logging at DEBUG level...

```python
import logging
import httpx

logging.basicConfig(
    format="%(levelname)s [%(asctime)s] %(name)s - %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
    level=logging.DEBUG
)

httpx.get('https://www.example.com')
```

Output:

```
DEBUG [2023-01-05 15:26:12] httpx - load_ssl_context verify=True cert=None trust_env=True http2=False
DEBUG [2023-01-05 15:26:12] httpx - load_verify_locations cafile='/Users/tomchristie/GitHub/encode/httpx/venv/lib/python3.10/site-packages/certifi/cacert.pem'
INFO [2023-01-05 15:26:13] httpx - HTTP Request: GET https://www.example.com "HTTP/1.1 200 OK"
```

I'd suggest that alongside this we'd also add `DEBUG` level logging into `httpcore`, at the points where we currently have `trace` extension.

Once we've done that you'd be able to have debug logging for both `httpcore` and `httpx`...

```python
import logging
import httpx

logging.basicConfig(
    format="%(levelname)s [%(asctime)s] %(name)s - %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
    level=logging.DEBUG
)

httpx.get('https://www.example.com')
```

Output:

```
DEBUG [2023-01-05 15:34:30] httpx - load_ssl_context verify=True cert=None trust_env=True http2=False
[...]

PR #2575, opened 2023-02-05
Drop private logger imports from tests/utils.py

florimondmanca:
More towards #2492

This one was a bit trickier.

I essentially moved the re-initialization logic to `get_logger()` itself, behind a special testing environment variable flag. The code paths when that is not enabled is not affected at all.


florimondmanca:
Dropping this given the upcoming #2547.

Issue #2186, opened 2022-04-21
Testing endless SSE endpoint with AsyncClient hangs forever


havardthom:
- [x] Initially raised as discussion #1787 

Current behaviour:
 - Running `uvicorn test_endless_sse:app --reload` then `python test_endless_sse.py` works.
 - Running `pytest test_endless_sse.py` hangs forever.

Expected behaviour:
- `pytest test_endless_sse.py` should pass

```
# test_endless_sse.py
import pytest
import httpx
import asyncio

from fastapi import FastAPI, Request
from sse_starlette.sse import EventSourceResponse

app = FastAPI()

@app.get("/endless_sse")
async def endless_sse(req: Request):
    async def event_publisher():
        i = 0
        try:
            while True:
                disconnected = await req.is_disconnected()
                if disconnected:
                    break
                i += 1
                yield dict(data=i)
                await asyncio.sleep(0.2)
        except asyncio.CancelledError as e:
            raise e

    return EventSourceResponse(event_publisher())

# This test should pass but it hangs forever and prints no output
@pytest.mark.asyncio
async def test_endless_sse():
    async with httpx.AsyncClient(app=app, base_url="http://test") as aclient:
        async with aclient.stream("GET", "/endless_sse") as response:
            async for line in response.aiter_lines():
                line_stripped = line.strip()
                print("line_stripped: ", line_stripped)
                if line_stripped == "data: 10":
                    break


if __name__ == "__main__":
    # Run: uvicorn test_endless_sse:app --reload

    # Check that we get output from the app running in uvicorn
    async def fetch_endless_sse():
        async with httpx.AsyncClient(base_url="http://127.0.0.1:8000") as aclient:
            async with aclient.stream("GET", "/endless_sse") as response:
                async for line in response.aiter_lines():
                    line_stripped = line.strip()
                    print("line_stripped: ", line_stripped)
                    if line_stripped == "data: 10":
                        break


    asyncio.run(fetch_endless_sse())
```
```
python 3.8.0

httpx==0.22.0
pytest==7.1.1
pytest-asyncio==0.18.3
fastapi==0.75.2
uvicorn==0.17.6
sse-starlette==0.10.3
```

Related: https://github.com/tiangolo/fastapi/issues/2006

tomchristie:
You've identified this correctly in your comment here... https://github.com/encode/httpx/discussions/1787#discussioncomment-1537543

The issue is that the ASGITransport doesn't stream responses.
[...]

PR #2587, opened 2023-02-13
Fix unclosed generator when running on trio

florimondmanca:
While taking a look at resolving #2186, I came across an issue in the trio variant of `test_asgi_streaming`.

It seems that our usage of generators for `__aiter__` doesn't satisfy trio's strict semantics.

AFAIU, when `__aiter__()` does more than delegating to an underlying async iterable, we should be implementing `__anext__()` so that `aclosing()` is applied to that streaming. Though my understanding is a bit limited so far in the night. 😃 

This comes with a companion PR on httpcore. https://github.com/encode/httpcore/pull/657

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -142,9 +142,8 @@ class BoundAsyncStream(AsyncByteStream):
         self._response = response
         self._timer = timer
 
-    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
-        async for chunk in self._stream:
-            yield chunk
+    def __aiter__(self) -> typing.AsyncIterator[bytes]:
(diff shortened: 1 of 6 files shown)


tomchristie:
> It seems that our usage of generators for `__aiter__` doesn't satisfy trio's strict semantics.

How does this present itself?
Can you demonstrate a warning that we're seeing with the existing implementation?

tomchristie:
I'm closing this off in line with https://github.com/encode/httpcore/pull/657 for now. I'm open to the suggestion that there *may* be an issue here, but we'd need to be able to clearly demonstrate that first.

PR #2588, opened 2023-02-13
Add support for streaming in ASGIDispatch

florimondmanca:
Refs #2186

Note that this is based on #2587

### DIFF
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -1,10 +1,17 @@
 import typing
+from contextlib import AsyncExitStack, asynccontextmanager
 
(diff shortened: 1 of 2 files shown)


httpx/_transports/asgi.py
```diff
@@ -35,12 +42,19 @@ def create_event() -> "Event":
         return asyncio.Event()
```
tomchristie, line 42: It looks to me like we can drop this `create_event()` function and the preceding `if typing.TYPE_CHECKING` block?

tomchristie:
Latest attempt at this is #3059

PR #2596, opened 2023-02-20
Add Response ok property like requests api

gzq0616:
The requests library has a simple way to determine whether the request is normal. In order to migrate from requests to httpx, keep the code unchanged, add ok property and perfect support

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -1005,6 +1005,17 @@ class Response:
             with request_context(request=self._request):
                 await self.stream.aclose()
 
+    @property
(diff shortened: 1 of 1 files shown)


tomchristie:
Declining to include this was a deliberate design decision...

See https://www.python-httpx.org/compatibility/#checking-for-success-and-failure-responses

> We don't support response.is_ok since the naming is ambiguous there, and might incorrectly imply an equivalence to response.status_code == codes.OK. Instead we provide the response.is_success property, which can be used to check for a 2xx response.

PR #2609, opened 2023-03-01
Add sslcontext to the asynclient docstring

bpoirriez:
ASyncClient docstring is not aligned with Client docstring. The sslcontext option for the verify parameter is not present. This PR fix this.


tomchristie (approved):
Well spotted, thanks.

PR #2622, opened 2023-03-18
Add caching for responses with the 301 status code

karosis88:
When a _uri_ that should redirect to another with the 301 status code is requested once, we can **remember** the location where it's redirecting and do not make an **unnecessary** HTTP request to the _url_ a second time, because **we know** it's been redirected once with the 301 status code.


tomchristie:
Alrighty - this is a nice piece of work, but I'm not convinced we want to handle caching in the client, since HTTP caching rules are actually pretty complex.

In terms of development process let's aim for starting these kind of functionality changes as discussions, that can then be raised to issues if accepted.

My standard response for something like this would *normally* be: "looks great, can we start it off as a third party package?" - Perhaps by implementing this as a custom transport? Also: are there any existing third party packages for `httpx` that handle HTTP caching?

*Having said all that, I'm also... "hrm perhaps yeah, handling cacheable redirects is quite a neat bit of functionality"*

karosis88:
I don't think it's about the HTTP complex caching described in RFC 9111, such as the 'Cache-Control', 'Age', 'Pragma', and 'Expires' headers.
The 301 response behavior is also defined in 9110 HTTP Semantics, but not in 9111 HTTP Caching, which only suggests remembering and reusing permanent redirects.

RFC 9110

> The server is suggesting that a user agent with link-editing capability can permanently replace references to the target URI with one of the new references sent by the server.

In another line
> The user agent **MAY** use the Location field value for automatic redirection. The server's response content usually contains a short hypertext note with a hyperlink to the new URI(s).

karosis88:
I completely agree with you; HTTP caching is very complex, and the best option is to create a third-party package for caching httpx requests; however, I don't believe httpx requires caching, and I believe it should only be supported in browsers rather than http libraries.

PR #2628, opened 2023-03-21
simplify the client states, allow reopen after close()

Hearty07:


Hearty07:
What is the different between state UNOPENED and CLOSED?  Why it can't call get() again after close()? My purpose is to call close() and disconnect to the server before the TCP stream id reach the maximum that the server allow, so that I can call get() to connect to the server again.

tomchristie:
we did allow this in the past yep, and then made it a little stricter.

imposing constraints here rules out a slice of race-conditions that developers might be able to otherwise introduce (is the client open here or not). we want to help developers treat a client instance as a single clearly scoped resource.

if you need different lifetimes you can re-instantiate a new client.

PR #2630, opened 2023-03-21
More robust check for upload files in binary mode

lkuchenb:
This is a minor patch proposal fixing fixing an issue with a recently introduced check for binary file objects in multipart uploads. The currently used check
https://github.com/encode/httpx/blob/f1157dbc4102ac8e227a0a0bb12a877f592eff58/httpx/_multipart.py#L125-L128
misses out on file objects that read and write byte-like objects, but have the `mode` attribute set to `r` such as [ZipFile.open()](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open) which is binary-only.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -122,13 +122,13 @@ class FileField:
             # requests does the opposite (it overwrites the header with the 3rd tuple element)
             headers["Content-Type"] = content_type
 
-        if isinstance(fileobj, io.TextIOBase):
-            raise TypeError(
-                "Multipart file uploads must be opened in binary mode, not text mode."
-            )
         if isinstance(fileobj, io.StringIO):
             raise TypeError(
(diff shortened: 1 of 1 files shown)


lkuchenb:
The `StringIO` check is now somewhat obsolete, I've changed the order so that the more specific test comes first, but they could probably also be merged into one with a meaningful combined message.

tomchristie:
do we think it's worth adding an extra test for this case or we happy to accept the pr as-is?

tomchristie (approved):

tomchristie:
Thanks!

lkuchenb:
Thanks for the review @tomchristie - I didn't think covering this with a test would make much sense since the test would involve intentionally making false assumptions about standard library behavior, not something one would typically want to test IMHO.

PR #2637, opened 2023-03-29
Improve typing for 'method' parameters

karosis88:
The parameter method previously had the type 'str,' which has now been changed to 'typing. Literal["GET", "POST",..., "DELETE"] | str'

old_signature
``` python
method: str,
```
new_signature
``` python
method: typing.Union[MethodTypes, str],
# where
MethodTypes = Literal[
    "GET", "HEAD", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE"
]
```

The text editor previously displayed method as a type of str, but now it displays acceptable values such as "GET", "POST", and so on.

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -9,6 +9,7 @@ from ._types import (
     CertTypes,
     CookieTypes,
     HeaderTypes,
+    MethodTypes,
     ProxiesTypes,
     QueryParamTypes,
     RequestContent,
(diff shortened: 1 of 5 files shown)


tomchristie:
interesting. it *is* valid to use other strings here, although having a type-check that would cause non-standard cases to be a linting failure might still be a UX improvement. anyone want to give their 👍/👎 on how we'd consider this change?

PR #2645, opened 2023-04-03
fix-2625: bump httpcore max version to 0.17.0

### LINKED ISSUE #2625: RFE: please provide support for `httpcore` >= 0.17
https://github.com/encode/httpx/blob/85c5898d8e48011ab7b660fe8aac58910e45c84b/pyproject.toml#L30-L32

DavidBord:
Resolves #2625

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.15.0,<0.17.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.15.0,<0.17.0",
```
tomchristie, line 32: If we're bumping, then bumping to <0.18.0 would make sense.

PR #2648, opened 2023-04-04
Use ruff instead of flake8, autoflake and isort

Kludex:
Ref.: https://github.com/encode/httpx/discussions/2607

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,7 @@ text = "\n---\n\n[Full changelog](https://github.com/encode/httpx/blob/master/CH
 pattern = 'src="(docs/img/.*?)"'
 replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 
+# https://beta.ruff.rs/docs/configuration/#using-rufftoml
 [tool.ruff]
 select = ["E", "F", "I", "B", "PIE"]
 ignore = ["B904", "B028"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,3 +90,11 @@ text = "\n---\n\n[Full changelog](https://github.com/encode/httpx/blob/master/CH
 [[tool.hatch.metadata.hooks.fancy-pypi-readme.substitutions]]
 pattern = 'src="(docs/img/.*?)"'
 replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
+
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -19,19 +19,13 @@ build==0.10.0
...
-flake8==3.9.2
-flake8-bugbear==23.1.20
-flake8-pie==0.16.0; python_version>='3.7'
-importlib-metadata==4.13.0; python_version>='3.7'
-isort==5.11.4; python_version<'3.8'
-isort==5.12.0; python_version>='3.8'
 mypy==1.0.1
 types-certifi==2021.10.8.2
 pytest==7.2.2
+ruff==0.0.260
```
tomchristie, line 28: This is nice

pyproject.toml
```diff
@@ -90,3 +90,11 @@ text = "\n---\n\n[Full changelog](https://github.com/encode/httpx/blob/master/CH
 [[tool.hatch.metadata.hooks.fancy-pypi-readme.substitutions]]
 pattern = 'src="(docs/img/.*?)"'
 replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
+
+[tool.ruff]
+select = ["E", "F", "I", "B", "PIE"]
+ignore = ["B904", "B028"]
+line-length = 120
```
tomchristie, line 97: Should we comment above here with a pointer to a relevant part of the docs?

tests/client/test_async_client.py
```diff
@@ -84,7 +84,7 @@ async def test_access_content_stream_response(server):
 
     assert response.status_code == 200
     with pytest.raises(httpx.ResponseNotRead):
-        response.content
+        response.content  # noqa: B018
```
tomchristie, line 87: How can I track down what a B018 is?

Kludex: It's on the rules page: https://beta.ruff.rs/docs/rules/#flake8-bugbear-b

Should I add a comment here or ignore it globally?

pyproject.toml
Kludex: I've added a reference to the docs

Kludex: Is it enough?

tomchristie: 'tis

tomchristie (approved):

PR #2669, opened 2023-04-18
Fix exception suppression in asgi transport

Nnonexistent:
Bug description:

If httpx client uses asgi transport with `raise_app_exceptions=False` and response is not complete, exception is still raised.

This is mostly come up in tests where you want to check how your application behave in error states. But currently it works only with non-streaming responses, which is quite limiting.

I propose simple fix for such non-complete responses to be able to comply with the `raise_app_exceptions=False`.

Test case included.


tests/test_asgi.py
```diff
@@ -191,3 +192,12 @@ async def read_body(scope, receive, send):
...
     assert disconnect
+
+
+@pytest.mark.anyio
+async def test_asgi_exc_no_raise():
+    transport = ASGITransport(app=raise_exc, raise_app_exceptions=False)
+    async with httpx.AsyncClient(app=raise_exc, transport=transport) as client:
+        response = await client.get("http://www.example.org/")
+
+        assert response.status_code == 500
```
tomchristie, line 203: What's the behaviour of this test before the code change?

tests/test_asgi.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 
 import httpx
+from httpx._transports.asgi import ASGITransport
```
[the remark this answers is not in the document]
Nnonexistent: thanks, fixed

tomchristie (approved):
Seems reasonable, yup.

madkinsz (approved):

tomchristie:
Okay, let's also add this to the `CHANGELOG`.
What's the best user-facing description we can add here?

> Return `500` error response instead of exceptions when `raise_app_exceptions=False` is set on `ASGITransport`.

Perhaps?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+* Return `500` error response instead of exceptions when `raise_app_exceptions=False` is set on `ASGITransport`. (#2669)
```
tomchristie, line 15: Looks good, tho needs to be under a new `## development` header, instead of included in the 0.24.1 release notes.

tomchristie (changes requested):

CHANGELOG.md
Nnonexistent: Oh, my bad, sorry

PR #2686, opened 2023-04-28
Move configuration of tools to pyproject.toml

Secrus:
# Summary

# Checklist

- [**No changes needed**] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. 
- [**No changes needed**] I've updated the documentation accordingly.

I have noticed that the tools configured by `setup.cfg` file are the ones that can be configured in `pyproject.toml` which already held some of the tools configurations. Since `setup.cfg` is inherently connected with `setuptools` which is not used in this project. I didn't see a need for a discussion in this case.


Kludex (approved):

requirements.txt
```diff
@@ -20,7 +20,7 @@ twine==4.0.2
 
 # Tests & Linting
 black==23.3.0
-coverage==7.2.2
+coverage[toml]==7.2.2
```
Kludex, line 23: Why did you add this?

Kludex:
Since we've removed flake8 (which was the only project holding this change), I think this is a good change. 👍

Kludex:
Same PR welcome on other encode projects. 👀 🙏

Secrus:
> Same PR welcome on other encode projects. eyes pray

Sure, give me a moment ;)

requirements.txt
Secrus: exactly. from what I have seen, coverage requires `toml` extra to work with `pyproject.toml` configs.

PR #2708, opened 2023-05-18
Ensure all WSGITransport environs have a SERVER_PROTOCOL

lazorchakp:
# Summary

- I noticed that WSGI environs sent to applications with `httpx.Client(app=app)` did not contain a SERVER_PROTOCOL
- [PEP-3333](https://peps.python.org/pep-3333/#environ-variables) indicates that this key must be present in the WSGI environ, and popular WSGI frameworks (such as [webob](https://github.com/Pylons/webob)) assume it will be as well
- It looks like httpx defaults to using `HTTP/1.1` for a nearly identical case in the ASGITransport, so I did the same here
- I'm more than happy to discuss any of this further!

# Checklist

  - _to my knowledge, this does not require a documentation update_

### DIFF
--- a/httpx/_transports/wsgi.py
+++ b/httpx/_transports/wsgi.py
@@ -102,6 +102,7 @@ class WSGITransport(BaseTransport):
             "QUERY_STRING": request.url.query.decode("ascii"),
             "SERVER_NAME": request.url.host,
             "SERVER_PORT": str(port),
+            "SERVER_PROTOCOL": "HTTP/1.1",
             "REMOTE_ADDR": self.remote_addr,
         }
         for header_key, header_value in request.headers.raw:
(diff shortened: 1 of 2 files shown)


madkinsz (approved):
LGTM! Thanks for contributing.

Another maintainer may weigh in but it seems unlikely that this was intentionally excluded.

lazorchakp:
Great, thanks! I'll give it a day or so before merging to see if anyone has other thoughts

tomchristie:
valid.

tomchristie:
Process improvements we can make here...

* @tomchristie - Thanks for the quick merge, though you don't need to always play that role. If you could allow other maintainers more space here too that'd probs be good for the team as a whole.
* @encode/maintainers - We ought to start making sure we include `CHANGELOG` changes as part of functional or bugfix PRs. That'll help ensure that we're properly describing the change from a user perspective, and will also make the release process simpler and less error-prone. We should probably update [the pull request template](https://github.com/encode/httpx/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to include this as an item?
* @lazorchakp would you consider opening a PR to add this to the `CHANGELOG` under a new "development" heading?

Kludex:
Yes. 👍

PR #2714, opened 2023-05-23
Support context manager in Response object

Zheaoli:
# Summary

For now, the context manager has been supported in requests lib, FYI https://github.com/psf/requests/blob/main/requests/models.py#L706-L711

So the people can use it more easy like

```python
with requests.get('https://httpbin.org/get') as r:
        # Do things with the response here.
        r.raise_for_status()
```

So I think it would be great to support context manager in httpx

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -441,6 +441,7 @@ class Request:
         self.extensions = {}
         self.stream = UnattachedStream()
 
+U = typing.TypeVar("U", bound="Response")
 
 class Response:
     def __init__(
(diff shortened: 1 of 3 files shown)


tomchristie:
Non-streaming case...

```python
r = httpx.get('https://httpbin.org/get')
```

Streaming case...

```python
with httpx.stream('GET', 'https://httpbin.org/get') as r:
        # Do things with the response here.
        r.raise_for_status()
```

Zheaoli:
> Member

I know, but the requests lib keeps the stream response and non-stream response to reach a consensus style. I think it's better to keep the API consensus.

tomchristie:
> I think it's better to keep the API consensus.

would you like to open a discussion for this proposal?

Zheaoli:
> > I think it's better to keep the API consensus.
> 
> would you like to open a discussion for this proposal?

Sure

Zheaoli:
https://github.com/encode/httpx/discussions/2717

PR #2716, opened 2023-05-23
Add `socket_options` argument to `httpx.HTTPTransport` class

### LINKED ISSUE #2715: Add `socket_options` to `httpx.HTTPTransport`.
Since [version 0.17.2](https://github.com/encode/httpcore/pull/697#issuecomment-1559059612) the `httpcore` package has support for a `socket_options` parameter.
We should add a corresponding parameter to `httpx.HTTPTransport`, which can be passed through to the connection pool.

This will resolve the use-case raised in discussion https://github.com/encode/httpx/discussions/2635...

```python
>>> import httpx
>>> socket_options = [(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, b"ETH999")]
>>> transport = httpx.HTTPTransport(socket_options=socket_options)
>>> client = httpx.Client(transport=transport)
```

karosis88:
closes #2715

### DIFF
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -28,6 +28,7 @@ import typing
 from types import TracebackType
 
 import httpcore
+from httpcore.backends.base import SOCKET_OPTION
 
 from .._config import DEFAULT_LIMITS, Limits, Proxy, create_ssl_context
 from .._exceptions import (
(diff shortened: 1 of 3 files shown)


karosis88:
These changes have no place in the documentation; hopefully, the `Transports` documentation will soon be improved.

httpx/_transports/default.py
```diff
@@ -28,6 +28,7 @@
 from types import TracebackType
 
 import httpcore
+from httpcore.backends.base import SOCKET_OPTION
```
tomchristie, line 31: Let's not import that type definition, it's not documented public API.

It's be okay to instead define the socket option types here.

```python
SOCKET_OPTION = typing.Union[
    typing.Tuple[int, int, int],
    typing.Tuple[int, int, typing.Union[bytes, bytearray]],
    typing.Tuple[int, int, None, int],
]
```

tomchristie (approved):

Issue #2721, opened 2023-05-25
Use percent-encoding for `/` in query params, for better robustness.


marban:
### Discussed in https://github.com/encode/httpx/discussions/2661

<div type='discussions-op-text'>

<sup>Originally posted by **marban** April 15, 2023</sup>
Haven't gotten into the details but by upgrading to 0.24 and using a regular dict as GET params that contains e.g. an url like {"q": "http://example.com"} will somehow break the request to, in this case, Twitter's API. If if urllib.parse.quote_plus the "http://example.com", everything works fine (The slash is the culprit). 
If I urlencode the whole dict manually before passing it to httpx and call the r.url after the request the / is again not encoded, although it previously was.
https://api.twitter.com/1.1/search/tweets.json?q=url%3Ahttp%3A//google.com

I assume it has to do with
- Query parameter encoding switches from using + for spaces and %2F for forward slash, to instead using %20 for spaces and treating forward slash as a safe, unescaped character. This differs from requests, but is in line with browser behavior in Chrome, Safari, and Firefox. Both options are RFC valid. (#2543)

</div>

tomchristie:
Let me nudge you in the right direction here...

https://github.com/encode/httpx/blob/733595037ac26d272f74ee6d702150b017913d11/httpx/_urlparse.py#L264

BjoernPetersen:
@tomchristie just so you know: this broke more than the Twitter API. I have a project in which I want to request a signed URL to an image (from the OpenAI API) using httpx. Some signature query params contain forward slashes, and the server evidently doesn't accept it if those are encoded as `%2F`. 

I want to note that the URL works in Firefox, Chrome, and using `requests.get(image_url)`, but httpx's encoding breaks it.

It might be true that encoding the slashes is technically "more correct", but that doesn't really help if the server just won't accept it.

ttys0dev:
> Some signature query params contain forward slashes, and the server evidently doesn't accept it if those are encoded as `%2F`.

Yeah, hit this as well in #2883 when migrating from requests to httpx, unconditionally precent-encoding `/`'s is clearly not the correct behavior IMO if servers treat it differently.

alexmaurizio:
Jumping in to signal we also had to revert httpx from 0.25.0 to 0.24.1 due to this PR. 
This change is causing issues with double encoding and a whole lot of services.

PR #2723, opened 2023-05-26
Always encode forward slashes as `%2F` in query parameters

madkinsz:
# Summary

Closes https://github.com/encode/httpx/issues/2721

Encodes `/` as `%2F` in query parameters for more robust behavior matching browser implementations. See issue for discussion.

# Checklist

### DIFF
--- a/httpx/_urlparse.py
+++ b/httpx/_urlparse.py
@@ -260,8 +260,11 @@ def urlparse(url: str = "", **kwargs: typing.Optional[str]) -> ParseResult:
     # For 'path' we need to drop ? and # from the GEN_DELIMS set.
     parsed_path: str = quote(path, safe=SUB_DELIMS + ":/[]@")
     # For 'query' we need to drop '#' from the GEN_DELIMS set.
+    # We also exclude '/' because it is more robust to replace it with a percent
+    # encoding despite it not being a requirement of the spec and include '%' to
+    # prevent duplicate encoding of previously quoted items.
     parsed_query: typing.Optional[str] = (
-        None if query is None else quote(query, safe=SUB_DELIMS + ":/?[]@")
+        None if query is None else quote(query, safe=SUB_DELIMS + ":?[]@%")
     )
     # For 'fragment' we can include all of the GEN_DELIMS set.
     parsed_fragment: typing.Optional[str] = (
--- a/httpx/_urlparse.py
+++ b/httpx/_urlparse.py
@@ -454,11 +454,11 @@ def urlencode(items: typing.List[typing.Tuple[str, str]]) -> str:
     #
     # https://github.com/python/cpython/blob/b2f7b2ef0b5421e01efb8c7bee2ef95d3bab77eb/Lib/urllib/parse.py#L926
     #
-    # Note that we use '%20' encoding for spaces, and treat '/' as a safe
-    # character. This means our query params have the same escaping as other
-    # characters in the URL path. This is slightly different to `requests`,
-    # but is the behaviour that browsers use.
+    # Note that we use '%20' encoding for spaces. and '%2F  for '/'.
+    # This is slightly different than `requests`, but is the behaviour that browsers use.
     #
-    # See https://github.com/encode/httpx/issues/2536 and
-    # https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
-    return "&".join([quote(k) + "=" + quote(v) for k, v in items])
+    # See
+    # - https://github.com/encode/httpx/issues/2536
+    # - https://github.com/encode/httpx/issues/2721
+    # - https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
+    return "&".join([quote(k, safe="") + "=" + quote(v, safe="") for k, v in items])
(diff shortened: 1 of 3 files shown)


httpx/_urlparse.py
```diff
@@ -260,8 +260,11 @@ def urlparse(url: str = "", **kwargs: typing.Optional[str]) -> ParseResult:
     # For 'path' we need to drop ? and # from the GEN_DELIMS set.
     parsed_path: str = quote(path, safe=SUB_DELIMS + ":/[]@")
     # For 'query' we need to drop '#' from the GEN_DELIMS set.
+    # We also exclude '/' because it is more robust to replace it with a percent
+    # encoding despite it not being a requirement of the spec and include '%' to
+    # prevent duplicate encoding of previously quoted items.
     parsed_query: typing.Optional[str] = (
-        None if query is None else quote(query, safe=SUB_DELIMS + ":/?[]@")
+        None if query is None else quote(query, safe=SUB_DELIMS + ":?[]@%")
```
madkinsz, line 267: I was surprised that I needed to include `%` but it was needed to to prevent ` ` converted to `%20` from being changed to `%2520` in tests.

madkinsz: Let me know if there's additional test coverage that would make sense for this or if we've revealed another issue.

tomchristie: Hrm. Can you issue this PR without that change?

tomchristie: I'll rephrase: Can you drop that bit from this pull request? It seems unrelated so it'd make sense to consider it separately.

madkinsz: Yeah I can but then the _other_ change will make the tests fail i.e. this pull request will introduce a incorrect behavior. I don't understand why yet.

tomchristie: > I can but then the other change will make the tests fail i.e. this pull request will introduce a incorrect behavior.

Okay. That would be worthwhile, we can review from there.

madkinsz: Here are the failing tests on https://github.com/encode/httpx/pull/2723/commits/5bbacceb39d6faf885e4741633474853e6d1da43

```
_____________________________________________________________________________________________________ test_param_with_existing_escape_requires_encoding ______________________________________________________________________________________________________

    def test_param_with_existing_escape_requires_encoding():
        url = httpx.URL("http://webservice", params={"u": "http://example.com?q=foo%2Fa"})
>       assert str(url) == "http://webservice?u=http%3A%2F%2Fexample.com%3Fq%3Dfoo%252Fa"
E       AssertionError: assert 'http://webse...3Dfoo%25252Fa' == 'http://webse...q%3Dfoo%252Fa'
E         - http://webservice?u=http%3A%2F%2Fexample.com%3Fq%3Dfoo%252Fa
E         + http://webservice?u=http%253A%2F%2Fexample.com%253Fq%253Dfoo%25252Fa
E         ?                          ++                    ++    ++         ++

tests/test_urlparse.py:144: AssertionError
__________________________________________________________________________________________________________________ test_url_query_encoding ___________________________________________________________________________________________________________________

    def test_url_query_encoding():
        """
        URL query parameters should use '%20' to encoding spaces,
        and should treat '/' as a safe character. This behaviour differs
        across clients, but we're matching browser behaviour here.
    
        See https://github.com/encode/httpx/issues/2536
        and https://github.com/encode/httpx/discussions/2460
        """
        url = httpx.URL("https://www.example.com/?a=b c&d=e/f")
        assert url.raw_path == b"/?a=b%20c&d=e%2Ff"
    
        url = httpx.URL("https://www.example.com/", params={"a": "b c", "d": "e/f"})
>       assert url.raw_path == b"/?a=b%20c&d=e%2Ff"
E       AssertionError: assert b'/?a=b%2520c&d=e%2Ff' == b'/?a=b%20c&d=e%2Ff'
E         At index 7 diff: b'5' != b'0'
E         Use -v to get more diff

tests/models/test_url.py:366: AssertionError
```

tomchristie:
Okay, so we also need to change this...

```python
def urlencode(items: typing.List[typing.Tuple[str, str]]) -> str:
    # We can use a much simpler version of the stdlib urlencode here because
    # we don't need to handle a bunch of different typing cases, such as bytes vs str.
    #
    # https://github.com/python/cpython/blob/b2f7b2ef0b5421e01efb8c7bee2ef95d3bab77eb/Lib/urllib/parse.py#L926
    #
    # Note that we use '%20' encoding for spaces, and treat '/' as a safe
    # character. This means our query params have the same escaping as other
    # characters in the URL path. This is slightly different to `requests`,
    # but is the behaviour that browsers use.
    #
    # See https://github.com/encode/httpx/issues/2536 and
    # https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
    return "&".join([quote(k) + "=" + quote(v) for k, v in items])
```

To not use the default `safe="/"` value...

```python
def urlencode(items: typing.List[typing.Tuple[str, str]]) -> str:
    # We can use a much simpler version of the stdlib urlencode here because
    # we don't need to handle a bunch of different typing cases, such as bytes vs str.
    #
    # Note that we use '%20' encoding for spaces, instead of "+".
    return "&".join([quote(k, safe="") + "=" + quote(v, safe="") for k, v in items])
```

madkinsz:
Thanks for the hint! I'm pretty confused how it manifested that way still :D

tomchristie (approved):

ttys0dev:
This needs to be reverted see [this comment](https://github.com/encode/httpx/issues/2721#issuecomment-1732273653) and #2883.

zanieb:
Hi @ttys0dev — I appreciate the sentiment but that's not a great way to collaborate with us. We're considering reverting this and the existing discussion is the place to weigh in with details of why this matters to you.

ttys0dev:
> existing discussion

Oh, I had looked for existing open issues and didn't originally find one tracking this issue yet, I hadn't realized this project often puts issues under discussions(I've never really used the discussions feature before) which is a bit different from other projects.

PR #2727, opened 2023-05-31
async recommendations

trondhindenes:
# Summary
Update docs regarding usage of client context manager inside hot loops, since we know that it negatively affects performance

# Checklist

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -54,7 +54,7 @@ async with httpx.AsyncClient() as client:
 ```
 
 !!! warning
-Note that there is some overhead to instantiating and terminating the client object, so it's recommended to avoid building the client using `async with` repeatedly inside a "hot loop". This can be achieved by having a single global client instance which will provide automatic connection pooling.
+In order to get the most benefit from connection pooling, make sure you're not instantiating multiple client instances. This can be achieved either by having a single scoped client that's passed throughout wherever it's needed, or by having a single global client instance.
 
 Alternatively, use `await client.aclose()` if you want to close a client explicitly:
 
--- a/docs/async.md
+++ b/docs/async.md
@@ -52,7 +52,9 @@ Use `async with httpx.AsyncClient()` if you want a context-managed client...
 async with httpx.AsyncClient() as client:
     ...
 ```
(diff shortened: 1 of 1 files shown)


docs/async.md
```diff
@@ -52,6 +52,7 @@ Use `async with httpx.AsyncClient()` if you want a context-managed client...
 async with httpx.AsyncClient() as client:
     ...
 ```
+Note that there is some overhead to instantiating and terminating the client object, so it's recommended to avoid building the client using `async with` repeatedly inside a "hot loop". This can be achieved either by having a single scoped client that's passed throughout wherever it's needed, or by having a single global client instance.
```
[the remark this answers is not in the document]
trondhindenes: good point. Feel free to suggest, I'll make a new commit with a suggestion too.

docs/async.md
```diff
@@ -53,6 +53,9 @@ async with httpx.AsyncClient() as client:
     ...
 ```
 
+!!! warning
+Note that there is some overhead to instantiating and terminating the client object, so it's recommended to avoid building the client using `async with` repeatedly inside a "hot loop". This can be achieved by having a single global client instance which will provide automatic connection pooling.
```
tomchristie, line 57: I've made a suggestion based on your first pass, but with a positive phrasing...

```suggestion
In order to get the most benefit from connection pooling, make sure you're not instantiating multiple client instances. This can be achieved either by having a single scoped client that's passed throughout wherever it's needed, or by having a single global client instance.
```

What do you think?

trondhindenes: Hm I think it's really import to be explicit about not using async with in hot loops. Thinking back to when I started with python async, the suggested text here would probably not be enough for me to understand clearly that I was going against recommended practices. How about:

```
In order to get the most benefit from connection pooling, make sure you're not instantiating multiple client instances - for example by using `async with` repeatedly inside a "hot loop".
```
Best of both worlds?

tomchristie (approved):
This seems good to me, yup.

trondhindenes:
@tomchristie nice thanks! I don't have merge access, so hopefully a maintainer will merge this in at some point. Thanks for helping out with this PR!

karosis88 (approved):

PR #2740, opened 2023-06-08
Fix URLPattern examples

karosis88:


karosis88:
This `if` would raise an exception if we just use `all`
```python
        if pattern and ":" not in pattern:
            raise ValueError(
                f"Proxy keys should use proper URL forms rather "
                f"than plain scheme strings. "
                f'Instead of "{pattern}", use "{pattern}://"'
            )
```

tomchristie (approved):

PR #2741, opened 2023-06-08
Fix environment proxies

karpetrosyan:
Now the httpx parses `NO_PROXY: http://google.com` to `{'all://*http://google.com': None}` but expected `{'http://google.com': None}`

# TODO
- [x] Write a failing test
- [x] Make the test pass


karosis88:
Docs
```
export HTTP_PROXY=http://my-external-proxy.com:1234
export NO_PROXY=http://127.0.0.1,python-httpx.org

# As in the previous example, this request will be sent through the proxy
python -c "import httpx; httpx.get('http://example.com')"

# These requests will be sent directly, bypassing the proxy
python -c "import httpx; httpx.get('http://127.0.0.1:5000/my-api')"  <- # This is not correct!
python -c "import httpx; httpx.get('https://www.python-httpx.org')"
```

tomchristie (approved):

tomchristie:
Excellent, yep. Apologies for the delay.

Let's get a CHANGELOG entry and pull this in. 👍

(Maybe this?... *"Handle `NO_PROXY` envvar cases when a fully qualified URL is supplied as the value."*)

karpetrosyan:
Done!

PR #2768, opened 2023-07-12
Use Mozilla documentation instead of `httpstatuses.com` for HTTP error reference

zanieb:
Discussion at https://github.com/encode/httpx/discussions/2766

httpstatuses.com redirects to https://www.webfx.com/web-development/glossary/http-status-codes/ which has advertisements and is now generally an unpleasant resource — I'll spare you the visit:

httpstatuses

Instead we use the Mozilla documentation which looks like this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200


zanieb:
CI is failing because of a mypy/click bug; ref https://github.com/pallets/click/pull/2559 https://github.com/pallets/click/pull/2562 https://github.com/python/mypy/issues/13449

Kludex (approved):

tomchristie (approved):

PR #2769, opened 2023-07-12
Add upper bound to click version to fix mypy error

zanieb:


Kludex:
It looks like a release in click will happen soon, fyi

zanieb:
I tried the linked pull request locally and it didn't work — figure we might as well be unblocked in the meantime but don't have strong feelings.

tomchristie:
Ah well caught.

We'd probably want to resolve a CI failure by pinning in `requirements.txt`, rather than in our packaging.

Issue #2794, opened 2023-07-28
Change the type of `Extensions` from `Mapping` to `MutableMapping`.


karosis88:
### Discussed in https://github.com/encode/httpx/discussions/2793

<div type='discussions-op-text'>

<sup>Originally posted by **karosis88** July 28, 2023</sup>
I'm working on a library that implements HTTP Caching for httpx and httpcore (it provides transports and connection pools), and I'd like to add an extension that simply indicates whether or not the response was taken from the cache.

Unfortunately, the type of extension is Mapping, so this is an error for mypy.

```python
response = httpx.Response(200)
response.extensions['my_custom_extension'] = 'something'
```

OUTPUT 
```
error: Unsupported target for indexed assignment ("Mapping[Str, Any]")  [index]
```

The solution is to simply change the extension type from `Mapping` to `MutableMapping`, allowing us to add custom extensions after the response has been created.

[See also this pr](https://github.com/karosis88/hishel/pull/4)</div>

---

I believe the only change needed is in the "_models.py" file.

PR #2776, opened 2023-07-15
Make `raise_for_status` chainable

trim21:
# Summary
make it possible to write code like `r = httpx.get('...').raise_for_status().json()`

Refs https://github.com/encode/httpx/discussions/2775

# Checklist

- [X] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.

PR #2702, opened 2023-05-16
Version 0.24.1

epenet:
### Added

* Provide additional context in some `InvalidURL` exceptions. (#2675)

### Fixed

* Fix optional percent-encoding behaviour. (#2671)
* More robust checking for opening upload files in binary mode. (#2630)
* Properly support IP addresses in `NO_PROXY` environment variable. (#2659)
* Set default file for `NetRCAuth()` to `None` to use the stdlib default. (#2667)
* Set logging request lines to INFO level for async requests, in line with sync requests. (#2656)
* Fix which gen-delims need to be escaped for path/query/fragment components in URL. (#2701)

PR #2801, opened 2023-08-01
Version 0.25.0

trim21:
### Removed

* Drop support for Python 3.7. (#2813)

### Added

* Support HTTPS proxies. (#2845)
* Change the type of `Extensions` from `Mapping[Str, Any]` to `MutableMapping[Str, Any]`. (#2803)
* Add `socket_options` argument to `httpx.HTTPTransport` and `httpx.AsyncHTTPTransport` classes. (#2716)
* The `Response.raise_for_status()` method now returns the response instance. For example: `data = httpx.get('...').raise_for_status().json()`. (#2776)

### Fixed

* Return `500` error response instead of exceptions when `raise_app_exceptions=False` is set on `ASGITransport`. (#2669)
* Ensure all `WSGITransport` environs have a `SERVER_PROTOCOL`. (#2708)
* Always encode forward slashes as `%2F` in query parameters (#2723)
* Use Mozilla documentation instead of `httpstatuses.com` for HTTP error reference (#2768)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.25.0 (7th August, 2023)
+## 0.25.0
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.25.0 (3rd August, 2023)
+## 0.25.0 (7th August, 2023)
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## 0.25.0 (1st August, 2023)
+
 ### Added
 
 * Add `socket_options` argument to `httpx.HTTPTransport` and `httpx.AsyncHTTPTransport` classes. (#2716)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
 ## 0.25.0 (1st August, 2023)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 * Return `500` error response instead of exceptions when `raise_app_exceptions=False` is set on `ASGITransport`. (#2669)
-* Ensure all WSGITransport environs have a SERVER_PROTOCOL. (#2708)
+* Ensure all `WSGITransport` environs have a `SERVER_PROTOCOL`. (#2708)
+* Always encode forward slashes as `%2F` in query parameters (#2723)
+* Use Mozilla documentation instead of `httpstatuses.com` for HTTP error reference (#2768)
 
 ## 0.24.1 (17th May, 2023)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
```
karosis88, line 7:
```suggestion
## Unreleased
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
```
karosis88, line 8:
```suggestion

```

tomchristie:
Okay, let's...

* Ensure that the PR description matches the style we're using [for release PRs](https://github.com/encode/httpx/pull/2702).
* Justify to ourselves if we should be issuing a minor point release or a media point release? Which one and why?
* Check if we have [accidentally missed any *behavioural* changes from the CHANGELOG](https://github.com/encode/httpx/commits/master)?

CHANGELOG.md
karosis88: We're adding the unreleased or development tag with the first pr, so let's stick with that.

trim21: Looks like your suggestion doesn't change any code, I'm a little bit confusing what you are asking.

trim21:
> * Ensure that the PR description matches the style we're using [for release PRs](https://github.com/encode/httpx/pull/2702).

does this mean to simply copy changelog of this version as PR description?

CHANGELOG.md
karosis88: It just seems strange to me to leave the unreleased tag empty; it would be better to remove it and add it with the first pull request.

trim21: OK

karosis88:
Can we assume that the current version of httpcore is the last one that we will be able to support in this release? Because httpcore changed the response.extensions type in https://github.com/encode/httpcore/pull/762 I'm guessing we'll have type issues with the next httpcore release.

tomchristie:
> I'm guessing we'll have type issues with the next httpcore release.

I'm... not sure if the `Mapping` -> `MutableMapping` switch-up will cause a continuous integration issue or not.
It's not an issue in *practise*, because we're actually using dictionary instances throughout anyways. But it might cause a typing problem?

You could issue a draft PR where `httpcore` installs from the repo source if you wanted to double check.

* Shall we consider rolling an `httpcore` release prior to releasing this one for `httpx`?
* Should we resolve https://github.com/encode/httpx/issues/2794 for this release?

trim21:
> For the 0.x series we're using median point bumps for releases that include API changes. We've only got additions and fixes here, so we probably want a `0.24.2` release?

I'll update the version number according your decision 🙂

trim21:
> > I'm guessing we'll have type issues with the next httpcore release.
> 
> I'm... not sure if the `Mapping` -> `MutableMapping` switch-up will cause a continuous integration issue or not. It's not an issue in _practise_, because we're actually using dictionary instances throughout anyways. But it might cause a typing problem?
> 
> You could issue a draft PR where `httpcore` installs from the repo source if you wanted to double check.
> 
> * Shall we consider rolling an `httpcore` release prior to releasing this one for `httpx`?
> * Should we resolve [Change the type of `Extensions` from `Mapping` to `MutableMapping`. #2794](https://github.com/encode/httpx/issues/2794) for this release?

I think if we want to include #2794, we should create a `httpcore` release (`>=0.18.0`) first, require master branch of httpcore is not a good idea.

tomchristie:
> I think if we want to include https://github.com/encode/httpx/issues/2794, we should create a httpcore release (>=0.18.0) first

Indeed, yep.

- [ ] Drop Python 3.7 from `httpcore`.
- [ ] Resolve https://github.com/encode/httpcore/pull/757
- [ ] Issue `httpcore` 0.18.0 release. https://github.com/encode/httpcore/pull/768
- [ ] Resolve #2794.
- [ ] Resolve #2813
- [ ] Issue `httpx` 0.25.0 release.

trim21:
> Drop Python 3.7 from `httpcore`.

This doesn't need to be in todo list I think

Pliner:
@trim21 Could you please include #2813 into this release? I merged it to master today morning.

trim21:
> For the 0.x series we're using median point bumps for releases that include API changes. We've only got additions and fixes here, so we probably want a `0.24.2` release?

Now we removed python3.7 support, I think it's definitely `0.25.0` not `0.24.2`

PR #2802, opened 2023-08-01
Test response extensions typing

karosis88:
Discussed in #2801


karosis88:
We have type problems if I did everything correctly.

We can most likely release httpcore and update the minimum version to that new one.

tomchristie:
Okay yup, I guess we can close this one off now.

PR #2803, opened 2023-08-01
Change extensions type

### LINKED ISSUE #2794: Change the type of `Extensions` from `Mapping` to `MutableMapping`.

karosis88:
Closes #2794

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.17.2,<0.18.0",
+    "httpcore>=0.18.0<0.19.0",
     "idna",
     "sniffio",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
+    "httpcore>=0.17.2,<0.18.0",
     "idna",
     "sniffio",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.18.0<0.19.0",
+    "httpcore>=0.18.0,<0.19.0",
     "idna",
     "sniffio",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.17.2,<0.18.0",
     "idna",
     "sniffio",
 ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+* Change the type of `Extensions` from `Mapping[Str, Any]` to `MutableMapping[Str, Any]`. (#2803)
 * Add `socket_options` argument to `httpx.HTTPTransport` and `httpx.AsyncHTTPTransport` classes. (#2716)
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


tomchristie:
> We need httpcore release to resolve this type error.

Okay, shall we add `httpcore @ git+https://github.com/encode/httpcore@master` to the `requirements.txt` in this PR to demo that this'd resolve [the type issues](https://github.com/encode/httpx/pull/2802).

pyproject.toml
```diff
@@ -29,7 +29,6 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.17.2,<0.18.0",
```
trim21, line 32: In my understanding, this is runtime dependencies, remove `httpcore` from here will cause users running `pip intsall httpx` not installing `httpcore`

karosis88: So the user installs httpx but is unable to use it because it also requires the installation of httpcore? 
Is there any motivation to do so?

trim21: > So the user installs httpx but is unable to use it because it also requires the installation of httpcore?

Yes, users will find they can't use `httpx` because `httpx` import `httpcore`, but `httpx` didn't specify `httpcore` as dependencies. 

`requirements.txt` are actually only for developing and testing.

> Is there any motivation to do so?

I didn't understand this question...

trim21: > I mean, what problem does it solve?

`httpx` require `httpcore`, and remove this line also remove `httpcore` from install_requires

you can try this:
```shell
python -m venv .venv
pip install  https://github.com/karosis88/httpx/archive/refs/heads/change-response-extensions-type.zip # this is change-response-extensions-type branch of your git repo
python -c 'import httpx'
```

karosis88: Sorry, but I don't understand why we should do that. Why should we remove httpcore as a dependency and install it independently?

trim21: It's not "we should do", it's "what you have done in this PR"

trim21: Your PR (this deletion line I comment on) remove `httpcore` as a dependency.

karosis88: > In my understanding, this is runtime dependencies, remove `httpcore` from here will cause users running `pip intsall httpx` not installing `httpcore`

Sorry, I thought you were suggesting that httpcore be removed here :pray:

karosis88: > I'm saying "don't do it".

Of course, we won't do that; it was just a temporary commit that will be reverted in the future; we do it to test with the httpcore master.

karosis88: > I'm saying "don't do it".

Of course, we won't do that; it was just a temporary commit that will be reverted in the future; we do it to test with the httpcore master.

trim21: > > I'm saying "don't do it".
> 
> Of course, we won't do that; it was just a temporary commit that will be reverted in the future; we do it to test with the httpcore master.

you can still add `httpcore @ git+https://github.com/encode/httpcore@master` to requirements.txt for testing, it will override the `pyproject.toml#dependencies` and install master branch of httpcore.

trim21:
looks like we can continue on this issue？

zanieb (approved):

PR #2840, opened 2023-09-02
Update Typeinfo for extensions in Response class

KalleDK:
To remove the unknown dict type info

(variable) extensions: ResponseExtensions | dict[Unknown, Unknown]

# Summary

Simple add typing information for the extensions variable, as the default dict is "unknown"

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -467,7 +467,7 @@ class Response:
         # the client will set `response.next_request`.
         self.next_request: typing.Optional[Request] = None
 
-        self.extensions = {} if extensions is None else extensions
+        self.extensions: ResponseExtensions = {} if extensions is None else extensions
         self.history = [] if history is None else list(history)
 
         self.is_closed = False


tomchristie:
I don't have a problem with merging this, since yep the typing here is correct.

> (variable) extensions: ResponseExtensions | dict[Unknown, Unknown]

What tooling are you using that's showing you this type error? (We don't have a `mypy` issue here in our CI)

tomchristie (approved):

KalleDK:
> I don't have a problem with merging this, since yep the typing here is correct.
> 
> > (variable) extensions: ResponseExtensions | dict[Unknown, Unknown]
> 
> What tooling are you using that's showing you this type error? (We don't have a `mypy` issue here in our CI)

Pylance (Pyright) with strict mode. They really don't like empty dicts/lists without typeinfo

KalleDK:
And thanks for merging it - one squiggly line less and one type ignore less for me :)

PR #2845, opened 2023-09-11
HTTPS proxies support

### LINKED ISSUE #1434: HTTPS proxies support
Refs #1424, #1428 

## Terminology

* HTTP proxy: a proxy server which supports connecting to it via HTTP. HTTP requests are forwarded, HTTPS requests are tunneled (via HTTP `CONNECT`). — HTTPX has good support for those, no questions asked. ✅ 
* HTTPS proxy: a proxy server that supports connecting to it via HTTPS. Only HTTPS requests are supported, and _must_ (?) be tunneled (TLS-in-TLS). — This is what seems to be missing still. :x:

## Problem statement

We've seen reports of issues recently such as #1424 and #1428 that reveal that our proxies implementation does not properly support HTTPS proxies yet.

My understanding right now is that supporting this requires implementing a technique known as "TLS-in-TLS" (or perhaps "nested TLS"). Here's how that works:

1. HTTPX issues a `CONNECT` request to the proxy, at `https://proxy.org`. This may use a dedicated `proxy_ssl_context` with proxy-specific certs, that I'll mark as TLS(p) ("p" as "proxy").
2. The proxy establishes a TCP tunnel to the `target`. The HTTPX-proxy "half" of the TCP connection is over TLS(p), and the other one is not TLS-enabled yet.
3. HTTPX must perform a TLS handshake with the target server, so that proxy-server "half" of the TCP connection becomes encrypted over TLS(t) ("t" as "target"). I'm not 100% certain I understand how that works. Right now I assume TLS handshake packets would be sent TLS(p)-encrypted to the proxy, which """decrypts""" them and sends them to the server for us. The server responds with its SERVER HELLO. The proxy """encrypts"" them back over TLS(p) and HTTPX sees them. The second pass of the handshake follows the same pattern. (I've put quotation marks because this is actually done without the proxy being able to actually intercept those packets — anyone could confirm?)

✅ Right now we can do steps 1/ and 2/, with the nuance that we have a single `ssl_context` option that's used for both proxy CONNECT and the handshake (we'd want to have `proxy_ssl_context` and `ssl_context`).

:x: What is definitely missing is step 3/.

Right now we attempt to do `start_tls()`, as if we were tunneling over a standard HTTP connection with the proxy — and generally that fais in a variety of ways depending on sync / async, async library, custom certs, HTTP/1.1 vs HTTP/2, proxy server implementation, etc.

## To reproduce

Right now the following would fail:

```python
proxies = {"https": "https://proxy.org:443"}
with httpx.Client(proxies=proxies) as client:
    response = client.get("https://example.org")
```

**TODO**: full `pproxy` setup (or perhaps `proxy.py`, which seems to support HTTPS proxying fully), full sample tracebacks.

## Additional context

Marked this as "requests-compat" because this is now supported in `urllib3` as of 1.26. It landed via this pull request: https://github.com/urllib3/urllib3/pull/1923. AFAICT they had to implement TLS-in-TLS themselves, overriding the standard `http.client` connection implementation because that one doesn't support TLS-in-TLS.

Marked this as "httpcore" because our proxy implementation lives there: https://github.com/encode/httpcore

karosis88:
Closes #1434

Because HTTPS proxies are now supported by httpcore, this pull request simply provides an API for that feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+* Support HTTPS proxies. (#2845)
 * Add `socket_options` argument to `httpx.HTTPTransport` and `httpx.AsyncHTTPTransport` classes. (#2716)
 
 ### Fixed
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -326,6 +326,7 @@ class Proxy:
         self,
         url: URLTypes,
         *,
+        ssl_context: typing.Optional[ssl.SSLContext] = None,
         auth: typing.Optional[typing.Tuple[str, str]] = None,
         headers: typing.Optional[HeaderTypes] = None,
     ):
@@ -343,6 +344,7 @@ class Proxy:
         self.url = url
         self.auth = auth
         self.headers = headers
+        self.ssl_context = ssl_context
 
     @property
     def raw_auth(self) -> typing.Optional[typing.Tuple[bytes, bytes]]:
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -156,6 +156,7 @@ class HTTPTransport(BaseTransport):
                 proxy_auth=proxy.raw_auth,
                 proxy_headers=proxy.headers.raw,
                 ssl_context=ssl_context,
+                proxy_ssl_context=proxy.ssl_context,
                 max_connections=limits.max_connections,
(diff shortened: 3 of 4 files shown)


tomchristie (approved):

tomchristie:
Oooh... docs.

https://www.python-httpx.org/advanced/#http-proxying

I notice our docs yet fully aren't up to pace with the `httpx.Proxy(...)` configuration.
I'd suggest we resolve that without blocking this pull request.

Issue #2788, opened 2023-07-26
DigestAuthentication not working


robin-blanchard:
### Discussed in https://github.com/encode/httpx/discussions/2787

<div type='discussions-op-text'>

<sup>Originally posted by **robin-blanchard** July 25, 2023</sup>
Hello,
It seems to be an issue with DigestAuthentication.  Even with a basic HTTP server, I can't manage to correctly authenticate a client with httpx.  

Here's the server implementation with Flask:
```python
from flask import Flask
from flask_httpauth import HTTPDigestAuth

app = Flask(__name__)
app.config['SECRET_KEY'] = 'secret key here'
auth = HTTPDigestAuth()

users = {
    "john": "hello",
    "susan": "bye"
}


@auth.get_password
def get_pw(username):
    if username in users:
        return users.get(username)
    return None


@app.route('/', methods=['GET', 'POST'])
@auth.login_required
def index():
    return "Hello, {}!".format(auth.username())


if __name__ == '__main__':
    app.run(host="0.0.0.0", debug=True)
```

And here's the small client:
```python
import httpx

auth = httpx.DigestAuth("susan", "bye")
print(httpx.get("http://192.168.2.140:5000/", auth=auth).status_code)
```
This code returns 401.

httpx version: 0.24.1

Does anyone has a solution ? I plan on using httpx to send asynchronous requests to a server with Digest Authentication. Only httpx seems to provide Digest Authentication with async support.

Many thanks</div>

robin-blanchard:
It seems that manually adding the cookie to the second request does the job. For example by adding:
`Cookies(response.cookies).set_cookie_header(request)` right before this line: https://github.com/encode/httpx/blob/master/httpx/_auth.py#L220

robin-blanchard:
I think there are other things to change to make it work with `qop=auth`
https://github.com/encode/httpx/blob/master/httpx/_auth.py#L288 : We should also enter here `if qop == "auth"`
https://github.com/encode/httpx/blob/master/httpx/_auth.py#L288: Incase qop is indeed auth, the response should only be `digest(key_digest)`. Line https://github.com/encode/httpx/blob/master/httpx/_auth.py#L280 should probably be changed also to make it cleaner

karosis88:
> It seems that manually adding the cookie to the second request does the job. For example by adding: `Cookies(response.cookies).set_cookie_header(request)` right before this line: https://github.com/encode/httpx/blob/master/httpx/_auth.py#L220

[...]

PR #2846, opened 2023-09-13
Add cookies to the retried request when performing digest authentication.

musale:
# Summary
Adding the cookies from the response in the first request into the follow up request. The server expects a cookie to be set. This is however not highlighted in the RFC. Supporting [SO answer](https://stackoverflow.com/a/25743837/4444629) and discussions by @robin-blanchard https://github.com/encode/httpx/issues/2788#issuecomment-1677015264.

# Checklist

### DIFF
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -217,7 +217,8 @@ class DigestAuth(Auth):
         request.headers["Authorization"] = self._build_auth_header(
             request, self._last_challenge
         )
-        Cookies(response.cookies).set_cookie_header(request=request)
+        if response.cookies:
+            Cookies(response.cookies).set_cookie_header(request=request)
         yield request
 
     def _parse_challenge(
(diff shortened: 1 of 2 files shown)


tomchristie:
Nicely done, thanks.

The test suite is failing because there's no coverage for this line... https://github.com/encode/httpx/blob/8302ab2765d2e7a4eb1caf5b430592bb9605ec92/httpx/_auth.py#L221

Do you want to try adding a test case for this?

musale:
> Nicely done, thanks.
> 
> The test suite is failing because there's no coverage for this line... https://github.com/encode/httpx/blob/8302ab2765d2e7a4eb1caf5b430592bb9605ec92/httpx/_auth.py#L221
> 
> Do you want to try adding a test case for this?

Yes. I'll add a test for it.

tomchristie (approved):

PR #2850, opened 2023-09-16
test: Drop private imports from `test_proxies.py`

YDX-2147483647:
# Summary

Refactor a test case that imports from private namespace, as suggested in #2492.

The original version constructs `URLPattern`s to compare. This PR destructs them.

# Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (This is a test refactor)
- [x] I've updated the documentation accordingly. (No code changed)

### DIFF
--- a/tests/client/test_proxies.py
+++ b/tests/client/test_proxies.py
@@ -2,7 +2,6 @@ import httpcore
 import pytest
 
 import httpx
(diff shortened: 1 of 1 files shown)


tomchristie (approved):
Thanks. 😊

PR #2851, opened 2023-09-17
Drop private imports on `test_utils.py`

T-256:
# Summary

closes https://github.com/encode/httpx/issues/2492

`get_ca_bundle_from_env` moved to where it's calling.

# Checklist


tests/test_utils.py
```diff
@@ -1,21 +1,14 @@
...
-    URLPattern,
-    get_ca_bundle_from_env,
-    get_environment_proxies,
-    guess_json_utf,
-    is_https_redirect,
-    obfuscate_sensitive_headers,
-    parse_header_links,
-    same_origin,
-)
+from httpx._utils import guess_json_utf  # has not any trace in public api
```
T-256, line 11: It has not any trace in public api.

tests/test_utils.py
```diff
@@ -3,18 +3,13 @@
...
-    URLPattern,
-    get_ca_bundle_from_env,
-    get_environment_proxies,
-    guess_json_utf,
-    is_https_redirect,
-    obfuscate_sensitive_headers,
-    parse_header_links,
-    same_origin,
+    get_ca_bundle_from_env,  # catching `ssl.SSLError` is insufficient.
+    guess_json_utf  # has not any trace in public api.
```
T-256, line 12:
```suggestion
    guess_json_utf,  # has not any trace in public api.
```

T-256:
Done

karpetrosyan:
Please optimize the number of git commits. 
You can even do fewer commits or force push them at the end so that we only see a few commits rather than the commit army.
I believe it will make the pull request more attractive and easier to review.

T-256:
> but switch to using private _accessors_.

previous PR like [this](https://github.com/encode/httpx/pull/2850/files#diff-d0fbc1167934498ad6894c6a6c6281f3bb91e979cf12502cf0c93090219acc98R37) also used private accessor, so I thought it's no problem using them in tests at all.

T-256:
close in favor of separating to chunks...

PR #2852, opened 2023-09-17
Raise ValueError on `Response.encoding` being set after `Response.text` has been accessed

xzmeng:
Discussed in #2041

# Summary

After `Response.text` is accessed, if you try to change `Response.encoding`, nothing happens. So just raise an Exception on change to `encoding` after `text` is accessed.

# Checklist

### DIFF
--- a/tests/models/test_responses.py
+++ b/tests/models/test_responses.py
@@ -298,6 +298,22 @@ def test_response_force_encoding():
     assert response.encoding == "iso-8859-1"
 
 
+def test_response_force_encoding_after_text_accessed():
+    response = httpx.Response(
+        200,
+        content=b"Hello, world!",
+    )
+    assert response.status_code == 200
+    assert response.reason_phrase == "OK"
+    assert response.text == "Hello, world!"
+    assert response.encoding == "utf-8"
+
+    response.encoding = "UTF8"
+
+    with pytest.raises(ValueError):
+        response.encoding = "iso-8859-1"
+
+
 def test_read():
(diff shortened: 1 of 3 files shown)


tomchristie:
Okay yes, as discussed this is probably a marginal improvement in behaviour. Thanks.

tests/models/test_responses.py
```diff
@@ -298,6 +298,22 @@ def test_response_force_encoding():
...
+    response = httpx.Response(
+        200,
+        content=b"Hello, world!",
+    )
+    assert response.status_code == 200
+    assert response.reason_phrase == "OK"
+    assert response.text == "Hello, world!"
+    assert response.encoding == "utf-8"
+
+    response.encoding = "UTF8"
```
tomchristie, line 311: I'd expect the `ValueError` to be raised here. I don't think we need the conditional "don't raise if the encoding is being set but the resulting codec won't change".

Simplicity over complexity where possible.

tomchristie (changes requested):

tomchristie:
What's a more precise title for this PR?

xzmeng:
Modified. I'm not proficient in English, please change it if it's not appropriate.

karosis88 (approved):

tomchristie (commented):
All looking great, thanks. 😊

I think we just need to update the `CHANGELOG.md`, then this is good to go.

tomchristie (approved):

PR #2813, opened 2023-08-03
Drop support for Python 3.7

Pliner:
Have to open a new PR with changes from #2744, because it has been forcefully closed by GH.

PR #2854, opened 2023-09-19
Add support for Python 3.12

hugovk:
# Summary

Follow on from https://github.com/encode/httpcore/pull/807.

The [third and final Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc3-released-honestly-the-final-release-candidate-i-swear/34093?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Python 3.12.0 final is due out in two weeks: [2023-10-02](https://peps.python.org/pep-0693/).

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

# Checklist

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import ssl
-import sys
 import typing
 from pathlib import Path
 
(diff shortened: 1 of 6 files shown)


httpx/_config.py
```diff
@@ -128,11 +127,10 @@ def load_ssl_context_verify(self) -> ssl.SSLContext:
 
         # Signal to server support for PHA in TLS 1.3. Raises an
         # AttributeError if only read-only access is implemented.
-        if sys.version_info >= (3, 8):  # pragma: no cover
-            try:
-                context.post_handshake_auth = True
-            except AttributeError:  # pragma: no cover
-                pass
```
Kludex, line 135: Why are you removing 3.8 conditionals?

T-256: Python 3.7 no longer supported https://github.com/encode/httpx/pull/2813

hugovk: Yep, with 3.8+, the condition `sys.version_info >= (3, 8)` is always true.

tomchristie (approved):

Issue #947, opened 2020-05-12
Version 1.0


tomchristie:
I think we're ready to start homing in on a 1.0 release, so I'd like to start a discussion to figure out what remaining aspects we need to finalise before we can cut a release there.

A key point to note here is that "1.0" doesn't mean that we can't *add* further API functionality, but rather any public API we *are* exposing won't easily be changed.

Here's the key point I can think about for this...

* Promote the low-level "transport" API into a public API, together with some renaming, for "Transport" rather than "Dispatch".
* Ensure exceptions are all comprehensively addressed - make the exception hierarchy public API, revisit any naming decisions, look at if we can make request/response mandatory arguments etc.
* Revisit the URL API, since I think it needs a little finessing.

Am I missing any areas?

florimondmanca:
I’d like to add a discussion point about our position wrt type hints would be. There’s been a couple of issues asking for them to be public API mainly for the use of third party library authors. I don’t have a strong position on this yet, but wondering if type hints (and more generally third party lib authoring experience, although that’s pretty broad) should be in our scope for 1.0?

tomchristie:
Good call yup, we need to discuss that for sure!

yeraydiazdiaz:
We have a [version 1.0 milestone](https://github.com/encode/httpx/milestone/1). Should we formalize these points into issues and add them?

tomchristie:
The proxy DEFAULT_MODE discussion needs to be part of this too.

tomchristie:
@yeraydiazdiaz Sure thing, yup.

pikeas:
Having recently used httpx, I was very surprised that there's no rate-limiting (requests per second, not just max connections) built in. This greatly reduces the utility of the async mode.

tomchristie:
Okay. Sure we can have a discussion about that, tho it's not strictly a 1.0 blocker.

florimondmanca:
Something else to potentially consider for 1.0 that came to my mind: #302, aka dropping `certifi` in favor of the system defaults. I have a different approach in mind that what was originally suggested in the issue, will comment there to see if it makes sense.

bwright1558:
I've had my eyes on this repository for a while and it looks very promising. The only thing preventing me from using this library is the lack of support for socks5 proxies. Gotta have this for connecting to Tor. It's also a lot easier to spin up a socks proxy server than it is an http proxy server (in my opinion).
There's already an open issue: #203. There's also a PR in httpcore: encode/httpcore#51.

[...]

PR #2858, opened 2023-09-20
Make async dependencies optional.

tomchristie:
See the ["Version 1.0, Minimising project dependancies"](https://github.com/encode/httpx/issues/947#issuecomment-1715532202) discussion for background, as well as [the proposed PR for `httpcore` version 1.0](https://github.com/encode/httpcore/pull/809).

This pull request would changes our default install footprint from...

* `httpcore`
  * `certifi`
  * `h11`
  * `sniffio`
  * `anyio`
* `idna`

To this...

* `httpcore`
  * `certifi`
  * `h11`
* `idna`

We do also have scope to make `idna` support optional, but this pull request doesn't address that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,10 +97,12 @@ Install with pip:
 $ pip install httpx
 ```
 
-Or, to include the optional HTTP/2 support, use:
+There are also a number of optional dependancies.
+
+For example to include asyncio and HTTP/2 support, use:
 
 ```shell
-$ pip install httpx[http2]
+$ pip install 'httpx[asyncio,http2]'
 ```
 
 HTTPX requires Python 3.8+.
@@ -129,15 +131,16 @@ The HTTPX project relies on these excellent libraries:
   * `h11` - HTTP/1.1 support.
 * `certifi` - SSL certificates.
 * `idna` - Internationalized domain name support.
-* `sniffio` - Async library autodetection.
 
 As well as these optional installs:
 
-* `h2` - HTTP/2 support. *(Optional, with `httpx[http2]`)*
-* `socksio` - SOCKS proxy support. *(Optional, with `httpx[socks]`)*
-* `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
-* `click` - Command line client support. *(Optional, with `httpx[cli]`)*
-* `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
+* `anyio`, `sniffio` - Async support for `asyncio`. *(Optional, with `httpx['asyncio']`)*
+* `trio`, `sniffio` - Async support for `trio`. *(Optional, with `httpx['trio']`)*
+* `h2` - HTTP/2 support. *(Optional, with `httpx['http2']`)*
+* `socksio` - SOCKS proxy support. *(Optional, with `httpx['socks']`)*
+* `rich` - Rich terminal support. *(Optional, with `httpx['cli']`)*
+* `click` - Command line client support. *(Optional, with `httpx['cli']`)*
+* `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx['brotli']`)*
 
 A huge amount of credit is due to `requests` for the API layout that
 much of this work follows, as well as to `urllib3` for plenty of design
--- a/docs/async.md
+++ b/docs/async.md
@@ -10,6 +10,24 @@ long-lived network connections such as WebSockets.
 If you're working with an async web framework then you'll also want to use an
 async client for sending outgoing HTTP requests.
 
+## Enabling Async support
+
+To enable async support you'll need to install some additional dependencies:
+
+If you're using Python's [standard `asyncio` support](https://docs.python.org/3/library/asyncio.html) then:
+
+```shell
+$ pip install httpx['asyncio']
+```
+
+Or, if you're working with the [`trio` third party package](https://trio.readthedocs.io/en/stable/):
+
+```shell
+$ pip install httpx['trio']
+```
+
+We highly recommend `trio` for async support. The `trio` project [pioneered the principles of structured concurrency](https://en.wikipedia.org/wiki/Structured_concurrency), and has a more carefully constrained API against which to work from.
+
 ## Making Async requests
 
 To make asynchronous requests, you'll need an `AsyncClient`.
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -64,7 +64,7 @@ SOCKET_OPTION = typing.Union[
 def map_httpcore_exceptions() -> typing.Iterator[None]:
     try:
         yield
-    except Exception as exc:  # noqa: PIE-786
+    except Exception as exc:
         mapped_exc = None
 
         for from_exc, to_exc in HTTPCORE_EXC_MAP.items():
@@ -269,6 +269,13 @@ class AsyncHTTPTransport(AsyncBaseTransport):
         retries: int = 0,
         socket_options: typing.Optional[typing.Iterable[SOCKET_OPTION]] = None,
     ) -> None:
+        try:
+            import sniffio  # noqa: F401
+        except ImportError:  # pragma: nocover
+            raise RuntimeError(
+                "Using httpx in async mode, but neither httpx['asyncio'] or asyncio['trio'] is installed."
+            )
+
         ssl_context = create_ssl_context(verify=verify, cert=cert, trust_env=trust_env)
 
         if proxy is None:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,9 +28,8 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.18.0,<0.19.0",
+    "httpcore>=1.0.0,<2.0.0",
     "idna",
-    "sniffio",
 ]
 dynamic = ["readme", "version"]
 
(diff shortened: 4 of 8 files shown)


tomchristie:
*Tests currently blocked on `httpcore` 1.0.*

tomchristie:
> *Tests currently blocked on `httpcore` 1.0.*

Resolved

aminalaee (approved):

pyproject.toml
```diff
@@ -46,10 +45,16 @@ cli = [
...
 socks = [
-    "socksio==1.*",
+    "httpcore[socks]",
+]
+asyncio = [
+    "httpcore[asyncio]"
+]
+trio = [
+    "httpcore[trio]"
 ]
```
tomchristie, line 58: This is interesting... currently we're including the `httpcore` dependencies explicitly.

Eg. `pip install httpx[socks]` uses the dependency `socksio==1.*` instead of the dependency `httpcore[socks]`.

The second style makes more sense, because that's what we're *actually* trying to effect... but I'm a bit hesitant that *some* Python install tooling might incorrectly fail to install `httpcore[socks]` as an optional dependency when the `httpcore` dependancy is already included. Is my hesitancy well founded here, or are we okay to use this new style throughout?

README.md
```diff
@@ -129,15 +131,16 @@ The HTTPX project relies on these excellent libraries:
...
 
 As well as these optional installs:
 
-* `h2` - HTTP/2 support. *(Optional, with `httpx[http2]`)*
-* `socksio` - SOCKS proxy support. *(Optional, with `httpx[socks]`)*
-* `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
-* `click` - Command line client support. *(Optional, with `httpx[cli]`)*
-* `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
+* `anyio`, `sniffio` - Async support for `asyncio`. *(Optional, with `httpx['asyncio']`)*
+* `trio`, `sniffio` - Async support for `trio`. *(Optional, with `httpx['trio']`)*
```
T-256, line 138:
```suggestion
* `anyio` - Async support for `asyncio`. *(Optional, with `httpx['asyncio']`)*
* `trio` - Async support for `trio`. *(Optional, with `httpx['trio']`)*
```

It's not in `pyproject.toml` and httpcore also doesn't mention it in readme. Actually, it's dependency of dependencies and if you want to mention it here, `trio` also have other dependencies can be here.

uuip:
remove anyio, good,very good

abersheeran:
If async is optional, should we improve the unit test so that it can skip some tests when certain optional dependencies are not installed? And can we consider conducting separate tests in CI for both scenarios: when optional dependencies are not installed and when they are already installed?

tomchristie:
@abersheeran I don't know. We don't have any need for that in CI, but perhaps some folks would appreciate fast runs in local dev. Perhaps spin it out into a new discussion?

zanieb:
> We don't have any need for that in CI

It's probably worthwhile to make sure the project _works_ as intended in some basic form during CI without the optional dependencies.

tomchristie:
[😀 💬](https://github.com/encode/httpx/discussions/new?category=general)

zanieb:
I'm not sure it's fair to say we should have a separate discussion when we need something like it for reasonable coverage of the changes introduced in _this_ pull request. 

I think "skip tests when the dependencies are not available for better developer experience" makes sense as a new discussion, but "ensure the library works without the async dependencies installed" is critical here. Otherwise, we're just relying on manual testing to ensure the default installation of the library isn't broken. If you have a strong preference for a follow-up pull request, I can get behind that, but I think we need automated testing before release.

tomchristie:
@zanieb Ah interesting. I've perhaps been missing your intent here.

I had "should we ensure the tests can run without all dependencies installed?".
I didn't have "should we run the tests without all dependencies installed (in CI)?".

zanieb:
👍 

> I didn't have "should we run the tests without all dependencies installed (in CI)?".

Yeah I think this is the idea. We should run some subset of the tests (i.e. the synchronous API) in CI with the default install dependencies to ensure things are working as intended now (and in the future).

httpx/_transports/default.py
```diff
@@ -272,6 +272,13 @@ def __init__(
         retries: int = 0,
         socket_options: typing.Optional[typing.Iterable[SOCKET_OPTION]] = None,
     ) -> None:
+        try:
+            import sniffio  # noqa: F401
+        except ImportError:  # pragma: nocover
+            raise RuntimeError(
+                "Using httpx in async mode, but neither httpx['asyncio'] or asyncio['trio'] is installed."
+            )
```
tomchristie, line 280:
```suggestion
            raise RuntimeError(
                "Using httpx in async mode, but neither "
                "httpx['asyncio'] or asyncio['trio'] is installed."
            )
```

tomchristie:
@zanieb Do we want to block this on *"run the tests without all dependencies installed"* or not?
(My take would probably be... we could consider it but as a follow-up... since we *already* have some optional dependencies)

There's a chunk of work there, that I'd suggest would need to be taken on incrementally if we want to deal with it.

docs/async.md
```diff
@@ -10,6 +10,24 @@ long-lived network connections such as WebSockets.
...
 async client for sending outgoing HTTP requests.
 
+## Enabling Async support
+
+To enable async support you'll need to install some additional dependencies:
+
+If you're using Python's [standard `asyncio` support](https://docs.python.org/3/library/asyncio.html) then:
+
+```shell
+$ pip install httpx['asyncio']
```
zanieb, line 20: Should we quote this so zsh users do not encounter an error? e.g. 

```
❯ pip install httpx['asyncio']
zsh: no matches found: httpx[asyncio]
```

zanieb:
> @zanieb Do we want to block this on _"run the tests without all dependencies installed"_ or not? (My take would probably be... we could consider it but as a follow-up... since we _already_ have some optional dependencies)
> 
> There's a chunk of work there that I'd suggest would need to be taken on incrementally if we want to deal with it.

I'm okay with merging this and addressing as a follow-up!

docs/async.md
tomchristie: I don't know, maybe? Could you show me the `±suggestion`?
We're probably using this style in a few places, right?

zanieb:
```suggestion
$ pip install "httpx[asyncio]"
```

tomchristie: We haven't released this.

zanieb: `pip` doesn't fail on unknown extras.

It fails with a parse error

```
    raise InvalidRequirement(
pip._vendor.packaging.requirements.InvalidRequirement: Parse error at ""['asynci"": Expected string_end
```

T-256 (approved):
Let's add changelog.

README.md
```diff
@@ -129,15 +131,16 @@ The HTTPX project relies on these excellent libraries:
...
-* `socksio` - SOCKS proxy support. *(Optional, with `httpx[socks]`)*
-* `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
-* `click` - Command line client support. *(Optional, with `httpx[cli]`)*
-* `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
+* `anyio` - Async support for `asyncio`. *(Optional, with `httpx['asyncio']`)*
+* `trio` - Async support for `trio`. *(Optional, with `httpx['trio']`)*
+* `h2` - HTTP/2 support. *(Optional, with `httpx['http2']`)*
+* `socksio` - SOCKS proxy support. *(Optional, with `httpx['socks']`)*
+* `rich` - Rich terminal support. *(Optional, with `httpx['cli']`)*
+* `click` - Command line client support. *(Optional, with `httpx['cli']`)*
```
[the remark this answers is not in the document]
tomchristie: Not a direct dependency.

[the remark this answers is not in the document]
tomchristie: Oh apologies... Yes it is a direct dependency for `httpx[cli]`. Hmmm.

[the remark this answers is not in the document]
tomchristie: Yep, we should mention pygments here. PRs welcome.

README.md
```diff
@@ -129,15 +131,16 @@ The HTTPX project relies on these excellent libraries:
...
 
-* `h2` - HTTP/2 support. *(Optional, with `httpx[http2]`)*
-* `socksio` - SOCKS proxy support. *(Optional, with `httpx[socks]`)*
-* `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
-* `click` - Command line client support. *(Optional, with `httpx[cli]`)*
-* `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
+* `anyio` - Async support for `asyncio`. *(Optional, with `httpx['asyncio']`)*
+* `trio` - Async support for `trio`. *(Optional, with `httpx['trio']`)*
+* `h2` - HTTP/2 support. *(Optional, with `httpx['http2']`)*
+* `socksio` - SOCKS proxy support. *(Optional, with `httpx['socks']`)*
```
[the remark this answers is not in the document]
tomchristie: 🤷🏼 Welcome to raise a discussion against httpcore.

httpx/_transports/default.py
```diff
@@ -272,6 +272,14 @@ def __init__(
         retries: int = 0,
         socket_options: typing.Optional[typing.Iterable[SOCKET_OPTION]] = None,
     ) -> None:
+        try:
+            import sniffio  # noqa: F401
+        except ImportError:  # pragma: nocover
+            raise RuntimeError(
+                "Using httpx in async mode, but neither "
+                "httpx['asyncio'] or asyncio['trio'] is installed."
+            )
```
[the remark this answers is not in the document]
tomchristie: ASGITransport doesn't have a dependency on `anyio` so it's a bit different.

PR #2860, opened 2023-09-23
use built-in idna instead of thrid party

karpetrosyan:
# Summary

Use built-in idna instead of third party library.

# Checklist

### DIFF
--- a/httpx/_urlparse.py
+++ b/httpx/_urlparse.py
@@ -19,8 +19,6 @@ import ipaddress
 import re
 import typing
 
-import idna
-
 from ._exceptions import InvalidURL
 
 MAX_URL_LENGTH = 65536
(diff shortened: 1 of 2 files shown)


karosis88:
Thinking about why the output is not the same.

``` python
import encodings.idna

print('Königsgäßchen.de'.encode('idna'), encodings.idna.ToASCII('Königsgäßchen.de').decode('ascii'))
```

out:

`.encode` b'xn--knigsgsschen-lcb0w.de' 
`encodings.idna.ToASCII` xn--knigsgsschen.de-5kb41a

httpx/_urlparse.py
```diff
@@ -324,8 +322,8 @@ def encode_host(host: str) -> str:
 
     # IDNA hostnames
     try:
-        return idna.encode(host.lower()).decode("ascii")
-    except idna.IDNAError:
+        return host.lower().encode("idna").decode("ascii")
+    except Exception:
```
karosis88, line 326:
```suggestion
    except UnicodeError:
```

T-256:
perhaps we could still support 3dparty idna as optional dependency.

jruere:
@karpetrosyan, the built-in IDNA encoding implements an outdated standard. It should not be used.

See the RFC each implement:
- https://docs.python.org/3/library/codecs.html#text-encodings
- https://pypi.org/project/idna/ (also explains this difference)

AFAIU, Python will not update the encoding as it would be a breaking change.

PR #2885, opened 2023-10-09
Support newer versions of httpcore

### LINKED ISSUE #2884: Support httpcore >= 1.0.0
Please add support for httpcore >= 1.0.0, which was released last week.

Currently, installs fail like so:
```
The conflict is caused by:
    The user requested httpcore==1.0.0
    httpx 0.25.0 depends on httpcore<0.19.0 and >=0.18.0
```

paulschreiber:
# Summary
Support newer versions of httpcore. httpcore 1.0.0 was release October 6, 2023.

Fixes #2884.


pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.18.0,<0.19.0",
+    "httpcore",
```
tomchristie, line 32: Since `httpcore 1.0` no longer includes `anyio` by default, we'd currently need the following...

```suggestion
    "httpcore",
    "anyio",
```

(Although the more comprehensive approach will be to pull in https://github.com/encode/httpx/pull/2858, which is what we'd like for httpx 1.0)

pyproject.toml
```diff
@@ -29,7 +29,8 @@ classifiers = [
...
+    "anyio",
```
karpetrosyan, line 33:
```suggestion

```

We need anyio only for tests. We can probably move this to requirements.txt.

T-256: If we are going to support httpcore 1.0 in httpx 0.26.0 I think it can have async default installation. httpx-async users don't get `anyio` not found error when they normally uses `pip install httpx`.
I think we can either not supporting httpcore 1.0 until httpx 1.0 (which done in #2858) OR don't drop httpcore's async dependencies.

T-256: If we are going to support httpcore 1.0 in httpx 0.26.0, I think it can have async default installation. it prevents httpx-async users get `anyio` not found error when they normally use `pip install httpx`.
I think we can either not supporting httpcore 1.0 until httpx 1.0 (which done in #2858) OR don't drop httpcore's async feature i.e include anyio.

karpetrosyan: How did `pip install httpx` raise `anyio` not found?

tomchristie:
Referring back to my previous review :)

This pull request should require both 'httpcore' and 'anyio'. (Effectively the same as requiring 'httpcore[asyncio]')
We could choose to roll a release from that point if we wanted too. 0.25.1 should be fine - we don't have any API breaking changes here, just a dependency upgrade.

Actually what we really want is PR #2858, which addresses this more comprehensively with the async dependencies no longer included by default. Tho we can treat that as a follow-up aimed at the upcoming 1.0 release.

See https://github.com/encode/httpx/issues/947#issuecomment-1715532202

pyproject.toml
T-256: Try install httpx + httpcore 1.0 then remove anyio. Error:
```
(.venv) E:\hcore1>python app.py
  File "E:\hcore1\app.py", line 6, in main
    r = await client.get('https://www.example.com/')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1757, in get
    return await self.request(
           ^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1530, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1617, in send
    response = await self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1645, in _send_handling_auth
    response = await self._send_handling_redirects(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1682, in _send_handling_redirects
    response = await self._send_single_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 1719, in _send_single_request
    response = await transport.handle_async_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_transports\default.py", line 366, in handle_async_request
    resp = await self._pool.handle_async_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_async\connection_pool.py", line 224, in handle_async_request
    async with self._pool_lock:
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_synchronization.py", line 62, in __aenter__
    self.setup()
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_synchronization.py", line 55, in setup
    raise RuntimeError(
RuntimeError: Running with asyncio requires installation of 'httpcore[asyncio]'.

  File "E:\hcore1\app.py", line 8, in <module>
    asyncio.run(main())
  File "C:\Users\Tester\.rye\py\cpython@3.12.0\install\Lib\asyncio\runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\Tester\.rye\py\cpython@3.12.0\install\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Tester\.rye\py\cpython@3.12.0\install\Lib\asyncio\base_events.py", line 664, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "E:\hcore1\app.py", line 5, in main
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_client.py", line 2003, in __aexit__
    await self._transport.__aexit__(exc_type, exc_value, traceback)
  File "E:\hcore1\.venv\Lib\site-packages\httpx\_transports\default.py", line 345, in __aexit__
    await self._pool.__aexit__(exc_type, exc_value, traceback)
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_async\connection_pool.py", line 327, in __aexit__
    await self.aclose()
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_async\connection_pool.py", line 312, in aclose
    async with self._pool_lock:
  File "E:\hcore1\.venv\Lib\site-packages\httpcore\_synchronization.py", line 67, in __aenter__
    await self._anyio_lock.acquire()
          ^^^^^^^^^^^^^^^^
AttributeError: 'AsyncLock' object has no attribute '_anyio_lock'

(.venv) E:\hcore1>
```

tomchristie (changes requested):
My suggestion here would be...

* Either add `anyio` to the dependencies list, or use `httpcore[asyncio]`.
* Add "update httpcore dependency" to the CHANGELOG.

And then we follow up with a new minor point release.

pyproject.toml
tomchristie, line 32:
```suggestion
    "httpcore",
    "anyio",
```

tomchristie: (Or `"httpcore[asyncio]"`)

tomchristie:
I'm happy to roll a minor point release with this if the above suggestions are resolved.
Alternatively we can pass on this and merge https://github.com/encode/httpx/pull/2858 with a new major release planned. 

(We'll do that either way, but we can have an interim release if we want.)

tomchristie:
Okay I think if we update the `CHANGELOG.md` as part of this pull request, then we should merge this and roll an 0.25.1 release.

tomchristie (approved):

pyproject.toml
```diff
@@ -29,7 +29,8 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore>=0.18.0,<0.19.0",
+    "httpcore",
```
T-256, line 32: Don't we need to pin version for httpcore and anyio here?

tomchristie: If we want pinning here then...

* We should probably pin anyio to the `>=4.0.0,<5.0.0` range.
* We can pin httpcore to the `>=1.0.0,<2.0.0` range. (Actually I think we can go lower than that, but...)

Issue #2875, opened 2023-10-02
Slow performance when merging cookies


gyula-lakatos:
- [X] Initially raised as discussion #2874 

There is a quite significant performance degradation when cookies are being used with requests.

`httpx._client.BaseClient._merge_cookies` calls into the cookiejar here:

```
        if cookies or self.cookies:
            merged_cookies = Cookies(self.cookies)
            merged_cookies.update(cookies)
            return merged_cookies
        return cookies
```

When either `cookies` or `self.cookies` is not None, then `httpx._models.Cookies.__bool__` will call into a method called `deepvalues` in http.cookiejar.CookieJar` by indirectly calling http.cookiejar.CookieJar.__iter__`.

`httpx._models.Cookies.__bool__`:

```
    def __bool__(self) -> bool:
        for _ in self.jar:
            return True
        return False
```

`http.cookiejar.CookieJar.__iter__`:

```
    def __iter__(self):
        return deepvalues(self._cookies)
```

Deepvalues is a significant performance hog because it does a lot of things recursively.

A suggested solution is to check `cookies` and `self.cookies` to None instead of using `__bool__`.

I dropped a performance snapshot to illustrate the problem:
<img width="1012" alt="image" src="https://github.com/encode/httpx/assets/126086412/618c7dce-6f1b-493c-8035-69bcd6e135f9">

Here `deepvalues` takes up almost 17% of the CPU time.

tomchristie:
Pulling in the stdlib `http.cookiejar` and `urllib.request` modules brings in quite a lot of submerged complexity... I'd wonder if we really need that for the `httpx.Cookies` implementation.

* Could we instead be using a plain dictionary format under-the-hood for the cookie persistence?
* Which cookie attributes does `httpx` honor? For example, does "expires" actually take affect?
* Can we point to a really clear easy-to-read functional description of http cookies?

vergeev:
> Could we instead be using a plain dictionary format under-the-hood for the cookie persistence?

I think the biggest obstacle here is the fact that the `jar` attribute of `Cookies` class is a part of a public interface: https://github.com/encode/httpx/blob/fbe35add82032d119d60c7f4de9bce2ccb12f6a1/docs/api.md?plain=1#L154 

[...]

PR #2894, opened 2023-10-14
perf: using None checking cookies instead of bool(#2875)

Tunglies:
# Summary
Performance improve. Related #2875
Before
```
         16002280 function calls (14002219 primitive calls) in 10.641 seconds

   Ordered by: internal time
   List reduced from 225 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  2000002    2.431    0.000    2.898    0.000 cookiejar.py:1679(set_cookie)
  1000001    2.243    0.000    6.808    0.000 _models.py:1057(set)
  1000001    1.289    0.000    1.599    0.000 copy.py:66(copy)
  1000001    1.219    0.000    2.914    0.000 cookiejar.py:761(__init__)
3000009/1000003    1.089    0.000    1.089    0.000 cookiejar.py:1227(deepvalues)
        1    0.532    0.532    0.532    0.532 perf_test.py:11(<dictcomp>)
        1    0.386    0.386   10.056   10.056 _models.py:1147(update)
        3    0.381    0.127    7.189    2.396 _models.py:1024(__init__)
  2000003    0.277    0.000    0.277    0.000 {method 'acquire' of '_thread.RLock' objects}
  1000001    0.207    0.000    0.207    0.000 {method 'copy' of 'dict' objects}
```
Now
```
         2245 function calls (2190 primitive calls) in 0.539 seconds

   Ordered by: internal time
   List reduced from 221 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.505    0.505    0.505    0.505 perf_test.py:11(<dictcomp>)
        3    0.029    0.010    0.029    0.010 {method 'load_verify_locations' of '_ssl._SSLContext' objects}
       14    0.001    0.000    0.001    0.000 {built-in method __new__ of type object at 0x00007FF9B50B8F90}
        3    0.001    0.000    0.001    0.000 {built-in method nt.stat}
        3    0.000    0.000    0.000    0.000 {method 'set_ciphers' of '_ssl._SSLContext' objects}
        6    0.000    0.000    0.000    0.000 enum.py:1374(_missing_)
        1    0.000    0.000    0.000    0.000 {built-in method winreg.OpenKey}
        5    0.000    0.000    0.000    0.000 _urlparse.py:149(urlparse)
      126    0.000    0.000    0.000    0.000 <frozen os>:674(__getitem__)
        3    0.000    0.000    0.032    0.011 default.py:119(__init__)
```
Benchmark
```
import io
import pstats
import cProfile
import httpx

merge_cookies_times = 1_000_000

pr = cProfile.Profile()
pr.enable()

httpx.Client(cookies={"default":"none"})._merge_cookies(cookies={f"{k}":f"{v}" for k,v in zip(range(merge_cookies_times), range(merge_cookies_times))})

pr.disable()
s = io.StringIO()
ps = (pstats.Stats(pr, stream=s)
      .strip_dirs()
      .sort_stats("tottime"))
ps.print_stats(10)
print(s.getvalue())
```
Github Action Workflow test all passed.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,19 +11,19 @@ chardet==5.2.0
 # Documentation
 mkdocs==1.5.3
 mkautodoc==0.2.0
-mkdocs-material==9.5.6
+mkdocs-material==9.5.12
 
 # Packaging
-build==1.0.3
-twine==4.0.2
+build==1.1.1
+twine==5.0.0
 
 # Tests & Linting
-coverage[toml]==7.4.1
-cryptography==42.0.4
+coverage[toml]==7.4.3
+cryptography==42.0.5
 mypy==1.8.0
-pytest==8.0.0
-ruff==0.1.15
+pytest==8.0.2
+ruff==0.3.0
 trio==0.24.0
 trio-typing==0.10.0
 trustme==1.1.0
-uvicorn==0.27.0.post1
+uvicorn==0.27.1


httpx/_client.py
```diff
@@ -397,7 +397,7 @@ def _merge_cookies(
         Merge a cookies argument together with any cookies on the client,
         to create the cookies used for the outgoing request.
         """
-        if cookies or self.cookies:
+        if cookies is None or self.cookies is None:
```
tomchristie, line 400: 🧐 This... seems to be the inverse of the logic before?

httpx/_client.py
```diff
@@ -397,7 +397,7 @@ def _merge_cookies(
         Merge a cookies argument together with any cookies on the client,
         to create the cookies used for the outgoing request.
         """
-        if cookies is None or self.cookies is None:
+        if not (cookies is None or self.cookies is None):
```
zanieb, line 400: This is not the same either... `not (x or y)` is not the same is `not x or not y`

For example, `x is true` and `y is false` then `not (x or y)` is `false` and `not x` or `not y` is `true`.

Additionally, `self.cookies` is _always_ not `None` so this will not work as intended https://github.com/encode/httpx/blob/ae4d72745c9aaf1d58db5ef5e6f19eeee83f7fe3/httpx/_client.py#L183

httpx/_client.py
```diff
@@ -397,7 +397,7 @@ def _merge_cookies(
         Merge a cookies argument together with any cookies on the client,
         to create the cookies used for the outgoing request.
         """
-        if cookies or self.cookies:
+        if not (cookies is None or self.cookies is None):
```
T-256, line 400:
```suggestion
        if not (cookies is None and self.cookies is None):
```
Then I think we can merge it after changelog added.

zanieb: No, we cannot. `self.cookies` is always not `None`.

zanieb (changes requested):
See https://github.com/encode/httpx/pull/2894/files#r1360817089

Marking this to avoid confusion re the review status.

PR #2901, opened 2023-10-25
Use ruff format

karpetrosyan:
[Ruff](https://docs.astral.sh/ruff/) introduced a new formatter that can completely replace black, allowing us to use ruff instead of ruff + black.

Ruff formatter, like ruff linter, is extremely fast.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,10 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 [tool.ruff]
 select = ["E", "F", "I", "B", "PIE"]
 ignore = ["B904", "B028"]
-line-length = 120
+line-length = 88
+
+[tool.ruff.pycodestyle]
+max-line-length = 120
 
 [tool.ruff.isort]
 combine-as-imports = true
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,13 +19,12 @@ build==0.10.0
 twine==4.0.2
 
 # Tests & Linting
-black==23.9.1
 coverage[toml]==7.3.0
 cryptography==41.0.4
 mypy==1.5.1
 types-certifi==2021.10.8.2
 pytest==7.4.3
-ruff==0.0.291
+ruff==0.1.3
 trio==0.22.2
 trio-typing==0.9.0
 trustme==1.1.0
--- a/scripts/check
+++ b/scripts/check
@@ -9,6 +9,6 @@ export SOURCE_FILES="httpx tests"
 set -x
 
 ./scripts/sync-version
-${PREFIX}black --check --diff $SOURCE_FILES
+${PREFIX}ruff format $SOURCE_FILES --diff
 ${PREFIX}mypy $SOURCE_FILES
 ${PREFIX}ruff check $SOURCE_FILES
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -64,7 +64,7 @@ SOCKET_OPTION = typing.Union[
 def map_httpcore_exceptions() -> typing.Iterator[None]:
     try:
         yield
-    except Exception as exc:  # noqa: PIE-786
+    except Exception as exc:
         mapped_exc = None
 
         for from_exc, to_exc in HTTPCORE_EXC_MAP.items():
--- a/scripts/lint
+++ b/scripts/lint
@@ -9,4 +9,4 @@ export SOURCE_FILES="httpx tests"
 set -x
 
 ${PREFIX}ruff --fix $SOURCE_FILES
-${PREFIX}black $SOURCE_FILES
+${PREFIX}ruff format $SOURCE_FILES


pyproject.toml
```diff
@@ -94,8 +94,9 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 # https://beta.ruff.rs/docs/configuration/#using-rufftoml
 [tool.ruff]
 select = ["E", "F", "I", "B", "PIE"]
-ignore = ["B904", "B028"]
-line-length = 120
```
karpetrosyan, line 98: Previously, we had a 120 limit on the ruff linter and an 88 limit on the black formatter, so our entire code base had an 88 limit.

Kludex:
Announcement: https://astral.sh/blog/the-ruff-formatter

scripts/check
```diff
@@ -9,6 +9,6 @@ export SOURCE_FILES="httpx tests"
 set -x
 
 ./scripts/sync-version
-${PREFIX}black --check --diff $SOURCE_FILES
+${PREFIX}ruff format $SOURCE_FILES --check
```
T-256, line 12:
```suggestion
${PREFIX}ruff format $SOURCE_FILES --diff
```

T-256 (approved):

scripts/check
karpetrosyan: Does ruff format support --diff option?

```
ruff format test.py --check --diff
error: unexpected argument '--diff' found

Usage: ruff format <FILES|--check|--config <CONFIG>|--respect-gitignore|--no-respect-gitignore|--force-exclude|--no-force-exclude|--line-length <LINE_LENGTH>|--isolated|--stdin-filename <STDIN_FILENAME>|--preview|--no-preview>

```

T-256: > Does ruff format support --diff option?

Yes.
```
>ruff --version
ruff 0.1.2

>ruff help format
Run the Ruff formatter on the given files or directories

Usage: ruff format [OPTIONS] [FILES]...

Arguments:
  [FILES]...  List of files or directories to format

Options:
      --check
          Avoid writing any formatted files back; instead, exit with a non-zero status code if any files would have been modified, and zero otherwise
      --diff
          Avoid writing any formatted files back; instead, exit with a non-zero status code and the difference between the current file and how the formatted file would look like

...
```

musale (approved):
This looks good to me.

pyproject.toml
```diff
@@ -94,8 +94,9 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 # https://beta.ruff.rs/docs/configuration/#using-rufftoml
```
zanieb, line 94: Fyi the documentation moved to docs.astral.sh/ruff

pyproject.toml
zanieb: Not technically true, as Black's limit is used to attempt to split lines but sometimes it cannot. Ruff would raise E501 on lines exceeding 120 which was the "true" line limit. Now you've disabled E501 so lines can exceed the maximum length without warning.

Personally I think having line-length enforcement in addition to the formatter is overkill, but this is a change.

karpetrosyan: I believe we should keep the real line-length limit at 120, as we did previously, and not ignore E501 so ruff can warn us about long strings that ruff format cannot fix.

Kludex (changes requested):
Change to 88 back, please.

Or at least let's discuss before.

karpetrosyan:
> Change to 88 back, please. At least let's discuss first.

It was 120 originally, and I just reverted that change :confused: .

> Also, what's the time black takes to format httpx?

I don't think it matters; we just have an opportunity to get rid of the extra dependency.

Kludex:
> > Change to 88 back, please. At least let's discuss first.
> 
> It was 120 originally, and I just reverted that change 😕 .

It was 120 on linting... It was 88 on format.

karpetrosyan:
> It was 120 on linting... It was 88 on format.
> Change to 88 back, please. At least let's discuss first.

How do you want me to do it? Both the formatter and the linter have a single line length. I'm not sure what the "change back to 88" event means here.

Setting the formatter line length equal to the linter line length feels more natural to me.

pyproject.toml
T-256: I think for now you can keep current behavior by this configuration:

```toml
# https://docs.astral.sh/ruff/configuration/#using-rufftoml
[tool.ruff]
ignore = ["B904", "B028"]

[tool.ruff.pycodestyle]
max-line-length = 120

[tool.ruff.isort]
combine-as-imports = true
```
Ref: https://docs.astral.sh/ruff/settings/#pycodestyle-max-line-length
Use default `line-length` for formatting and warn about long strings via `max-line-length`.

requirements.txt
```diff
@@ -19,13 +19,12 @@ build==0.10.0
...
 
 # Tests & Linting
-black==23.9.1
 coverage[toml]==7.3.0
 cryptography==41.0.4
 mypy==1.5.1
 types-certifi==2021.10.8.2
 pytest==7.4.2
-ruff==0.0.291
+ruff==0.1.2
```
T-256, line 27:
```suggestion
ruff==0.1.3
```

pyproject.toml
tomchristie: > I think for now you can keep current behavior by this configuration

That seems helpful, thanks.

Their announcement includes this...

> "ruff format achieves >99.9% compatibility with Black, as measured by changed lines. When migrating an existing project to Ruff’s formatter, the vast majority of lines will be formatted identically"

We ought to aim for this PR having as minimal a change footprint as possible.

Kludex: > I'm not sure why the formatter limit is 88 and the linter limit is 120.

The obvious guess is the formatter is an ecosystem replacement to black, and the linter is a replacement for flake8.

We aim to have the same setup/configuration/organization across the projects of encode, and the decision here will affect the others.

Also, we always have a policy of minimal changes on each PR.

karpetrosyan: > We aim to have the same setup/configuration/organization across the projects of encode, and the decision here will affect the others.

Yes, it makes sense now why we might want to continue using this approach. 
Compared to the request to "change back to 88," this explanation is far more detailed.

Anyway, if you like, this pull request could be a good starting point for migrating to the "same limitation for linter and formatter" approach across the encode projects.

tomchristie: > Also, we always have a policy of minimal changes on each PR.

👍 - I'm okay with reviewing this pr if/when it doesn't substantially affect the existing formatting.

T-256: > Anyway, if you like, this pull request could be a good starting point for migrating to the "same limitation for linter and formatter" approach across the encode projects

As you also pointed there are some cases (e.g. long strings) formatter unable to keep line length so as POV of formatter the code is formatted, but linter will raise error on that line.
Here are all lines raises `E501` with line length > 88:
```shell
> ruff check .
httpx\_client.py:1267:89: E501 Line too long (96 > 88)
httpx\_client.py:1983:89: E501 Line too long (96 > 88)
httpx\_decoders.py:262:89: E501 Line too long (89 > 88)
httpx\_exceptions.py:316:89: E501 Line too long (99 > 88)
httpx\_exceptions.py:326:89: E501 Line too long (98 > 88)
httpx\_main.py:66:89: E501 Line too long (95 > 88)
httpx\_main.py:78:89: E501 Line too long (92 > 88)
httpx\_models.py:361:89: E501 Line too long (93 > 88)
httpx\_multipart.py:51:89: E501 Line too long (96 > 88)
httpx\_multipart.py:99:89: E501 Line too long (94 > 88)
httpx\_multipart.py:103:89: E501 Line too long (93 > 88)
httpx\_multipart.py:121:89: E501 Line too long (94 > 88)
httpx\_transports\default.py:193:89: E501 Line too long (108 > 88)
httpx\_transports\default.py:331:89: E501 Line too long (108 > 88)
httpx\_urlparse.py:363:89: E501 Line too long (93 > 88)
httpx\_urlparse.py:379:89: E501 Line too long (97 > 88)
httpx\_urlparse.py:458:89: E501 Line too long (89 > 88)
httpx\_urls.py:54:89: E501 Line too long (91 > 88)
httpx\_urls.py:60:89: E501 Line too long (89 > 88)
httpx\_urls.py:67:89: E501 Line too long (89 > 88)
httpx\_urls.py:118:89: E501 Line too long (93 > 88)
httpx\_urls.py:308:89: E501 Line too long (92 > 88)
httpx\_urls.py:345:89: E501 Line too long (106 > 88)
tests\client\test_auth.py:599:89: E501 Line too long (94 > 88)
tests\models\test_cookies.py:97:89: E501 Line too long (105 > 88)
tests\test_config.py:104:89: E501 Line too long (98 > 88)
tests\test_multipart.py:177:89: E501 Line too long (106 > 88)
tests\test_utils.py:112:89: E501 Line too long (100 > 88)
Found 28 errors.
```

You can fix them in separate PR then remove `max-line-length = 120`.

karpetrosyan:
Okay, I used the solution that @T-256 suggested, which was to use the max-line-length provided by the `pycodestyle` settings to keep this pull request as simple as possible.

Later, I'll start a discussion about enforcing limits to the same number.

Kludex (approved):

PR #2909, opened 2023-10-30
Drop unneccessary `binascii` import

lovelydinosaur:
This is a minor refactoring that removes an unnecessary import.
We should aim to remove imports and dependencies wherever possible, and aim for a low-overall complexity.

I'm not including this in the `CHANGELOG` since it's a refactoring rather than a user-visible functional change.

### DIFF
--- a/httpx/_multipart.py
+++ b/httpx/_multipart.py
@@ -199,7 +199,7 @@ class MultipartStream(SyncByteStream, AsyncByteStream):
         boundary: typing.Optional[bytes] = None,
     ) -> None:
         if boundary is None:
-            boundary = os.urandom(16).hex()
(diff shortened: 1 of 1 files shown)


httpx/_multipart.py
```diff
@@ -200,7 +199,7 @@ def __init__(
         boundary: typing.Optional[bytes] = None,
     ) -> None:
         if boundary is None:
-            boundary = binascii.hexlify(os.urandom(16))
+            boundary = ''.join([f'{byte:02x}' for byte in(os.urandom(16)])
```
T-256, line 202: Oh, I see python 2 user 😄 
```suggestion
            boundary = os.urandom(16).hex()
```

zanieb (approved):

PR #2920, opened 2023-11-02
Support legacy WSGI write functions

Cito:
# Summary

HTTPX already supports [calling into WSGI apps](https://www.python-httpx.org/advanced/#calling-into-python-web-apps).

However, this support does not include legacy (WSGI 1) apps and frameworks that use a [write callable](https://peps.python.org/pep-3333/#the-write-callable). This small PR adds support for these apps and frameworks, without changing the way how the WSGI support works otherwise. It also works as specified in the PEP when both write callable and iterable output are used.

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion.
     (see #2921)
     (doesn't need documentation - just makes a special case work that didn't work before)


tomchristie:
> However, this support does not include legacy (WSGI 1) apps and frameworks that use a [write callable](https://peps.python.org/pep-3333/#the-write-callable).

Interesting thanks ☺️

Could you show me a real-world scenario to demo this against?

Cito:
Hi Tom. A real-world example would be [Webware for Python](https://webware-for-python-3.readthedocs.io/en/latest/) which uses WSGI with write callables. You can create and run a demo with Webware for Python as follows:

```sh
# create and activate virtual env
python -m venv venv
source venv/bin/activate
# install web framework
pip install "Webware-for-Python[dev]>=3"
# create and run a demo app
webware make DemoApp
cd DemoApp
webware serve -b
```

In the web browser, you should see a simple demo application, served by a development web server.

We now want to fetch pages with httpx from this demo app using WSGI directly, i.e. without running the web server.

So you can stop the running app, but keep the virtualenv activated and install httpx with this patch in it. In the same directory DemoApp, create the following Python script:

```python
### Create webware application

import webware
webware.addToSearchPath()
from Application import Application

settings = {'PrintConfigAtStartUp': False}
app = Application(".", settings=settings, development=True)

### Test run the application

import httpx

print("Home page fetched via WSGI:")
base_url = "http://localhost:8080/"
client= httpx.Client(app=app, base_url=base_url)
print(client.get(base_url).text)
```

If you run this script, you should now see in the output the HTML of the start page of the demo app. The point here is that this works without a web server, i.e. using the WSGI application in-process. Without the patch in this PR, this would not work.

In the PR, I have also created a test case with a demo app `echo_body_with_write_callable` that is using the write callable feature.

tomchristie:
Thanks @Cito. I think we'll leave that particular appendix in the past.

Cito:
Ok, thanks for triaging this, although of course I am a bit disappointed since the implementation is so simple.

PR #2922, opened 2023-11-02
Ruff linter: Use the default line-length

T-256:
A prototype to avoid using strings and comments longer than 88 (the default).
88 line length also used by the formatter.

Follow-up #2901

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,10 +95,6 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 [tool.ruff]
 select = ["E", "F", "I", "B", "PIE"]
 ignore = ["B904", "B028"]
-line-length = 88
-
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -95,10 +95,6 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 [tool.ruff]
 select = ["E", "F", "I", "B", "PIE"]
 ignore = ["B904", "B028"]
-line-length = 88
```
T-256, line 98: Do we need mention the default value?

pbelskiy: Explicit better than implicit

T-256: I didn't find in Ruff's docs that prefers default values to use explicitly.
As this PR also removes the `max-line-length` so there aren't any line configs at all.

PR #2927, opened 2023-11-04
Add tests and requirements.txt to sdist

mtelka:
# Summary

This adds `tests` and `requirements.txt` to sdist.

# Checklist

See https://github.com/encode/httpx/discussions/2926#discussioncomment-7472927

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,8 @@ include = [
     "/httpx",
     "/CHANGELOG.md",
     "/README.md",
+    "/tests",
+    "/requirements.txt",
 ]
 
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -70,6 +70,8 @@ include = [
     "/httpx",
     "/CHANGELOG.md",
     "/README.md",
+    "/tests",
+    "/requirements.txt",
```
tomchristie, line 74:
```suggestion
    "/tests",
```

mtelka: Why without `requirements.txt`?  The file is needed to get list of requirements needed to run tests.

karpetrosyan: Just to enforce the `sdist` files that we already have in `httpcore`.

Any other changes should go through the traditional path, which is `discussion -> issue -> pull request`.

karpetrosyan (approved):

PR #2929, opened 2023-11-08
hotfix: `urlparse` keep the existing escape string in query

lloydzhou:
# Summary

https://github.com/encode/httpx/discussions/2844#discussioncomment-7505285

# Checklist


tomchristie:
Superseeded by #2990

PR #2935, opened 2023-11-10
Inline Brotli samples in tests

### LINKED ISSUE #2906: Support brotlicffi in tests
While the package itself can work with either `brotli` or `brotlicffi`, `tests/test_decoders.py` explicitly requires `brotli`. We're currently working on having all packages support `brotlicffi` in Gentoo, since `brotli` doesn't work reliably on PyPy3. Could you please consider making the test accept `brotlicffi` as well?

I suppose the simplest way would be to reuse the existing logic, i.e.:

    from httpx._compat import brotli

I can submit a PR for that.

---

- [x] Initially raised as discussion #2903

mgorny:
# Summary

Inline the compressed Brotli samples in tests to make them independent of Brotli implementation.  This makes it possible to run the test suite both against Brotli and brotlicffi.

Fixes #2906

# Checklist

- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (n/a)
- [ ] I've updated the documentation accordingly. (n/a)


tomchristie (approved):

tomchristie:
Thanks!

mgorny:
Thank you!

PR #2938, opened 2023-11-12
add missing type hints to __init__(...)

pbelskiy:
Hello! Thanks for awesome package! 🔥

First of all, I though about create an issue, but this is too small fix for that, dear maintainers
should I create an Issue or Discussion for that super small fixes?

# Summary
Here I did little work for fixing of missed type hints of few of __init__()

According to https://peps.python.org/pep-0484/
`(Note that the return type of __init__ ought to be annotated with -> None.`

So type checkers (mypy, pyright) will work fine after this PR.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+* Add missing type hints to few `__init__()` methods
+
 ## 0.25.1 (3rd November, 2023)
 
 * Add support for Python 3.12. (#2854)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-* Add missing type hints to few `__init__()` methods
+* Add missing type hints to few `__init__()` methods. (#2938)
 
 ## 0.25.1 (3rd November, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
 * Add missing type hints to few `__init__()` methods. (#2938)
 
 ## 0.25.1 (3rd November, 2023)
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -126,7 +126,7 @@ class BasicAuth(Auth):
 
     def __init__(
         self, username: typing.Union[str, bytes], password: typing.Union[str, bytes]
-    ):
+    ) -> None:
         self._auth_header = self._build_auth_header(username, password)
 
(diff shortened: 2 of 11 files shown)


karpetrosyan:
Hi @pbelskiy! Thanks for the pull request :pray: 

> dear maintainers should I create an Issue or Discussion for that super small fixes?

We prefer **discussion -> issue -> pull request** path rather than code-change first, but I am okay reviewing this small change.

Yeah, this change makes sense, also because in some places we do not use annotation for `__init__`, though I found some classes where we use it.

Demo:

``` shell
grep -irn __init__ tests

tests/test_multipart.py:385:        def __init__(self, iterator: typing.Iterator[bytes]) -> None:
tests/test_content.py:68:        def __init__(self, content: bytes) -> None:
tests/client/test_async_client.py:183:        def __init__(self) -> None:
tests/client/test_async_client.py:215:        def __init__(self, name: str):
tests/client/test_client.py:232:        def __init__(self) -> None:
tests/client/test_client.py:263:        def __init__(self, name: str):
tests/client/test_auth.py:27:    def __init__(self, auth_header: str = "", status_code: int = 200) -> None:
tests/client/test_auth.py:38:    def __init__(
tests/client/test_auth.py:96:    def __init__(self, repeat: int):
tests/client/test_auth.py:123:    def __init__(self, token: str) -> None:
tests/client/test_auth.py:142:    def __init__(self) -> None:
```

karpetrosyan:
Can you also add a changelog for this pr?

pbelskiy:
@karpetrosyan 

Okay, I will add today later, thanks!

pbelskiy:
@karpetrosyan 

From CHANGELOG.md
`All notable changes to this project will be documented in this file.`

Is adding few type hints are really notable?

karpetrosyan:
I think we should have a changelog, even if it's a little type change, so the user having type issues with the new release can understand what could be the reason.
We do not add a changelog if it's a tool change or the user cannot interact with that change.

Other changelog that we have:

> Change the type of Extensions from Mapping[Str, Any] to MutableMapping[Str, Any]. (#2803)

Anyway, if we are sure that there are no regression chances, we can skip the changelog for this change.

pbelskiy:
Thanks, agree with you. So I've added info to changelog.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+* Add missing type hints to few `__init__()` methods
```
karpetrosyan, line 9:
```suggestion
* Add missing type hints to few `__init__()` methods. (#2938)
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+* Add missing type hints to few `__init__()` methods. (#2938)
```
karpetrosyan, line 9:
```suggestion
### Added
* Add missing type hints to few `__init__()` methods. (#2938)
```

karpetrosyan (approved):

tomchristie (approved):

PR #2976, opened 2023-12-04
Pre encoded path/params/fragments should be kept

### LINKED ISSUE #2883: ULR parser percent encoding corrupting query params

elupus:
# Summary

URL with pre-encoded segments would end up double encoded when parsed by the URL parser. We must consider the % as a safe value, to avoid encoding it here.

Related to comments in #2883
Related #2723
Duplicate #2929

# Checklist

### DIFF
--- a/tests/test_urlparse.py
+++ b/tests/test_urlparse.py
@@ -150,6 +150,26 @@ def test_param_with_existing_escape_requires_encoding():
     assert str(url) == "http://webservice?u=http%3A%2F%2Fexample.com%3Fq%3Dfoo%252Fa"
 
 
+def test_param_with_existing_escape():
+    url = httpx.URL("https://webservice/?u=/%3D%26&v=1%202")
+    assert str(url) == "https://webservice/?u=%2F%3D%26&v=1%202"
+    assert url.params["u"] == "/=&"
+    assert url.params["v"] == "1 2"
+
+
+def test_param_nested_urls_in_query():
+    src = "special;string with:reserved?cha%20raca/ters&d"
+    data = str(httpx.URL("http://webservice", params={"u": src}))
+    data = str(httpx.URL("http://webservice", params={"u": data}))
+    data = str(httpx.URL("http://webservice", params={"u": data}))
+
+    url = httpx.URL(data)
+    url = httpx.URL(url.params["u"])
+    url = httpx.URL(url.params["u"])
+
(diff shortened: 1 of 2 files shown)


T-256:
I don't think this PR fixes #2883. That was about encodes forward slash.
Does [this comment](https://github.com/encode/httpx/pull/2976/files#diff-78d8d93b5dd4c77d99c3e2b46b7286ba71a8fd60e92d8bd4eee5fb200b4f87bfR256) need to be updated accordingly?
Is this behavior change?

elupus:
> I don't think this PR fixes #2883. That was about encodes forward slash. Does [this comment](https://github.com/encode/httpx/pull/2976/files#diff-78d8d93b5dd4c77d99c3e2b46b7286ba71a8fd60e92d8bd4eee5fb200b4f87bfR256) need to be updated accordingly? Is this behavior change?

Hmm. Well not the initial report there indeed. But the the receiver of that url should url decode. Not doing that would be a bug in its parser.

It might only fix the followup comments with similar issues.

tests/test_urlparse.py
```diff
@@ -150,6 +150,26 @@ def test_param_with_existing_escape_requires_encoding():
     assert str(url) == "http://webservice?u=http%3A%2F%2Fexample.com%3Fq%3Dfoo%252Fa"
 
 
+def test_param_with_existing_escape():
+    url = httpx.URL("https://webservice/?u=/%3D%26&v=1%202")
+    assert str(url) == "https://webservice/?u=%2F%3D%26&v=1%202"
```
ttys0dev, line 155: I don't think this will work, it's still corrupting the parameter by encoding `/` to `%2F` which is something not all servers can handle.

elupus:
Moving to draft n favor off https://github.com/encode/httpx/pull/2980

zanieb:
While I think we may want to revert the behavior that encodes `/`, I think this change doesn't need to be conflated with that. We should not double-encode already encoded characters — that's a clear bug from my perspective. I'd rather review the changes in isolation so we can have a discussion about one thing at a time.

tomchristie:
Thanks for taking a look at this @elupus, I've been planning to address this for a while and have taken a fresh pass on it at #2990. Reviews appreciated.

tomchristie:
Superseeded by #2990

PR #2980, opened 2023-12-04
Differenctiate between encoded form and element form for url

### LINKED ISSUE #2883: ULR parser percent encoding corrupting query params

elupus:
# Summary

* Allow quote and is_safe to take sets as input to speed up calculations
* Differenciate between reencoding raw url data, and piece wise constructed query parameters
* Revert always encoding of forward slashes in query

Fixes #2883
Pertially reverts: #2721

Related #2723, #2929

# Checklist


T-256:
Hi @elupus
Is #2990 also solves your needs at here?

PR #2981, opened 2023-12-05
Reorganise tests in 'test_url.py'

tomchristie:
I would like to deal with https://github.com/encode/httpx/issues/2883.

In order to tackle this comprehensively I'm suggesting we *start by tidying up our existing URL test cases*.
There are two parts to this:

* Reorganise the existing tests in `models/test_url.py` - The current ordering of the tests isn't coherent.
* Remove `test_urlparse.py`, moving test cases into `models/test_url.py` - All the test cases there are now properly against the public `httpx.URL()` interface, tackled previously as part of the #2492.

This pull request deals with the first of these two bullet-points.
I'll follow up with the second part once this is in.

Here's the explanation of the proposed change (assuming I've got it correct 😅)...

* Tests are re-ordered, these no removal, addition, or modifying of test cases.
* Some docstrings have been added.
* Some comments helping claify the structure have been added.

The ordering for the test cases here is...

* General tests for `httpx.URL`.
* General tests for `httpx.QueryParams`.
* Tests for `httpx.URL.join()`.
* Tests for `httpx.URL.copy_with()`.
* Tests for `httpx.URL.copy_<...>.params()` methods.
* Tests for IDNA hostname support.
* Tests for IPv6 hostname support.

### DIFF
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -3,76 +3,6 @@ import pytest
 import httpx
 
 
-@pytest.mark.parametrize(
-    "given,idna,host,raw_host,scheme,port",
-    [
-        (
-            "http://中国.icom.museum:80/",
-            "http://xn--fiqs8s.icom.museum:80/",
-            "中国.icom.museum",
-            b"xn--fiqs8s.icom.museum",
-            "http",
-            None,
-        ),
-        (
-            "http://Königsgäßchen.de",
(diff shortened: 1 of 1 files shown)


Kludex (approved):

michaeloliverx (approved):

tomchristie:
Thanks @Kludex, @michaeloliverx. 🙏🏼

PR #2982, opened 2023-12-05
Moving test cases into 'test_url.py'

tomchristie:
Second part of #2981.

Move test cases from `tests/test_urlparse.py` into `tests/models/test_url.py`.

A comprehensive review might double check the number of tests (or names of tests) that are being run.
Or you could eyeball this.


tomchristie:
Hm, I'm missing coverage here. What have I missed?...

michaeloliverx (approved):

PR #2990, opened 2023-12-07
Cleanup URL percent-encoding behavior.

### LINKED ISSUE #2883: ULR parser percent encoding corrupting query params

tomchristie:
Closes #2883

#### Cleanup test cases

* The first two commits here cleanup three test cases into an exactly equivalent parameterised test case.
* The next commit supplements the test cases with some extra cases *(using with the current behaviour for the assertions)*.
* Drop a test case that's now been superseeded.

#### Determine correct behaviour

For each case here I'm listing...

* The URL under test.
* The URL once it's been percent-escaped by chrome.
* The URL we're *currently* percent-escaping into.

**Test 1: Correct**

```python
# URL with unescaped chars in path.
https://example.com/!$&'()*+,;= abc ABC 123 :/[]@
https://example.com/!$&'()*+,;=%20abc%20ABC%20123%20:/[]@
https://example.com/!$&'()*+,;=%20abc%20ABC%20123%20:/[]@
```

**Test 2: Correct**

```python
# URL with escaped chars in path.
https://example.com/!$&'()*+,;=%20abc%20ABC%20123%20:/[]@
https://example.com/!$&'()*+,;=%20abc%20ABC%20123%20:/[]@
https://example.com/!$&'()*+,;=%20abc%20ABC%20123%20:/[]@
```

**Test 3: Incorrect. We're double-escaping here.**

```python
# URL with mix of unescaped and escaped chars in path.
https://example.com/ %97%98%99
https://example.com/%20%97%98%99
https://example.com/%20%2597%2598%2599
```

**Test 4: Incorrect. We don't need to escape `/` in the query portion. We do need to percent escape `'`.**

```python
# URL with unescaped chars in query.
https://example.com/?!$&'()*+,;= abc ABC 123 :/[]@?
https://example.com/?!$&%27()*+,;=%20abc%20ABC%20123%20:/[]@?
https://example.com/?!$&'()*+,;=%20abc%20ABC%20123%20:%2F[]@?
```

**Test 5: Correct.**

```python
# URL with escaped chars in query.
https://example.com/?!$&%27()*+,;=%20abc%20ABC%20123%20:%2F[]@?
https://example.com/?!$&%27()*+,;=%20abc%20ABC%20123%20:%2F[]@?
https://example.com/?!$&%27()*+,;=%20abc%20ABC%20123%20:%2F[]@?
```

**Test 6: Correct.**

```python
# URL with mix of unescaped and escaped chars in query.
https://example.com/?%20%61%62%63
https://example.com/?%20%61%62%63
https://example.com/?%20%61%62%63
```

**Test 7: Correct.**

```python
# URL encoding characters in fragment.
# (The fragment isn't sent as part of the HTTP request.)
https://example.com/#!$&'()*+,;= abc ABC 123 :/[]@?#
https://example.com/
https://example.com/
```

#### Resolve issue

* Update test cases to reflect desired behaviour instead of existing behaviour.
* Update behaviour so that double-escaping is not applied.
* Update behaviour so that `/` is treated as a safe character in the query portion of the URL.


tomchristie:
At this point I've change the expectation in "test 3" to reflect the desired behaviour, not the existing behaviour.

tomchristie:
This is once I've fixed the double escaping.

tomchristie:
And now updated to also fix test 4.

(Well, almost except we're not encoding `'` to `%27` - spec seems to imply we're getting this correct here vs. chrome)

httpx/_urlparse.py
```diff
@@ -464,4 +489,9 @@ def urlencode(items: typing.List[typing.Tuple[str, str]]) -> str:
     - https://github.com/encode/httpx/issues/2721
     - https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
     """
-    return "&".join([quote(k, safe="") + "=" + quote(v, safe="") for k, v in items])
+    return "&".join(
+        [
+            percent_encoded(k, safe="") + "=" + percent_encoded(v, safe="")
+            for k, v in items
+        ]
+    )
```
tomchristie, line 497: Comment to the gallery...

We've switch from `quote` to `percent_encoded` here, because we *always* want to percent encode items that are being provided as form inputs, with `params = {}`.

httpx/_urlparse.py
```diff
@@ -449,6 +448,32 @@ def quote(string: str, safe: str = "/") -> str:
     )
 
 
+def quote(string: str, safe: str = "/") -> str:
+    """
+    Use percent-encoding to quote a string, omitting existing '%xx' escape sequences.
+    """
```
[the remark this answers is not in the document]
T-256: I see code comments more difficult to read than that simple straightforward description.
Perhaps `safe` parameter needs more info here(?)

httpx/_urlparse.py
```diff
@@ -432,13 +432,12 @@ def is_safe(string: str, safe: str = "/") -> bool:
...
 
 
-def quote(string: str, safe: str = "/") -> str:
+def percent_encoded(string: str, safe: str = "/") -> str:
     """
-    Use percent-encoding to quote a string if required.
+    Use percent-encoding to quote a string.
     """
     if is_safe(string, safe=safe):
         return string
```
T-256, line 443: I suggest use minimal filtering instead of separated one-time function.
```py
    if all(char in NON_ESCAPED_CHARS for char in string):
        return string
```

I also noticed `NON_ESCAPED_CHARS` name in `is_safe` function has different value than it is in `percent_encoded`.

httpx/_urlparse.py
```diff
@@ -449,6 +448,32 @@ def quote(string: str, safe: str = "/") -> str:
...
+        # Add the '%xx' escape sequence.
+        parts.append(matched_text)
+        current_position = end_position
+
+    # Add any text after the final '%xx' escape sequence.
+    if current_position != len(string):
+        trailing_text = string[current_position:]
+        parts.append(percent_encoded(trailing_text, safe=safe))
+
+    return "".join(parts)
```
T-256, line 474: `parts` could be string. `parts +=` instead of `parts.append`

T-256 (approved):
I think we need minor version release for this merge, because it is behavior change.

elupus:
This unnecessarily encode slashes if you pass them as query params. Which will not solve cases with servers needing stuff like: `https://host?mimetype=application/data`.

The library will not mutate and break an existing string like that, which is good. But you can not construct it using params={"mimetime": "application/data"}

These server are in theory broken since they should expect possible encoding inside the arguments, but feels like we should avoid overly encoding that. There is quite a lot that in theory is safe here: https://github.com/encode/httpx/pull/2980/files#diff-78d8d93b5dd4c77d99c3e2b46b7286ba71a8fd60e92d8bd4eee5fb200b4f87bfR51

elupus:
Ps. Since it does solve most cases where the URL is in a pre-encoded form, its absolutely a good change. So i think something like this could go in to start with.

I think some optimization should be done with the use of sets for the reserved characters like i did in my change, since now it's doing linear scans of strings many times over during encoding, but its much less important.

httpx/_urlparse.py
tomchristie: Sure. I'd like to treat performance improvements separately to this.
I'll followup with some further work once we've got the behavioural changes in.

httpx/_urlparse.py
tomchristie: It could, yes. I've gone with this pattern because in the past we've seen performance issues with the `+=` pattern, that a `.append() ... "".join()` pattern resolves. Might not be the case in this context, since we've got strict limits on the allowable string lengths in URLs. But... let's handle any performance related considerations separately to the behavioural fixes.

httpx/_urlparse.py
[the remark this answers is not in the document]
tomchristie: Updated

tomchristie:
> I think we need minor version release for this merge, because it is behavior change.

We can commit to that yep.

tomchristie:
@T-256 Thanks for the review. ☺️
If you'd like to be added as a maintainer ping me [a private message on gitter](https://app.gitter.im/#/room/#encode_community:gitter.im).

tomchristie:
Okay, I think we're good here now.

Just pending a review from @encode/maintainers. 🙏🏼

httpx/_transports/asgi.py
```diff
@@ -1,7 +1,5 @@
 import typing
 
-import sniffio
```
T-256, line 3: `_utils.py` also imports sniffio as top-level.
Could we consider add httpcore's [`current_async_library`](https://github.com/encode/httpcore/blob/2fcd062df71555cc7de55774c6dc137551eb8692/httpcore/_synchronization.py#L21) in httpx?

T-256:
> @T-256 Thanks for the review. ☺️ If you'd like to be added as a maintainer ping me [a private message on gitter](https://app.gitter.im/#/room/#encode_community:gitter.im).

Thanks, I sent PM in gitter.

httpx/_urlparse.py
```diff
@@ -263,7 +263,7 @@ def urlparse(url: str = "", **kwargs: typing.Optional[str]) -> ParseResult:
     # We also exclude '/' because it is more robust to replace it with a percent
     # encoding despite it not being a requirement of the spec.
```
jkseppan, line 264: Doesn't this comment become incorrect with this change?

tomchristie, line 264:
```suggestion
```

tomchristie: Good catch, thanks.

httpx/_urlparse.py
```diff
@@ -449,6 +446,39 @@ def quote(string: str, safe: str = "/") -> str:
     )
 
 
+def quote(string: str, safe: str = "/") -> str:
```
karpetrosyan, line 449: I believe we should also add unit tests for these functions, rather than simply testing them with `httpx.URL`. 
This would be a more robust approach, in my opinion.

T-256: IMO current approach is fine.

> This is clearly a code-smell, because our test cases ought to be tests against our public API, rather than testing implementation details. Perhaps there's some cases where it's a necessary hack, but... perhaps not?


from https://github.com/encode/httpx/issues/2492#issue-1478857204

T-256 (approved):
Thanks, LGTM!

httpx/_urlparse.py
zanieb: I agree testing the public API should be sufficient unless something private is particularly expensive to test via the public API.

karpetrosyan: It's a matter of preference, but if we encounter regression in our `httpx.URL` tests, we will go through these functions and find the method that isn't working properly, which is why unit tests are useful.

karpetrosyan (approved):

tomchristie:
thanks @karpetrosyan 🙏🏼

PR #2997, opened 2023-12-09
Fixing missed arguments for pools when init transports

NewUserHa:
# Summary

https://github.com/encode/httpx/blob/5b5f6d8e172910371b5901f6e21717097274c513/httpx/_transports/default.py#L258-L330
eg.
* in `AsyncHTTPTransport`, `AsyncConnectionPool` has the argument of retries passed in, but `AsyncHTTPProxy` and `AsyncSOCKSProxy` have not;
* in `HTTPTransport`, it has the same issue.

discussion: https://github.com/encode/httpx/discussions/2988

This also requires a fix in https://github.com/encode/httpcore

# Checklist

### DIFF
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -140,9 +140,9 @@ class HTTPTransport(BaseTransport):
                 keepalive_expiry=limits.keepalive_expiry,
                 http1=http1,
                 http2=http2,
-                uds=uds,
-                local_address=local_address,
                 retries=retries,
+                local_address=local_address,
+                uds=uds,
                 socket_options=socket_options,
             )
         elif proxy.url.scheme in ("http", "https"):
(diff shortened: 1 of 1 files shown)


NewUserHa:
The PR requires https://github.com/encode/httpcore/pull/857 PR to pass checks

tomchristie:
> The PR requires https://github.com/encode/httpcore/pull/857 PR to pass checks

Can we just start by considering `retries` here? That's already supported by `httpcore`>1.0, right?

NewUserHa:
Ok, done.

Should open a new PR for the rest missed arguments?

tomchristie:
Okay, first up let's just have a sanity check... do connection retries actually get applied when connecting to proxies?

NewUserHa:
The `SocksProxy`?

PR #2998, opened 2023-12-11
Fixed `iter_text` adding an empty string

jamesbraza:
# Summary

Fixes `TextChunker().decode("")` not handling empty string, as exposed in https://github.com/encode/httpx/discussions/2995

# Checklist


jamesbraza:
Looks like `coverage` checks are failing because this bug was ensuring code got covered. I will adjust now

zanieb (approved):

jamesbraza:
@tomchristie so since `LineDecoder` is only used after `iter_text`, I think it's no longer possible to get empty string to be the input to `decode`. To test a null input to `LineDecoder.decode`, I would have to directly import and test it.  What do you think of that?

tests/test_decoders.py
```diff
@@ -262,6 +273,11 @@ def test_line_decoder_crnl():
     assert list(response.iter_lines()) == ["12345", "foo bar baz"]
 
 
+@pytest.mark.parametrize(["text", "expected"], [("", [])])
+def test_line_decoding_edge_cases(text: str, expected: typing.List[str]) -> None:
+    assert httpx._decoders.LineDecoder().decode(text) == expected
```
jamesbraza, line 278: This was the only way I could figure out how to cover an empty input to `LineDecoder.decode`.

jamesbraza:
Alright, done! Ready for review again @tomchristie and @zanieb 👌

tomchristie:
Looks great. Nice tidy lil PR, this. One query on the usage of chain, could probs approve either ways tho.

jamesbraza:
Thanks!  Yeah, I added the `itertools.chain` to uphold code coverage. After fixing this bug, it meant the `chunker.decode(decoder.flush())` iterable was empty, so the first `yield` didn't get covered.  By using `itertools.chain`, it basically combines the two lines so code coverage remains intact.

Would you like me to roll the `itertools.chain` out to all `iter_` methods in the `_decoders` module? I think it's a bit cleaner

tomchristie:
Not sure. Perhaps we could consider `iter_chain()` separately?

zanieb:
@jamesbraza I think the suggestion is use a no cover pragma then follow with a separate pull request that explores using `itertools.chain` consistently throughout the code.

jamesbraza:
Sounds good, agree two PRs is better. Switched to `pragma: no cover` comments

tomchristie:
Wonderful, thanks.

(Note to self: CHANGELOG.md updates as part of the PR before ya merge in the future, doofus.)

jamesbraza:
I will get around to `itertools.chain` change this weekend, and will have a `CHANGELOG.md` entry for that 👍

PR #3001, opened 2023-12-12
Enabling `ruff` C416

jamesbraza:
# Summary

I decided to decompose my PR on incorporating `itertools.chain` a bit, and some of it can be automated by [C416](https://docs.astral.sh/ruff/rules/unnecessary-comprehension/). So this PR enables [C416](https://docs.astral.sh/ruff/rules/unnecessary-comprehension/) in `ruff` to simplify code.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,7 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 
 # https://beta.ruff.rs/docs/configuration/#using-rufftoml
 [tool.ruff]
-select = ["E", "F", "I", "B", "PIE"]
+select = ["E", "F", "I", "B", "PIE", "C416"]
 ignore = ["B904", "B028"]
 
 [tool.ruff.isort]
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -95,7 +95,7 @@ replacement = 'src="https://raw.githubusercontent.com/encode/httpx/master/\1"'
 
 # https://beta.ruff.rs/docs/configuration/#using-rufftoml
 [tool.ruff]
-select = ["E", "F", "I", "B", "PIE"]
+select = ["E", "F", "I", "B", "PIE", "C416"]
```
tomchristie, line 98: I'm neither for or against these code changes, it's probably a neater style yeah.

Tho I'd prefer we stick with minimal confic where possible, so perhaps we can aim for a minimal `select` hete.

tomchristie, line 98:
```suggestion
select = ["E", "F", "I", "B", "PIE"]
```

tomchristie (approved):
Sure thing, that's neater.

tomchristie:
I've left the config simple (the less config options the better) but included the suggested changes.
Thanks. 🙏

jamesbraza:
Okay sounds good, thanks!

PR #3004, opened 2023-12-13
Using `itertools.chain` and `yield from` in `Response.iter_`

jamesbraza:
# Summary

Minor refactor of `_models` to use `yield from` and `itertools.chain` where possible.  This has two benefits:

- Removes several `pragma: no cover` comments
- Reduces `_models` by about 20 lines

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -1,5 +1,6 @@
 import datetime
 import email.message
+import itertools
 import json as jsonlib
 import typing
 import urllib.request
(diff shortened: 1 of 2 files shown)


jamesbraza:
Cc @tomchristie @zanieb here's the `itertools.chain` integration as promised. I am not attached to this PR or anything, so feel free to say you don't like the more compressed styling it introduces

tomchristie:
> feel free to say you don't like the more compressed styling it introduces

*skirts around the issue self-consciously*

tomchristie:
cue definitive action

PR #2986, opened 2023-12-06
Add test cases for userinfo in URL.

tomchristie:
Refs https://github.com/encode/httpx/issues/2985.

PR #2879, opened 2023-10-06
Add the 'proxy' parameter and deprecate 'proxies'.

karpetrosyan:
# Summary

HTTPX supports `proxies` and `mounts` arguments, which are not the same, but `mounts` can simply replace `proxies` for complex routing, which is not always used.

So for simplicity, this PR deprecates `proxies` and adds `proxy` argument.
You can still use `proxies`, but it will warn you that it is deprecated and that you should use the new API.

Use...

- `proxy` if you want to use a proxy server for all your requests.
- `mounts` if you want to use a proxy server for requests matching the URL pattern you specified.

## Proxy

```python
import httpx

# instead of 
# httpx.Client(proxies="http://127.0.0.1:80")
# and
# httpx.Client(proxies={"all://": "http://127.0.0.1:80"})
# do ...
client = httpx.Client(proxy="http://127.0.0.1:80")
```

## Mounts

```python
import httpx

# instead of
# client = httpx.Client(
#    proxies={"http://": "http://127.0.0.1:80", "https://": "https://127.0.0.1:443"}
# )
# do ...
client = httpx.Client(
    mounts={
        "http://": httpx.HTTPTransport(proxy="http://127.0.0.1:80"),
        "https://": httpx.HTTPTransport(proxy="https://127.0.0.1:443"),
    }
)
```

PR #3009, opened 2023-12-18
Version 0.26.0

karpetrosyan:
### Added

* The `proxy` argument was added. You should use the `proxy` argument instead of the deprecated `proxies`, or use `mounts=` for more complex configurations. (#2879)

### Deprecated

* The `proxies` argument is now deprecated. It will still continue to work, but it will be removed in the future. (#2879)

### Fixed

* Fix cases of double escaping of URL path components. Allow / as a safe character in the query portion. https://github.com/encode/httpx/pull/2990.
* Handle `NO_PROXY` envvar cases when a fully qualified URL is supplied as the value. (#2741)
* Allow URLs where username or password contains unescaped '@'. (#2986)
* Ensure ASGI `raw_path` does not include URL query component. (#2999)
* Ensure `Response.iter_text()` cannot yield empty strings. (#2998)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+* Fix cases of double escaping of URL path components. Allow / as a safe character in the query portion. (#2990)
 * Handle `NO_PROXY` envvar cases when a fully qualified URL is supplied as the value. (#2741)
 * Allow URLs where username or password contains unescaped '@'. (#2986)
 * Ensure ASGI `raw_path` does not include URL query component. (#2999)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## 0.26.0 (TODO)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.26.0 (20th December, 2023)
 
-### Changed
+### Added
 
-* The `proxies` argument is now deprecated. You should use the `proxy` argument instead, or use `mounts=` for more complex configurations. (#2879)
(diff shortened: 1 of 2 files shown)


karpetrosyan:
Here is the release draft: https://github.com/encode/httpx/releases/tag/untagged-117424d33cff82b70e02

T-256:
Don't forget https://github.com/encode/httpx/pull/2990

tomchristie:
Thanks @T-256 - apologies for missing the `CHANGELOG.md` entry on that.

I'd suggest...

* Fix cases of double escaping of URL path components. Allow `/` as a safe character in the query portion. #2990.

karpetrosyan:
Thanks @T-256 

I think this one is better

> Fix cases of double escaping of URL path components. Allow `/` as a safe character in the query portion. #2990.

tomchristie:
> I think this one is better

Ok. Let's go with that.

karpetrosyan:
Should I schedule it for today?

tomchristie:
That would be fine, yep.
Thanks so much, @karpetrosyan. 😄

karpetrosyan:
Here is the new one: https://github.com/encode/httpx/releases/edit/untagged-893baae112e60ca1d83c

CHANGELOG.md
```diff
@@ -4,14 +4,15 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## 0.26.0 (TODO)
```
karpetrosyan, line 7:
```suggestion
## 0.26.0 (20th December, 2023)
```

tomchristie (approved):
We might want to make it clear the `proxies=...` does still continue to work, but will raise a warning?

Looking great, thanks.

karpetrosyan:
I'd recommend something like this.

## 0.26.0 (TODO)

### Added

* You should use the `proxy` argument instead of the deprecated `proxies`, or use 'mounts=' for more complex configurations. (#2879)

### Deprecated

* The `proxies` argument is now deprecated. It will still continue to work, but it will be removed in the future. (#2879)

tomchristie:
Neat.

karpetrosyan:
The final one: https://github.com/encode/httpx/releases/edit/untagged-8838a3cfc61a09bc35d6

tomchristie:
Hrm... we don't seem to have triggered a `publish` event?

https://github.com/encode/httpx/actions/workflows/publish.yml

tomchristie:
Ah okay, done.

PR #3013, opened 2023-12-20
Adding ability to raise_for_status with response text rather than generic text

cgood92:
# Summary

If a response has information in it that is helpful for the client, then I think that text from the response should be the message in the exception that is raised, not some generic "4xx and 5xx are bad" messages.

For example, today, as I'm using `httpx` against a service I own, I keep getting `422` errors.  In the response is the information that I need to debug, so I want to see that in the error thrown.  For example, a response might say:

```
422: startDate is required
```

```
429: You have surpassed your quota for the day.  If you feel this is in error, please see: xyz.com
```

I recognize that I can get this information if I don't want to use `raise_for_status`, but it is my pattern to write code like:

```python
httpx.get(url).raise_for_status().json()
```

And that is just so convenient.  After this MR, in order to get the text from the response, I can just:

```python
httpx.get(url).raise_for_status(True).json()
```

I have kept the default behavior as-is, with the ability to change it with this boolean flag.  I would think a lot, if not most, people would want this behavior.

# Checklist

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -70,7 +70,7 @@
   * The amount of time elapsed between sending the request and calling `close()` on the corresponding response received for that request.
   [total_seconds()](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds) to correctly get
   the total elapsed seconds.
-* `def .raise_for_status()` - **Response**
+* `def .raise_for_status([use_response_text])` - **Response**
 * `def .json()` - **Any**
 * `def .read()` - **bytes**
 * `def .iter_raw([chunk_size])` - **bytes iterator**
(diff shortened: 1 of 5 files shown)


cgood92:
@tomchristie would you mind reviewing this PR?  Thanks!

tomchristie:
> would you mind reviewing this PR? Thanks!

https://github.com/encode/httpx/blob/1d526a018060d158e7548366216491c413756872/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L10

Probably needs to start as a... [😀 💬](https://github.com/encode/httpx/discussions/new?category=general)

There's quite a lot of potentially hidden complexities here...

* Do we want extra API configuration.
* Response may or may not have been read.
* Response content might raise a decode error.
* Response content might not be text.
* Response content might be large.

I'm going to close the pull request off for now, in favour of a discussion.

PR #3015, opened 2023-12-21
`types-certifi` and `types-chardet` are no longer needed

mtelka:
# Summary

Both `certifi` and `chardet` now contains typing info so we no longer need `types-certifi` and `types-chardet`.
See also https://github.com/encode/httpx/discussions/3014

# Checklist


tomchristie (approved):
Good call, thanks.

PR #3020, opened 2023-12-21
Tidy up import

tomchristie:
Tidy up an unnecessary import.

### DIFF
--- a/tests/test_asgi.py
+++ b/tests/test_asgi.py
@@ -3,7 +3,6 @@ import json
 import pytest
 
 import httpx
-from httpx import ASGITransport
 
 
 async def hello_world(scope, receive, send):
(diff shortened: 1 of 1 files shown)


karpetrosyan:
I believe this is not the only place where we do this. Maybe we could decide on a single import style and enforce it with a [ruff rules](https://docs.astral.sh/ruff/rules/)?

karpetrosyan:
For example, I found a [useful rule](https://docs.astral.sh/ruff/rules/banned-import-from/) that can prevent `from` imports for specific paths.

Configuration:

``` toml
[tool.ruff]
select = ["ICN003"]
    
[tool.ruff.lint.flake8-import-conventions]
banned-from = ["httpx"]
```

This would prevent any import with the "from ... import ..." syntax with "httpx" in the left side.

Examples:

``` python
from httpx import ASGITransport  # Error
from httpx._models import Request # Ok
from httpx import Request # Error
```

blag (approved):

tomchristie:
> I believe this is not the only place where we do this. 

I picked this one in isolation, since I think it's the only place where we've got two different types of import against the same module.

(Yes we *could* have a strict import style throughout, interesting point.)

karpetrosyan (approved):

PR #3022, opened 2023-12-25
Add `httpx.SSLContext` configuration.

karpetrosyan:
Discussed in https://github.com/encode/httpx/discussions/3007

TODO:

- [ ] Add changelog
- [x] Add more migration examples

### Enabling and disabling verification

``` python
import httpx

# instead of 
httpx.Client(verify=False)

# do
context = httpx.SSLContext(verify=False)
httpx.Client(ssl_context=context)
```

### Use a different set of certificates

``` python
import httpx

# instead of
httpx.Client(verify="path/to/certs.pem")

# do
context = httpx.SSLContext(verify="path/to/certs.pem")
httpx.Client(ssl_context=context)
```

## Client side certificates

```python
import httpx

# instead of
httpx.Client(cert="path/to/client.pem")

# do
context = httpx.SSLContext(cert="path/to/client.pem")
httpx.Client(ssl_context=context)
```

## Using alternate SSL contexts

```python
import ssl
import httpx

context = ssl.create_default_context()

# instead of
client = httpx.Client(verify=context)

# do
client = httpx.Client(ssl_context=context)
```


T-256:
Removing `verify` and `cert` is too aggressive, I'd recommend deprecate them first.

karpetrosyan:
Yes, I agree that removing them completely is not what the user expects. 
However, I believe migration is simple enough to skip that stage in order to avoid making that part of the code overly complicated.

Or should we deprecate them anyway? Not obvious to me.

T-256:
There are dropping from top-level API, I think deprecation policy should be applied here.

Also, we have other deprecated drops those pending to remove in v1.0.0 (IIUC):
- `proxies=`
- `data=` (non-Mapping objects)
- `cookies=` (per-request)

httpx/_config.py
```diff
@@ -42,150 +40,102 @@ class UnsetType:
...
-        trust_env: bool = True,
-        http2: bool = False,
-    ) -> None:
-        self.cert = cert
-        self.verify = verify
-        self.trust_env = trust_env
-        self.http2 = http2
-        self.ssl_context = self.load_ssl_context()
+        cert: typing.Optional[CertTypes] = None,
+    ) -> "SSLContext":
```
tomchristie, line 51: I'm curious why we're overriding `__new__` instead of `__init__` here?

httpx/_config.py
```diff
@@ -42,150 +40,102 @@ class UnsetType:
...
-        elif Path(self.verify).exists():
-            ca_bundle_path = Path(self.verify)
+        elif Path(verify).exists():
+            ca_bundle_path = Path(verify)
         else:
             raise IOError(
                 "Could not find a suitable TLS CA certificate bundle, "
-                "invalid path: {}".format(self.verify)
+                "invalid path: {}".format(verify)
             )
```
tomchristie, line 95: We confuse the flow here by having this section of code separated from where `ca_bundle_path` is actually used. It'd be more clear if we pushed this block down to just before `if ca_bundle_path.is_file():`.

tomchristie: (Tho perhaps we address clean-ups separately to the main API behavioral changes here?)

httpx/_config.py
karpetrosyan: Yes, it should definitely be in __init__. I believe I encountered some issues and have temporarily moved it to __new__ to see how it looks in general.

httpx/_config.py
karpetrosyan: Separating them will make it easier to review this PR, which contains an important API change.

karpetrosyan: Another important question is whether we need to deprecate previous arguments or simply delete them.

httpx/_client.py
```diff
@@ -1455,6 +1455,9 @@ def _init_proxy_transport(
     ) -> AsyncBaseTransport:
         return AsyncHTTPTransport(
             ssl_context=ssl_context,
+            verify=verify,
+            cert=cert,
+            http1=http1,
```
tomchristie:
```suggestion
            http1=http1,
```

httpx/_config.py
T-256: > (Tho perhaps we address clean-ups separately to the main API behavioral changes here?)

Agree with tom, you could add `ssl_context=` and `httpx.SSLContext` in current PR and decide to deprecate/remove in separated PR.

tomchristie:
Looking good.

> Update documentation

I'd suggest you don't spin time on that until we've had a clear discussion around the "API finessing" and "Documentation refresh" areas mentioned in https://github.com/encode/httpx/issues/947#issuecomment-1727389402. I like this change, but we'll want to make it as part of a coherent & consistent 1.0 refresh.

httpx/_config.py
```diff
@@ -138,6 +134,14 @@ def _load_client_certs(self, cert: typing.Optional[CertTypes] = None) -> None:
                     password=cert[2],  # type: ignore
                 )
 
+    def __new__(
+        cls,
+        protocol: ssl._SSLMethod = ssl.PROTOCOL_TLS_CLIENT,
+        *args: typing.Any,
+        **kwargs: typing.Any,
+    ) -> "SSLContext":
+        return super().__new__(cls, protocol, *args, **kwargs)
```
karpetrosyan, line 143: We need this to avoid "DeprecationWarning: ssl.SSLContext() without protocol argument is deprecated."

tomchristie:
Okay, shall we take the docs from this comment... https://github.com/encode/httpx/discussions/3007#discussioncomment-7872119 and include them in the PR?

karpetrosyan:
> Okay, shall we take the docs from this comment... [#3007 (comment)](https://github.com/encode/httpx/discussions/3007#discussioncomment-7872119) and include them in the PR?

I like it

karpetrosyan:
Have we decided to remove or deprecate the old arguments?

T-256:
> Have we decided to remove or deprecate the old arguments?

deprecate warning on 0.27
raise error on 1.0
remove at some point later on 1.x

This is what I can be happy with.

karpetrosyan:
Is there a reason we should continue to support this API?

As far as I can tell, the primary reason for deprecating is that we are adding new functionality and fixing bugs in applications that use the old API.
Do we want to deprecate them only for 1-2 releases before the 1.0 version?

T-256:
My contextual reason got from https://astral.sh/blog/ruff-v0.2.0


> Do we want to deprecate them only for 1-2 releases before the 1.0 version?

Actually, those releases are for users using version pinning (`httpx < 1.0.0`) before major migrating. they might be decided that they aren't ready to stop using old api. so, they can cope with warnings and use latest versions of which deprecated apis allowed to use.

karpetrosyan:
> they might be decided that they aren't ready to stop using old api. so, they can cope with warnings and use latest versions of which deprecated apis allowed to use.

Actually, we are very close to HTTPX 1.0, so I believe there will be just 1-2 releases before then, in which we would primarily introduce the new API, so there will be no features that we would want to ship to our users who are still using the old API.

Also, this migration is rather easy, and I don't believe we need to complicate this PR. Client configurations are frequently found in 1-2 places across the source code, therefore let's make this PR concise and avoid dealing with both new and old APIs.

T-256:
Then, we can merge this PR for v1.0 but not 0.27. (breaking api)


> raise error on 1.0
> remove at some point later on 1.x

The error could describe the reason for the drop. This could be removed sometimes later.
I think it's better than confusing `TypeError: 'app' is an invalid keyword argument for httpx.Client()` message.

tomchristie:
> Then, we can merge this PR for v1.0 but not 0.27.

Yep, that's what we'll do.

> The error could describe the reason for the drop

We might choose not to handle the error cases for this: it requires leaving dead code across a *lot* of function calls.
I'm okay with us just making this change outright with a 1.0 release, so long as it's clearly documented in the release.

tomchristie:
Related... https://github.com/encode/httpx/discussions/3126

tomchristie:
Looks wonderful... some conflicts that need resolving, other than that I think we ought to get this merged.

karpetrosyan:
Do we have any blockers here?

tomchristie:
Just a changelog entry I think? And then we figure out what we want to do version-wise.

karpetrosyan:
Seems like we should merge this directly into 1.0 because we do have breaking changes here.
Ready to spend some time this weekend getting PRs for [1.0](https://github.com/encode/httpx/issues/947#issuecomment-1881173378) ready.

httpx/_config.py
```diff
@@ -46,151 +44,113 @@ class UnsetType:
...
+            cert,
         )
 
-        if self.verify:
-            return self.load_ssl_context_verify()
-        return self.load_ssl_context_no_verify()
+        if verify:
+            self.load_ssl_context_verify(cert, verify)
+        else:
+            self.load_ssl_context_no_verify(cert)
```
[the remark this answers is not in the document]
tomchristie: Nicely observed. otoh I do like that @karpetrosyan is keeping the churn low.

httpx/_config.py
```diff
@@ -46,151 +44,113 @@ class UnsetType:
...
-        cert=cert, verify=verify, trust_env=trust_env, http2=http2
-    ).ssl_context
-
-
-class SSLConfig:
-    """
-    SSL Configuration.
-    """
-
+class SSLContext(ssl.SSLContext):
```
graingert, line 47: If possible I'd like this not to subclass ssl.SSLContext, so we can use pyopenssl with http3

graingert (changes requested):

httpx/_config.py
tomchristie: Ah, so...

The *type* required by the client methods should be `ssl.SSLContext`.
For example in correctly typecheck against this documented usage...

https://github.com/encode/httpx/blob/998f445c83efb140b7e9ae05c993e8f36663cba2/docs/advanced/ssl.md?plain=1#L148-L149

I think it's okay for us to have an `httpx.SSLContext()` subclass.

tomchristie: @graingert Interesting...

Can we progress that separately by opening a design discussion in `httpcore` for adding `pyopenssl` support?
I'd like to understand that individual part better without risking this PR getting sidetracked.

tomchristie:
Switching to an explicitly configured SSL context is a fantastic change, and we should prioritise getting this in.

I'd like to talk through a *further* design discussion here. Since we're making this change, we've also got an opportunity to  drop the `requests` approach to the `verify` and `cert` arguments...

* Drop the overloaded parameter style on `verify` accepting either a `bool` or `str`, in favour of a `bool` plus configuration on the SSL context.
* Drop the overloaded parameter style on `cert` accepting either a `str` or `(str, str)` or `(str, str, str)`, in favour of configuration on the SSL context.

# ChangeLog

* The `verify` and `cert` arguments were in place to mirror `requests` API, these are now removed in favour of passing an explicit `ssl.SSLContext` instance.
* The `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables are no longer automatically used. If required this behaviour can be explicitly enabled.

## Using the default context

The default SSL context is a subclass of Python's `ssl.SSLContext` and supports all the standard API available there for configuring the SSL context.

```
ssl_context = httpx.SSLContext()
client = httpx.Client(ssl_context=ssl_context)
```

## Disabling verification

SSL verification can be bypassed, in order to deal with loading websites which have invalid SSL credentials.

```
ssl_context = httpx.SSLContext(verify=False)
```

## Loading custom verify locations

By default, `httpxSSLContext` will use `certifi` for certificate verification. If you want to use a custom location you can disable this behaviour and instead explicitly [configure the SSL context](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations)...

```
ssl_context = httpx.SSLContext(load_certifi=False)
ssl_context.load_verify_locations(cafile=...)
```

```
ssl_context = httpx.SSLContext(load_certifi=False)
ssl_context.load_verify_locations(capath=...)
```

## Enabling `SSL_CERT_FILE` and `SSL_CERT_DIR` support

Unlike `requests`, the `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables are not automatically loaded. This behaviour can be enabled explicitly...

```
# Use `SSL_CERT_FILE` or `SSL_CERT_DIR` if available,
# otherwise default to using certifi.
cafile = os.environ.get("SSL_CERT_FILE", certifi.where())
capath = os.environ.get("SSL_CERT_DIR")

# Prevent certifi from loading automatically,
# then load the configured certificates.
ssl_context = httpx.SSLContext(load_certifi=False)
ssl_context.load_verify_locations(cafile=cafile, capath=capath)
```

## Passing a custom certificate chain

Custom certificate chains can be loaded by [configuring the SSL context](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain)...

```
ssl_context = httpx.SSLContext()
ssl_context.load_cert_chain(certfile=...)
```

```
ssl_context = httpx.SSLContext()
ssl_context.load_cert_chain(certfile=..., keyfile=...)
```

```
ssl_context = httpx.SSLContext()
ssl_context.load_cert_chain(certfile=..., keyfile=..., password=...)
```

## Using alternate SSL contexts

Any SSL context can be used, for example using the default SSL context from the Python standard library without any additional automatic configuration...

```
import ssl

context = ssl.create_default_context()
```

Using [the `truststore` package](https://truststore.readthedocs.io/en/latest/) to use native system certificate stores, instead of relying on the `certifi` certificate bundle.

```
import truststore
import ssl

context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
```

---

We'll also want to tie in...

* Instantiating a singleton default SSL context when `ssl_context` is left unset with `None`.
* Configuring HTTP/2 on the SSL context, instead of on the client, tho I'm going to defer that conversation to be discussed separately.

tomchristie:
@karpetrosyan Would you like to issue this against the `version-1.0` branch? Then we can get it merged as-is, and consider any further design discussion as a follow-up PR?

karpetrosyan:
> @karpetrosyan Would you like to issue this against the `version-1.0` branch? Then we can get it merged as-is, and consider any further design discussion as a follow-up PR?

Yes, of course. I'll do some tidying up and then change the target branch.

tomchristie:
Wonderful thanks @karpetrosyan. 💯

Version 1.0 pull request now in the works... #3319

httpx/_api.py
```diff
@@ -49,8 +47,7 @@ def request(
     proxy: ProxyTypes | None = None,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
     follow_redirects: bool = False,
-    verify: VerifyTypes = True,
-    cert: CertTypes | None = None,
+    ssl_context: typing.Optional[SSLContext] = None,
```
tomchristie, line 50: We should be using `typing.Optional[ssl.SSLContext]` here. (Yes it's a bit subtle)

tomchristie:
There's some follow-ups that I'd like to make, tho I think we should deal with those incrementally, and start with just getting this PR in.

I'd suggest we constrain *this* PR to be exactly...

* `verify` and `cert` are removed and replaced with `ssl_context = httpx.SSLContext(verify, certs)`.

...and then we can discuss from there. I think the review points required to get us to that constrained place are...

* Revert the `SSLKEYLOGFILE` behavior change. (Let's just scrub it from the docs, and ensure that we *are* setting in the context. We can then consider if we do/don't want that change in isolation.)
* Fix up the typing to be `ssl.SSLContext`.

Seem like a reasonable approach?

karpetrosyan:
I think we got rid of verify and cert and are now using the new ssl_context everywhere. As for changing the typing, I thought it should be Optional everywhere. Am I wrong?

I also...

- Reverted the `SSLKEYLOGFILE` change, restored the trust_env variable for the SSL configuration, and added a test.
- Tidied up environment variable documentation, removing the old SSL-related environment variable entries.
- Moved `SSLKEYLOGFILE` documentation from the environment variables section to ssl.md. (I think we should eliminate documenting environment variables in a separate file and aim to keep them closer to the main documentation.)

tests/test_config.py
```diff
@@ -9,48 +9,46 @@
...
-def test_load_ssl_config_verify_existing_file():
-    context = httpx.create_ssl_context(verify=certifi.where())
-    assert context.verify_mode == ssl.VerifyMode.CERT_REQUIRED
-    assert context.check_hostname is True
-
+@pytest.mark.parametrize(
+    "trust_env, expected_keylog_filename",
+    [
+        pytest.param(True, "test", id="With trasting env"),
+        pytest.param(False, None, id="Without trasting env"),
```
tomchristie, line 26:
```suggestion
        pytest.param(True, "test", id="With trusting env"),
        pytest.param(False, None, id="Without trusting env"),
```

😎

tomchristie:
> As for changing the typing, I thought it should be Optional everywhere. Am I wrong?

Yes, it should be `Optional[ssl.SSLContext]` everywhere.

(Not `Optional[httpx.SSLContext]` everywhere... any SSLContext is valid.)

karpetrosyan:
FIxed 👍 Do we need any other changes?

docs/environment_variables.md
```diff
@@ -8,66 +8,6 @@ Environment variables are used by default. To ignore environment variables, `tru
 
 Here is a list of environment variables that HTTPX recognizes and what function they serve:
 
-## `SSLKEYLOGFILE`
-
-Valid values: a filename
```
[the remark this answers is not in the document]
karpetrosyan: It was moved to the ssl documentation

docs/compatibility.md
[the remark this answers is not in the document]
karpetrosyan: fixed, thanks

httpx/_api.py
```diff
@@ -49,8 +48,7 @@ def request(
     proxy: ProxyTypes | None = None,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
     follow_redirects: bool = False,
-    verify: VerifyTypes = True,
-    cert: CertTypes | None = None,
+    ssl_context: typing.Optional[ssl.SSLContext] = None,
```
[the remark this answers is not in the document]
karpetrosyan: Let's add things incrementally, maybe it should be resolved in another PR?

docs/compatibility.md
```diff
@@ -171,12 +171,10 @@ Also note that `requests.Session.request(...)` allows a `proxies=...` parameter,
 
 ## SSL configuration
 
-When using a `Client` instance, the `trust_env`, `verify`, and `cert` arguments should always be passed on client instantiation, rather than passed to the request method.
+When using a `Client` instance, the ssl configurations should always be passed on client instantiation, rather than passed to the request method.
 
 If you need more than one different SSL configuration, you should use different client instances for each SSL configuration.
 
-Requests supports `REQUESTS_CA_BUNDLE` which points to either a file or a directory. HTTPX supports the `SSL_CERT_FILE` (for a file) and `SSL_CERT_DIR` (for a directory) OpenSSL variables instead.
-
```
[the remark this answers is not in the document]
karpetrosyan: We removed those support, do we need to mention them?

httpx/_api.py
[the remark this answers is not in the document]
karpetrosyan: Yes, please. I don't even remember a lot of things from this PR 😄
One year old PR is a bad practice... always!

docs/environment_variables.md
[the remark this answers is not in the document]
karpetrosyan: I think it would be better to get rid of environment_variables.md. It's better to show env variables in each section, and only related ones.

tomchristie:
> FIxed 👍 Do we need any other changes?

Yes, tho we can probably deal with them as follow-ups...

* We'll want to instantiate a global default, although we should treat this as a follow-up PR. (Bit fiddly)
* I'd like us to simplify the parameters, we can be neater now. We don't really need `cert` since we can use the existing `.load_cert_chain()` interface. We probably don't want to support the `verify=<str>` form either, since we can use the existing `.load_verify_locations()` interface.
* We are going to need `http2` configuration *be on the SSL instance*. I'll explain why, tho we can deal with this as a follow-up. This will need a bit of attention to detail.<sup>*</sup>

<sup>* Currently `httpcore` will [modify the SSL context if HTTP/2 support is enabled](https://github.com/encode/httpcore/blob/4ee1ca25b6b242caec2d96782ed25e81ad7b4ecb/httpcore/_async/connection.py#L145). We don't want to mutate the SSL context, eg. because it might be a single global instance.</sup>

httpx/_api.py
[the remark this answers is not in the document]
tomchristie: We should use `None` here.
We should use with instantiating a singleton default when `None` is passed to the transport.
Tho we can deal with that as a follow-up.

docs/advanced/ssl.md
```diff
@@ -1,100 +1,225 @@
...
+Example:
 
 ```python
-cert = ("path/to/client.pem", "path/to/client.key", "password")
-client = httpx.Client(cert=cert)
-response = client.get("https://example.org")
+# test_script.py
+import httpx
+
+with httpx.AsyncClient() as client:
```
tomchristie, line 188:
```suggestion
with httpx.Client() as client:
```

docs/environment_variables.md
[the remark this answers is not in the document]
tomchristie: `I think it would be better to get rid of environment_variables.md` - I think that'll work out best yep. Let's work on that as a follow-up.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+* Introduced a new SSL configuration API.
```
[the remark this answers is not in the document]
tomchristie: Yeah we need to fill this out. I'm okay with us dealing with that separately in this particular case.

httpx/_client.py
```diff
@@ -695,20 +683,18 @@ def __init__(
 
     def _init_transport(
         self,
-        verify: VerifyTypes = True,
-        cert: CertTypes | None = None,
+        ssl_context: typing.Optional[ssl.SSLContext],
```
tomchristie, line 686:
```suggestion
        ssl_context: ssl.SSLContext | None = None,
```

httpx/_client.py
```diff
@@ -718,16 +704,14 @@ def _init_transport(
     def _init_proxy_transport(
         self,
         proxy: Proxy,
-        verify: VerifyTypes = True,
-        cert: CertTypes | None = None,
+        ssl_context: typing.Optional[ssl.SSLContext] = None,
```
tomchristie, line 707:
```suggestion
        ssl_context: ssl.SSLContext | None = None,
```

httpx/_client.py
```diff
@@ -1415,43 +1388,36 @@ def __init__(
...
         self,
-        verify: VerifyTypes = True,
-        cert: CertTypes | None = None,
+        ssl_context: ssl.SSLContext | None = None,
         http1: bool = True,
         http2: bool = False,
         limits: Limits = DEFAULT_LIMITS,
         transport: AsyncBaseTransport | None = None,
-        trust_env: bool = True,
+        app: typing.Callable[..., typing.Any] | None = None,
```
tomchristie, line 1396:
```suggestion
```

tests/test_config.py
```diff
@@ -1,5 +1,5 @@
-import os
 import ssl
+import typing as t
```
tomchristie, line 2:
```suggestion
import typing
```

tests/test_config.py
```diff
@@ -9,48 +9,46 @@
...
+@pytest.mark.parametrize(
+    "trust_env, expected_keylog_filename",
+    [
+        pytest.param(True, "test", id="With trusting env"),
+        pytest.param(False, None, id="Without trusting env"),
+    ],
+)
+def test_load_ssl_with_keylog(
+    monkeypatch: t.Any, trust_env: bool, expected_keylog_filename: t.Union[str, None]
+) -> None:
```
tomchristie, line 31:
```suggestion
def test_load_ssl_with_keylog(
    monkeypatch: typing.Any,
    trust_env: bool,
    expected_keylog_filename: typing.Union[str, None]
) -> None:
```

httpx/_api.py
```diff
@@ -141,8 +132,7 @@ def stream(
...
```
tomchristie, line 135:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_api.py
```diff
@@ -187,8 +176,7 @@ def get(
     auth: AuthTypes | None = None,
     proxy: ProxyTypes | None = None,
     follow_redirects: bool = False,
-    cert: CertTypes | None = None,
-    verify: VerifyTypes = True,
+    ssl_context: typing.Optional[ssl.SSLContext] = None,
```
tomchristie, line 179:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_api.py
```diff
@@ -225,8 +212,7 @@ def options(
...
```
tomchristie, line 215:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_api.py
```diff
@@ -348,8 +329,7 @@ def put(
...
```
tomchristie, line 332:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_api.py
```diff
@@ -391,8 +370,7 @@ def patch(
...
```
tomchristie, line 373:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_api.py
```diff
@@ -430,9 +407,8 @@ def delete(
     auth: AuthTypes | None = None,
     proxy: ProxyTypes | None = None,
     follow_redirects: bool = False,
-    cert: CertTypes | None = None,
-    verify: VerifyTypes = True,
     timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
+    ssl_context: typing.Optional[ssl.SSLContext] = None,
```
tomchristie, line 411:
```suggestion
    ssl_context: ssl.SSLContext | None = None,
```

httpx/_client.py
```diff
@@ -695,20 +683,18 @@ def __init__(
...
         self,
-        verify: VerifyTypes = True,
-        cert: CertTypes | None = None,
+        ssl_context: typing.Optional[ssl.SSLContext],
         http1: bool = True,
         http2: bool = False,
         limits: Limits = DEFAULT_LIMITS,
-        transport: BaseTransport | None = None,
         trust_env: bool = True,
+        transport: typing.Optional[BaseTransport] = None,
```
tomchristie, line 691:
```suggestion
        transport: BaseTransport | None = None,
        trust_env: bool = True,
```

httpx/_config.py
```diff
@@ -46,151 +45,103 @@ class UnsetType:
...
     DEFAULT_CA_BUNDLE_PATH = Path(certifi.where())
 
     def __init__(
         self,
-        *,
-        cert: CertTypes | None = None,
         verify: VerifyTypes = True,
+        cert: CertTypes | None = None,
         trust_env: bool = True,
```
tomchristie, line 55:
```suggestion
```

httpx/_config.py
```diff
@@ -46,151 +45,103 @@ class UnsetType:
...
-        http2: bool = False,
     ) -> None:
-        self.cert = cert
         self.verify = verify
+        set_minimum_tls_version_1_2(self)
+        self.options |= ssl.OP_NO_COMPRESSION
+        self.set_ciphers(DEFAULT_CIPHERS)
         self.trust_env = trust_env
```
tomchristie, line 61:
```suggestion
```

httpx/_config.py
```diff
@@ -46,151 +45,103 @@ class UnsetType:
...
-        self.http2 = http2
-        self.ssl_context = self.load_ssl_context()
 
-    def load_ssl_context(self) -> ssl.SSLContext:
+        keylogfile = os.environ.get("SSLKEYLOGFILE")
+        if keylogfile and self.trust_env:
```
tomchristie, line 64:
```suggestion
        if keylogfile:
```

tests/test_config.py
```diff
@@ -9,48 +9,48 @@
...
+    ],
+)
+def test_load_ssl_with_keylog(
+    monkeypatch: typing.Any,
+    trust_env: bool,
+    expected_keylog_filename: typing.Union[str, None],
+) -> None:
+    monkeypatch.setenv("SSLKEYLOGFILE", "test")
+    context = httpx.SSLContext(trust_env=trust_env)
+    assert context.keylog_filename == expected_keylog_filename
```
tomchristie, line 36:
```suggestion
def test_load_ssl_with_keylog(
    monkeypatch: typing.Any,
    expected_keylog_filename: typing.Union[str, None],
) -> None:
    monkeypatch.setenv("SSLKEYLOGFILE", "test")
    context = httpx.SSLContext()
    assert context.keylog_filename == expected_keylog_filename
```

httpx/_client.py
```diff
@@ -664,25 +656,21 @@ def __init__(
...
 
         self._transport = self._init_transport(
-            verify=verify,
-            cert=cert,
+            ssl_context=ssl_context,
             http1=http1,
             http2=http2,
             limits=limits,
             transport=transport,
-            trust_env=trust_env,
```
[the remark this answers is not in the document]
tomchristie: This is in the right direction... dropping `trust_env` out of the `httpx.SSLContext()` API.

tests/test_config.py
```diff
@@ -9,48 +9,40 @@
...
-    context = httpx.create_ssl_context(verify=certifi.where())
-    assert context.verify_mode == ssl.VerifyMode.CERT_REQUIRED
-    assert context.check_hostname is True
+def test_load_ssl_with_keylog(
+    monkeypatch: typing.Any,
+    expected_keylog_filename: typing.Union[str, None],
+) -> None:
+    monkeypatch.setenv("SSLKEYLOGFILE", "test")
+    context = httpx.SSLContext()
+    assert context.keylog_filename == expected_keylog_filename
```
tomchristie, line 28:
```suggestion
def test_load_ssl_with_keylog(monkeypatch: typing.Any) -> None:
    monkeypatch.setenv("SSLKEYLOGFILE", "test")
    context = httpx.SSLContext()
    assert context.keylog_filename == "test"
```

tomchristie:
Thanks so much @karpetrosyan - going to merge this into "version-1.0" now, and then apply remaining changes incrementally.

karpetrosyan:
Sorry, I couldn't find time to finish all the tidy-ups.
Thank you for the help! 🙏

tomchristie:
Thank you so much for your time & dedication @karpetrosyan. 🙌

PR #3025, opened 2023-12-28
Sketching out 1.0 docs refresh.

tomchristie:
**...**

*Hmmm... I've approached this in a non-review friendly way, sketching out the re-write as I go as a thinking process. Might need to take a different tack, to be discussed. Still. Throwing it up for some public visibility.*


T-256:
Can you publish preview docs?

Also, I think we can automate docs preview publish on PRs.
We could create new workflow for docs and exclude it from _test suite_ workflow (no duplicated docs builds). the new workflow also would automatically preview-publish to gh pages if there were any modifications to docs.

docs/community/code_of_conduct.md
T-256, line 1: I'd recommend move to repo's main directory.

docs/community/contributing.md
T-256, line 1: same

tomchristie:
> Can you publish preview docs?

I'm not sure, how would I do that?

PR #3042, opened 2024-01-05
Keep clients in sync

karpetrosyan:
Except for the async/await syntax, `AsyncClient` and `Client` are supposed to be the same.

I isolated them into separate files and used `git diff --no-index --word-diff asyncclient.py syncclient.py` to find and fix differences.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -598,6 +598,8 @@ class Client(BaseClient):
     to authenticate the client. Either a path to an SSL certificate file, or
     two-tuple of (certificate file, key file), or a three-tuple of (certificate
     file, key file, password).
+    * **http2** - *(optional)* A boolean indicating if HTTP/2 support should be
+    enabled. Defaults to `False`.
     * **proxy** - *(optional)* A proxy URL where all the traffic should be routed.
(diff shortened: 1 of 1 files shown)


httpx/_client.py
```diff
@@ -1311,6 +1313,8 @@ class AsyncClient(BaseClient):
     An asynchronous HTTP client, with connection pooling, HTTP/2, redirects,
     cookie persistence, etc.
 
+    It can be shared between threads.
```
karpetrosyan, line 1316:
```suggestion
    It can be shared between tasks.
```

httpx/_client.py
```diff
@@ -1544,6 +1548,15 @@ async def request(
 
         [0]: /advanced/#merging-of-configuration
         """
+
+        if cookies is not None:
```
karpetrosyan, line 1552:
```suggestion
        if cookies is not None:  # pragma: no cover
```

tomchristie:
neat!

tomchristie (approved):

tomchristie:
(ready to merge. reminder to team, we're okay without a `CHANGELOG.md` entry on refactorings that don't represent a change in user-facing behavior.)

PR #3045, opened 2024-01-06
Fix RFC 2069 mode digest authentication

the-ress:
# Summary

Fixes a typo that breaks old RFC 2069 mode digest auth.

The issue is that HA1 is added both on line 283, and again on line 293. ([According to RFC, it should be there only once.](https://datatracker.ietf.org/doc/html/rfc2069#section-2.1.2))

Also mentioned in https://github.com/encode/httpx/issues/2788#issuecomment-1677260605

# Checklist

### DIFF
--- a/httpx/_auth.py
+++ b/httpx/_auth.py
@@ -280,17 +280,18 @@ class DigestAuth(Auth):
 
         qop = self._resolve_qop(challenge.qop, request=request)
         if qop is None:
+            # Following RFC 2069
             digest_data = [HA1, challenge.nonce, HA2]
         else:
-            digest_data = [challenge.nonce, nc_value, cnonce, qop, HA2]
-        key_digest = b":".join(digest_data)
+            # Following RFC 2617/7616
+            digest_data = [HA1, challenge.nonce, nc_value, cnonce, qop, HA2]
 
         format_args = {
             "username": self._username,
             "realm": challenge.realm,
             "nonce": challenge.nonce,
             "uri": path,
-            "response": digest(b":".join((HA1, key_digest))),
+            "response": digest(b":".join(digest_data)),
             "algorithm": challenge.algorithm.encode(),
         }
         if challenge.opaque:
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -140,3 +140,168 @@ def test_digest_auth_setting_cookie_in_request():
     )
     with pytest.raises(StopIteration):
         flow.send(response)
+
+
+def test_digest_auth_rfc_2069():
+    # Example from https://datatracker.ietf.org/doc/html/rfc2069#section-2.4
+    # with corrected response from https://www.rfc-editor.org/errata/eid749
+
+    auth = httpx.DigestAuth(username="Mufasa", password="CircleOfLife")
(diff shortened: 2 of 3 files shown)


httpx/_auth.py
```diff
@@ -280,7 +280,7 @@ def digest(data: bytes) -> bytes:
 
         qop = self._resolve_qop(challenge.qop, request=request)
         if qop is None:
-            digest_data = [HA1, challenge.nonce, HA2]
+            digest_data = [challenge.nonce, HA2]
         else:
             digest_data = [challenge.nonce, nc_value, cnonce, qop, HA2]
```
tomchristie, line 285: Looking at the curl code, seems like this change should be the other way around(?)...

```python
        if qop is None:
            digest_data = [HA1, challenge.nonce, HA2]
        else:
            digest_data = [HA1, challenge.nonce, nc_value, cnonce, qop, HA2]
```

And then...

```python
        format_args = {
            "username": self._username,
            "realm": challenge.realm,
            "nonce": challenge.nonce,
            "uri": path,
            "response": digest(key_digest),
            "algorithm": challenge.algorithm.encode(),
        }
```

tomchristie:
I appreciated your "this branch is RFC ..." comment in https://github.com/encode/httpx/pull/2600#issuecomment-1879883701 - perhaps we should include those?

Also the link [to the `curl` code](https://github.com/curl/curl/blob/8edcfedc1a144f438bd1cdf814a0016cbe678aaf/lib/vauth/digest.c#L793-L799) is super helpful, thanks.

httpx/_auth.py
the-ress: Yeah, initially I wanted to avoid touching the other branch (so I don't break it accidentally), but I agree the curl way is more readable. I also add the comments.

the-ress:
I made the discussed changes and added a few tests.

tests/test_auth.py
```diff
@@ -140,3 +140,168 @@ def test_digest_auth_setting_cookie_in_request():
...
+    )
+    assert "qop=auth" in request.headers["Authorization"]
+    assert (
+        'opaque="FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS"'
+        in request.headers["Authorization"]
+    )
+    assert (
+        'response="8ca523f5e9506fed4657c9700eebdbec"'
+        in request.headers["Authorization"]
+    )
```
tomchristie, line 242: These tests are fantastic, thanks!

Are we able to do a literal `assert request.headers["Authorization"] == ...` test here or is the ordering that we're returning different to the example in the RFC?

tomchristie:
Looking wonderful, thanks. ✨😊✨

You'll also want to update the `CHANGELOG.md` with an entry pointing at this PR.

tests/test_auth.py
the-ress: I wouldn't want to depend on the exact ordering since it'd make the test more fragile with little benefit.

the-ress:
> You'll also want to update the `CHANGELOG.md` with an entry pointing at this PR.

Done.

T-256 (approved):

tomchristie (approved):

PR #3050, opened 2024-01-11
Deprecate `app=...` in favor of explicit `WSGITransport`/`ASGITransport`.

tomchristie:
*Deprecate the `httpx.Client(app=...)` and `httpx.AsyncClient(app=...)` shortcuts.*

---

Instead of...

```python
client = httpx.Client(app=app)
```

Use this...

```python
transport = httpx.WSGITransport(app=app)
```

---

Instead of...

```python
client = httpx.AsyncClient(app=app)
```

Use this...

```python
transport = httpx.ASGITransport(app=app)
client = httpx.AsyncClient(transport=transport)
```

---

The rationale for this is described in https://github.com/encode/httpx/issues/947#issuecomment-1727389402.

**TODO**:

- [x] Docs.
- [x] `CHANGELOG.md`.
- [x] ~Ensure we have a clear section where we outline 1.0 version changes in one place.~ See https://github.com/encode/httpx/pull/3069


httpx/_client.py
```diff
@@ -682,7 +679,13 @@ def __init__(
             if proxy:
                 raise RuntimeError("Use either `proxy` or 'proxies', not both.")
 
-        allow_env_proxies = trust_env and app is None and transport is None
+        if app:
+            raise RuntimeError(
```
T-256, line 683: I don't know what's deprecation policy but are you sure raise error here? isn't it breaking change for those using `app`?

I'd recommend keep docs, type-hints and use warning instead. then we can schedule the removal version and mention it in warning message (?)

T-256: > I don't know what's deprecation policy but are you sure raise error here? isn't it breaking change for those using `app`?

It looks good to me if this breaking change is for 1.0. But do we want to keep these additional errors in 1.0?

tomchristie: > It looks good to me if this breaking change is for 1.0. But do we want to keep these additional errors in 1.0?

So...  we'll need to have consistency around this.

Options might be...

* Include old argument styles for 1.0 with explicit errors, to aid migrations. Clean up later and drop the errors in a 1.1 release.
* Don't include old argument styles.

T-256: At there, I'm more fan of what done for `proxies`:
https://github.com/encode/httpx/blob/419d3a9d80d0c4072f6cb58eeb306148ae89e2e9/httpx/_client.py#L676-L681

So, what I can suggest at here is:
1. deprecate and mention scheduled removal if any (e.g. "The 'proxies' argument is now deprecated and it will remove in v1.0.0")
2. minor/patch release
3. clear all deprecations in major release

T-256: I think now is good time to discuss about deprecation policy, since there are few `1.0 proposal` PRs those trying to directly remove APIs without any deprecation.

tomchristie: https://github.com/encode/httpx/pull/3069 would be a good place for us to discuss how we'd like to approach this.

T-256: Via #3071 and since v1.0 is major upgrade, I'm now more interested to directly drop instead of deprecation.

Kludex:
This is not deprecation of `app=...`, it's a removal directly. Is it ok? We usually raise warnings when deprecating paramters, don't we?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Removed
+
+* The `app=...` shortcut has been removed. Use the explicit style of `transport=httpx.WSGITransport()` or `transport=httpx.ASGITransport()` instead.
```
T-256, line 11: Since this PR is version 1.0 proposal, why don't you track its changelog in #3069 ?

tomchristie: https://github.com/encode/httpx/pull/3069/files#diff-75bb3ae57865eb28e99c3e4fff3efc72a5d49e3f14ac324195b405058cb9731aR13

tomchristie:
> We usually raise warnings when deprecating paramters, don't we?

Yep, so... https://github.com/encode/httpx/pull/3069

There's a proposal for several related changes, that end up with a stricter simpler more consistent API.
We can land these in a 1.0, and need to take a call on what we think is best wrt. hard-change vs. soft deprecation path.

CHANGELOG.md
T-256: Then do we need to re-mention it in this PR?

tomchristie: They should do before approval.

karpetrosyan:
Is there any reason why we should deprecate things?
Deprecation is good when we want to allow users to use the latest features and updates for software without changing anything. HTTPX 1.0 will change a lot of things, and I think we shouldn't strive to support all the HTTPX/1.x updates for 0.x syntax.

Instead, we can have a well-defined migration guide that will help our users migrate easily.
Also, all of the changes are mainly in the client, which is very easy to migrate because they are not so much in the codebase.

tomchristie:
> This is not deprecation of app=..., it's a removal directly. Is it ok? We usually raise warnings when deprecating parameters, don't we?

Based, on this I've updated the PR to be just a deprecation at this point.

Once this is merged I'll look at rolling a 0.27 version with the updated docs.

ofer-pd:
There's one more section of docs that needs to be updated - the API reference for [Client](https://www.python-httpx.org/api/#client) and AsyncClient don't yet mention that the `app` param is deprecated.

PR #3051, opened 2024-01-11
Add `httpx.Version` configuration.

tomchristie:
Discussion: https://github.com/encode/httpx/discussions/3026#discussioncomment-7980059

I don't think we're clear on this yet, but drafting up as a proposal.

**TODO**:

- [ ] Docs.
- [ ] `CHANGELOG.md`.
- [ ] Ensure we have a clear section where we outline 1.0 version changes in one place.


karpetrosyan:
I think we are going too far with encapsulating arguments :d

tomchristie:
Valid perspective yes.

PR #3052, opened 2024-01-12
Add `httpx.NetworkOptions` configuration.

lovelydinosaur:
Refs https://github.com/encode/httpx/issues/947#issuecomment-1881173378

Here's how the proposed documentation looks...

---

# Network Options

There are several advanced network options that are made available through the `httpx.NetworkOptions` configuration class.

```python
# Configure an HTTPTransport with some specific network options.
network_options = httpx.NetworkOptions(
    connection_retries=1,
    local_address="0.0.0.0",
)
transport = httpx.HTTPTransport(network_options=network_options)

# Instantiate a client with the configured transport.
```

## Configuration

The options available on this class are...

### `connection_retries`

Configure a number of retries that may be attempted when initially establishing a TCP connection. Defaults to `0`.

### `local_address`

Configure the local address that the socket should be bound too. The most common usage is for enforcing binding to either IPv4 `local_address="0.0.0.0"` or IPv6 `local_address="::"`.

### `socket_options`

Configure the list of socket options to be applied to the underlying sockets used for network connections.
For example, you can use it to explicitly specify which network interface should be used for the connection in this manner:

```python
import httpx

socket_options = [(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, b"ETH999")]

network_options = httpx.NetworkOptions(
    socket_options=socket_options
)
```

### `uds`

Connect to a Unix Domain Socket, rather than over the network. Should be a string providing the path to the UDS.

---

Design options that we have here...

* We *could* allow a `httpx.Client(network_options=...)` shortcut, in the same way we do for eg. `ssl_context=...`, `proxy=...`, `limits=...`, `version=...`.
* We *could* have `Limits` and `NetworkOptions` all be part of the same single config class. Including `max_keepalive_connections`, `max_connections`, `keepalive_expiry` on this class.

### DIFF
--- a/docs/advanced/network-options.md
+++ b/docs/advanced/network-options.md
@@ -26,7 +26,18 @@ Configure the local address that the socket should be bound too. The most common
 
 ### `socket_options`
 
-*TODO: Example*
+Configure the list of socket options to be applied to the underlying sockets used for network connections.
+For example, you can use it to explicitly specify which network interface should be used for the connection in this manner:
+
+```python
+import httpx
+
+socket_options = [(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, b"ETH999")]
+
+network_options = httpx.NetworkOptions(
+    socket_options=socket_options
+)
+```
 
 ### `uds`
 
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -14,7 +14,7 @@ from ._types import CertTypes, HeaderTypes, TimeoutTypes, URLTypes, VerifyTypes
 from ._urls import URL
 from ._utils import get_ca_bundle_from_env
 
-__all__ = ["Limits", "Proxy", "Timeout", "create_ssl_context"]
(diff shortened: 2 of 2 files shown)


jeffsilverm:
I would like to ask that something be added to the documentation at [advanced usage](https://www.python-httpx.org/advanced/#usage_1). I would like you to add something such as: **"This is how you force httpx to use IPv4 even if it can use IPv6 (otherwise, it will use IPv6 if it can and IPv4 if it cannot)"** That way if I search for "IPv4", it will be found. Currently, it is very hard to find until I was told what to look for.  Of course, once I was told what to look for, it was immediately obvious.

karpetrosyan:
[Here is](https://github.com/encode/httpx/discussions/2635#discussioncomment-5986869) the `socket_options` example that we might want to use here.

karpetrosyan:
I have added an example for the `socket_options` parameter. I believe we are now ready to merge this PR.

karpetrosyan (approved):

karpetrosyan:
Maybe we should also add a changelog

PR #3053, opened 2024-01-12
fix(ssl): add lru_cache for create_ssl_context

shepilov-vladislav:
# Summary

# Checklist

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -2,6 +2,7 @@ import logging
 import os
 import ssl
 import typing
+from functools import lru_cache
 from pathlib import Path
 
 import certifi
@@ -42,6 +43,7 @@ class UnsetType:
 UNSET = UnsetType()
 
 
+@lru_cache(None)
 def create_ssl_context(
     cert: typing.Optional[CertTypes] = None,
     verify: VerifyTypes = True,


T-256:
Thanks for the PR, I like the solution. This is high memory usage I encountered when using tons of Client instance in an app (I need them for several reasons).

But perhaps you would like to see and wait for #3022 

@karpetrosyan could new `httpx.SSLContext` cachable?

karpetrosyan:
Are we really capable of doing something like this? What if the SSL context is modified? Why should any subsequent function call return the modified SSL context?

I mean...

```python
import ssl
from functools import cache

@cache
def get_ssl_context():
    return ssl.create_default_context()

first_ssl = get_ssl_context()
first_ssl.check_hostname = False

second_ssl = get_ssl_context()
print(f.check_hostname) # False
```

shepilov-vladislav:
@T-256, @karpetrosyan First of all thank you for your quick response and help.

The thing is that we make a lot of requests using http and we tried to replace `httpx` with `aiohttp` and saw that CPU consumption decreased many times, after that I went to compare the code base and saw that `aiohttp` just caches the creation of SSLContext - [see here](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L903).

After adding caching - CPU consumption dropped by more than 4x on our Pods. If there is another more correct way to do it - please advise. I haven't found a way to do it without changing the `httpx` codebase.

Perhaps this [PR](https://github.com/encode/httpx/pull/3022) will provide that opportunity, but I'm not sure

If you still find my changes useful, just tell me, I'll fix the tests and we'll merge this one out ✌️

karpetrosyan:
We should probably wait until #3022 is marged before we consider adding cache at that layer. I like this idea, because configuring SSL context is a very expensive job.

Ref: https://github.com/karpetrosyan/hishel/issues/139

tomchristie:
> After adding caching - CPU consumption dropped by more than 4x on our Pods.

Prob indicates that you're instantiating clients repeatedly, and losing out on the benefits of connection pooling.

> We should probably wait until https://github.com/encode/httpx/pull/3022 is marged before we consider adding cache at that layer.

Would be simpler to consider then yep. (Anyone want to open a "SSL Cache" discussion?)

PR #3061, opened 2024-01-16
test `parse_header_links` via public api

T-256:
# Summary

chunk of #2851
# Checklist


tomchristie (approved):

tomchristie:
Neat, yep!

PR #3062, opened 2024-01-16
test `same_origin` via public api

T-256:
# Summary
chunk of #2851


tomchristie:
Thanks, same thoughts as #3064.

PR #3064, opened 2024-01-16
test `is_https_redirect` via public api

T-256:
# Summary
chunk of #2851


tests/test_utils.py
```diff
@@ -234,21 +233,39 @@ def test_not_same_origin():
...
-    url = httpx.URL("http://example.com")
-    location = httpx.URL("https://example.com")
-    assert is_https_redirect(url, location)
+    url = httpx.URL("https://example.com")
+    request = httpx.Request(
+        "GET", "http://example.com", headers={"Authorization": "empty"}
+    )
+
+    client = httpx.Client()
+    headers = client._redirect_headers(request, url, "GET")
```
tomchristie, line 242: Hrm. Is there a way around that allows us to test this against public API?

tomchristie: Might be that we're okay with leaning on implementation details a little bit, certainly not a terrible trade-off to make. (I think we already have some instances of this in our test cases right?)

T-256: Also, we already use private-accessors in another test:
https://github.com/encode/httpx/blob/4f6edf36e93fd9f83ff95b065718fd6bd0c4d3c5/tests/client/test_proxies.py#L333-L338

tests/test_utils.py
```diff
@@ -11,7 +11,7 @@
     URLPattern,
     get_ca_bundle_from_env,
     get_environment_proxies,
-    is_https_redirect,
+    parse_header_links,
```
T-256, line 14:
```suggestion
```

PR #3065, opened 2024-01-17
Add persistent cookies option to Client

### LINKED ISSUE #2992: Add option to disable cookie persistence in Client/AsyncClient
Initially raised as discussion #1533 

I've been using `httpx`'s `AsyncClient` as a web spider, however I have noticed that cookies are automatically persisted and there's no easy way to disable this.   My workaround has been to subclass python's `http.cookiejar.CookieJar` like so:

```python
from http.cookiejar import CookieJar

class NullCookieJar(CookieJar):
    """A CookieJar that rejects all cookies."""

    def extract_cookies(self, *_):
        """For extracting and saving cookies.  This implementation does nothing"""
        pass

    def set_cookie(self, _):
        """Normally for setting a cookie.  This implementation does nothing"""
        pass
```

Would it be possible to:
1. add an option to disable cookie persistence to `Client`/`AsyncClient` or
2. include this implementation of `NullCookieJar` in `httpx` as a utility class?

Thanks!

MarkWine:
# Summary

Closes #2992 

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -164,6 +164,7 @@ class BaseClient:
         params: typing.Optional[QueryParamTypes] = None,
         headers: typing.Optional[HeaderTypes] = None,
         cookies: typing.Optional[CookieTypes] = None,
+        persistent_cookies: bool = False,
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
         follow_redirects: bool = False,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
@@ -182,6 +183,7 @@ class BaseClient:
         self._params = QueryParams(params)
         self.headers = Headers(headers)
         self._cookies = Cookies(cookies)
+        self._persistent_cookies = persistent_cookies
         self._timeout = Timeout(timeout)
         self.follow_redirects = follow_redirects
         self.max_redirects = max_redirects
@@ -590,6 +592,8 @@ class Client(BaseClient):
     sending requests.
     * **cookies** - *(optional)* Dictionary of Cookie items to include when
     sending requests.
+    * **persistent_cookies** - *(optional) A boolean indicating if cookies should
+    persist. Defaults to `False`.
     * **verify** - *(optional)* SSL certificates (a.k.a CA bundle) used to
     verify the identity of requested hosts. Either `True` (default CA bundle),
     a path to an SSL certificate file, an `ssl.SSLContext`, or `False`
@@ -626,6 +630,7 @@ class Client(BaseClient):
         params: typing.Optional[QueryParamTypes] = None,
         headers: typing.Optional[HeaderTypes] = None,
         cookies: typing.Optional[CookieTypes] = None,
+        persistent_cookies: bool = False,
         verify: VerifyTypes = True,
         cert: typing.Optional[CertTypes] = None,
         http1: bool = True,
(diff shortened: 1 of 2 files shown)


MarkWine:
Happy to change argument name in api implementation.

For tests, not explicitly testing the persistence as True option, but the existing persistence test fails if default set to False (as expected). Also not directly testing AsyncClient, but that is true for other cookie tests as well.

rafalkrupinski:
Cool, the issue just came to me too.

I think the naming convention here would be `persist_cookies`, similar to `follow_redirects`. I'm asking the client to persist them, not informing it that I'm doing the persistence.
Parameters and fields are more or less logically grouped (data, config). Do you think moving it somewhere near `follow_redirects` would make sense?

Can you add an override parameter to all request methods (request, send, get, post, etc...), with the default value `USE_CLIENT_DEFAULT`?

You've added test for making a request without cookie persistence; is there a test that specifically tests cookie persistence?

There's a deprecation warning about cookie persistence in `Client.request`. IMO with your changes it would no longer apply.

rafalkrupinski:
Just a thought, doesn't persistence suggest that cookies survive the client itself? Like storing to a file?
How about `merge_response_cookies`?

rafalkrupinski:
> Would it be better to explicitly add the parameter value to True, leave as default, or do both in different tests?

Just have a look how `follow_redirects` is tested.

> For the deprecation warning:

You're right, that deprecation warning is related to keeping cookies on redirect, nothing to do with your patch. One more reason not to call this one persist_cookies :)

MarkWine:
https://github.com/encode/httpx/pull/3071 Given that cookie params are being removed from the request methods (following the deprecation warning noted above), should this option also only exist at the client level?

rafalkrupinski:
I sure [hope](https://github.com/encode/httpx/pull/3071#issuecomment-1916538878) [they](https://github.com/encode/httpx/discussions/2931#discussioncomment-8302308) [don't](https://github.com/encode/httpx/pull/3071#issuecomment-1917414375) :)

rafalkrupinski (approved):

rafalkrupinski:
I'm alright with the code, but it probably should be mentioned somewhere in the documentation.
I think right now the only mention is `Cookie persistence across requests.` in extra feature section.

I'm looking at `follow_redirects` documentation... it's covered in the quick start, but merge_response_cookies applies only to client usage, not covered there. So I guess it should go to advanced, a new section on cookies.

WDYT @tomchristie ?

rafalkrupinski:
> #3071 Given that cookie params are being removed from the request methods (following the deprecation warning noted above), should this option also only exist at the client level?

Come to think about it, even if per-request cookies were removed, it still made sense to include this feature.
The flag `merge_response_cookies` tells whether the response object is allowed to modify the client state or not, at least regarding the cookies.

rafalkrupinski:
BTW
#1533
#2271

One mentioned workaround is to pass custom no-op CookieJar. I guess it forbids any Cookies, unless you set them manually in headers.

MarkWine:
I added an internal pull request with all the method overrides. It's a bit on the big/boring side because of all the public methods available. I can merge if it looks good.

https://github.com/MarkWine/httpx/pull/1

Not sure on documentation. It might be an esoteric enough option that for now the client docstrings might be enough? I can take a shot if new section is warranted, or probably defer to someone more familiar with the documentation style for httpx.

jschlyter (approved):
This seems like to most reasonable approach to do able to do this on a per-request basis.

PR #3066, opened 2024-01-17
Mention NO_PROXY environment variable on Advanced Usage page

RichieB2B:
# Summary

The NO_PROXY environment variable is not mentioned at all on the Advanced Usage page, only on the Environment Variables page. This PR adds it so it is clear right away this environment variable is used by httpx.

# Checklist


zanieb (approved):
Seems reasonable to me.

PR #3069, opened 2024-01-22
Draft 1.0 release notes

tomchristie:
I'm drafting up here what our 1.0 release notes *could* look like, so that we've a single point of reference for deciding what does and does not make the cut.

Let's take a look...

### Simplified transport configuration

As described at https://github.com/encode/httpx/issues/947#issuecomment-1881173378

Includes PRs...

* #3052 
* #3051 
* #3050
* #3022
* #2879
* TODO: Mounts

### Stricter parameters

We don't necessarily need these changes, and I've not discussed them elsewhere. However they do align with a stricter more explicit approach. Zen of Python etc. I'd be perfectly happy dropping these, but might be worthwhile. Can discuss the motivation more in due course.

### Minor cleanups

* I have some planned documentation on the URL API that ties in with the "All components...".
* `Response.elapsed` is a tidy one to remove, and can replaced with an example of how to handle it using event hooks. Discussion at #3012

### Environment variables

The changes in "simplified transport configuration" end up removing all auto-environment-variable based configuration by default. There's easy migrations for re-enabling the existing functionalities.

### Dependencies

* This is PR #2858

### DIFF
--- a/docs/1.0-release.md
+++ b/docs/1.0-release.md
@@ -1,0 +1,40 @@
+# HTTPX 1.0
+
+The `httpx` 1.0 release includes a number of API changes intended to present a simpler, more consistent interface throughout.
+
+If you're migrating from previous versions, you'll need to be aware of the following:
+
+## Simplified transport configuration
+
+* The `verify` and `cert` arguments no longer exist. Use `ssl_context = httpx.SSLContext(verify=..., cert=...)` instead. See [the SSL documentation](#) for details.
+* The `http1` and `http2` arguments no longer exist. Use `version = httpx.Version("HTTP/1.1", "HTTP/2")` instead. See [the HTTP version documentation](#) for details.
+* The `proxies` argument no longer exists. Use `proxy = httpx.Proxy(...)` instead. See [the Proxy documentation](#) for details.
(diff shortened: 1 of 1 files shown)


T-256 (approved):
LGTM 🎉
Is **`Removed`** section missed?

_From https://github.com/encode/httpx/issues/3022#issuecomment-1869489189:_

> 
> Also, we have other deprecated drops those pending to remove in v1.0.0 (IIUC):
> - `proxies=`
> - `data=` (non-Mapping objects)
> - `cookies=` (per-request)
>

tomchristie:
@encode/maintainers How do we feel about this?
My inclination is that we should probably just push on with deprecations, and commit ourselves to a 1.0 release.

It's a more coherent, slightly stricter take on the API which all fits together really neatly.
Seems worth it.

T-256:
since each PR should [contain changelog](https://github.com/encode/httpx/pull/3050?notification_referrer_id=NT_kwDOB-BRl7Q5MDQ2NzE0NTE0OjEzMjE0MTQ2Mw#discussion_r1467491476), how about creating new branch and rebase `1.0 proposal` PRs to it? it'd be easier to manage approved PRs.

PR #3070, opened 2024-01-25
Add `httpx.Mounts` transport class.

tomchristie:
Discussed in https://github.com/encode/httpx/issues/947#issuecomment-1881173378

This PR adds the `Mounts` and `AsyncMounts` transport classes.

Currently in draft... switch `mounts=...` out of the `Client` class directly.

---

Here's the synopsis...

We currently have `mounts=...` behaviour backed directly into the client, like so...

```python
client = httpx.Client(mounts=...)
```

However it'd be neater to keep that complexity encapsulated within a transport class that supports routing, so...

```python
transport = httpx.Mounts(...)
```

At this point in time the pull request replicates the `mounts=...` functionality in a transport class, but doesn't yet remove it from the client. A follow-up would deal with removing it.


httpx/_transports/mounts.py
```diff
@@ -0,0 +1,218 @@
...
+            and self.host_regex is not None
+            and not self.host_regex.match(other.host)
+        ):
+            return False
+        if self.port is not None and self.port != other.port:
+            return False
+        return True
+
+    @property
+    def priority(self) -> typing.Tuple[int, int, int]:
```
karpetrosyan, line 103:
```suggestion
    def priority(self) -> tuple[int, int, int]:
```

Now we support this :)

tomchristie: Ahhh... better 💃🏻

httpx/_transports/mounts.py
```diff
@@ -0,0 +1,218 @@
...
+from .._models import Request, Response
+from .._urls import URL
+from .base import AsyncBaseTransport, BaseTransport
+
+
+class MountNotFound(RequestError):
+    pass
+
+
+class URLPattern:
```
T-256, line 19: why duplicating it from `httpx._utils`?

tomchristie: Well caught... because this pull request *will* also include removing `mounts=...` from the client, but I haven't gotten to that part yet.

It'll make more sense for all the implementation to be together in this module once the functionality is fully encompassed by `httpx.Mounts()`.

httpx/_transports/mounts.py
karpetrosyan: Do we need CI check for this? Check out https://github.com/frostming/fix-future-annotations

karpetrosyan: @Kludex what do you think?

PR #3071, opened 2024-01-25
Drop deprecates

T-256:
# Summary
version 1.0 proposal
Resolves requirements for https://github.com/encode/httpx/pull/3069#pullrequestreview-1843345129

# Checklist


httpx/_client.py
```diff
@@ -210,24 +208,18 @@ def _enforce_trailing_slash(self, url: URL) -> URL:
...
-            for key, value in proxies.items():
-                proxy = Proxy(url=value) if isinstance(value, (str, URL)) else value
-                new_proxies[str(key)] = proxy
-            return new_proxies
-        else:
-            proxy = Proxy(url=proxies) if isinstance(proxies, (str, URL)) else proxies
-            return {"all://": proxy}
+
+        proxy = Proxy(url=proxy) if isinstance(proxy, (str, URL)) else proxy
+        return {"all://": proxy}
```
T-256, line 222: variable name could be improved

tests/client/test_proxies.py
```diff
@@ -14,27 +14,14 @@ def url_to_origin(url: str) -> httpcore.URL:
...
 @pytest.mark.parametrize(
-    ["proxies", "expected_proxies"],
+    ["proxy", "expected_proxies"],
     [
         ("http://127.0.0.1", [("all://", "http://127.0.0.1")]),
-        ({"all://": "http://127.0.0.1"}, [("all://", "http://127.0.0.1")]),
-        (
-            {"http://": "http://127.0.0.1", "https://": "https://127.0.0.1"},
-            [("http://", "http://127.0.0.1"), ("https://", "https://127.0.0.1")],
-        ),
```
T-256, line 24: these units removed because of `proxy` doesn't support mapping.

tomchristie:
Great. Needs a `CHANGELOG.md` entry.
Once we've got that I think we're good to merge this.

T-256:
The purpose of this is for v1.0
If next release is not 1.0 then I'd recommend don't merge this soon.

I thought #3069 will include changelogs. For example for this PR:

> ## Removed
> - `proxies` parameter from client.
> - `data` parameter no longer accepts non mapping objects.
> - `cookies` parameter no longer accepts per requests.



@tomchristie Do any docs need to be updated?

tomchristie:
> The purpose of this is for v1.0

My feeling is that [we should push on with 1.0](https://github.com/encode/httpx/pull/3069#issuecomment-1910060209).

(Tho I also think that this particular change would be acceptable even in 0.27 release, since it's only removing features that are already formally deprecated.)

.github/CONTRIBUTING.md
```diff
@@ -211,9 +211,10 @@ this is where our previously generated `client.pem` comes in:
 ```
 import httpx
 
-proxies = {"all": "http://127.0.0.1:8080/"}
+proxy = "http://127.0.0.1:8080"
```
tomchristie, line 214: This example looks incorrect to me.
I'd assume the proxy address should be `https://`?

tomchristie:
Neat.

> Do any docs need to be updated?

For this PR just the `CHANGELOG.md`.

rafalkrupinski:
#2931

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * Respect the `http1` argument while configuring proxy transports. (#3023)
 * Fix RFC 2069 mode digest authentication. (#3045)
 
+### Removed
+
+* Recently deprecated `proxies` argument has been removed. (#3071)
+* Per request `cookies` argument is not supporting anymore. (#3071)
+* Non-Mapping objects for `data` argument is no longer allowed. (#3071)
```
tomchristie, line 18: I'd suggest we limit the pull request to *just* one line item here, and deal with the `proxies` case.

tomchristie:
Targeting *just* the `proxies=...` change is the thing to do here.
Let's keep PRs sharply targeted. 🎯

T-256:
I ended up to not drop any public api [before major release](https://github.com/encode/httpx/pull/3022#issuecomment-1923970421).
So, closing this for now.

vmatt:
FYI, one uscase: for my unit tests, i need to have dynamic cookie setting (as I impersonate different users. Legacy system auth is based on cookies not bearer)

PR #3080, opened 2024-02-02
Extensions docs

### LINKED ISSUE #2599: Document the `extensions` API.
We have some [documentation on the `extensions` API in the `httpcore` docs](https://www.encode.io/httpcore/extensions/), but I don't see anywhere that we're referencing it in `httpx`.

Maybe we need an extra section in the "[Advanced Usage](https://www.python-httpx.org/advanced/)" docs?

---

### Prompted by https://github.com/encode/httpx/discussions/2552

<sup>Originally posted by **al3x4kov** January 22, 2023</sup>
Hi!

i have this code:
`    async with httpx.AsyncClient(verify=False if not ca_bundle else ca_bundle,
                                 timeout=timeout) as client:`
 `           rsp = await client.request(
            method="GET",
            url=url,
            headers=headers
        )`
 
ca_bundle is a path like /tmp/ca_bundle
timeout = 30.0 

Everything is ok, i get response but the main question is what i should add in my code to have in log strings like:
`console.print(f"* SSL established using {version!r} / {cipher[0]!r}")`
This line from def trace from httpx/_main.py

I found that in client.request i can add parameter `extensions`, but i can't really understand what to pass in

I tried to use TRACE logging type - doesn't work for me. I see only this lines.
image

</div>

tomchristie:
Closes #2599.

Mostly cribbed from the `httpcore` documentation.

* I've re-ordered the request `timeout` to reflect it as a lower priority in this case. Users should be using the `timeout=...` API with `httpx`.
* The `network_stream` docs don't currently include a `CONNECT` example. In order to support this we'd need to be able to specify the request target explicitly, since we currently only support requests that use the standard [HTTP/1.1 abs-path](https://www.rfc-editor.org/rfc/rfc2616#section-5.1.2). Supporting a `target` request extension would allow us to support `OPTIONS *` requests, proxy absolute-url requests, non-standard path quoting and `CONNECT` requests.
* The `network_stream` docs don't currently include an `Upgrade` example. [This discussion](https://github.com/encode/httpcore/discussions/871 ) correctly identifies that we need to handle the `h11` tailing_data in order to properly support that. See https://github.com/encode/httpcore/issues/872


docs/advanced/extensions.md
```diff
@@ -0,0 +1,201 @@
...
+* `"http11.response_closed"`
+
+**HTTP/2 events**
+
+* `"http2.send_connection_init"`
+* `"http2.send_request_headers"`
+* `"http2.send_request_body"`
+* `"http2.receive_response_headers"`
+* `"http2.receive_response_body"`
+* `"http2.response_closed"`
```
karpetrosyan, line 96: Maybe we shouldn't list all of the supported events here. We should remember to update this list every time we introduce a new event to httpcore. We could simply mention them in httpcore and provide a link here.

docs/advanced/extensions.md
```diff
@@ -0,0 +1,201 @@
...
+
+* `read(max_bytes, timeout = None) -> bytes`
+* `write(buffer, timeout = None)`
+* `close()`
+* `start_tls(ssl_context, server_hostname = None, timeout = None) -> NetworkStream`
+* `get_extra_info(info) -> Any`
+
+This API can be used as the foundation for working with HTTP proxies, WebSocket upgrades, and other advanced use-cases.
+
+See the [network backends documentation](network-backends.md) for more information on working directly with network streams.
```
karpetrosyan, line 177: I think you meant this here.

```suggestion
See the [network backends documentation](https://www.encode.io/httpcore/network-backends/) for more information on working directly with network streams.
```

docs/advanced/extensions.md
Kludex: (an automatic workflow here would also help - mkdocs plugins are easy to implement)

karpetrosyan (approved):

PR #3086, opened 2024-02-06
Handle cookies on redirect.

rafalkrupinski:
# Summary

Following #2931

When users passes cookies to a `client.request()` as a dict or list, they're stored as a `Cookies` instance, with domain taken from the URL, and `secure` set when the `url.scheme` is https. This object is kept between redirects, and used to build the redirect request.
When cookies are passed as an instance of `Cookies` or `CookieJar`, they're stored as-is.

New optional parameter `secure` is added to `Cookies.get` and `set`

In effect
- cookies follow redirects to the same host, even with intermediate redirects to other hosts
- cookies don't cross domains, including subdomains
- cookies don't follow downgrade redirect https->http

# Checklist

# To do
- decide whether the new function could be disabled with a flag
- how to handle pickling? Cookies objects are not pickable, bc its inner CookieJar has a lock object
- add documentation (if this feature is tentatively accepted)


rafalkrupinski:
Tests fail bc they don't fully cover ...tests 🤷

povilasb:
Thank you for the PR 🙇 

I'm interested in these changes:

> cookies follow redirects to the same host, even with intermediate redirects to other hosts

Is there anyway I could help to get this merged?

rafalkrupinski:
Hi team!
Any thoughts on this?

PR #3096, opened 2024-02-15
Support cookies with value `None`

tducret:
# Summary

There are 3 kinds of request cookie headers.
```py
request = httpx.Request("GET", "https://www.example.org", cookies=cookies)
request.headers["cookie"]  # <== This is the request cookie header we are talking about here
```

### `"name=value"`

This is the classic case, already supported through : 
```py
cookies.set(name="name", value="value", domain="example.org")
```

### `"name="`

Please note the trailing `=` sign here.
This is also supported, by passing an empty string as the cookie value :
```py
cookies.set(name="name", value="", domain="example.org")
```

### `"name"`

Please note there is **no** trailing `=` sign here.
This is not officially supported (yet), but can be achieved with :
```py
cookies.set(name="name", value=None, domain="example.org")
```

It works because the [cookie jar implementation](https://github.com/python/cpython/blob/0e4f73b8e457b1efa57b735205e8e85a3d11d9f2/Lib/http/cookiejar.py#L1335-L1336) in the standard library already deals with it.
Contrary to the `httpx.Cookies.set` method, the standard `Cookie` [doesn't define strict typing](https://github.com/python/cpython/blob/0e4f73b8e457b1efa57b735205e8e85a3d11d9f2/Lib/http/cookiejar.py#L761-L772) for the `value`.

So this PR changes the `httpx._models` in order to support a cookie `value` which is `None`.
Otherwise the typing is incorrect, forcing the `httpx` to add a `# type: ignore` instruction.

The getter functions had to be updated as well, as the `None` was considered as a not found value.

2 new test cases are added in this PR to ensure the proper behavior.

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -1010,7 +1010,7 @@ class Response:
                 await self.stream.aclose()
 
 
-class Cookies(typing.MutableMapping[str, str]):
+class Cookies(typing.MutableMapping[str, str | None]):
     """
     HTTP Cookies, as a mutable mapping.
     """
@@ -1048,7 +1048,9 @@ class Cookies(typing.MutableMapping[str, str]):
         urllib_request = self._CookieCompatRequest(request)
         self.jar.add_cookie_header(urllib_request)
 
-    def set(self, name: str, value: str, domain: str = "", path: str = "/") -> None:
+    def set(
+        self, name: str, value: str | None, domain: str = "", path: str = "/"
+    ) -> None:
         """
         Set a cookie value by name. May optionally include domain and path.
         """
@@ -1080,23 +1082,26 @@ class Cookies(typing.MutableMapping[str, str]):
         default: str | None = None,
         domain: str | None = None,
         path: str | None = None,
+        raise_when_not_found: bool = False,
     ) -> str | None:
         """
         Get a cookie by name. May optionally include domain and path
         in order to specify exactly which cookie to retrieve.
         """
-        value = None
+        value = default
+        found = False
         for cookie in self.jar:
             if cookie.name == name:
                 if domain is None or cookie.domain == domain:
                     if path is None or cookie.path == path:
-                        if value is not None:
+                        if found:
                             message = f"Multiple cookies exist with name={name}"
                             raise CookieConflict(message)
                         value = cookie.value
+                        found = True
 
-        if value is None:
-            return default
+        if not found and raise_when_not_found:
+            raise KeyError(name)
         return value
 
     def delete(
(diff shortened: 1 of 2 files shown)


httpx/_models.py
```diff
@@ -1008,7 +1008,7 @@ async def aclose(self) -> None:
                 await self.stream.aclose()
 
 
-class Cookies(typing.MutableMapping[str, str]):
+class Cookies(typing.MutableMapping[str, typing.Optional[str]]):
```
T-256, line 1011:
```suggestion
class Cookies(typing.MutableMapping[str, str | None]):
```

T-256 (approved):

httpx/_models.py
tducret: Hi @T-256 ! Actually, That was my initial attempt, but the test workflow [failed](https://github.com/encode/httpx/actions/runs/7914423735/job/21604123090) with this notation.

T-256: Interesting, no problem then...

tomchristie:
Do `name=` and `name` have the same behavior?...

tducret:
> Do name= and name have the same behavior?...

According to the [RFC for HTTP protocol](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3) :

> If the name-value-pair string lacks a %x3D ("=") character, then the name string is empty, and the value string is the value of name-value-pair.

So it may depend on the server implementation, but I can confirm a case where the server considered `name=` and `name` as different.
Although this was already possible to create both kinds of cookies, this PR confirms this usage by changing the type hint.

httpx/_models.py
```diff
@@ -1080,23 +1082,26 @@ def get(  # type: ignore
         default: str | None = None,
         domain: str | None = None,
         path: str | None = None,
+        raise_when_not_found: bool = False,
```
tomchristie, line 1085: Could we avoid adding this API as part of the pull request?

tests/models/test_cookies.py
```diff
@@ -30,6 +30,16 @@ def test_cookies_update():
...
 
 
+def test_cookie_value_can_be_none():
+    cookies = httpx.Cookies()
+    more_cookies = httpx.Cookies()
+    more_cookies.set("no-value", None, domain="example.com")
+
+    cookies.update(more_cookies)
+    assert dict(cookies) == {"no-value": None}
+    assert cookies.get("no-value", domain="example.com") is None
```
tomchristie, line 40: It's not absolutely clear to me that allowing `None` here is an API improvement. 🤔

httpx/_models.py
tducret: Here, I wanted to preserve this responsibility of raising the `KeyError` exception to the `__getitem__()` method.

However, if it's safe to raise it whenever it's not found through `get()`, then we could remove `raise_when_not_found` parameter.

What do you think?

tests/models/test_cookies.py
tducret: I changed the test to `assert cookies["no-value"] is None`.
I also added `test_get_missing_cookie_raises_exception` to highlight the difference between getting a missing cookie (raises exception) and getting a cookie whose value is `None`.

Is it better?

tomchristie:
Thanks @tducret. Appreciate your time. I think we'll pass on this.
I think we should just treat `name` and `name=` identically.
Ie. both should be the empty string.

tducret:
Alright @tomchristie, no worries.

tomchristie:
(Apologies for the delay)

PR #3097, opened 2024-02-16
Update CHANGELOG.md

Rainkenstein:
wrong year I think? I'm new to github so idk if I'm doing this right

# Summary

# Checklist


tomchristie:
@Rainkenstein We like to make sure you're paying attention. 😎

Rainkenstein:
ahaha this is the second reaction I've gotten like this, I think I'm gonna like github

PR #3103, opened 2024-02-21
✨  obey retry after delay

ndaWefox:
# Summary

Obey retry after as defined by the [specifications](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1414,9 +1414,8 @@ class AsyncClient(BaseClient):
         follow_redirects: bool = False,
         limits: Limits = DEFAULT_LIMITS,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
-        event_hooks: None | (
-            typing.Mapping[str, list[typing.Callable[..., typing.Any]]]
-        ) = None,
+        event_hooks: None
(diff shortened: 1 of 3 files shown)


ndaWefox:
This issue was raised on [Retry requests](https://github.com/encode/httpx/issues/108).

Proposition was to add a transport layer but it will not suffice, as if the redirection is done by the transport layer, then the history will be lost due to the behavior of `Client._send_handling_redirects`

tomchristie:
Ah thanks for getting involved.
We'd really need this to be started as a discussion first, tho it *is* helpful to see how you've resolved it.

> Proposition was to add a transport layer but it will not suffice, as if the redirection is done by the transport layer, then the history will be lost due to the behavior of Client._send_handling_redirects

That's interesting... can we instead resolve that issue? Ie... if a transport returns a `Response` with a non-empty history, then the full history is retained, even when redirects occur?

tomchristie:
(I'm going to close this off as a pull request, but very much open to this discussion progressing... I think you've probably identified a valid issue here let's talk it through)

PR #3106, opened 2024-02-21
Define and expose the API from the same place

karpetrosyan:
This is a little refactor of our import strategy.

Now the main `__init__.py` file contains a lot of imports to deliver the public API that we want to expose in httpx.
It's actually working perfectly, but we can make it even simpler by splitting that logic and exposing the API from each module.

So in each module, we should define the variable `__all__`, which will contain the names of the classes that we want to expose at one level higher.

And each time we want to introduce a new public API, we will just modify the variable `__all__` from the module that class belongs to, rather than modifying the main `__init__.py` module of the httpx library.

### DIFF
--- a/tests/test_exported_members.py
+++ b/tests/test_exported_members.py
@@ -3,16 +3,6 @@ import httpx
 
 def test_all_imports_are_exported() -> None:
     included_private_members = ["__description__", "__title__", "__version__"]
-    print(
-        sorted(
-            (
-                member
-                for member in vars(httpx).keys()
-                if not member.startswith("_") or member in included_private_members
-            ),
-            key=str.casefold,
-        )
-    )
     assert httpx.__all__ == sorted(
         (
             member
--- a/tests/test_exported_members.py
+++ b/tests/test_exported_members.py
@@ -3,6 +3,16 @@ import httpx
 
 def test_all_imports_are_exported() -> None:
     included_private_members = ["__description__", "__title__", "__version__"]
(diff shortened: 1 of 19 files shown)


tests/test_exported_members.py
```diff
@@ -3,6 +3,16 @@
...
+    print(
+        sorted(
+            (
+                member
+                for member in vars(httpx).keys()
+                if not member.startswith("_") or member in included_private_members
+            ),
+            key=str.casefold,
+        )
+    )
```
karpetrosyan, line 15:
```suggestion
```

T-256:
> And each time we want to introduce a new public API, we will just modify the variable `__all__` from the module that class belongs to, rather than modifying the main `__init__.py` module of the httpx library.

I can't see that `__all__` removed from [`httpx/__init__.py`](https://github.com/encode/httpx/pull/3106/files#diff-af9c02a399b9f1d59373788a508a0e21dee694f48dac7681c0c1a2b8c0e12cd1), is it mean that for exposing new api we should modify two `__all__`?
atm, we have only a top-level `__all__` and I prefer to keep it.

tomchristie:
@karpetrosyan - I think this is a neat little refactor yep.

Similarly... our docs could possibly do with a note on importing... we deliberately keep the modules private, the only public API is directly `import httpx`. We could also choose to recommend `import httpx` instead of `from httpx import ...` for a consistent style.

PR #3109, opened 2024-02-22
Use the type.MutableMapping instead of Dict in ASGIApp types

jonfinerty:
MutableMapping is a slightly more permissible type (allowing the previous Dict type) but matches up to Starlettes tpyes

# Summary
Hello! When upgrading to the latest version of Httpx we found that our test client was producing types errors (but fine at run time):

`AsyncClient(transport=ASGITransport(app=app), base_url="http://test")`
```
Argument of type "FastAPI" cannot be assigned to parameter "app" of type "_ASGIApp" in function "__init__"
  Type "FastAPI" cannot be assigned to type "_ASGIApp"
    Parameter 3: type "_Send" cannot be assigned to type "Send"
      Type "_Send" cannot be assigned to type "Send"
        Parameter 1: type "Message" cannot be assigned to type "Dict[str, Any]"
          "Message" is incompatible with "Dict[str, Any]"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```
This might be a bug in pylance not understanding the nested mutable mappings and their compatibililty with dicts, but this change makes httpx more flexible with other ASGI apps, and matches for example Starlette's ASGI app types: https://github.com/encode/starlette/blob/7533b61a34f8b0dcb5fc35c717d458439f0baf18/starlette/types.py#L10 

So the consistency might be worth it?

# Checklist

- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. (covered by existing tests)


tomchristie:
Yep. Seems reasonable enough... possibly worth a `CHANGELOG.md` item. (Tho I think it'd be okay either way)

jonfinerty:
changelog added!

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+* Update ASGIApp types to allow FastApi/Starlette app transports. (#3109)
```
tomchristie, line 11:
```suggestion
* Fix `app` type signature in `ASGITransport`. (#3109)
```

tomchristie (approved):

tomchristie:
Thanks, let's get this in.

antonwnk:
Just curious when this gets shipped into a version. Couldn't find any info on that.

Meanwhile I'll be using the deprecated `AsyncClient(app=app...`

jamestrousdale:
Is there any plan to release this change soon? Still getting hit with deprecation warnings, but if we update then `mypy` and `pyright` complain.

PR #3116, opened 2024-02-23
Fix client.send() timeout when used with a new Request instance

alexgmin:
# Summary

Discussion here https://github.com/encode/httpx/discussions/3110

On `httpx.client.send()`, now the timeout of the request is set if not set already.
This is for cases in which `httpx.client.send()` is used with `httpx.Request()` instead of `client.build_request()`

I expected timeout tests for the non-async client too. But since there are only async tests I didn't create a non-async test.

I didn't think a documentation change was needed since this is (for me) the expect behavior.

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -562,6 +562,15 @@ class BaseClient:
 
         return request.stream
 
+    def _set_timeout(self, request: Request) -> None:
+        if "timeout" not in request.extensions:
+            timeout = (
+                self.timeout
+                if isinstance(self.timeout, UseClientDefault)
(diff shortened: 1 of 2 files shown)


tomchristie (approved):
Nicely done. 😄

nwnorris:
Thank you for this! I spent the last day trying to figure out why my `client.send()` calls weren't timing out as I expected them to. I struggled to find any other discussions of this bug.


I'll use `client.build_request` from here on out, as that seems to be the idiomatic way to construct a request object, but I'm happy that future users won't have to figure this out 😄

PR #3123, opened 2024-02-28
Docs: fix 404ing /advanced/#<anchor> links -> /advanced/clients/#<anchor>

nick2i:
# Summary

A couple of the documentation links in httpx/_client.py docstrings were out of date and led to 404's:
https://www.python-httpx.org/advanced/#request-instances
https://www.python-httpx.org/advanced/#merging-of-configuration

These have been updated to:

https://www.python-httpx.org/advanced/clients/#request-instances
https://www.python-httpx.org/advanced/clients/#merging-of-configuration

I hope this doesn't require a prior discussion 😄 

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.

PR #3124, opened 2024-02-28
Fix broken links in docs/contributing.md and CHANGELOG.md

nick2i:
# Summary

Similarly to https://github.com/encode/httpx/pull/3123, some links to the /advanced/ section of the docs were out of date.
There is also one instance of link text changing from `Writing custom transports` -> `Custom transports` to match the section header.

There is an outstanding broken link in README.md that I wasn't sure about, maybe it should point to https://www.python-httpx.org/advanced/transports/#wsgi-transport?
```
[. . .] directly to [WSGI applications](https://www.python-httpx.org/advanced/#calling-into-python-web-apps) or [. . .]
```

# Checklist


tomchristie (approved):
Thanks! ☺️

PR #3139, opened 2024-03-10
Add support for zstd decoding

mbeijen:
# Summary

This adds support for zstd decoding using the python package zstandard. This is similar to how it is implemented in urllib3. I also chose the optional installation option httpx[zstd] to mimic the same option in urllib3.

zstd decoding is similar to brotli, but in benchmarks it is supposed to be even faster. The zstd compression is described in RFC 8878.

See https://github.com/encode/httpx/discussions/1986

# Checklist


httpx/_decoders.py
```diff
@@ -140,6 +140,41 @@ def flush(self) -> bytes:
...
+            raise ImportError(
+                "Using 'ZStandardDecoder', ..."
+                "Make sure to install httpx using `pip install httpx[zstd]`."
+            ) from None
+
+        self.decompressor = zstd.ZstdDecompressor().decompressobj()
+
+    def decode(self, data: bytes) -> bytes:
+        try:
+            data_parts = [self.decompressor.decompress(data)]
```
Zaczero, line 162: Perfect use case for BytesIO

httpx/_decoders.py
```diff
@@ -140,6 +140,41 @@ def flush(self) -> bytes:
...
+            while self.decompressor.eof and self.decompressor.unused_data:
+                unused_data = self.decompressor.unused_data
+                self.decompressor = zstd.ZstdDecompressor().decompressobj()
+                data_parts.append(self.decompressor.decompress(unused_data))
```
Zaczero, line 166: I may be wrong about this one. But wouldn't read_across_frames=True simplify this logic (since it will allow to read multiple frames)?

httpx/_compat.py
```diff
@@ -17,6 +18,23 @@
...
+else:
+    # The package 'zstandard' added the 'eof' property starting
+    # in v0.18.0 which we require to ensure a complete and
+    # valid zstd stream was fed into the ZstdDecoder.
+    # See: https://github.com/urllib3/urllib3/pull/2624
+    _zstd_version = tuple(
+        map(int, re.search(r"^([0-9]+)\.([0-9]+)", zstd.__version__).groups())  # type: ignore[union-attr]
+    )
+    if _zstd_version < (0, 18):  # Defensive:
+        zstd = False
```
Zaczero, line 35: This is just out of my curiosity, feel free to not answer it.
Why do we need this code if we already specify zstandard>=0.18.0, wouldn't package manager guarantee it already?

mbeijen: zstandard is an optional install. So if your `requirements.txt` just has `httpx` and not `httpx[zstd]`, and you have `zstandard==[old version]` somehow, this would prevent breakage.

Zaczero: That's so smart. Thank you!

httpx/_decoders.py
mbeijen: Thanks, I've added BytesIO here

httpx/_decoders.py
mbeijen: I'm not exactly sure what happens, but if I try to use `read_across_frames` I get decoding errors. Also, please note I've copied this logic from urllib3 so it is "Best Practice" :-) or at least proven to be working.

I would be open for suggestions of course!

docs/index.md
```diff
@@ -138,10 +139,11 @@ Or, to include the optional HTTP/2 support, use:
 $ pip install httpx[http2]
 ```
 
-To include the optional brotli decoder support, use:
+To include the optional brotli and zstandard decoder support, use:
 
 ```shell
 $ pip install httpx[brotli]
+$ pip install httpx[zstd]
```
Zaczero, line 146: I think it would be better to merge those 2 into a single command. New developers sometimes don't know how to merge multiple optional deps and the index.md lacks such an example. Also when using "and" it indicates the presence of both decoders, while running those 2 commands sequentially will end up with just 1 decoder (2nd command will uninstall brotli).

```suggestion
To include the optional brotli and zstandard decoders support, use:

```shell
$ pip install httpx[brotli,zstd]
```

Zaczero (approved):
Overall I like those changes!

httpx/_compat.py
```diff
@@ -17,6 +20,24 @@
...
+else:
+    # The package 'zstandard' added the 'eof' property starting
+    # in v0.18.0 which we require to ensure a complete and
+    # valid zstd stream was fed into the ZstdDecoder.
+    # See: https://github.com/urllib3/urllib3/pull/2624
+    _zstd_version = tuple(
+        map(int, re.search(r"^([0-9]+)\.([0-9]+)", zstd.__version__).groups())  # type: ignore[union-attr]
+    )
+    if _zstd_version < (0, 18):  # Defensive:
+        zstd = None
```
tomchristie, line 38: This might be overly defensive? Can we pin >= 0.18 in the requirements instead?

tomchristie:
This is a lovely bit of functionality. Missing some coverage at the moment.

httpx/_compat.py
Zaczero: I asked a similar question already and got this answer:

https://github.com/encode/httpx/pull/3139#discussion_r1518815892

> zstandard is an optional install. So if your `requirements.txt` just has `httpx` and not `httpx[zstd]`, and you have `zstandard==[old version]` somehow, this would prevent breakage.

Basically the idea is that the optional dependency pin is optional and does not guarantee that the given dependency version will be installed. This check would only be redundant if we made zstandard a hard dependency (one not requiring httpx[zstandard]).

The same check [exists](https://github.com/urllib3/urllib3/blob/733f638a2faa02b4ff8a9f3b5668949d39396b8b/src/urllib3/response.py#L28-L43) in urllib3.

Personally, I +1 the extra safety at the cost of minimal startup delay. :slightly_smiling_face:

mbeijen:
> This is a lovely bit of functionality. Missing some coverage at the moment.

Thanks. I will see if I can add tests for full coverage

httpx/_compat.py
tomchristie: Okay yep. Main thing is coverage needs addressing.

mbeijen:
I've added full test coverage now, again by borrowing a test inspired from urllib3
@tomchristie please consider my PR again

tomchristie:
> please consider my PR again

Sure thing. (thanks) 

* Are we crediting the `urllib3` team sufficiently/appropriately here, perhaps a code comment in places where we're drawing from their work?
* Could you provide an example URL in this thread to test against?
* Let's get a `CHANGELOG.md` entry added.

mbeijen:
>     * Are we crediting the `urllib3` team sufficiently/appropriately here, perhaps a code comment in places where we're drawing from their work?

I've added a code comment in the appropriate places. In the docs and README there is already a credit given to urllib3 as a source for much of the lower level details, this is not much different. It is fair to add some extra credit here as well as a pointer for a reader to check the urllib3 implementatition in case of changes.

>     * Could you provide an example URL in this thread to test against?

Sure! Support for zstd is not really common on web sites yet. libcurl has support for some years now. Chrome has it now enabled by default so I will assume this might help? Also, Mozilla changed its stance on zstd from 'defer' to 'positive' in October, ref: https://github.com/mozilla/standards-positions/issues/775#issuecomment-1754635490 -- it is particularly well-suited for compressing JSON data.

Sites that support it include instagram.com, facebook.com, caddyserver.com 

```
httpx -v --http2 https://caddyserver.com/  | grep encoding:
accept-encoding: gzip, deflate, br, zstd
content-encoding: zstd
``

>     * Let's get a `CHANGELOG.md` entry added.

I've added that

**Review by tomchristie — APPROVED** [review]
Really nice work ☺️

```
$ httpx -v https://caddyserver.com
* Connecting to 'caddyserver.com'
* Connected to '165.227.20.207' on port 443
* SSL established using 'TLSv1.3' / 'TLS_AES_128_GCM_SHA256'
* Selected ALPN protocol: 'http/1.1'
* Server certificate:
*   subject:
*     commonName: 'caddyserver.com'
*   issuer:
*     countryName: 'US'
*     organizationName: "Let's Encrypt"
*     commonName: 'R3'
*   version: 3
*   serialNumber: '0334F48787F76D10C8FF7D436F4AB8C56B0F'
*   notBefore: 'Feb 26 23:44:40 2024 GMT'
*   notAfter: 'May 26 23:44:39 2024 GMT'
*   subjectAltName:
*     DNS: 'caddyserver.com'
*   OCSP:
*     'http://r3.o.lencr.org'
*   caIssuers:
*     'http://r3.i.lencr.org/'
GET / HTTP/1.1
Host: caddyserver.com
Accept: */*
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
User-Agent: python-httpx/0.27.0

HTTP/1.1 200 OK
Alt-Svc: h3=":443"; ma=2592000
Content-Encoding: zstd
Content-Type: text/html; charset=utf-8
Server: Caddy
Vary: Accept-Encoding
Date: Thu, 21 Mar 2024 10:12:38 GMT
Transfer-Encoding: chunked
```

mrcatspoon:
@tomchristie Hello there 👋 Small question: When are you planning to release this feature?

tomchristie:
Useful question. Want to draft up a release PR?

Issue #3140, opened 2024-03-11
Httpx is not encoding "%" symbol while sending requests


heysarthak:
### Discussed in https://github.com/encode/httpx/discussions/3094

<div type='discussions-op-text'>

<sup>Originally posted by **heysarthak** February 14, 2024</sup>
Hi team, 

I am ussing httpx plugin, while sending httpx query "ngi%abc.com" .
ideally "%" symbol should be encoded as "%25" but it is still sending as the same "%".

can we please have a fix for this.</div>

deedy5:
`%` character has a special meaning in URLs, representing the start of a percent-encoded character. To include a literal `%` character in a URL, it must be percent-encoded as `%25`.

You can manually encode the query parameters in Python using the `urllib.parse` module:
```python3
from urllib.parse import quote
import httpx

query_param = "ngi%abc.com"
encoded_query_param = quote(query_param)

url = f"https://example.com/search?q={encoded_query_param}"
resp = httpx.get(url)

PR #3141, opened 2024-03-14
Fix httpx is not encoding with symbol '%s' (#3140)

Tunglies:
Fix httpx is not encoding with symbol '%s' (#3140)
Add url parsing test

# Summary

# Checklist


httpx/_urlparse.py
```diff
@@ -424,13 +424,23 @@ def is_safe(string: str, safe: str = "/") -> bool:
...
+            char == "%"
+            and i + 2 < len(string)
+            and string[i + 1 : i + 3].isalnum()
+            and len(string[i + 1 : i + 3]) == 2
+        ):
+            i += 3
+        else:
             return False
-
     return True
```
T-256, line 444:
```suggestion

    # All characters must already be non-escaping or valid percent-encoded
    for index, char in enumerate(string):
        if char not in NON_ESCAPED_CHARS and not PERCENT_ENCODED_REGEX.match(
            string[index : index + 3]
        ):
            return False

    return True
```

Tunglies: Thanks for review! I have applied the suggestion.

T-256 (approved):
Thanks, I think it deserves a note in changelog.

heysarthak:
Hi @T-256,
will the code change encode every "%" sent into uri as "%25" ?

T-256:
> Hi @T-256, will the code change encode every "%" sent into uri as "%25" ?

Yes, if it is invalid percent-encoded.

heysarthak:
when can we merge this changes? @tomchristie

jhominal:
Hello, could you look at if this PR fixes the cases of bad `%` character encoding that are raised in https://github.com/encode/httpx/discussions/3135?

sileht (approved):

tomchristie:
Thanks. I think this is superseeded by #3187.

heysarthak:
I tried the same on http 0.27.0, still having the same issue.

tomchristie:
Thanks everyone.
Remaining issues here should now all fully be addressed in #3371 and #3373.

PR #3143, opened 2024-03-14
pydantic initial integration

07pepa:
# Summary
i wanted to use URL to be  usable in pydantic setting directly see https://github.com/pydantic/pydantic-extra-types/pull/156 this is mainly to entice discussion and is in no way shape or form final (if this may be accepted)

# Checklist

- [-] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [-] I've updated the documentation accordingly.

### DIFF
--- a/httpx/_urls.py
+++ b/httpx/_urls.py
@@ -4,9 +4,12 @@ import typing
 from urllib.parse import parse_qs, unquote
 
 import idna
+from pydantic import GetCoreSchemaHandler, GetJsonSchemaHandler
+from pydantic_core import core_schema
 
+from ._exceptions import InvalidURL
 from ._types import QueryParamTypes, RawURL, URLTypes
-from ._urlparse import urlencode, urlparse
+from ._urlparse import MAX_URL_LENGTH, urlencode, urlparse
 from ._utils import primitive_value_to_str
 
 __all__ = ["URL", "QueryParams"]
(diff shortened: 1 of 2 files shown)


tomchristie:
Can you describe from a docs pov what you're trying to do?

07pepa:
I am tryiing to acheave compatibility so url can be in setting class (pulled from .env files or enviroment or other sources)

```
from pydantic_settings import BaseSettings
from httpx import URL


class Settings(BaseSettings):
    url: URL 



CONFIG = Settings()
AsyncClient = AsyncClient(base_url=CONFIG.url)
```

alternative is to support `AnyUrl` from pydantic

tomchristie:
Okay, can you provide an example of what the above *does* and then show why `httpx` would need to be modified in order to accomodate it?

07pepa:
> Okay, can you provide an example of what the above _does_ and then show why `httpx` would need to be modified in order to accomodate it?

Trick is it does not need to  modified... it just would be nice to do it Here (and can be done other way around (extending URL from HTTPX in pydantic) it is mostly for better convinience


demo without change
```python
import os

from pydantic_settings import BaseSettings
from pydantic import AnyUrl
from httpx import AsyncClient


class Settings(BaseSettings):
    url: AnyUrl


# assume this is in .env file
os.environ["url"] = "http://localhost"

CONFIG = Settings()

client = AsyncClient(base_url=str(CONFIG.url))
```


here is demo with change
```python
import os

from pydantic_settings import BaseSettings
from httpx import URL, AsyncClient


class Settings(BaseSettings):
    url: URL


# assume this is in .env file
os.environ["url"] = "http://localhost"

CONFIG = Settings()

client = AsyncClient(base_url=CONFIG.url)
```

PR #3155, opened 2024-03-29
✨ maintain transport response history

ndaWefox:
# Summary

Maintain the response history when using transport layer that retrieves their own response with history

# Checklist


spumer:
Any updates?

PR #3172, opened 2024-04-17
tests: fix win32

trim21:
# Summary

There are several test cases broken on windows.

# Checklist


tests/test_utils.py
```diff
@@ -142,6 +144,7 @@ def test_logging_ssl(caplog):
     ]
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Path separator problem")
```
tomchristie, line 147: What's the issue here?

trim21: tests will assert file path with `/`

tomchristie: Okay, so let's fix up the test case so that we don't do that.

Presumably that's this line?...

```
assert ca_bundle is not None and ca_bundle.endswith("tests/test_utils.py")
```

Can we fix up those cases so we're not using literal `/` seperators.

tests/client/test_async_client.py
```diff
@@ -18,7 +19,9 @@ async def test_get(server):
     assert response.http_version == "HTTP/1.1"
     assert response.headers
     assert repr(response) == "<Response [200 OK]>"
-    assert response.elapsed > timedelta(seconds=0)
+    if sys.platform != "win32":
+        # flaky on windows
+        assert response.elapsed > timedelta(seconds=0)
```
tomchristie, line 24: Thanks for highlighting flaky areas on our tests, windows.

I think this is a bit of code smell either way. Perhaps we switch it to this?...

```suggestion
    assert response.elapsed >= timedelta(seconds=0)
```

trim21: isn't this a no-op?

tests/test_multipart.py
```diff
@@ -371,6 +372,9 @@ def test_multipart_encode_files_raises_exception_with_StringIO_content() -> None
         httpx.Request("POST", url, data={}, files=files)  # type: ignore
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="TemporaryFile on windows is binary mode"
+)
 def test_multipart_encode_files_raises_exception_with_text_mode_file() -> None:
     url = "https://www.example.com"
     with tempfile.TemporaryFile(mode="w") as upload:
```
tomchristie, line 380: Perhaps we can resolve the test case like this?...

```suggestion
    # TemporaryFiles are always binary mode on windows.
    # For this test case where we need a text-mode file.
    # We make sure to create-and-write in binary mode first,
    # and then open the file in text mode.
    with tempfile.NamedTemporaryFile() as temp:
        temp.write(b'xxxxxxxxxx')
        with open(upload.name, mode="r") as upload:
            ...
```

Or else use a mock file-like object perhaps?

tests/client/test_async_client.py
tomchristie: It's an odd and flaky thing to test against.
I think it only ended up in there because of code coverage.
We could comment it with something like this perhaps?

```python
# Sanity test the property access and return type.
```

tests/test_multipart.py
trim21: this doesn't look right? what's `upload.name`

trim21: do you mean `temp.name`? it's not possible to do this on windows.

trim21: so I think another way is to use a `io.StringIO`

trim21:
coverage failed because this:

https://github.com/encode/httpx/blob/fa6dac8383541669f42086d4f07c71b0f8d6a279/httpx/_multipart.py#L127-L134

`io.StringIO` is a subclass of `io.TextIOBase`

trim21:
not sure should I change code in source file or code in test file

trim21:
I just use pytest's tmp_path to create a real file on disk. pytest will clean it up after test.

tests/test_utils.py
```diff
@@ -122,6 +123,7 @@ def test_logging_redirect_chain(server, caplog):
     ]
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="os path sep escaped on windows")
```
tomchristie, line 126: We don't need this `skipif` now, right?

tests/test_timeouts.py
```diff
@@ -12,6 +14,9 @@ async def test_read_timeout(server):
             await client.get(server.url.copy_with(path="/slow_response"))
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="time related tests are flaky on win32"
+)
```
tomchristie, line 19: I assume we could drop this, and this particular test *would actually* pass okay on windows?

(Also, I think we might want to raise a related refactoring issue to remove these `/slow_response` tests completely.)

.github/workflows/test-suite.yml
```diff
@@ -32,3 +37,4 @@ jobs:
         run: "scripts/test"
       - name: "Enforce coverage"
         run: "scripts/coverage"
+        if: "matrix.os == 'ubuntu'"
```
tomchristie, line 40: We probably don't need this conditional once we've fixed up all the tests to pass okay on windows, right?

.github/workflows/test-suite.yml
```diff
@@ -7,13 +7,18 @@ on:
...
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
-    runs-on: "ubuntu-latest"
+    name: "Python ${{ matrix.python-version }} (${{ matrix.os }})"
+    runs-on: "${{ matrix.os }}-latest"
 
     strategy:
       matrix:
+        os: ["ubuntu", "windows"]
```
tomchristie, line 21: I'm not sure - I'm pleased to see us fix the test cases up for supporting windows, tho do we want to double the time our test runs take to complete?

tests/test_utils.py
trim21: we still need this skipif because path seprator `/` is double escaped on this tests, don't know why....

tests/test_timeouts.py
trim21: this test **may** pass on windows.

trim21: it's a time accuracy on windows. we could increate sleep time in `/slow_response` to several seconds to fix this.

tests/test_utils.py
tomchristie: Can you remove the `skipif`, so we can see the exact error in CI?

trim21:
```text
E
E         At index 1 diff: ('httpx', 10, "load_verify_locations cafile='C:\\\\Users\\\\Trim21\\\\proj\\\\httpx\\\\.venv\\\\Lib\\\\site-packages\\\\certifi\\\\cacert.pem'") != ('httpx', 10, "load_verify_locations cafile='C:\\Users\\Trim21\\proj\\httpx\\.venv\\Lib\\site-packages\\certifi\\cacert.pem'")
E
E         Full diff:
E           [
E               (
E                   'httpx',
E                   10,
E                   'load_ssl_context verify=True cert=None trust_env=True http2=False',
E               ),
E               (
E                   'httpx',
E                   10,
E                   'load_verify_locations '
E         -         "cafile='C:\\Users\\Trim21\\proj\\httpx\\.venv\\Lib\\site-packages\\certifi\\cacert.pem'",
E         +         "cafile='C:\\\\Users\\\\Trim21\\\\proj\\\\httpx\\\\.venv\\\\Lib\\\\site-packages\\\\certifi\\\\cacert.pem'",
E         ?                      ++     ++          ++    ++         ++       ++   ++                 ++       ++
E               ),
E           ]

```

.github/workflows/test-suite.yml
trim21: scritps/coverage only works on unix

.github/workflows/test-suite.yml
Kludex: I agree with keeping it simple, and not add windows.

PR #3175, opened 2024-04-20
[fix] Use proxy ssl context consistently

manav-a:
# Summary

SSL context from httpx.Proxy is not used consistently across the proxy connections created in in HTTP transport this PR changes that. Haven't really added a test as I haven't found another test that check for proxy with custom ssl context but would love guidance around if/where one should be added here.

No specific documentation seems to be needed to be updated this should just get the api consistent.

# Checklist


manav-a:
Miss clicked to closed by mistake

karpetrosyan:
My apologies for the delayed review. Yes, we do need to pass the SSL context of the proxy to the HTTPCore pool.

karpetrosyan (approved):

PR #3186, opened 2024-05-02
Fix doc links for making requests directly to WSGI/ASGI apps

kiendang:
# Summary

Fix doc links for making requests directly to WSGI/ASGI apps.

# Checklist


tomchristie (approved):
Thanks. ☺️

PR #3187, opened 2024-05-02
Resolve queryparam quoting

### LINKED ISSUE #3140: Httpx is not encoding "%" symbol while sending requests

tomchristie:
* Percent-encoding in `params={}` should match browser form behavior. (Always percent-encode.)
* Percent-encoding in `url={}` should match browser URL bar behavior. (Percent-encode, preserving existing escape sequences.)

Closes #3135.
Closes #3140.


jhominal (approved):
@tomchristie I think you zeroed in on the right root change, that is, to always escape values from parsed query parameters.

Thank you very much for writing this PR as fast as you did!

tomchristie:
Oops, I should have included a CHANGELOG.md entry.

Perhaps we should update the PR template to include that as a checklist item?

eli-bl:
FWIW, this is a fairly critical bug for some of my users - they're making requests to an endpoint where percent characters can and do appear in the data being queried. Whenever they don't happen to be querying such a value, the automatic escaping of the parameters works as expected. And if they add an explicit "url-encode the param value" step now to handle this case, then their code will break again as soon as this fix is released.

In other words, it'd be great if there could be a release with this fix before too long, so we can just tell people to use a version >= 0.28.

PR #3188, opened 2024-05-03
Add URL parsing tests from WHATWG

tomchristie:
Add 563 URL parsing test cases from https://url.spec.whatwg.org/

Test cases are from...

* https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json
* We're only testing the non-failure cases here.

TODO:

- [ ] We've got a private import here, which we're avoiding in all our other tests case.
- [ ] Add `username`/`password` checks.
- [x] We've two failing cases here.

*Reminder to self follow-ups:*

* *Comparison of WHATWG percent-escape set vs. chrome percent-escape set.*
* *Exceptions for invalid domain names.* <sup>*<sup>

<sup>* eg. see [this comment](https://github.com/whatwg/url/issues/245#issuecomment-1708219206) for useful context</sup>


tomchristie:
Not complete, but suitable for preliminary review.

karpetrosyan (approved):

tomchristie:
Thanks @karpetrosyan. Let's pass on the initial TODO items here, more of a priority to just get this in.

PR #3189, opened 2024-05-03
Fast path returns for normalize_path cases

tomchristie:
Return early for cases where there are no `.` or `..` components in the path.


zanieb (approved):

zanieb:
Did you do any microbenchmarks?

tomchristie:
@zanieb good question.

karpetrosyan:
It's a very small change, I think we can merge this without any microbenchmarks.

tomchristie:
Okay, let's do that then.
It is still a good question. ☺️

PR #3192, opened 2024-05-05
Revert "Removed leading $ from cli code blocks"

tomchristie:
Reverts encode/httpx#3174

Tried the style without the leading `$ ` prompts, and not sure it's quite right.


T-256:
> Reverts #3174
> 
> Tried the style without the leading `$ ` prompts, and not sure it's quite right.

Can you give more details about why it's not?

tomchristie:
For example on the README, our console style "no leading prompt" now isn't consistent with our Python REPL style "leading >>>" prompt.

T-256 (approved):

karpetrosyan (approved):

zanieb:
We did https://github.com/astral-sh/uv/pull/5397 to solve the copy/paste problem over in uv.

PR #3199, opened 2024-05-11
Provide request_class, response_class for httpx.Client 

### LINKED ISSUE #717: Custom JSON library

93578237:
# Summary
Closes #717 

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -641,6 +641,7 @@ class Client(BaseClient):
         self,
         *,
         request_class: type[Request] | None = None,
+        response_class: type[Response] | None = None,
         auth: AuthTypes | None = None,
         params: QueryParamTypes | None = None,
         headers: HeaderTypes | None = None,
@@ -706,6 +707,8 @@ class Client(BaseClient):
         allow_env_proxies = trust_env and app is None and transport is None
         proxy_map = self._get_proxy_map(proxies or proxy, allow_env_proxies)
 
+        self._response_class = response_class or Response
+
         self._transport = self._init_transport(
             verify=verify,
             cert=cert,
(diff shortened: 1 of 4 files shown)


T-256 (approved):
Lets have it ✌️

httpx/_client.py
```diff
@@ -164,6 +164,7 @@ class BaseClient:
     def __init__(
         self,
         *,
+        request_class: type[Request] | None = None,
```
tomchristie, line 167: Presumably also `response_class`, right?
Also I guess move this to the end of the parameters.

q0w: > Presumably also response_class, right?

I thought about that, but baseclient has no logic related with responses, so response_class is now added only in Client for passing to transports. I can added response_class, if needed.

trim21:
some methods of `httpx.Reponse` need to return `Self` instead of `Response`, which is added in py311.

So we need typing_extensions.Self in lower version.

q0w:
> generic type for request and response type

https://github.com/encode/httpx/issues/717#issuecomment-2118083691

trim21:
return type of httpx.Reponse.raise_for_status still need to update

thomasht86:
Hi!
Just wondering what's the status on this? 
Is there any intention to get it merged?

oelhammouchi:
Hello, thanks to all involved for the great work! Are there any updates with respect to this PR? It would be very useful atm :sweat_smile:

MarcBresson:
any news for this PR?

lovelydinosaur:
@MarcBresson I'm not progressing this because the work on the `httpx` 1.0 prerelease deals with this much more comprehensively. (To the point that implementing a custom client class is really simple to implement)

Eg. There's a worked through example on https://www.encode.io/httpnext/clients "Custom client implementations" - contrast that with the existing `Client` implementation and signatures if you want to start getting a feel for how much sharper and neater the 1.0 typing is.

*And, yes... it's taking a little time for me to start presenting the work fully... it's on the way*

PR #3210, opened 2024-05-30
Update asgi.py docstring

michaelfeil:
# Summary

Update deprecated use of ASGITransport.

# Checklist


tomchristie (approved):
Thanks :)

Issue #3170, opened 2024-04-17
add more type checking in params/form-data


trim21:
- [ ] Initially raised as discussion #3168

# Problem
we should check input value types. Current behavoir which simply convert input value to `str(v)` is not very ideal.

for example:
```py
import httpx


r = httpx.post(
    "https://httpbin.org/post",
    data={
        "content": bytes.fromhex(
            "4f5f13d16da9f10d013933673a1ef4e7b974dbe3"
        ),  # random non-printable bytes
    },
)
print(r.json()["form"])
```

will get `{'content': "b'O_\\x13\\xd1m\\xa9\\xf1\\r\\x0193g:\\x1e\\xf4\\xe7\\xb9t\\xdb\\xe3'"}`

# Proposal

add type checking and more coerce.

for example:

## form

- convert `int`/`float` to str with `str(v)`
- `bool` (Represent as JSON-style "true"/"false")
- `None` (Represent as empty string)
- `Enum` (Use .value and then as above.)
- `TypeError` for anything else.

## query
except about, query should support `bytes` with percent encoding and do not throw.

## multipart_data

not sure how to handle this....

This will cause breaking changes, so we should consider do a minor version bump.

---

related: https://github.com/encode/httpx/issues/1500 https://github.com/encode/httpx/pull/1539 https://github.com/encode/httpx/pull/1608

https://github.com/encode/httpx/blob/4b85e6c3898b94e686b427afd83138c87520b479/httpx/_content.py#L136-L149

https://github.com/encode/httpx/blob/4b85e6c3898b94e686b427afd83138c87520b479/httpx/_utils.py#L56-L68

trim21:
there is also another issue, query string support bytes natively.

for example:

```python
import httpx
import urllib.parse

r = httpx.get(
    "https://httpbin.org/get",
    params={"q": bytes.fromhex("E1EE0E2734986F5419BB6CB6252BD9377183440E")},
)

print(urllib.parse.quote(bytes.fromhex("E1EE0E2734986F5419BB6CB6252BD9377183440E")))

print(r.text)
```

expected url should be `https://httpbin.org/get?q=%E1%EE%0E%274%98oT%19%BBl%B6%25%2B%D97q%83D%0E`, instead of ValueError or TypeError

tomchristie:
> query string should support bytes natively.

I'm not convinced by that. Why?

trim21:
> > query string should support bytes natively.
> 
> I'm not convinced by that. Why?

[...]

PR #3212, opened 2024-05-31
feat: support bytes as query value

trim21:
# Summary

ref: https://github.com/encode/httpx/issues/3170

bytes as query string are supported by url spec but not httpx, this PR add support for bytes query value types.

it also fix a `QueryParams` mutable problem when using `set`/`add`

# Checklist


tests/test_content.py
```diff
@@ -200,7 +200,7 @@ async def test_urlencoded_content():
 
 @pytest.mark.anyio
 async def test_urlencoded_boolean():
-    request = httpx.Request(method, url, data={"example": True})
+    request = httpx.Request(method, url, data={"example": True, "e2": False})
```
trim21, line 203: for coverage

PR #3220, opened 2024-06-11
Drop `anyio` dependency

T-256:
It is currently used indirectly by httpcore's _asyncio_ feature.
There is no usage of `anyio` in httpx's codebase.


rafalkrupinski:
There's some code that tries to use trio falling back to asyncio. Why not use anyio for this? less code to maintain...

PR #3226, opened 2024-06-20
Add An Importable `default_user_agent` Function

MichaelYochpaz:
# Summary
This PR adds a `default_user_agent` function (corresponding to [requests.utils.default_user_agent](https://github.com/psf/requests/blob/main/src/requests/utils.py#L891-L897)) and allows it to be imported using `from httpx import default_user_agent` to allow different projects to check what user-agent should be expected by default from the clients they're using (for unit-tests, for example).

This value existed as a constant variable named `USER_AGENT` (accessible only using `httpx._client.USER_AGENT`) which I didn't necessarily have to convert to a function, but I thought it would be better as a function both because it will be implemented similarly to `requests`, and because it will allow more flexibility in the future, in case it will need to be dynamically calculated (changing it to a function from a variable if needed in the future will break backwards compatibility).

Don't think this change requires any unit-tests or docs to be updated (didn't find any matching section), but if you think it does, let me know and I'll try to add those.

# Checklist

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -48,6 +48,7 @@ __all__ = [
     "Cookies",
     "create_ssl_context",
     "DecodingError",
+    "default_user_agent",
     "delete",
     "DigestAuth",
     "get",
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -58,7 +58,7 @@ from ._utils import (
     same_origin,
 )
 
-__all__ = ["USE_CLIENT_DEFAULT", "AsyncClient", "Client"]
+__all__ = ["USE_CLIENT_DEFAULT", "AsyncClient", "Client", "default_user_agent"]
 
 # The type annotation for @classmethod and context managers here follows PEP 484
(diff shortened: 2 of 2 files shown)


tomchristie:
Reasonable request, tho we'll prioritise minimising API surface area.

Perhaps try this?...

```python
ua = httpx.Client().headers['User-Agent']
```

MichaelYochpaz:
> Reasonable request, tho we'll prioritise minimising API surface area.
> 
> Perhaps try this?...
> 
> ```python
> ua = httpx.Client().headers['User-Agent']
> ```

@tomchristie this solution will initialize a new `Client` object just to get this constant string. Seems very wasteful to me, compared to the simple solution of just allowing access to this variable through the API (which, as I mentioned, `requests` allows).

tomchristie:
That’s a valid point.
Alternatives here…

* Set the User-Agent explicitly on the client instance, so that it is known.
* Use `f”python-httpx/{ httpx.__version__ }”`

Should the UA format be covered by our versioning contract once we’re at 1.0?

PR #3244, opened 2024-07-19
make 'anyio' optional requirement.

xiacunshun:
'anyio' is only used in tests, so there is no need to require it in main process.

# Summary

# Checklist


waketzheng:
#3220

Issue #3176, opened 2024-04-21
Requests for new public APIs


karpetrosyan:
Ref: https://github.com/encode/starlette/pull/2534#issuecomment-2067634124

This issue was opened to resolve all kinds of problems where third-party packages may need to use private imports. We should definitely expose them to avoid problems like https://github.com/encode/httpx/pull/3130#issuecomment-1974778017.

I will include the package name along with its private imports (only starlette for now), so we can clearly identify what needs to be exposed.

Note that:
- [ ] -> Not yet publicly exposed
- [x] -> Publicly exposed

We can also add other packages as well to track all of them in this issue.

# Starlette

- [ ] [httpx._types.CookieTypes](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L431)
- [ ] [httpx._client.UseClientDefault](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L475C17-L475C47)
- [ ] [httpx._client.USE_CLIENT_DEFAULT](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L477)
- [ ] [httpx._types.URLTypes](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L497C14-L497C35)
- [ ] [httpx._types.RequestContent](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L499C18-L499C45)
- [ ] [httpx._types.RequestFiles](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L501C16-L501C41)
- [ ] [httpx._types.QueryParamTypes](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L503C17-L503C45)
- [ ] [httpx._types.HeaderTypes](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L504C18-L504C42)
- [ ] [httpx._types.CookieTypes](https://github.com/encode/starlette/blob/96c90f26622c8f243ad965371eae2c2028a518de/starlette/testclient.py#L505C18-L505C42)
[...]

PR #3245, opened 2024-07-23
Clean up URL signature.

tomchristie:
Refs https://github.com/encode/httpx/issues/3176#issuecomment-2068231852

Our public methods currently include a shortcut type `URLTypes`, which is not public API.
This refactoring removes the shortcut type in favor of explicit typing.

There are also other similar cases, tho one this is an easy refactoring to demonstrate.

### DIFF
--- a/httpx/_api.py
+++ b/httpx/_api.py
@@ -18,9 +18,9 @@ from ._types import (
     RequestData,
     RequestFiles,
     TimeoutTypes,
-    URLTypes,
     VerifyTypes,
 )
+from ._urls import URL
 
 __all__ = [
     "delete",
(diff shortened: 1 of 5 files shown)


httpx/_api.py
```diff
@@ -18,9 +18,9 @@
     RequestData,
     RequestFiles,
     TimeoutTypes,
-    URLTypes,
     VerifyTypes,
 )
+from ._urls import URL
```
T-256, line 23: could it be under `typing.TYPE_CHECKING`?

T-256 (approved):

httpx/_api.py
tomchristie: I'd only use that if it was required to avoid a cyclical dependency.

zanieb (approved):

PR #3262, opened 2024-08-09
fix HTTP/2 docs

ArkadyRudenko:
Note on forcing HTTP/2 to be enabled by disabling HTTP/1.1 for the client

### DIFF
--- a/docs/http2.md
+++ b/docs/http2.md
@@ -58,6 +58,13 @@ requests and responses will be transported over HTTP/2, since both the client
 *and* the server need to support HTTP/2. If you connect to a server that only
 supports HTTP/1.1 the client will use a standard HTTP/1.1 connection instead.
 
+To force HTTP/2 to be enabled, disable HTTP/1.1. Otherwise, the client will try to use HTTP/1.1
+```python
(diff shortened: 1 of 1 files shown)


docs/http2.md
```diff
@@ -58,6 +58,13 @@ requests and responses will be transported over HTTP/2, since both the client
 *and* the server need to support HTTP/2. If you connect to a server that only
 supports HTTP/1.1 the client will use a standard HTTP/1.1 connection instead.
 
+To force HTTP/2 to be enabled, disable HTTP/1.1. Otherwise, the client will try to use HTTP/1.1
```
tomchristie, line 61: This isn't correct. Here's the deal...

* Default. `http1=True, http2=False` - HTTP/1.1 will be used.
* HTTP/2 support enabled. `http1=True, http2=True` - Either HTTP/1.1 or HTTP/2 will be used. The protocol negotiation (ALPN) occurs during the SSL setup.
* Force HTTP/2. `http1=False, http2=True` - HTTP/2 will be used, without negotiation. Not a typically useful configuration, tho allows HTTP/2 over plain HTTP.

PR #3270, opened 2024-08-16
fix(decoder): decode multi-member gzip data

lizeyan:
# Summary

https://github.com/encode/httpx/discussions/3269

Allow GZipDecoder to decode multi-member gzip data

# Checklist

### DIFF
--- a/httpx/_decoders.py
+++ b/httpx/_decoders.py
@@ -7,6 +7,7 @@ See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
 from __future__ import annotations
 
 import codecs
+import enum
 import io
 import typing
 import zlib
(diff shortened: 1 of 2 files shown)


tests/test_decoders.py
```diff
@@ -9,6 +9,7 @@
 import zstandard as zstd
 
 import httpx
+from httpx._decoders import GZipDecoder
```
tomchristie, line 12: Rather than using a private import, could we switch this to use the same style as the other test cases here.

httpx/_decoders.py
```diff
@@ -64,21 +65,37 @@ def flush(self) -> bytes:
...
+    OTHER_MEMBERS = 1
+    SWALLOW_DATA = 2
 
-    See: https://stackoverflow.com/questions/1838699
-    """
 
+class GZipDecoder(ContentDecoder):
     def __init__(self) -> None:
         self.decompressor = zlib.decompressobj(zlib.MAX_WBITS | 16)
+        self.state = GzipDecoderState.FIRST_MEMBER
```
tomchristie, line 77: I'm wondering if there's an implementation using something like...

```python
self._buffer = io.BytesIO()
self._reader = gzip.GzipFile(fileobj=self._buffer, mode='r')
```

Would using `GzipFile` be more robust here / catch other cases?

lizeyan: It seems that zlib is slightly faster than gzip

```python
import gzip
import zlib
import time
import random
import string

def generate_random_data(size):
    return ''.join(random.choices(string.ascii_letters + string.digits, k=size)).encode()

def test_gzip_decompress(data, iterations):
    compressed = gzip.compress(data)
    start_time = time.time()
    for _ in range(iterations):
        gzip.decompress(compressed)
    end_time = time.time()
    return end_time - start_time

def test_zlib_decompress(data, iterations):
    compressed = zlib.compress(data)
    start_time = time.time()
    for _ in range(iterations):
        zlib.decompress(compressed)
    end_time = time.time()
    return end_time - start_time

# 测试参数
data_sizes = [1000, 10000, 100000]
iterations = 10000

print("Testing gzip and zlib decompression performance:")
print("------------------------------------------------")

for size in data_sizes:
    print(f"\nTesting with data size: {size} bytes")
    data = generate_random_data(size)
    
    gzip_time = test_gzip_decompress(data, iterations)
    zlib_time = test_zlib_decompress(data, iterations)
    
    print(f"gzip decompression time: {gzip_time:.4f} seconds")
    print(f"zlib decompression time: {zlib_time:.4f} seconds")
    
    if gzip_time < zlib_time:
        print(f"gzip is faster by {(zlib_time / gzip_time - 1) * 100:.2f}%")
    else:
        print(f"zlib is faster by {(gzip_time / zlib_time - 1) * 100:.2f}%")

print("\nTest completed.")
```

```
Testing gzip and zlib decompression performance:
------------------------------------------------

Testing with data size: 1000 bytes
gzip decompression time: 0.0503 seconds
zlib decompression time: 0.0386 seconds
zlib is faster by 30.47%

Testing with data size: 10000 bytes
gzip decompression time: 0.2463 seconds
zlib decompression time: 0.2337 seconds
zlib is faster by 5.38%

Testing with data size: 100000 bytes
gzip decompression time: 2.3218 seconds
zlib decompression time: 2.3094 seconds
zlib is faster by 0.54%

Test completed.
```

tests/test_decoders.py
lizeyan: I changed the style of the test case

tomchristie:
Thanks for your time looking into this.

We'll prefer following Chrome/Safari behavior here... https://github.com/encode/httpx/discussions/3269#discussioncomment-10676315

rafalkrupinski:
> We'll prefer following Chrome/Safari behavior here.

seriously, why?

tomchristie:
Because introducing complexity to support a use-case that browsers don't support is a trade-off I'd rather not make.

(More broadly speaking... `httpx` should prefer a design aesthetic of minimal complexity, where possible.)

PR #3250, opened 2024-07-25
Improve InvalidURL error message.

tomchristie:
Prompted by https://github.com/encode/httpx/discussions/3248.

*before*...

```python
>>> import httpx
>>> httpx.get("https://www.example.com\n")
Traceback (most recent call last):
  ...
  File "/Users/tomchristie/GitHub/encode/httpx/httpx/_urlparse.py", line 163, in urlparse
    raise InvalidURL(error)
httpx.InvalidURL: Invalid non-printable ASCII character in URL.
```

*after*...

```python
>>> import httpx
>>> httpx.get("https://www.example.com\n")
Traceback (most recent call last):
  ...
  File "/Users/tomchristie/GitHub/encode/httpx/httpx/_urlparse.py", line 168, in urlparse
    raise InvalidURL(error)
httpx.InvalidURL: Invalid non-printable ASCII character in URL, '\n' at position 23.
```

PR #3275, opened 2024-08-21
Version 0.27.1

mbeijen:
As suggested in https://github.com/encode/httpx/pull/3139#issuecomment-2299677531

## Added

* Support for `zstd` content decoding using the python `zstandard` package is added. Installable using `httpx[zstd]`. (#3139)

### Fixed

* Improved error messaging for `InvalidURL` exceptions. (#3250)
* Fix `app` type signature in `ASGITransport`. (#3109)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## 0.27.1 (22st August, 2024)
 
 ## Added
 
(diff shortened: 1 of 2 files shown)


tomchristie:
Thanks. ☺️

Lets use a minor release bump here. (No removals or deprecations)

mbeijen:
> Lets use a minor release bump here. (No removals or deprecations)

You mean to bump the last digit, which makes the version 0.27.1 instead of 0.28.0? (in 'semantic versioning', the last digit would be the patch level and the middle digit would be the minor level, ref https://semver.org/ - but I guess you're not using semver)

I've just updated the PR so now it's 0.27.1  - if there is anything else, please let me know

T-256 (approved):

tomchristie:
> I guess you're not using semver

No and yes, in the lead up to a 1.0.

tomchristie (approved):

tomchristie:
@encode/maintainers Could I get a review on this draft release https://github.com/encode/httpx/releases/tag/untagged-1f3c9bdb863e201ab647 - are we good to publish?

PR #3280, opened 2024-08-25
chore(deps): Update dependencies

reneleonhardt:
# Summary

### 📦 Chores
* Update dependencies
* Run Dependabot weekly

### Notes
* Please check why Dependabot is not updating requirements.txt, maybe disabled or a bug?
* Python 3.13 will be finished soon, can you add it to your test-suite to prepare for this big release? 😅

# Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "monthly"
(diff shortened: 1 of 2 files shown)


tomchristie:
Looks to me like @dependabot is running okay... https://github.com/encode/httpx/commits/master/requirements.txt

Let's stay with the monthly schedule. Don't really want to be getting spammed.

PR #3281, opened 2024-08-26
Add check_ascii_printable function to reduce duplicated code

waketzheng:
# Summary

Add check_ascii_printable function to reduce duplicated code.

# Checklist


T-256 (approved):
Thank you!

PR #3288, opened 2024-08-27
Reintroduce `URLTypes`.

tomchristie:
See https://github.com/encode/httpx/discussions/3287.

I guess we want to roll a minor point release based on this.


zanieb (approved):
Makes sense

PR #3294, opened 2024-09-03
add `stacklevel=2` to Client warnings

trim21:
# Summary

# Checklist

see this rule: https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

without stacklevel=2, warnings show on lines in httpx module, not in user code, where `Client` is created


trim21:
I also recommend httpx enable this ruff rule

tomchristie:
Thanks for this. Now obsoleted both of these occurrences in `version-1.0` branch.

PR #3296, opened 2024-09-04
HTTPTransport: Pass along http2 bool to `create_ssl_context(..., http2=http2)`

ar45:
# Summary
Client(http2=True) should pass the truth value to create_ssl_context

HTTPTransport: Pass along http2 bool to `create_ssl_context(..., http2=http2)`

### DIFF
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -136,7 +136,9 @@ class HTTPTransport(BaseTransport):
         retries: int = 0,
         socket_options: typing.Iterable[SOCKET_OPTION] | None = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


karpetrosyan:
`http2` attribute will be removed with [this pr](https://github.com/encode/httpx/pull/3022/files#diff-c73be11bcd157d39ddbd8afeeac32c7d503a67a5a8603008247996f2bafaa685), because all the http2 configurations are already handled by [httpcore](https://www.encode.io/httpcore/)

karpetrosyan:
I think it's not a bugfix, we are removing it because it's not used anywhere. All that configurations are handled by httpcore

PR #3300, opened 2024-09-11
Add support for custom JSON encoders/decoders.

AndreCimander:
# Summary
Added customizable JSON encoder/decoder.

Working example (also provided in docs):

```python
import httpx
import msgspec
import typing


encoder = msgspec.json.Encoder()
def custom_json_encoder(json_data: typing.Any) -> bytes:
    return encoder.encode(json_data)

httpx.register_json_encoder(custom_json_encoder)


decoder = msgspec.json.Decoder()
def custom_json_decoder(json_data: bytes, **kwargs: Any) -> bytes:
    return decoder.decode(json_data)

httpx.register_json_decoder(custom_json_decoder)
```

# Checklist

### DIFF
--- a/docs/advanced/json.md
+++ b/docs/advanced/json.md
@@ -1,0 +1,27 @@
+# JSON encoding/decoding
+
+You can set a custom json encoder and decoder, e.g. if you want to use `msgspec` instead or python's `json` implementation.
+
+Your custom function needs to return `bytes`. 
+
+
+**Custom msgspec implementation:**
+```python
+import httpx
+import msgspec
+import typing
+
+
+encoder = msgspec.json.Encoder()
+def custom_json_encoder(json_data: typing.Any) -> bytes:
+    return encoder.encode(json_data)
+
+httpx.register_json_encoder(custom_json_encoder)
+
+
+decoder = msgspec.json.Decoder()
+def custom_json_decoder(json_data: bytes, **kwargs) -> bytes:
+    return decoder.decode(json_data)
+
+httpx.register_json_decoder(custom_json_decoder)
+```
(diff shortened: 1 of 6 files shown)


AndreCimander:
It's getting late over here, still need to figure out how to satisfy mypy with Protocol usage, as `mypy_extension.KwArg` and `mypy_extenions.Arg` are deprecated and I don't want to import those (even with TYPE_CHECKING).

AndreCimander:
After wrangling with correct type definitions for **kwargs, I concede defeat and decided to use `# type: ignore`.

If there is a better solution for this I'd would be more than happy to learn, this is a little frustrating :)

tomchristie:
Appreciate that there's clearly demand for something like this, tho have been clear that a global registry approach isn't going to be accepted. Can we close this off, and continue pushing for design-discussion-first approachs?

AndreCimander:
Ah, sorry, haven't seen that comment, should have searched more diligently.

Bummer, well, yeah, I can understand your point. But I really would have favored a global registry call in my projects :)  

A customized Client solution is fine as well (but it feels a bit off needing to customize 3 classes just for a simple json callable change).

I'll close this PR then.

tomchristie:
Thanks @AndreCimander this one is proving a bit of a thorny issue.

gtors:
> A customized Client solution is fine as well (but it feels a bit off needing to customize 3 classes just for a simple json callable change).


@AndreCimander  Try httpj 😎  It's my fork with similar functionality:

```python
import httpj
import msgspec
import typing


encoder = msgspec.json.Encoder()
def custom_json_encoder(json_data: typing.Any) -> bytes:
    return encoder.encode(json_data)

decoder = msgspec.json.Decoder()
def custom_json_decoder(json_data: bytes, **kwargs: Any) -> bytes:
    return decoder.decode(json_data)

client = httpj.Client(json_serialize=custom_json_encode, json_deserialize=custom_json_decoder)
```

Issue #3301, opened 2024-09-12
_load_client_certs() should raise an error if `cert` could not be handled.


zopyx:
The code in `_load_client_certs`  does not handle a cert path of type `pathlib.Path`:

```
    def _load_client_certs(self, ssl_context: ssl.SSLContext) -> None:
        """
        Loads client certificates into our SSLContext object
        """
        if self.cert is not None:
            if isinstance(self.cert, str):
                ssl_context.load_cert_chain(certfile=self.cert)
            elif isinstance(self.cert, tuple) and len(self.cert) == 2:
                ssl_context.load_cert_chain(certfile=self.cert[0], keyfile=self.cert[1])
            elif isinstance(self.cert, tuple) and len(self.cert) == 3:
                ssl_context.load_cert_chain(
                    certfile=self.cert[0],
                    keyfile=self.cert[1],
                    password=self.cert[2],
                )
```

Rather than silently discarding an unhandled type for `self.cert` here, it would be better to raise an exception (e.g. `NotImplementedError` for  an unsupported type for cert.

Background: we switched our code from `str` to `pydantic.FilePath` and suddenly the authentication part via a PEM did no longer work..and it took a long time to figure out that this was caused by the unhandled case here.  An exception is more explicit than implicitly ignoring the problem (without reporting it).

tomchristie:
Okay. Would you like to issue a pull request that'd raise a TypeError for that case?

zopyx:
Happy to work on PR. Do you have a quick pointer where a related test should be added? I am not familar with the httpx codebase.

tomchristie:
👉 https://github.com/encode/httpx/blob/master/tests/test_config.py

tomchristie:
Obsoleted by #3022 and #3335.

PR #3304, opened 2024-09-14
Fix for #3301: raise an explicit TypeError if `cert` type is unsupported

zopyx:
Fix for 

https://github.com/encode/httpx/issues/3301


tomchristie:
Thanks @zopyx - this will be obsoleted by #3335

PR #3312, opened 2024-09-23
Clarified error when header value is None

nomilkinmyhome:
Previously, when the header value was None, a misleading error was thrown: "AttributeError: Object 'NoneType' does not have attribute 'encode'." Now it looks better.

image


.gitignore
```diff
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+.idea
```
zanieb, line 13: This probably belongs in your user-level `.gitignore` — there are no editor-specific ignore entries here.

zanieb (approved):
Thanks this seems reasonable to me, the error message is much more friendly. I want to make sure another maintainer signs-off that we want to have these kinds of checks in general though.

.gitignore
zanieb:
```suggestion
```

tests/test_utils.py
```diff
@@ -300,6 +301,24 @@ def test_url_matches(pattern, url, expected):
...
+)
+def test_normalize_header_value(value, expected):
+    assert normalize_header_value(value) == expected
+
+
+def test_normalize_header_incorrect_value():
+    with pytest.raises(
+        TypeError, match=f"Header value must be str or bytes, not {type(None)}"
+    ):
+        normalize_header_value(None)  # type: ignore
```
tomchristie, line 319: Could we have these tests in `tests/client/test_headers.py` with the testing against `httpx.Header(...)` instead of against the internal API?

tomchristie (changes requested):
Brilliant, thanks. I've requested a change to the tests, let me know if needs any more explanation.

(We should always be testing against public API. There's a couple of places left to clean up, tho we're almost there now)

tomchristie (approved):
👌

PR #3327, opened 2024-09-27
Add `httpdbg` to third party packages.

T-256:
# Summary
I think it deserve to mention...


karpetrosyan:
Yep, hiding user credentials at the lowest layer and preventing them from being passed higher makes perfect sense.

Now it's exposed in URL.__str__, but not in URL.__repr__, which is a bit weird. __str__ is supposed to show more user-related data, while __repr__ is more for debugging and development. So, hiding it in __repr__ would make more sense. However, I think preventing it entirely is the most secure way.

karpetrosyan (approved):

PR #3331, opened 2024-10-02
Fix client response stalling, exception handling, and disconnects in ASGI Transport

### LINKED ISSUE #2186: Testing endless SSE endpoint with AsyncClient hangs forever

AviNoah:
ASGI transport layer would stall indefinitely due to improper stop signaling of the send stream, now close them when client closes response or stream is exhausted.
It also would not raise exceptions correctly from run_app.

This happens because the receive stream expects to receive more values after send stream stopped producing values, and the async for loop stalls - leading the client to wait for the response content indefinitely.
To amend this, we now close the send stream after more body is False.

Whenever the response is closed, cancel the run_app task if still running, and aclose the send and receive memory streams.

in the case of an exception from run_app, raise it, and then also aclose the send and receive memory streams.

Added tests for exhausting the stream, interrupting it and disconnecting the response, and interrupting it without disconnecting the response and continuing to consume it

# Checklist

Fixes #2186

### DIFF
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -174,18 +174,11 @@ async def run_asgi(
     response_complete = anyio.Event()
 
     send_stream, receive_stream = anyio.create_memory_object_stream()
-    disconnected = anyio.Event()
 
-    async def watch_disconnect(cancel_scope: anyio.CancelScope) -> None:
-        await disconnected.wait()
-        cancel_scope.cancel()
-
-    async def run_app(cancel_scope: anyio.CancelScope) -> None:
+    async def run_app() -> None:
         try:
             await app(scope, receive, send)
         except Exception:  # noqa: PIE-786
-            cancel_scope.cancel()
-
             if raise_app_exceptions or not response_complete.is_set():
                 raise
 
@@ -207,12 +200,6 @@ async def run_asgi(
 
     async def send(message: _Message) -> None:
         nonlocal status_code, response_headers
-
-        if disconnected.is_set():
-            await anyio.sleep(
-                0
-            )  # temp fix - yield control for disconnect watcher when loop is tight
-            return
 
         if message["type"] == "http.response.start":
             assert not response_started.is_set()
@@ -236,8 +223,7 @@ async def run_asgi(
 
     try:
         async with anyio.create_task_group() as tg:
-            tg.start_soon(watch_disconnect, tg.cancel_scope)
-            tg.start_soon(run_app, tg.cancel_scope)
+            tg.start_soon(run_app)
 
             await response_started.wait()
             assert status_code is not None
@@ -248,9 +234,11 @@ async def run_asgi(
                     yield chunk
 
             yield (status_code, response_headers, stream())
-            disconnected.set()
+            # Once the yielded value stops being used by the client cancel, cancel tasks
+            tg.cancel_scope.cancel()
     except ExceptionGroup as exc_group:
         raise exc_group.exceptions[0]  # only run_app should raise exceptions
     finally:
+        # Make sure memory streams are closed
         await send_stream.aclose()
         await receive_stream.aclose()
--- a/httpx/_transports/asgi.py
+++ b/httpx/_transports/asgi.py
@@ -187,7 +187,7 @@ async def run_asgi(
             await app(scope, receive, send)
         except Exception:  # noqa: PIE-786
             cancel_scope.cancel()
-            
+
             if raise_app_exceptions or not response_complete.is_set():
                 raise
 
@@ -211,6 +211,9 @@ async def run_asgi(
         nonlocal status_code, response_headers
 
         if disconnected.is_set():
+            await anyio.sleep(
+                0
+            )  # temp fix - yield control for disconnect watcher when loop is tight
             return
 
         if message["type"] == "http.response.start":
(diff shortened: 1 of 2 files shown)


tomchristie:
> ASGI transport layer would stall indefinitely due to improper stop signaling of the send stream

Okay, walk me through this... can you start by describing the potential issue, so the team can understand this more clearly...

(Team: we might need to be stricter about triage via discussion -> issue -> pr process. I really do feel it helps keep the project properly on track. Partly thinking aloud here.)

AviNoah:
Hey, thanks for taking the time to review this.
Let me explain, I wanted to test my ASGI app using the ASGI transport layer and noticed the one on the master branch simply awaits for the generator to stop yielding, saves it into a list of bytes, and then returns it all at once, this obviously defeats the purpose of streaming as it requires waiting until the response content finishes producing before you can even access the header or the request status - something I needed as soon as the request completed in my tests.

I had looked in the branches of the repository and noticed a branch that tried addressing the issue I was facing, when I ran it in the included tests I noticed it stalled seemingly forever, I decided to prod this a bit.

I noticed there weren't any indicator for the receive stream to stop expecting values and therefore exit the async for loop, which if a generator needs to be fully consumed by the client, will leave the client waiting indefinitely since there is never a stop signal, furthermore I noticed the disconnect flag is simply redundant as all we care for is whether the stream was fully consumed or the connection was disconnected (continuing from the yield point), which will then cancel the app task, and finally a close the memory streams to prevent leaks and more errors.

After i had done this I noticed the raise exc tests were still failing, and the problem was the fact I used TaskGroup since it returns an ExceptionGroup, since I run only run_app, there should always be only one exception which is directly what was raised from run_app.

AviNoah:
@tomchristie I edited in the issue tag for this problem

tomchristie:
>  noticed the one on the master branch simply awaits for the generator to stop yielding, saves it into a list of bytes, and then returns it all at once, this obviously defeats the purpose of streaming

Yep that's correct, `ASGITransport` does not (currently?) support streaming... https://github.com/encode/httpx/issues/2186

Thanks for the reminder there... I'd suggest the first thing for us to do would be open an issue to ensure that this limitation is documented. See https://www.python-httpx.org/advanced/transports/

AviNoah:
@tomchristie But there is an issue for that already, its #2186 , my pull request attempts to resolve that exact problem

tomchristie:
Thanks for your time here, @AviNoah. I'm going to close this one off.

Before accepting PRs dealing with "support streaming downloads with ASGITransport" I'm going suggest the first thing for us to do would be open an issue to ensure that this limitation is documented...

See https://www.python-httpx.org/advanced/transports/

We may or may not want to then address that limitation.

AviNoah:
@tomchristie I started a discussion about it at #3391

AviNoah:
> Thanks for your time here, @AviNoah. I'm going to close this one off.
> 
> Before accepting PRs dealing with "support streaming downloads with ASGITransport" I'm going suggest the first thing for us to do would be open an issue to ensure that this limitation is documented...
> 
> See https://www.python-httpx.org/advanced/transports/
> 
> We may or may not want to then address that limitation.

I have noticed my discussion is not gaining any attention at #3391 for a long while now
should I change something? is there anything I need to update to be more specific or something?

PR #3336, opened 2024-10-08
Improve docs.

T-256:
Ref: https://github.com/encode/httpx/pull/3022#discussion_r1787967524

# Summary
- Removed `docs/environment_variables.md` and moved its contents to `docs/transports.md` under _Environment Variables_ section.
- Some touchups to `docs/compability.md`.

# Checklist


zanieb:
I'm not a big fan of the moved environment variable docs. It's nice to have a document that enumerates the environment variables respected by the library. I think would have a hard time finding this in the "Transports" section as a user.

zanieb:
Ah helpful to link that context. It's Tom's choice of course. We've gotten a fair amount of positive feedback about a [dedicated environment variable page](https://docs.astral.sh/uv/configuration/environment/) in uv.

PR #3337, opened 2024-10-08
Cleanup `trust_env` as property of Client.

T-256:
# Summary
No need to have `trust_env` as property since it only used on initialization. I added new test for `trust_env` behavior.


zanieb:
Is there any reason someone would want to check if the client has this option set?

PR #3340, opened 2024-10-09
Default ssl contexts per http versions.

T-256:
# Summary
Ref: https://github.com/encode/httpx/pull/3022#pullrequestreview-2350615169

Instead of creating ssl context per client, we now re-use available default contexts for different http versions configurations.

### TODO
- [ ] add logging
- [ ] add tests
- [ ] make it thread safe


tomchristie:
Heya. Gonna suggest we have a design discussion before considering this an issue..

PR #3345, opened 2024-10-10
Cleanup `response.elapsed` implementation.

tomchristie:
More internal refactorings, keeping that codebase clean & sharp.


tomchristie:
@T-256 That's a valid question, tho let's keep this particular PR focused on this particular piece of internal refactoring.

PR #3353, opened 2024-10-21
Simplify and optimize the most common methods of 'Headers'

princerb:
This PR enhances the performance of the modified functions by significantly increasing their execution speed x100000. I used generators of course and I can add more of these kinds in the future if you like!
The current test (`tests/models/test_headers.py`) can handle this change so I didn't write any.

**Before:**
```bash
$ python -m timeit -r 10 -n 10 -c "from main import h" "h.items()"
10 loops, best of 10: 297 msec per loop
```

**After:**
```bash
$ python -m timeit -r 10 -n 10 -c "from main import h" "h.items()"
10 loops, best of 10: 660 nsec per loop
:0: UserWarning: The test results are likely unreliable. The worst time (4.01 msec) was more than four times slower than the best time (660 nsec).
```
_If these kind of PRs can be merged, I would like to contribute with like these optimised solutions again to speed up the lib. Or feel free to give a feedback)_

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -128,39 +128,37 @@ class Headers(typing.MutableMapping[str, str]):
         return [(raw_key, value) for raw_key, _, value in self._list]
 
     def keys(self) -> typing.KeysView[str]:
-        return {key.decode(self.encoding): None for _, key, value in self._list}.keys()
+        stream = set()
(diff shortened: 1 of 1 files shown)


princerb:
So the tests, I guess then although generators are faster, they are not for good presentation(

PR #3357, opened 2024-10-22
Global timeout

andybbruno:
# Summary
This pull request introduces a new, more straightforward way to configure a global default timeout for all clients and requests in `httpx` without needing to access the internal `_config` object. The enhancement is achieved by exposing the `DEFAULT_TIMEOUT_CONFIG` in the main `__init__.py` file (see discussion [here](https://github.com/encode/httpx/discussions/3356)).

This change simplifies the process, allowing developers to set a global timeout directly and elegantly, like so:
```python
import httpx
httpx.DEFAULT_TIMEOUT_CONFIG = httpx.Timeout(10.0)
```

# Checklist

### DIFF
--- a/docs/advanced/timeouts.md
+++ b/docs/advanced/timeouts.md
@@ -38,6 +38,15 @@ client = httpx.Client(timeout=10.0)  # Use a default 10s timeout everywhere.
 client = httpx.Client(timeout=None)  # Disable all timeouts by default.
 ```
 
+## Setting a global timeout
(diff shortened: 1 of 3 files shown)


tomchristie:
Thanks @andybbruno - do appreciate your time & interest on this.

The proposal (exposing a global constant as public API) doesn't really match the coding style in other parts of the project,
and I don't think we reached a resolution on #3356. We'd need a clearer consensus before accepting PRs in this direction.

PR #3359, opened 2024-10-23
typo: Reading a response expose response text, not request text

Colin-b:
# Summary

Typo in documentation about reading a response and updating a request field.

# Checklist


tomchristie (approved):
Thanks. ☺️

PR #3361, opened 2024-10-24
Docs minor fix

### LINKED ISSUE #3349: `mounts` kwarg to `AsyncClient` is not documented in the docstring
See

https://github.com/encode/httpx/blob/1bf1fc0ea8417c8e539914aeafdb48f15591ba75/httpx/_client.py#L641

Both on `main` and v0.27.2 `mounts` is not documented in the docstring above, unlike other arguments.

🙏

karpetrosyan:
Closes #3349 

Just a minor documentation fix referring to the non-existent `httpx.Mounts`

We also have some undocumented attributes and need to tidy up the API documentation. I believe this will be addressed within the scope of #3091


tomchristie:
Wonderful, thanks @karpetrosyan.

Issue #3363, opened 2024-10-24
Ensure JSON representation is compact.


BERRADA-Omar:
### Discussed in https://github.com/encode/httpx/discussions/3204

<div type='discussions-op-text'>

<sup>Originally posted by **alleml** May 17, 2024</sup>
httpx.Request blindly dumps provided json to ascii characters by invoking json.dumps without any parameters. Currently there is no way to provide `ensure_ascii=False` argument to the Request instance (or even better to the client).

It leads to the following scenario:
```
>>> httpx.Request(method="POST", url="http://localhost:8000/test", json={"field": "cześć"}).content
b'{"field": "cze\\u015b\\u0107"}'
````

when I need properly encoded unicode bytes: `b"cze\xc5\x9b\xc4\x87"`</div>

**Comment by BERRADA-Omar** [comment]
This bug have a huge impact on the openai package (1.52.0).
Openai package uses httpx to make it requests, and it turns out that when we make call in French, the GPT models receives text in the wrong encoding. The encoding should be UTF-8 ! 

For example : 

- input text in French : "salut ça va ? très bien ou pas bien ?"
- GPT receives in payload : "salut \u00e7a va ? tr\u00e8s bien ou pas bien ?"

This could have huge impact on text understanding by LLMs, and also could generate more costs for intesive llm users.

Please have this issue fixed.

**Comment by BERRADA-Omar** [comment]
For those who are looking for a workaround, you can override the method that causes the issue in the entry point of your app (main.py for example).

The patch would be : 

``` python
import httpx._content
from httpx import ByteStream
from json import dumps as json_dumps

def custom_httpx_encode_json(json: Any) -> tuple[dict[str, str], ByteStream]:
    # disable ascii for json_dumps 
    body = json_dumps(json, ensure_ascii=False).encode("utf-8") # <------------------------------------------------
    content_length = str(len(body))
    content_type = "application/json"
    headers = {"Content-Length": content_length, "Content-Type": content_type}
    return headers, ByteStream(body)

# fix encoding utf-8 bug
httpx._content.encode_json = custom_httpx_encode_json
```

tomchristie:
> Openai package uses httpx to make it requests, and it turns out that when we make call in French, the GPT models receives text in the wrong encoding.

We're compliant with [the JSON spec here](https://www.json.org/json-en.html), and parsers will be decoding text escape sequences into the correct unicode text. Tho you are correct that a more compact format would be sensible here. Would you like to submit a pull request resolving this?

[...]

PR #3367, opened 2024-10-26
Ensure JSON representation is compact. #3363

BERRADA-Omar:
# Summary

this PR solves the issues mentionned here #3363

> The following related issues could all be resolved together...
> 
> We should use ensure_ascii=False for more compact text representations.
> We should use separators = (',', ':') for more compact list and object representations.
> We should use allow_nan=False to disallow invalid Infinity and NaN representations.

# Checklist


.gitignore
```diff
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+.vscode/
```
tomchristie, line 13:
```suggestion
```

.gitignore
```diff
@@ -9,4 +9,4 @@ site/
 venv*/
 .python-version
 build/
-dist/
+dist/
```
tomchristie, line 12:
```suggestion
dist/

```

tomchristie (approved):
Yep, thx. ☺️

tests/test_content.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 
 import httpx
+from httpx._content import encode_json
```
tomchristie, line 7: I've just noticed this accidentally snuck in a private import.

@BERRADA-Omar would you be up for refactoring these tests?

We can test this against public API using `httpx.Request()`. Eg...

```python
data = {...}
req = httpx.Request("POST", "https://www.example.com/", json=data)
assert req.content == ...
assert req.headers = ...
```

Kludex:
This broke the test suite on Starlette. FYI

Kludex:
This broke the test suite on Starlette.

tests/test_content.py
BERRADA-Omar: Hello, i'll make the change.

BERRADA-Omar: @tomchristie i see that this is already fixed ?

tomchristie: Yes. Thanks, tho.

beliaev-maksim:
@tomchristie This PR not only adjusts the ASCII representation but also alters the default JSON separators. These changes lead to unnecessary modifications of JSON objects, resulting in altered validation hashes that broke some of our tests.

tomchristie:
Thoughts...

* CHANGELOG entry here [is insufficiently emphasised](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0280-28th-november-2024). Not an API change, but nonetheless changes to representation-on-the-wire *are* likely to negatively impact teams in the short run, as evidenced.
* Unclear how we should improve evaluation process of trade-offs such as this one.

beliaev-maksim:
@tomchristie More detailed CHANGELOG is a "must". While it won’t solve the technical issue, it will simplify the debugging.  

When moving forward a codebase, tough changes might be necessary, that is obvious. However, providing a grace period for users can ease the process. E.g., you could introduce a warning in the codebase and maintain it for two minor versions before enforcing the update.
As one of the options.

PR #3368, opened 2024-10-27
Lazy load httpcore, ssl, certifi

joemarshall:
This change makes httpcore, certifi and ssl only be loaded when the transport that uses it is actually instantiated.

As suggested in:
https://github.com/encode/httpx/pull/3330

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import logging
 import os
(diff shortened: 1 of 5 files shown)


joemarshall:
Hang on, I didn't rebase, will fix

tomchristie:
Thanks @joemarshall, would you be able to reissue this against the `master` branch.

PR #3371, opened 2024-10-28
Review URL percent escaping sets, from whatwg.

tomchristie:
Review and update which characters should be percent escaped when parsing & normalising URLs.
Based on the URL living standard... https://url.spec.whatwg.org/

Prompted by https://github.com/encode/httpx/discussions/3360

### DIFF
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -614,10 +614,10 @@ def test_url_copywith_userinfo_subcomponents():
     }
     url = httpx.URL("https://example.org")
     new = url.copy_with(**copy_with_kwargs)
-    assert str(new) == "https://tom%40example.org:abc123%40%20%25@example.org"
+    assert str(new) == "https://tom%40example.org:abc123%40%20%@example.org"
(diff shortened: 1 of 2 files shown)


tests/models/test_url.py
```diff
@@ -614,10 +614,10 @@ def test_url_copywith_userinfo_subcomponents():
     }
     url = httpx.URL("https://example.org")
     new = url.copy_with(**copy_with_kwargs)
-    assert str(new) == "https://tom%40example.org:abc123%40%20%25@example.org"
+    assert str(new) == "https://tom%40example.org:abc123%40%20%@example.org"
     assert new.username == "tom@example.org"
     assert new.password == "abc123@ %"
-    assert new.userinfo == b"tom%40example.org:abc123%40%20%25"
+    assert new.userinfo == b"tom%40example.org:abc123%40%20%"
```
elupus, line 620: This looks weird. I could be wrong. But should the userinfo field contain partial escapes? The trialing % sign change here seem odd.

PR #3377, opened 2024-10-28
made dependencies on certifi and httpcore only load when required

joemarshall:
As per #3330 this makes certifi and httpcore only load when required. 

For now I haven't done the same with ssl, because SSLContext derives from ssl, which makes it a pain to load on demand. I can make that behind an ImportError guard, with a dummy SSLContext class, but it is a bit hacky. This is what asyncio does. The alterative is dynamically reloading the class and importing ssl on first instantiation (or overriding `__new__` maybe), but again, it is a bit of a hack.

Personally I think there isn't a massive overhead to just making ssl a dependency even on platforms where it isn't being used, but I'm easy to go with the dummy class or something if you don't want that; I just worry that code may assume that the dummy SSLContext is a class derived from ssl.SSLContext and bad things will occur.

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -6,8 +6,6 @@ import sys
 import typing
 import warnings
 
-import certifi
-
 from ._models import Headers
 from ._types import HeaderTypes, TimeoutTypes
 from ._urls import URL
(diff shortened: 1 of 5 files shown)


joemarshall:
@tomchristie here's the change to lazy load httpcore and certifi

tomchristie (approved):
😎

PR #3378, opened 2024-10-29
Cleanup `Request` method parameter.

tomchristie:
The `Request()` class currently has a `method` parameter that's out of sync with `client.request(...)` and `httpx.request(...)`. This is nominally an API change, tho I think it's *actually* a bugfix in practice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Version 0.28.0
+## [Unreleased]
 
-Version 0.28.0 introduces an `httpx.SSLContext()` class and `ssl_context` parameter.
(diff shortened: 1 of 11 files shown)


karpetrosyan (approved):

karpetrosyan:
Oh, it seems this change was breaking and [affected the compatibility of another library with httpx](https://github.com/karpetrosyan/hishel/actions/runs/12241440026/job/34146523039). It looks like we should take changes that require a major release a bit more seriously.

This kind of change breaks specific versions of other projects, without giving them any chance to avoid it :c

tomchristie:
Thanks @karpetrosyan, good to see type checker checking types.

[Hishel](https://hishel.com/) is looking really smart. Most obvious action points...

* Reintroduce a clear "pinning" note on the `httpx` README and homepage. This might either recommend `==0.28` or `<1.0`.
* Do whatever we can this side communications-wise to help ensure our dependencies are neatly pinned.
* Ensure future updates have a clear transition approach. (pre-releases might be appropriate.)

Issue #3381, opened 2024-10-29
Cleanup `_utils.py` functions. ✨🧹✨


tomchristie:
Possibly a good issue for new contributors?...

The `_utils.py` module has a smorgasbord of utility functions. Many of these are only used in one place in `httpx` and it would be easier to comprehend if they were inlined, or refactored closer to where they're being used. ✨🧹✨

Pull requests for this should typically tackle one item at a time for easier review. 🤓

RafaelWO:
I'm happy to take a look at this 🙂 

> Pull requests for this should typically tackle one item at a time for easier review

I'm not fully sure if I understand this correctly. Do you mean per function or per "topic"?

tomchristie:
@RafaelWO Wonderful, thanks. 😊

> Do you mean per function or per "topic"?

Use your own judgement I guess, we can have a conversation if there's any awkward points.
As a guideline, simpler PRs tend to be more easily acceptable.

RafaelWO:
@tomchristie I think we can close this as completed, correct? :slightly_smiling_face:

tomchristie:
Complete enough for now, yep. ☺️

PR #3382, opened 2024-10-30
Move normalize header functions from _utils.py to _models.py

AstralScribe:
# Summary

Issue: #3381

This PR refactors  the `normalize_header_key` as an inline method and moves `normalize_header_value` function to _models.py from _utils.py.   

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -54,6 +54,13 @@ from ._utils import (
 __all__ = ["Cookies", "Headers", "Request", "Response"]
 
 
+def _normalize_header_key(key: str | bytes, encoding: str | None = None) -> bytes:
+    """
+    Coerce str/bytes into a strictly byte-wise HTTP header key.
+    """
+    return key if isinstance(key, bytes) else key.encode(encoding or "ascii")
+
+
 def _normalize_header_value(value: str | bytes, encoding: str | None = None) -> bytes:
     """
     Coerce str/bytes into a strictly byte-wise HTTP header value.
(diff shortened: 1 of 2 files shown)


tomchristie:
Thanks. ☺️

I'd suggest we should stay consistent and use `_normalize_header_key()` and `_normalize_header_value()` here.
There's perhaps some more cleanup we could do in this particular case, but that's probably the best starting point.

AstralScribe:
I have converted the inline method into a similar function.

AstralScribe:
I have done some more refactoring, but I am not sure if this simplifies or complicates it. I had to add some things due to mypy type checking.

```
def _convert_mapping_to_sequence(
    mappings: typing.Union[typing.Mapping[str, str], typing.Mapping[bytes, bytes]],
) -> typing.Sequence[typing.Tuple[str | bytes, str | bytes]]:
    return [(k, v) for k, v in mappings.items()]


def _normalize_headers(
    headers: HeaderTypes | None, encoding: str | None
) -> list[tuple[bytes, bytes, bytes]]:
    """
    Coerce str/bytes into a strictly byte-wise HTTP headers.
    """
    if headers is None:
        return []

    if isinstance(headers, Headers):
        return list(headers._list)

    normalized_list: list[tuple[bytes, bytes, bytes]] = []

    if isinstance(headers, typing.Mapping):
        n_headers = _convert_mapping_to_sequence(headers)
    elif isinstance(headers, typing.Sequence):
        n_headers = headers

    for key, value in n_headers:
        if not isinstance(value, (str, bytes)):
            raise TypeError(f"Header value must be str or bytes, not {type(value)}")
        b_key = key if isinstance(key, bytes) else key.encode(encoding or "ascii")
        b_value = (
            value if isinstance(value, bytes) else value.encode(encoding or "ascii")
        )
        normalized_list.append((b_key, b_key.lower(), b_value))

    return normalized_list
```

It replaces the whole init with:
```
self._list = _normalize_headers(headers, encoding)
```

tomchristie (approved):
Exactly like this, yep. 👍
Just helps keep things neat & localised.

tomchristie:
Thanks @AstralScribe. 😊

> I have done some more refactoring, but I am not sure if this simplifies or complicates it. I had to add some things due to mypy type checking.

Let's take a pass on that particular follow-up. What you've got here is great as it is.

PR #3385, opened 2024-10-30
Make it work without ssl

joemarshall:
As per #3330  here is a change to make the core library code work if the ssl library is not importable (assuming the transport being used doesn't require it obviously). This will make emscripten support more download / load efficient, because it won't have to depend on ssl.


joemarshall:
I think this doesn't really need to be here any more - ssl is much more isolated in master anyway now.

PR #3387, opened 2024-11-01
Move remaining utility functions from _utils.py to _models.py 

RafaelWO:
# Summary
Related issue: #3381

This PR moves the following remaining utility functions from `_utils.py` to `_models.py` that are only used in `_models.py` (except `to_bytes_or_str`):
* `is_known_encoding`
* `parse_header_links`
* `parse_content_type_charset`
* `obfuscate_sensitive_headers`

Related tests are also moved accordingly, i.e. from `tests/test_utils.py` to `tests/models/test_headers.py`.

---

Additionally, this PR fixes the script `scripts/lint` by using `ruff check` due to the following error:
```
error: `ruff <path>` has been removed. Use `ruff check <path>` instead.
```

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import codecs
 import datetime
 import email.message
 import json as jsonlib
+import re
 import typing
 import urllib.request
 from collections.abc import Mapping
@@ -44,14 +46,22 @@ from ._types import (
     SyncByteStream,
 )
 from ._urls import URL
-from ._utils import (
-    is_known_encoding,
-    obfuscate_sensitive_headers,
-    parse_content_type_charset,
-    parse_header_links,
-)
+from ._utils import to_bytes_or_str, to_str
 
 __all__ = ["Cookies", "Headers", "Request", "Response"]
+
+SENSITIVE_HEADERS = {"authorization", "proxy-authorization"}
+
+
+def _is_known_encoding(encoding: str) -> bool:
+    """
+    Return `True` if `encoding` is a known codec.
+    """
+    try:
+        codecs.lookup(encoding)
+    except LookupError:
+        return False
+    return True
 
 
 def _normalize_header_key(key: str | bytes, encoding: str | None = None) -> bytes:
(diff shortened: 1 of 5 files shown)


tomchristie (approved):

RafaelWO:
@tomchristie I first overlooked that I should also move some of the related tests. I did this now 🙂

tests/test_utils.py
RafaelWO, line 1: I'm note sure if `test_encoded`, `test_bad_utf_like_encoding`, and `test_guess_by_bom` are related to the moved function `is_known_encoding` and therefore should be moved into tests related to models?

httpx/_models.py
```diff
@@ -72,6 +82,60 @@ def _normalize_header_value(value: str | bytes, encoding: str | None = None) ->
...
+            try:
+                key, value = param.split("=")
+            except ValueError:
+                break
+            link[key.strip(replace_chars)] = value.strip(replace_chars)
+        links.append(link)
+    return links
+
+
+def _obfuscate_sensitive_headers(
```
tomchristie, line 130: Is now a good time to review this? Eg does flask provide similar behaviour onto its headers data structures? Does Django?

(Possibly too off topic??)

tomchristie (approved):
Thanks. ☺️ Prompted one review thought, tho let's get this in.

httpx/_models.py
RafaelWO: It seems that flask does not obfuscate header values for `repr` (see [werkzeug](https://github.com/pallets/werkzeug/blob/357681fc26dd46ec5b7c3c067732873f57db0bc8/src/werkzeug/datastructures/headers.py#L20))

```python
>>> from werkzeug.datastructures import Headers
>>>
>>> head = Headers({"authorization": "s3kr3t"})
>>> head
Headers([('authorization', 's3kr3t')])
>>> repr(head)
"Headers([('authorization', 's3kr3t')])"
```

Same for Django (see [HttpHeaders ](https://github.com/django/django/blob/611bf6c2e2a1b4ab93273980c45150c099ab146d/django/http/request.py#L468) or [ResponseHeaders](https://github.com/django/django/blob/611bf6c2e2a1b4ab93273980c45150c099ab146d/django/http/response.py#L33))

```python
>>> from django.http.response import ResponseHeaders
>>> head = ResponseHeaders({"Authorization": "s3kr3t"})
>>> repr(head)
"{'Authorization': 's3kr3t'}"
```

PR #3389, opened 2024-11-01
Move utility functions from _utils.py to _client.py

RafaelWO:
# Summary
Related issue: #3381

This PR moves the following remaining utility functions from `_utils.py` to `_client.py` that are only used in `_client.py`:
* `port_or_default`
* `same_origin`
* `is_https_redirect`

Related tests are also moved accordingly (to `tests/client/test_headers.py`)

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -46,12 +46,7 @@ from ._types import (
     TimeoutTypes,
 )
 from ._urls import URL, QueryParams
-from ._utils import (
-    URLPattern,
-    get_environment_proxies,
-    is_https_redirect,
-    same_origin,
-)
+from ._utils import URLPattern, get_environment_proxies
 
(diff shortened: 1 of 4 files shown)


RafaelWO:
I am not sure whether we should also move the class `URLPattern` - it is only used in `_client.py`. Maybe it should go into its own file or `_urls.py`?

tomchristie:
Ah this is an interesting one.

I'm going to suggest leaving the following in `_utils.py` for the moment...

* `URLPattern`
* `get_environment_proxies`
* `is_ipv4_hostname`
* `is_ipv6_hostname`

We'll likely be able to remove them entirely due to a different piece of incoming refactoring, tho will treat that seperately.

RafaelWO:
@tomchristie, I updated the PR accordingly 🙂

tomchristie (approved):
Thanks, yes. Neatly done. ☺️

Issue #3396, opened 2024-11-08
Allow Multipart Requests Without Attaching Files


laipz8200:
Currently, a multipart request can only be sent when the files parameter is non-empty. This restriction limits cases where users might want to send data using multipart mode without attaching any files, which is possible in tools like Postman.

Proposed Change:

Instead of requiring files to be non-empty, we could determine the need for a multipart request based on whether files is None rather than its emptiness. This would allow users to send multipart requests without attaching files if needed.

Example Use Case:

- Sending form data using multipart mode without the need for a file upload.

Additional Context:

This change has been discussed and approved in the Q&A section with a maintainer. I’d be happy to work on this change.

I'd like to try to implement this feature, please assign this issue to me.

---

- [x] Initially raised as discussion ##3395

sylann:
There is a way to solve this problem with current versions of httpx.
The documentation does not exactly demonstrate this use case, but it still gives the hint.

https://www.python-httpx.org/quickstart/#sending-multipart-file-uploads

```py
multipart_form_data = {
    "field1": (None, "data"),
    "field2": (None, "data"),
}
http.post("your-url", files=multipart_form_data)
```

None corresponds to the "filename" of the field, which does not exist in this case.

Or, supposing you already have some dict (be careful though, it must be `dict[str, str | bytes]`),
You could convert it to a proper format like so:
```py
data = {
    "field1": "data",
    "field2": "data",
}
multipart_form_data = {k: (None, v) for k, v in data.items()}
http.post("your-url", files=multipart_form_data)
```

Anton-Shutik:
Might be related - https://github.com/encode/httpx/issues/3522

PR #3397, opened 2024-11-08
feat(content): allow multipart requests without files.

laipz8200:
# Summary

related #3396.

Currently, a multipart request can only be sent when the files parameter is non-empty. This restriction limits cases where users might want to send data using multipart mode without attaching any files, which is possible in tools like Postman.

**Proposed Change**: Instead of requiring files to be non-empty, we could determine the need for a multipart request based on whether files is None rather than its emptiness. This would allow users to send multipart requests without attaching files if needed.

# Checklist

### DIFF
--- a/httpx/_main.py
+++ b/httpx/_main.py
@@ -487,7 +487,7 @@ def main(
                 params=list(params),
                 content=content,
                 data=dict(data),
-                files=files,  # type: ignore
+                files=files or None,  # type: ignore
                 json=json,
                 headers=headers,
                 cookies=dict(cookies),
--- a/tests/test_content.py
+++ b/tests/test_content.py
@@ -4,7 +4,8 @@ import typing
 import pytest
 
 import httpx
-from httpx._content import encode_json
+from httpx._content import encode_json, encode_request
+from httpx._types import SyncByteStream
 
 method = "POST"
 url = "https://www.example.com"
@@ -345,7 +346,7 @@ async def test_multipart_data_and_files_content():
 
 @pytest.mark.anyio
 async def test_empty_request():
-    request = httpx.Request(method, url, data={}, files={})
+    request = httpx.Request(method, url, data={})
     assert isinstance(request.stream, typing.Iterable)
     assert isinstance(request.stream, typing.AsyncIterable)
 
@@ -521,3 +522,12 @@ def test_allow_nan_false():
         ValueError, match="Out of range float values are not JSON compliant"
     ):
         encode_json(data_with_inf)
+
+
+def test_encode_request_with_data_and_empty_files():
+    headers, stream = encode_request(data={"key": "value"}, files={})
+    assert headers["Content-Type"].startswith("multipart/form-data; boundary=")
+    assert isinstance(stream, SyncByteStream)
+    stream_content = b"".join(stream)
+    assert b'Content-Disposition: form-data; name="key"' in stream_content
+    assert b"value" in stream_content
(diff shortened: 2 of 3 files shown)


tests/test_content.py
```diff
@@ -516,3 +517,16 @@ def test_allow_nan_false():
...
         httpx.Response(200, json=data_with_inf)
+
+
+def test_encode_request_with_data_and_empty_files():
+    request = httpx.Request(
+        url="https://www.example.com",
+        method="POST",
+        data={"key": "value"},
+        files={}
+    )
```
tomchristie, line 528:
```suggestion
    request = httpx.Request(url="https://www.example.com", method="POST", data={"key": "value"}, files={})
```

tests/test_content.py
```diff
@@ -516,3 +517,11 @@ def test_allow_nan_false():
         ValueError, match="Out of range float values are not JSON compliant"
     ):
         httpx.Response(200, json=data_with_inf)
+
+
+def test_encode_request_with_data_and_empty_files():
+    request = httpx.Request(url="https://www.example.com", method="POST", data={"key": "value"}, files={})
```
tomchristie, line 523:
```suggestion
    request = httpx.Request(
        url="https://www.example.com", method="POST", data={"key": "value"}, files={}
    )
```

tomchristie:
Thanks. I've updated the tests to avoid private imports.

Do we think this is an improved behaviour? It does seem nice that an explicit `files={}` will enforce multipart. (On the other hand, if `data` and `files` are determined dynamically as in the cli we need to explicitly switch to `None`)

laipz8200:
Thanks for the update!

My original plan was to make it easier to use multipart without having to carry files, which wasn’t as straightforward before.

I don’t use the cli much, so if this causes any issues, just let me know.

httpx/_main.py
```diff
@@ -483,7 +483,7 @@ def main(
                 params=list(params),
                 content=content,
                 data=dict(data),
-                files=files,  # type: ignore
+                files=files or None,  # type: ignore
```
sylann, line 486: I don't think this change is correct. It seems to go against the spirit of your main change in the httpx/_content.py file.
It seems to prevent the fix from working when using the cli.
Although I am not familiar with this usage and I didn't look too much into it, so I don't know if it's even possible to do this through the cli.

laipz8200: Thank you for the review, I'll take a look at the CLI.

ujayant:
@laipz8200 @tomchristie When do you plan to merge this feature? I am running into a similar issue

ujayant:
@tomchristie @laipz8200 Is there a workaround for this in the older versions?

laipz8200:
> Is there a workaround for this in the older versions?

@ujayant Here is a workaround: https://github.com/encode/httpx/issues/3396#issuecomment-2508142475. I apologize for not being able to dedicate time recently to address this issue.

ujayant:
> > Is there a workaround for this in the older versions?
> 
> @ujayant Here is a workaround: [#3396 (comment)](https://github.com/encode/httpx/issues/3396#issuecomment-2508142475). I apologize for not being able to dedicate time recently to address this issue.

I see this error when I try to use your example -` chunk = self.file.read(self.CHUNK_SIZE) AttributeError: 'dict' object has no attribute 'read'`. You will have to pass an empty file

laipz8200:
Hi @tomchristie,

Sorry for the late reply — I think this PR is ready to be merged.

Regarding the point @sylann raised, I see it as more of a decision question: do we want to offer the same functionality for the CLI as well? If yes, should it be done in this PR?

PR #3178, opened 2024-04-25
add socks5h proxy support

liubin:
In some case socks5h is needed to resolve the hostname.

# Summary

add socks5h proxy support.

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.

PR #3404, opened 2024-11-15
Version 0.28.0

tomchristie:
Hey team. 👋🏼

Alrighty we've got ourselves a release here. I think this strikes the right balance between pushing forwards towards an improved and simplified 1.0 release while minimising disruption in getting there.

---

## 0.28.0 (15th November, 2024)

The 0.28 release includes a limited set of backwards incompatible changes.

**Backwards incompatible changes**:

SSL configuration has been significantly simplified.

* The `verify` argument no longer accepts string arguments.
* The `cert` argument has now been removed.
* The `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables are no longer automatically used.

For users of the standard `verify=True` or `verify=False` cases this should require no changes.

For information on configuring more complex SSL cases, please see the [SSL documentation](docs/advanced/ssl.md).

**The following changes are also included**:

* The undocumented `URL.raw` property has now been deprecated, and will raise warnings.
* The deprecated `proxies` argument has now been removed.
* The deprecated `app` argument has now been removed.
* Ensure JSON request bodies are compact. (#3363)
* Review URL percent escape sets, based on WHATWG spec. (#3371, #3373)
* Ensure `certifi` and `httpcore` are only imported if required. (#3377)
* Treat `socks5h` as a valid proxy scheme. (#3178)
* Cleanup `Request()` method signature in line with `client.request()` and `httpx.request()`. (#3378)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.28.0 (15th November, 2024)
+## 0.28.0 (28th November, 2024)
 
 The 0.28 release includes a limited set of backwards incompatible changes.
 
@@ -617,7 +617,7 @@ See pull requests #1057, #1058.
 
 * Added dedicated exception class `httpx.HTTPStatusError` for `.raise_for_status()` exceptions. (Pull #1072)
 * Added `httpx.create_ssl_context()` helper function. (Pull #996)
-* Support for proxy exlcusions like `proxies={"https://www.example.com": None}`. (Pull #1099)
+* Support for proxy exclusions like `proxies={"https://www.example.com": None}`. (Pull #1099)
 * Support `QueryParams(None)` and `client.params = None`. (Pull #1060)
 
 ### Changed
@@ -845,7 +845,7 @@ We believe the API is now pretty much stable, and are aiming for a 1.0 release s
 
 ### Fixed
 
-- Fix issue with concurrent connection acquiry. (Pull #700)
+- Fix issue with concurrent connection acquisition. (Pull #700)
 - Fix write error on closing HTTP/2 connections. (Pull #699)
 
 ## 0.10.0 (December 29th, 2019)
@@ -1094,7 +1094,7 @@ importing modules within the package.
 
 ## 0.6.7 (July 8, 2019)
 
-- Check for connection aliveness on re-acquiry (Pull #111)
+- Check for connection aliveness on re-acquisition (Pull #111)
 
 ## 0.6.6 (July 3, 2019)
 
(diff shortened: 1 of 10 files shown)


zanieb (approved):

agrueneberg:
@tomchristie How would I supply a custom root certificate if a dependency is using httpx but does not allow a custom context to be passed? So far we have used `SSL_CERT_FILE` in those situations.

tomchristie:
@agrueneberg Thanks. You would need the dependency to support passing an ssl context through, if you wanted to use a custom context...

Can you describe your setup just a little more?

tomchristie:
@agrueneberg Thanks. You would need the dependency to support passing an ssl context through, if you wanted to use a custom context...

Good pointer towards why we might want to switch to truststore over certifi.

Perhaps you could take a look at #3409 and see if that'd resolve things for you.

agrueneberg:
Sorry for the late response. Our dependency (qdrant-client) does in fact pass additional arguments to httpx, so the problem is not as bad as I thought. I still liked the configuration-free approach that `SSL_CERT_FILE` offered, being able to run in different environments without a change in code (e.g., dev / prod environment).

Truststore appears to be a step in the right direction.

agrueneberg:
That said, I've never seen a code snippet where an explicit SSLContext is passed as an argument to `verify`. `SSL_CERT_FILE` is not universal (despite being mentioned in PEP 476) because requests only supports `REQUESTS_CA_BUNDLE` instead, but passing a string pointing to the CA bundle as an argument to `verify` is such a common solution to SSL problems that I'm not sure if breaking that behavior is worth it.

tomchristie:
> passing a string pointing to the CA bundle as an argument to verify is such a common solution to SSL problems that I'm not sure if breaking that behavior is worth it.

@agrueneberg That's a concession I'd be happy for us to take for a 0.28.0 release.

agrueneberg:
Godspeed, thanks for hearing me out :)

tomchristie:
@agrueneberg Okay I've taken another look at this...

I *do* think it's important for us to guide our users towards a more constrained API.

PR #3418 would moderate this by disallowing the `str` usage but being explicit about how to get the same behavior...

```
        msg = (
            "verify should be a boolean or SSLContext, since version 0.28. "
            "Use `verify=ssl.create_default_context(cafile=...)` "
            "or `verify=ssl.create_default_context(capath=...)`."
        )
        raise RuntimeError(msg)
```

It's a pain point, tho the trade off is that it helps us push towards a simpler & more comprehensible implementation. Ultimately that means a sharper API surface area, fewer potential bug points, and an easier-to-understand system.

That'd probably be *my* preference here as a middle-ground of sorts.

(Progressing towards a simpler implementation, while trying to minimise disruption. ?)

tomchristie:
Okay, as much as I'd like to push forward with this quickly we really need a deprecation path rather than hard breakages here.

So... #3419

PR #3409, opened 2024-11-21
Truststore

tomchristie:
# Summary

# Checklist

### DIFF
--- a/httpx/_config.py
+++ b/httpx/_config.py
@@ -22,10 +22,10 @@ UNSET = UnsetType()
 def create_ssl_context(verify: ssl.SSLContext | bool = True) -> ssl.SSLContext:
     import ssl
 
-    import certifi
+    import truststore
 
     if verify is True:
-        return ssl.create_default_context(cafile=certifi.where())
+        return  truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
     elif verify is False:
         ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
         ssl_context.check_hostname = False


docs/advanced/ssl.md
```diff
@@ -71,9 +50,9 @@ client = httpx.Client(verify=ctx)
 
 ### Working with `SSL_CERT_FILE` and `SSL_CERT_DIR`
 
-Unlike `requests`, the `httpx` package does not automatically pull in [the environment variables `SSL_CERT_FILE` or `SSL_CERT_DIR`](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_default_verify_paths.html). If you want to use these they need to be enabled explicitly.
+Unlike `requests`, the `httpx` package does not automatically pull in [the environment variables `SSL_CERT_FILE` or `SSL_CERT_DIR`](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_default_verify_paths.html). 
 
-For example...
+These environment variables shouldn't be required since they're obsoleted by `truststore`. They can be enabled if required like so...
```
tomchristie, line 55:
```suggestion
These environment variables shouldn't be necessary since they're obsoleted by `truststore`. They can be enabled if required like so...
```

docs/advanced/ssl.md
```diff
@@ -29,34 +31,11 @@ import certifi
 import httpx
 import ssl
```
tomchristie, line 32:
```suggestion
import ssl
import certifi
```

tomchristie:
Looking at PyPI stats we've currently still got [fairly significant 3.9 usage](https://pypistats.org/packages/httpx).

Perhaps a smart approach here would be...

* Plan on switching from `certifi` to `truststore` with a release aimed at 2025-10.
* At the same time as Python 3.9 becomes EOL.
* And Python 3.14 is released.

Secrus:
Maybe it could be done based on python version? `certifi` for <3.10 and `truststore` for the rest?

tomchristie:
@Secrus - Could phase it in gradually yep. Or eg... support it in Python 3.14.
I really don't have enough context to know how to take a call on this one.

zanieb:
I would be uncomfortable dropping support for Python 3.9 before EOL, but it is worth noting there are other ecosystem packages that do this, e.g., `numpy` has an aggressive Python support schedule and dropped Python 3.9 support earlier this year https://github.com/numpy/numpy/pull/26222

deathaxe:
Maybe I am missing something, but according to some experiences made with SSL module of python 3.8 on Windows, `ssl.create_default_context()` already loads OS level trust stores by calling `SSLContext.load_default_certs()`, if called without explicit `cafile`, `capath` and `cadata` arguments.

see: https://github.com/python/cpython/blob/39b2f82717a69dde7212bc39b673b0f55c99e6a3/Lib/ssl.py#L750

As of python 3.8 ssl module also performs host name verification by default. 

So why struggle with custom cert store libraries?

jmaillefaud:
@deathaxe Behavior of the stdlib `ssl` module is not consistent across python versions and sources on POSIX systems. For example on ubuntu, installing python through `uv` will have `ssl` bundled with [`webpki-roots`](https://crates.io/crates/webpki-roots) and will not load the openssl system certificates by default.

cclauss:
I suggest adding Python 3.14 to the test suite as in:
* #3645

.github/workflows/test-suite.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
cclauss, line 17:
```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```

CarliJoy:
Support for Python 3.9 ended more than 3 month ago.
Maybe it is time to revisit this MR and merge it? (Maybe rebase first).

I don't see a hold-up anymore or?

Any help required?

PR #3418, opened 2024-11-28
Error on verify as str.

tomchristie:
For users working with `verify="/path/to/certs"` point them explicitly towards the more explicit usage style.


httpx/_config.py
```diff
@@ -31,6 +31,14 @@ def create_ssl_context(verify: ssl.SSLContext | bool = True) -> ssl.SSLContext:
         ssl_context.check_hostname = False
         ssl_context.verify_mode = ssl.CERT_NONE
         return ssl_context
+    elif isinstance(verify, str):  # pagma: nocover
```
tomchristie, line 34:
```suggestion
    elif isinstance(verify, str):  # pragma: nocover
```

PR #3419, opened 2024-11-28
Version 0.28.0.

tomchristie:
Okay, here's version 0.28 with *nice clean deprecations* for our more constrained SSL API...

---

## 0.28.0 (28th November, 2024)

The 0.28 release includes a limited set of deprecations.

**Deprecations**:

We are working towards a simplified SSL configuration API.

*For users of the standard `verify=True` or `verify=False` cases, or `verify=<ssl_context>` case this should require no changes. The following cases have been deprecated...*

* The `verify` argument as a string argument is now deprecated and will raise warnings.
* The `cert` argument is now deprecated and will raise warnings.

Our revised [SSL documentation](docs/advanced/ssl.md) covers how to implement the same behaviour with a more constrained API.

**The following changes are also included**:

* The deprecated `proxies` argument has now been removed.
* The deprecated `app` argument has now been removed.
* JSON request bodies use a compact representation. (#3363)
* Review URL percent escape sets, based on WHATWG spec. (#3371, #3373)
* Ensure `certifi` and `httpcore` are only imported if required. (#3377)
* Treat `socks5h` as a valid proxy scheme. (#3178)
* Cleanup `Request()` method signature in line with `client.request()` and `httpx.request()`. (#3378)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -617,7 +617,7 @@ See pull requests #1057, #1058.
 
 * Added dedicated exception class `httpx.HTTPStatusError` for `.raise_for_status()` exceptions. (Pull #1072)
 * Added `httpx.create_ssl_context()` helper function. (Pull #996)
-* Support for proxy exlcusions like `proxies={"https://www.example.com": None}`. (Pull #1099)
+* Support for proxy exclusions like `proxies={"https://www.example.com": None}`. (Pull #1099)
 * Support `QueryParams(None)` and `client.params = None`. (Pull #1060)
 
 ### Changed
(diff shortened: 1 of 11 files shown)


tomchristie:
For comparison, SSL implementation [before](https://github.com/encode/httpx/blob/0.27.2/httpx/_config.py#L60-L193) and [after](https://github.com/encode/httpx/blob/version-0.28-deprecations-only/httpx/_config.py#L23-L70).
Much simpler and more comprehensible, even including deprecations.

Revised documentation is here... https://github.com/encode/httpx/blob/version-0.28-deprecations-only/docs/advanced/ssl.md

zanieb (approved):
Love deprecation warnings :)

CHANGELOG.md
```diff
@@ -4,25 +4,23 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.28.0 (...)
+## 0.28.0 (28th August, 2024)
```
zanieb, line 7: Presumably you mean November

zanieb:
I would suggest waiting until 1.0 to drop 3.8 support, since it's not really costing us anything.

tomchristie:
> I would suggest waiting until 1.0 to drop 3.8 support.

I can work with that, yep.

PR #3421, opened 2024-11-28
Change timeout values to be ints

vpetro:
# Summary

Timeout values are specified as `Timeout | int | None` in `get` and etc. `httpx.Timeout` however accepts value of `float` type.

# Checklist


tomchristie:
> Timeout values are specified as Timeout | int | None in get and etc.

Pointer?

PR #3423, opened 2024-11-29
HOTFIX: Keyword param "proxies" is missing.

Danarvelini:
Update _client.py

"proxies" param is missing if passed as keyword argument. Even tough it should be there, since it is mentioned in the docstring as an optional param.

# Summary

I am adding "proxies" keyword params to optional Client() params. It should already be there to deal with a scenario in which it is going as a keyword argument, even if it is None. I am facing this scenario at the moment and I thought of adding this missing part of the code so others won't need to face it as well.
Also, docstring of the class mentions it, but it is not there in the __init__ params.

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -651,6 +651,7 @@ class Client(BaseClient):
         http1: bool = True,
         http2: bool = False,
         proxy: ProxyTypes | None = None,
+        proxies: dict[str, ProxyTypes] | None = None,
         mounts: None | (typing.Mapping[str, BaseTransport | None]) = None,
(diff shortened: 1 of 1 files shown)


Danarvelini:
Will this be deprecated?

Shouldn't it be removed from class docstring as an acceptable optional parameter, then?
Right now, it says as acceptable, but actually, it is not supported when passed down to the class.

tomchristie:
It was deprecated and has now been removed.

You’re correct, it’s still accidentally included in the docstring.
If you’d like to issue a pull request resolving that it’d be welcome.

Danarvelini:
Ok, I just did that. Thank you :)

https://github.com/encode/httpx/pull/3426

PR #3426, opened 2024-11-29
Updating deprecated docstring Client() class

Danarvelini:
# Summary

Updating deprecated docstring Client() class. Keyword param "proxies" is no longer used and supported, but it was still "optional" by the docstring of the class. That docstring is now updated to avoid developers on passing deprecated params.

# Checklist


tomchristie (approved):

tomchristie:
Thanks.

Danarvelini:
Great, thank you :)

PR #3427, opened 2024-11-29
show how to send multipart form data without files

sylann:
# Summary

As it is, the documentation hints at the possibility of sending a request with a multipart form encoded  body, but it is not explicit about it.

Hence it takes more time to figure it out on your own, especially for a less experienced developer.

This PR only adds 1 example in the multipart section of the quickstart page to demonstrate the process.

# Checklist

- [ ] ~~I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.~~ Not applicable (only changed documentation)

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -169,7 +169,7 @@ Form encoded data can also include multiple values from a given key.
 }
 ```
 
-## Sending Multipart File Uploads
+## Sending Multipart File Uploads and/or Data
 
 You can also upload files, using HTTP multipart encoding:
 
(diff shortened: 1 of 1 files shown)


sylann:
I only have a doubt on the section title and the small sentence before the example does not mention that the example makes use of the `files` param instead of the `data` param.

Could improve this if needed.

ujayant:
@sylann I see this error when I try to use your example - ```chunk = self.file.read(self.CHUNK_SIZE) AttributeError: 'dict' object has no attribute 'read'```

PR #3438, opened 2024-12-02
fix: merge existing params in url, close #3433

### LINKED ISSUE #3433: Missing query params in url when params option is set
In version 0.28.0, query params are never merged.

```python
import httpx

client = httpx.Client()
request = client.build_request(
    "GET", "https://httpbin.org/get?param=default", params={"foo": "bar"}
)
print(request.url)  # This will be `https://httpbin.org/get?foo=bar`
```

The query string in the request url will be ignored. It seems this is caused by the pr #3364. Params are replaced instead of merged when parsing. https://github.com/encode/httpx/blob/15e21e9ea3cad4f06e22a7e704aabefdf43d2e29/httpx/_urlparse.py#L301

Related discussion https://github.com/encode/httpx/discussions/3428

GreyElaina:
# Summary

Revert #3364

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -396,7 +396,9 @@ class Request:
         extensions: RequestExtensions | None = None,
     ) -> None:
         self.method = method.upper()
-        self.url = URL(url) if params is None else URL(url, params=params)
+        self.url = URL(url)
+        if params is not None:
+            self.url = self.url.copy_merge_params(params=params)
         self.headers = Headers(headers)
         self.extensions = {} if extensions is None else dict(extensions)
 


1umine (approved):

Kludex (dismissed):

Kludex:
Can we add a test here, so we are sure this doesn't break again in the future?

GreyElaina:
> Can we add a test here, so we are sure this doesn't break again in the future?

no problem, will present later.

GreyElaina:
@Kludex tests are ready.

tests/models/test_requests.py
```diff
@@ -226,3 +226,27 @@ def content() -> typing.Iterator[bytes]:
...
+def test_with_params():
+    request = httpx.Request("GET", "http://example.com", params={"a": "1", "b": "2"})
+    assert str(request.url) == "http://example.com?a=1&b=2"
+
+
+def test_merge_existing_params():
+    request = httpx.Request(
+        "GET", "http://example.com?c=3", params={"a": "1", "b": "2"}
+    )
+    assert str(request.url) == "http://example.com?c=3&a=1&b=2"
```
Kludex, line 240: This is the only one that fails on main, right?

GreyElaina: let me test.

GreyElaina: yes, and `test_empty_params` too.

GreyElaina: yes, and `test_empty_params` too.

Kludex: can you just leave the tests that actually break on main on this PR?

GreyElaina: fine

Kludex: thanks

tomchristie:
Thanks, tho see https://github.com/encode/httpx/issues/3433#issuecomment-2514327255

My suggestion would be...

* Introduce tests for the 0.28.0 behaviour.
* Include a release note item in 0.28.0 documenting the behavior change.

GreyElaina:
> Thanks, tho see [#3433 (comment)](https://github.com/encode/httpx/issues/3433#issuecomment-2514327255).
> 
> My suggestion would be...
> 
>     * Introduce tests for the 0.28.0 behaviour.
> 
>     * Include a release note item in 0.28.0 documenting the behavior change.

fine, see you another pr
close is welcome.

PR #3450, opened 2024-12-12
add support for any type of URL as long as it can be stringified

MarcBresson:
# Summary

for reference : https://github.com/encode/httpx/discussions/3142

This PR brings compatibility with pydantic HttpURL object (or any object). 

I'm using pydantic HttpURL type in my project (for validation purposes) but had errors when I passed the pydantic http object directly to https.request function. If the type of the given request is not httpx.URL, then the object will be stringified and the string will be passed on `urlparse`.

# Checklist

### DIFF
--- a/httpx/_urls.py
+++ b/httpx/_urls.py
@@ -118,10 +118,7 @@ class URL:
         elif isinstance(url, URL):
             self._uri_reference = url._uri_reference.copy_with(**kwargs)
         else:
-            raise TypeError(
-                "Invalid type for url.  Expected str or httpx.URL,"
-                f" got {type(url)}: {url!r}"
-            )
+            self._uri_reference = urlparse(str(url), **kwargs)
 
     @property
     def scheme(self) -> str:
--- a/httpx/_urls.py
+++ b/httpx/_urls.py
@@ -113,9 +113,7 @@ class URL:
                 params = kwargs.pop("params")
                 kwargs["query"] = None if not params else str(QueryParams(params))
 
-        if isinstance(url, str):
-            self._uri_reference = urlparse(url, **kwargs)
-        elif isinstance(url, URL):
+        if isinstance(url, URL):
             self._uri_reference = url._uri_reference.copy_with(**kwargs)
         else:
             self._uri_reference = urlparse(str(url), **kwargs)
--- a/tests/models/test_url.py
+++ b/tests/models/test_url.py
@@ -452,19 +452,17 @@ def test_url_set():
     assert all(url in urls for url in url_set)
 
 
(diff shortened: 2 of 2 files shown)


docs/api.md
```diff
@@ -112,6 +112,12 @@ what gets sent over the wire.*
 >>> url = URL("https://example.org/")
 >>> url.host
 'example.org'
+
+from pydantic import HttpUrl
```
Kludex, line 116:
```suggestion
>>> from pydantic import HttpUrl
```

httpx/_urls.py
```diff
@@ -74,7 +74,7 @@ class URL:
       themselves.
     """
 
-    def __init__(self, url: URL | str = "", **kwargs: typing.Any) -> None:
+    def __init__(self, url: URL | str | typing.Any = "", **kwargs: typing.Any) -> None:
```
Kludex, line 77: This will allow any type, which makes the other 2 types in this union irrelevant.

```suggestion
    def __init__(self, url: URL | str = "", **kwargs: typing.Any) -> None:
```

Maybe what you want is something like this?
```py
class HasDunderStr(Protocol):
    def __str__(self): ...
```

But... Not sure this is useful.

MarcBresson: I quite like keeping the first two types. It makes it explicit that there is a great compatibility for them.

I like your idea to use protocol. I'm gonna implement it right away.

Kludex: Note the "But... Not sure this is useful." 👀 👍

KalleDK: I understand where you are coming from, but this would make all classes / objects "work" - more or less remove the static typing help. As in converting a static bug to a runtime bug.

MarcBresson: indeed. However, it s the fastest way to bring great compatibility at the (small) expense of having a runtime error if you pass something that does not resemble a URL into this object.

The alternative solution is to explicitly type every library that provides a URL object of some sort, which I'm not too keen on doing.

zyv (commented):
This would be fantastic, especially because of Pydantic's `HttpUrl`... 👍

MarcBresson:
@tomchristie Hello, would you mind looking over this small pull request when you will have some time please?

lovelydinosaur:
@MarcBresson Thanks so much for your contribution, it really is appreciated.

I tend to prefer the strict typing approach here.

---

* https://daniel.haxx.se/blog/2022/01/10/dont-mix-url-parsers/ - Relevant article from Daniel Stenberg.
* https://www.encode.io/httpnext/urls - WIP docs for URL and QueryParam classes in 1.0.

PR #3460, opened 2024-12-20
Add support for Python 3.13

hugovk:
# Summary

Python 3.13 was released in October:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0719/

Let's add it to the CI and classifiers.

Python 3.8 went EOL at the same time, shall we drop support for that too?

# Checklist

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -3,6 +3,7 @@ name: Test Suite
 
 on:
   push:
+    branches: ["master"]
   pull_request:
     branches: ["master", "version-*"]
 
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -3,7 +3,6 @@ name: Test Suite
 
 on:
   push:
-    branches: ["master"]
   pull_request:
     branches: ["master", "version-*"]
 
@@ -14,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: "actions/checkout@v4"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [


zanieb (approved):

.github/workflows/test-suite.yml
```diff
@@ -3,7 +3,6 @@ name: Test Suite
 
 on:
   push:
-    branches: ["master"]
```
zanieb, line 6: Why did this change? I don't think it should?

zanieb:
Some prior discussion of 3.8 in https://github.com/encode/httpx/pull/3419#issuecomment-2506228687

.github/workflows/test-suite.yml
hugovk: It was so I could run the CI on my feature branch and be confident it passes before opening a PR.

zanieb: 👍 could you revert it? I think this should be covered by the `pull_request` target.

zanieb: Separately, we could add a `workflow_dispatch` trigger so it can be invoked on forks manually?

hugovk: Yeah, happy to revert it. It just means next time I'll either make my changes directly on my `main` (not recommended) or not bother contributing (also not recommended :)

`pull_request` alone doesn't cover it: if I push changes to my fork, the CI won't run.

zanieb: Usually I just open a pull request in my fork for CI if I need to test it before opening a real pull request.

I'm not sure exactly what the conditions cause this, but having `push` without a filter _and_ `pull_request` can run jobs _twice_ on a pull request e.g. https://github.com/astral-sh/python-build-standalone/pull/445/commits/accb28c1005486c11b3830852f2618c284a9cfa6

Screenshot 2024-12-23 at 1 16 17 PM

hugovk: Yeah, creating a PR in your own fork is possible, it's still an extra impediment to contribution. This project's contribution guidelines shown how to test locally, which is good.

But we get all these free CI minutes, I'm a big fan of CIs, and my opinion is we should encourage people to test on CI and make it easy for them. No "works on my machine". Let me easily test all the Python version, all the operating systems, all the build steps in a clean CI environment. Let's have good "DX".

And it's possible to add config to prevent those double runs on upstream, I can look it up if you like.

Anyway, I've reverted this change.

zanieb: > And it's possible to add config to prevent those double runs on upstream, I can look it up if you like.

I'm open to a proposal. My main reactions are that it doesn't belong in this pull request / limiting to the main branch is the standard pattern in all the repositories I work in. I agree making it easy for developers to test in CI is nice though :)

Issue #2298, opened 2022-07-08
SSLContext get created unconditionally which can cause serious performance issues


kissgyorgy:
### Discussed in https://github.com/encode/httpx/discussions/2245

<div type='discussions-op-text'>

<sup>Originally posted by **kissgyorgy** May 25, 2022</sup>
Our use case is that we built an API Gateway, in which we use httpx as a client to connect to the backend service.

We have a multi-tenant system, so we make a new httpx client every time to isolate customer requests completely.
We have a serious performance issue this way, we measured it, and every request takes 100ms longer just because a `ssl.SSLContext` is created unconditionally, so we spend a lot of time building `SSLContext`s, which we don't need anyway, because we don't need an HTTPS connection to the backend, as the two services are on the same machines.

I tried to solve it by passing in a global `httpx.AsyncHTTPTransport`, but httpx closes the transport after finishing the request, and I was very confused by that so I didn't thought it could be solved this way.
The next thing I tried to implement a _global connection pool_, so the `SSLContext` would be created at application startup only once, but because of a serious bug in `httpcore` about corrupted connections, we couldn't do that either (we would have hit the same bug with the global transport implementation too).

A trivial implementation would be to only create `ssl.SSLContext` in case of HTTPS connections, which [httpcore is already doing correctly](https://github.com/encode/httpcore/blob/b76afce454aa3c90bbe7f98424ff2fc8a95d45b6/httpcore/_sync/connection.py#L135-L140), but for some reason, [httpx does this unconditionally](https://github.com/encode/httpx/blob/89cdd903b2f339c6ae319773c31b8c9352bd9cd8/httpx/_transports/default.py#L126).

Another implementation idea by @vlaci that there could be an `ssl_context_factory`, which could be implemented by users.

Here is a py-spy snapshot analyzed with [speedscope](https://www.speedscope.app/), which shows we spend half the time creating unused `SSLContext`s:
![image](https://user-images.githubusercontent.com/244203/170277618-7aa207d2-8376-4102-8c69-c32ebaf77e9b.png)
</div>

tomchristie:
I think you'd find the same behaviour here with `requests` or `urllib3`.

I'd suggest either reusing the same transport as @florimondmanca suggested in https://github.com/encode/httpx/discussions/2245. You could also just share a single ssl context by creating a global SSL context instance that you pass in to each client with `verify = ssl_context`.

For now I'd suggest we close this off, as this isn't an issue when you:

* Share a single client.
[...]

PR #3473, opened 2025-01-15
transport: cache default ssl context

karolinepauls:
# Summary

Creating a client takes 25ms of CPU time. This is a lot. Most of that time is spent initialising the SSL context. Since the context is private anyway, we can cache it. In order to avoid memory problems and privacy issues, only the default context is cached.

Prior discussion, when the problem was left for later (2022): https://github.com/encode/httpx/issues/2298#issuecomment-1199229281. Personally, I'm finding solving this problem in httpx to be a comparable amount of work to solving it in applications.

Before:
```
In [2]: timeit httpx.AsyncClient()
26.6 ms ± 364 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: timeit httpx.Client()
26.8 ms ± 268 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

After:
```
In [2]: timeit httpx.AsyncClient()
46.2 µs ± 224 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [3]: timeit httpx.Client()
46.5 µs ± 1.14 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

CPU Profile (only before):
image

# Checklist


deathaxe:
Maybe a caller should rather prefer to create a global Client instance for all its requests, then arbitrarily and implicitly caching internals.

karolinepauls:
> Maybe a caller should rather prefer to create a global Client instance for all its requests, then arbitrarily and implicitly caching internals.

The problem is that in practice this simply doesn't happen - instead, CPUs are spun, more instances are added, and energy is burned.

deathaxe:
Well, than probably re-considder your application's architecture.

Also note you simply being able to create your own global ssl context and pass it to each client instance on creation. It would at least be way more explicit and in control of a caller.

karolinepauls:
> Well, than probably re-considder your application's architecture.

In the last application I worked on I have solved this problem.

**The argument I'm making is about the ethics of engineering** - this library causes 100% CPU usage for 20ms with default settings. Given the popularity of Python and the typical quality of engineering, this could be gigawatthours of contribution to global warming every day.

deathaxe:
Well, those are arguments we could start a neverending debate on software development in general. I agree proper choise of programming languages, architectures and smart development decisions could save our earth 80% of CO2, but this would in a first step to abondon all those inefficient technologies which drive our modern web, such as bloated text based data transfers driven by inefficient resource eating scripting languages at all. 

However those are still non-arguments against explicity.

A Client instance should always be treated as an isolated object not sharing any resources with other instances by default, as it is completely unclear for a library such as httpx, in which context those clients are used. They may require different ssl contexts for security/isolation requirements.

For those who just want to use `httpx.get()` and `httpx.put()` it would probably be more straight forward for httpx to provide a shared client object, which lives forever, instead of creating new ones for each function call.

PR #3480, opened 2025-01-23
chore(docs): add httpx-retry to 3P packages

mharrisb1:
# Summary

Adds [httpx-retry](https://github.com/mharrisb1/httpx-retry) to third-party packages page in docs as suggested by [issue comment](https://github.com/encode/httpx/issues/108#issuecomment-1653298485) in #108 

# Checklist

- [ ] ~I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.~

### DIFF
--- a/docs/third_party_packages.md
+++ b/docs/third_party_packages.md
@@ -82,6 +82,12 @@ Allows consuming Server-Sent Events (SSE) with HTTPX.
 
 A library for scraping the web built on top of HTTPX.
 
+### httpx-retry
+
+[Github](https://github.com/mharrisb1/httpx-retry)
+
+Middleware for implementing retry policies with HTTPX
+
 ## Gists
 
 <!-- NOTE: this list is in alphabetical order. -->


mharrisb1:
@zanieb up to you. Just following suggestion from issue thread. Feel free to close

zanieb:
I don't mean to be discouraging — other people may feel like it's fine.

mharrisb1:
> surprised to see a week-old package with one release

This is valid. Maybe there should be a note on adoption threshold, minimum age, or something else to avoid this and other new projects that might become abandonware from being proposed or added.

Or maybe even something as simple as this:

```diff
# Third Party Packages
As HTTPX usage grows, there is an expanding community of developers building tools and libraries that integrate with 
- HTTPX, or depend on HTTPX. Here are some of them.
+ HTTPX, or depend on HTTPX. Here are some of the more popular ones.
```

I shilled the project in the comments so that should be fine for anyone landing there from a Google search that doesn't want to roll their own

jareks:
@zanieb please note that in this case not having any package for retries resulted in building two at same time: https://github.com/encode/httpx/issues/108#issuecomment-2614617035

Of course it's up to you to decide if you want to encourage 3rd party package or not but I just wanted to point out there's also a cost of not doing it.

zanieb:
It's not really up to me! @tomchristie has the final say and others on the encode team may have different opinions.

I just noted that the package was brand new because it wasn't mentioned in the pull request.

I do think it makes some sense to only recommend packages that have demonstrated some level of adoption and stability, but, as I said, I don't feel strongly. There's a "Gist" section so maybe the intent _is_ for it to be very inclusive.

PR #3486, opened 2025-01-30
Detect blocking calls in coroutines during tests using BlockBuster

cbornet:
# Summary

[BlockBuster](https://github.com/cbornet/blockbuster) is a library to detect harmful blocking calls done in the async event loop (when activated, it raises an error if a blocking call is done).
It is just a few lines to integrate with pytest.
Globally, `httpx` is a good citizen and the only blocking call it detected is `mimetypes.guess_type` in `Request` constructor. `mimetypes` has a global cache so this blocking call is only done once.
Solutions for this would be:
* defer the mimetype resolution to when the request is actually called and call `mimetypes.guess_type` in a thread
* or just accept that it blocks for a short time only once and add an exemption in blockbuster setup.
I've also seen blocking calls made when using HTTPS but there seems to be no unit tests for async+TLS.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,3 +28,4 @@ trio-typing==0.10.0
 trustme==1.1.0; python_version < '3.9'
 trustme==1.2.0; python_version >= '3.9'
 uvicorn==0.32.1
+blockbuster==1.5.23
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ import typing
 
 import pytest
 import trustme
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
Not looking to introduce additional tooling, so I'll close this off.

> Globally, httpx is a good citizen and the only blocking call it detected is mimetypes.guess_type in Request constructor. mimetypes has a global cache so this blocking call is only done once.

That's useful feedback, thanks.

PR #3489, opened 2025-01-31
Allow auth to be removed (be None)

KalleDK:
# Summary

Allow auth property on the BaseClient to be None (as a way to remove auth from the Client)

# Checklist


lovelydinosaur:
Thanks. Going to pass on changing this... you're not wrong here, however the mutable state on `Client` isn't a fantastic pattern and I'd prefer to minimise ongoing changes on 0.x.

v1 has immutability throughout.

PR #3500, opened 2025-02-12
Change return type annotation for Response.raise_for_status() to Self

viccie30:
# Summary

It always returns its self argument. This change makes subclasses easier to use.

# Checklist


tomchristie:
Interesting, thanks.

PR #3506, opened 2025-02-14
Update default proxy construction in HTTPTransport to include SSL context

DHerls:
# Summary
When a user created an HTTPTransport and supplied a URL for their proxy, the SSLContext was not passed to the created proxy. This caused some confusion when I ran into it the first time.

Now, the SSLContext is passed to the proxy when constructing by default.

# Checklist

### DIFF
--- a/httpx/_transports/default.py
+++ b/httpx/_transports/default.py
@@ -118,6 +118,15 @@ def map_httpcore_exceptions() -> typing.Iterator[None]:
         raise mapped_exc(message) from exc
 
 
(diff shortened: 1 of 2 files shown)


deathaxe:
Communication between client and local proxies is usually performed unencrypted

Coupling default SSL context by default would however result in httpx attempting to establish TLS connections to proxies by default.

A caller should better create a Prox() instance explicitly and pass it to Client() on instantiation, in case encrypted client<->proxy communication is required.

Also local proxies may require different, possibly self-signed?, special purpose certificates, which should probably not be leaked to default SSL context under all circumstances.

Issue #302, opened 2019-09-01
Drop certifi, use system trust store by default


sethmlarson:
This is an issue that many people have been trying to tackle for Python but hasn't been done yet. I think we're in a pretty good position to try to tackle this in a way that cam be available to everyone. 

- Linux/OpenSSL should use the default system CA bundle path compiled into OpenSSL.
- Windows should use SChannel which is capable of fetching updated CA certs if they're not in the trust store.
- macOS should use SecureTransport

This should probably be implemented as a separate library, maybe use a stripped-down oscrypto project as a starting point.

Reasons to do this:
- HTTPX will use the same CA certs as the system
- Easier configuration and deployment to corporate settings
- If a system is shipped with outdated OpenSSL (Windows, macOS) we don't care because the system is more likely to be up to date than whatever was shipped with Python. 
- Windows and macOS automatically update their certs and CRLs.

yeraydiazdiaz:
It's been a while since this was raised has been brought up recently a couple of times, could we go into detail of what's needed here?

For reference [oscrypto](https://github.com/wbond/oscrypto) lists [other related libraries](https://github.com/wbond/oscrypto#related-crypto-libraries), would any of those fit our needs? (I'd look myself but I just don't know enough about the subject 🙂 )

florimondmanca:
Hey! So considering the amount of :+1:'s on this topic I think it would make sense to decide on whether we should keep `certifi` or use system defaults as part of 1.0 - https://github.com/encode/httpx/issues/947#issuecomment-629698226

I had a striking moment of consciousness about this earlier today:

What if we just used [`ssl.create_default_context()`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context)

From the docs (emphasis mine):

> Return a new SSLContext object with default settings for the given purpose. The settings are chosen by the ssl module, and usually represent a higher security level than when calling the SSLContext constructor directly.
>
> cafile, capath, cadata represent optional CA certificates to trust for certificate verification, as in SSLContext.load_verify_locations(). If all three are None, **this function can choose to trust the system’s default CA certificates instead**.

TL;DR: it looks to me `ssl.create_default_context()` will* create an `ssl_context` that uses the system defaults. Is there anything else we need? Might draft a PR for this…

* The usage of "*can*" in the docs suggests that this might be "_may_ create", but I don't see why it wouldn't be the case. Are there cases when Python _wouldn't_ trust the default system CA?

tomchristie:
[...]

PR #3510, opened 2025-02-20
introduce HTTPX_CA_BUNDLE, fix docs

Yanpas:
# Summary

`HTTPX_CA_BUNDLE` env-variable was introduced (similar to `REQUESTS_CA_BUNDLE`). The motivation behind it: `SSL_CERT_FILE` affects other libraries/programs.

Alternatively I can introduce `HTTPX_SYSTEM_CA`.

Related to #302

# Checklist

### DIFF
--- a/docs/advanced/ssl.md
+++ b/docs/advanced/ssl.md
@@ -69,21 +69,11 @@ ctx.load_cert_chain(certfile="path/to/client.pem")  # Optionally also keyfile or
 client = httpx.Client(verify=ctx)
 ```
 
-### Working with `SSL_CERT_FILE` and `SSL_CERT_DIR`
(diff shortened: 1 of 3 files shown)


Yanpas:
Also current documentation is misleading

salman1993:
httpx docs suggest that they do not automatically pull in the environment variables SSL_CERT_FILE -
https://www.python-httpx.org/advanced/ssl/#working-with-ssl_cert_file-and-ssl_cert_dir

the code suggests that those env vars do get picked up: 
https://github.com/encode/httpx/blob/e70d0b08c929e5140e38aa92a419de0c6d494148/httpx/_config.py#L33-L37

PR #3512, opened 2025-02-24
Docs: Add `httpx.Proxy` to api.md

matusvalo:
# Summary

Documentation of httpx does not mention the parameters of `httpx.Proxy()`. They can be important e.g. in the case of custom `ssl_context`. This PR documents them in the API Reference.

# Checklist


tomchristie (approved):
Thanks.

PR #3513, opened 2025-02-25
Mask passwords in URLs when converting to a string

zanderxyz:
# Summary

As previously noted in [this GitHub discussion](https://github.com/encode/httpx/discussions/2765), this library by default leaks credentials which are included in URL strings (common for basic authentication). It can also raise exceptions which contain the credentials in the error string if a request fails (see `raise_for_status`).

This PR updates the `__str__` method on URLs to redact the password, something which was already implemented for `__repr__`. I believe this is the correct default behaviour for a library like this.

# Checklist

- [x] I've updated the documentation accordingly.  [I did not make any documentation changes as I do not think any are needed]


karpetrosyan (approved):

zanderxyz:
> Yep, hiding user credentials at the lowest layer and preventing them from being passed higher makes perfect sense.
> 
> Now it's exposed in URL.**str**, but not in URL.**repr**, which is a bit weird. **str** is supposed to show more user-related data, while **repr** is more for debugging and development. So, hiding it in **repr** would make more sense. However, I think preventing it entirely is the most secure way.

Just to clarify - after the change in this PR, the username & password are not exposed in either **str** or **repr✱. I do think this is the most secure implementation, and I don't think we really lose anything..

grahamwhiteuk:
Anyone have a view on when this might get merged (and released)?

grahamwhiteuk:
Seems to be taking an age to be reviewed/merge so in the meantime, if you find this and don't want to log secrets you could just alter the log level of httpx...

```python
import logging

httpx_logger = logging.getLogger("httpx")
httpx_logger.setLevel(logging.WARNING)
```

lovelydinosaur:
> This PR updates the __str__ method on URLs to remove the user & password details.

Thank you no.

PR #3520, opened 2025-02-27
Use unasync for tests

karpetrosyan:
This PR attempts to use the unasync strategy to keep both async and sync tests of the client synchronized. Previously, we encountered issues where something worked for the async client but not for the sync client. I tried using unasync to generate the synchronous client code, but it seems too risky to do so without first applying the unasync strategy to the tests. This way, in the next step, we can safely generate the implementation without any risks.

Ref: 3046

I tried to make the changes incrementally to make the review process easier. The changes include:

- Removing sync tests completely and replacing them with async ones (we don't need to write any sync code as it will be autogenerated).
- In async tests, avoiding test names that mention "async," such as "test_async_client_request." We don't need this, as it would only complicate our unasync code by requiring additional rules.
- Copying unasync code from httpcore and adding it to the lint/check scripts.


karpetrosyan:
What do you think, @tomchristie ? I would appreciate a review. These kinds of PRs are hard to keep in sync with the master branch.

PR #3551, opened 2025-04-20
Add `content_type` for Response

Sanchoyzer:
# Summary

I would like to suggest adding a little syntax sugar. It might be more convenient to use `response.content_type` instead of `response.headers.get("Content-Type")`.
I hope I haven't forgotten any places where changes need to be made

# Checklist

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -690,11 +690,10 @@ class Response:
         """
         Return the encoding, as specified by the Content-Type header.
         """
-        content_type = self.headers.get("Content-Type")
-        if content_type is None:
+        if self.content_type is None:
             return None
 
-        return _parse_content_type_charset(content_type)
+        return _parse_content_type_charset(self.content_type)
 
     def _get_content_decoder(self) -> ContentDecoder:
         """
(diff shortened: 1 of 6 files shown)


Sanchoyzer:
Hey, @tomchristie , what do you think of this suggestion?

httpx/_models.py
```diff
@@ -1075,6 +1074,14 @@ async def aclose(self) -> None:
             with request_context(request=self._request):
                 await self.stream.aclose()
 
+    @property
+    def content_type(self) -> str | None:
+        """
+        Return the Content-Type header.
+        """
+        result = self.headers.get("Content-Type")
+        return str(result) if result else None
```
Kludex, line 1083: Why the `str()` cast? Isn't it `str` already?

Kludex:
I think this is not needed, but we can see what other maintainers think about it.

httpx/_models.py
Sanchoyzer: It's not `str`, it's `Any`

```python
class Headers(typing.MutableMapping[str, str]):
    def get(self, key: str, default: typing.Any = None) -> typing.Any:
        ...
```

MinecraftFuns: Actually, `result` defaults to `None` if the key is not present in the HTTP headers due to how the `Headers.get` function works, so you may as well `return self.headers.get("Content-Type")`.

lovelydinosaur:
>> what do you think of this suggestion?

>  I think this is not needed, but we can see what other maintainers think about it.

Indeed. Appreciate the interest tho.

PR #3556, opened 2025-04-26
Python 3.14: handle `typing.Union` immutability

ncoghlan:
# Summary

`typing.Union` instances are becoming immutable in Python 3.14+: https://docs.python.org/3.14/whatsnew/3.14.html#typing

This trips up the code in `httpx.__init__` that hides the internal module details in top-level object metadata by setting their `__module__` attributes.

# Checklist

I don't believe test or documentation changes are needed (when the CI config is expanded to include Python 3.14, this will be covered automatically), but I've left those unchecked until that has been confirmed.

# Details

Attempting to [publish an experimental project](https://github.com/ncoghlan/tstrprocess/actions/runs/14682311907/job/41206490424#step:3:76) that will require Python 3.14's t-strings to run, I encountered the following exception from this code:

```
AttributeError: 'typing.Union' object has no attribute '__module__' and no __dict__ for setting new attributes. Did you mean: '__reduce__'?
```

I then traced that new failure back to the `typing.Union` change described in the summary.

The PR currently works by ignoring the exception raised (since that's the most comprehensive option), but an alternative approach would be to add a `hasattr` check so only values that already have a `__module__` attribute get modified.

### DIFF
--- a/httpx/__init__.py
+++ b/httpx/__init__.py
@@ -102,4 +102,9 @@ __all__ = [
 __locals = locals()
 for __name in __all__:
     if not __name.startswith("__"):
-        setattr(__locals[__name], "__module__", "httpx")  # noqa
+        try:
+            setattr(__locals[__name], "__module__", "httpx")  # noqa
+        except AttributeError:
+            # typing.Union instances are immutable in Python 3.14+,
+            # so don't fail if writing to "__module__" fails
+            continue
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks @ncoghlan. Are you able to determine which symbol(s) are breaking here?

ncoghlan:
Ah, trying to reproduce this locally shows that I misread the original traceback. This is actually a dupe of the `httpcore` failure that was already fixed in https://github.com/encode/httpcore/commit/b2a796c4ba093e10bb05184e69374191ae6ed1d2

`setup-pdm` must not have picked up the new version of `httpcore` yet, hence the failure. Sorry for the noise!

PR #3558, opened 2025-05-01
bumped minimum version of httpcore

kylecacciatore:
# Summary

Updating version criteria for a minimum version of `httpcore` to mitigate a downstream vulnerability.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore==1.*",
+    "httpcore>=1.0.9",
     "anyio",
     "idna",
 ]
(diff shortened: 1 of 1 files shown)


tomchristie:
Yep, so...

* `httpcore` v1.0.9 https://github.com/encode/httpcore/pull/1011 unpins the upper bound on `h11`.
* Ensuring that we don't block `h11` v0.16.0 https://github.com/python-hyper/h11/commit/9c9567f0a92d13a83a8d8ebdbc757c8c2d384536

We *could* issue a new version of httpx that forces httpcore>=1.0.9, tho I don't think that forcing the entire dependency chain to issue new releases all the way through would be necessary/helpful.

PR #3564, opened 2025-05-07
fix CVE-2025-43859

imredobos:
h11 accepts some malformed Chunked-Encoding bodies before 0.16.0, httpcore 1.0.9 ensure using at least 0.16.0

# Summary

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dependencies = [
     "certifi",
-    "httpcore==1.*",
+    "httpcore>=1.0.9",
     "anyio",
     "idna",
 ]


zanieb:
I'm actually not sure this is correct. I don't think `httpx` needs to enforce this for you. You can just upgrade that downstream package? I don't think `httpx` should say it is _incompatible_ with earlier versions of `h11`.

imredobos:
i have actually upgraded the downstream package to fix immediately, in my opinion, enforcing a version which solves a critical vulnerability is good thing, but i understand your point

LeoSL:
@zanieb thanks for replying :) 

I'm not too sure what are the deptree repercussions if I bump the h11 version while httpx's dependencies require a lower version. In my view, that leads to inconsistencies and messes up with the pip's dependency resolver.

Please educate me if I'm not seeing things straight.

zanieb:
> I'm not too sure what are the deptree repercussions if I bump the h11 version while httpx's dependencies require a lower version. In my view, that leads to inconsistencies and messes up with the pip's dependency resolver.

I'm not sure I follow, but httpx itself does not require the lower version — it allows any 1.x version. There shouldn't be repercussions to bumping h11. If some other dependency requires a _lower_ version of h11, and does not allow a newer version, that is indeed a problem and should be fixed in that package.

LeoSL:
Ah I see, I see.

The problem is within httpx's dependency `httpcore`, as we can see on this dep tree:
```
httpx==0.28.1
├── httpcore [required: ==1.*, installed: 1.0.5]
│   └── h11 [required: >=0.13,<0.15, installed: 0.14.0]
```

LeoSL:
Alright, after installing manually declaring `httpcore==1.0.9` (that requires h11's >= 0.16), this CVE seems to be fixed.

```
httpx==0.28.1
├── httpcore [required: ==1.*, installed: 1.0.9]
│   ├── certifi [required: Any, installed: 2025.1.31]
│   └── h11 [required: >=0.16, installed: 0.16.0]
```

danielfcollier:
If it is enforced, the ecosystem would gain awareness of the vulnerability. Not enforcing is going along with the problem.

zanieb:
@danielfcollier adding a lower bound to a new version of httpx does not "increase awareness" of a vulnerability, it just forces an upgrade of the package when upgrading httpx. In either case, you need to be performing upgrades to resolve the vulnerability. You should be using dedicated tooling, e.g., Dependabot, for awareness of CVEs.

danielfcollier:
It is just a fact that enforcing a higher version would raise the bar, enforce security, and raise awareness - some people are just not educated about vulnerabilities or do not have a security department to take care of these scans. But, ok, when someone reverse engineers the ecosystem to find projects with top vulnerabilities, someone might consider enforcing some updates to strategic dependencies.

PR #3592, opened 2025-06-26
Drop Python 3.8 support

agronholm:
# Summary

# Checklist


Kludex:
Based on @tomchristie 's comment on https://github.com/encode/httpcore/pull/1019#issuecomment-2996228576, I'm merging this.

PR #3593, opened 2025-06-26
Ensured explicit closing of async generators

agronholm:
# Summary

This change ensures that async generators are always explicitly closed, even if the task running the generator is cancelled or another exception is raised. This prevents unpredictable behavior and also avoids the asyncgen finalizer warnings on Trio.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [UNRELEASED]
 
+### Fixed
+
+* Explicitly close all async generators to ensure predictable behavior
+
 ### Removed
 
 * Drop support for Python 3.8
--- a/httpx/_content.py
+++ b/httpx/_content.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import inspect
 import warnings
+from collections.abc import AsyncGenerator
 from json import dumps as json_dumps
 from typing import (
     Any,
@@ -10,6 +11,7 @@ from typing import (
     Iterable,
     Iterator,
     Mapping,
+    NoReturn,
 )
 from urllib.parse import urlencode
 
@@ -23,7 +25,7 @@ from ._types import (
     ResponseContent,
     SyncByteStream,
 )
-from ._utils import peek_filelike_length, primitive_value_to_str
+from ._utils import peek_filelike_length, primitive_value_to_str, safe_async_iterate
 
 __all__ = ["ByteStream"]
 
@@ -35,7 +37,7 @@ class ByteStream(AsyncByteStream, SyncByteStream):
     def __iter__(self) -> Iterator[bytes]:
         yield self._stream
 
-    async def __aiter__(self) -> AsyncIterator[bytes]:
+    async def __aiter__(self) -> AsyncGenerator[bytes]:
         yield self._stream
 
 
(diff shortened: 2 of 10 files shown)


agronholm:
Note that this depends on another PR: #3592. I will mark as ready to review once that one has been merged.

agronholm:
I checked locally with Python 3.10 that the supposedly missing coverage cannot be reproduced. I'm seeing 100% coverage, even without the `httpcore` changes.

agronholm:
> I checked locally with Python 3.10 that the supposedly missing coverage cannot be reproduced. I'm seeing 100% coverage, even without the `httpcore` changes.

Never mind, I was on the wrong branch. Fixing.

agronholm:
@Kludex All good now.

httpx/_content.py
```diff
@@ -99,9 +102,8 @@ class UnattachedStream(AsyncByteStream, SyncByteStream):
     def __iter__(self) -> Iterator[bytes]:
         raise StreamClosed()
 
-    async def __aiter__(self) -> AsyncIterator[bytes]:
+    def __aiter__(self) -> NoReturn:
```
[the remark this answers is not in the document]
agronholm: It's not really necessary here. `__aiter__()` can return any arbitrary object that supports the `AsyncIterator` protocol, and the implementation here just raises `NotImplementedError`.

httpx/_utils.py
```diff
@@ -240,3 +245,19 @@ def is_ipv6_hostname(hostname: str) -> bool:
     except Exception:
         return False
     return True
+
+
+@asynccontextmanager
+async def safe_async_iterate(
```
tomchristie, line 251: Interesting. I'm curious about this vs. `aclosing`.

https://docs.python.org/3/library/contextlib.html#contextlib.aclosing

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [UNRELEASED]
 
+### Fixed
+
+* Explicitly close all async generators to ensure predictable behavior
```
tomchristie, line 11: 👍🏼 Neat.

httpx/_utils.py
agronholm: I originally went with `aclosing()` but it doesn't work very well with arbitrary async iterables.

httpx/_content.py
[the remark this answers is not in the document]
agronholm: On another note, is either one of you at EuroPython?

lovelydinosaur:
Thanks for your time on this @agronholm.

The 1.0 pre-release deals with this comprehensively. There's no I/O iteration or generators, just regular `.read()`/`.close()` for I/O, as well as a much simpler callstack.

ofek:
I haven't been following the developments. Are the changes you mentioned due to performance?

agronholm:
> I haven't been following the developments. Are the changes you mentioned due to performance?

No, but correctness. Async generators are supposed to be explicitly closed when you're done with them. But this PR is obsolete as httpx is transitioning to a new API in which I'm told this is a non-issue anyway.

ofek:
My apologies for not being explicit, I was referring to the removal of generators/the new design. I'm always interested in hearing about architectural changes to this library because at work we have a task I created a while back to deprecate our use of `requests` on customer machines. I'm no longer on that team, but I try to update the task description over time with tips or best practices for the eventual migration.

PR #3613, opened 2025-07-25
Use standard libary Zstandard for Python 3.14+

mollymorphous:
# Summary
[PEP 784](https://peps.python.org/pep-0784/) add a Zstandard implementation to the Python standard library under `compression.zstd`, and is scheduled for release in Python 3.14. This PR adapts `ZStandardDecoder` to work with either the standard library implementation or the implementation from the [`zstandard`](https://github.com/indygreg/python-zstandard) package.

This has the implication that Zstandard content decoding is available by default on Python 3.14 and later, without the need to install the `zstd` extra. 

# Checklist
  - Testing this requires Python 3.14, but the existing zstd unit tests pass with the standard library implementation. I'm always happy to add more tests if needed!

### DIFF
--- a/httpx/_decoders.py
+++ b/httpx/_decoders.py
@@ -26,11 +26,15 @@ except ImportError:  # pragma: no cover
         brotli = None
 
 
-# Zstandard support is optional
+# Zstandard support is avaible in the standard library from 3.14 and later,
+# or by an optional dependency on `zstandard`
 try:
-    import zstandard
+    import compression.zstd as zstandard
 except ImportError:  # pragma: no cover
-    zstandard = None  # type: ignore
+    try:
+        import zstandard
+    except ImportError:  # pragma: no cover
+        zstandard = None
 
 
 class ContentDecoder:
@@ -174,8 +178,15 @@ class ZStandardDecoder(ContentDecoder):
                 "Make sure to install httpx using `pip install httpx[zstd]`."
             ) from None
 
-        self.decompressor = zstandard.ZstdDecompressor().decompressobj()
+        self._new_decompressor()
         self.seen_data = False
+
+    def _new_decompressor(self) -> None:
+        decompressor = zstandard.ZstdDecompressor()
+        if hasattr(decompressor, "decompressobj"):
+            self.decompressor = decompressor.decompressobj()  # prgama: no cover
+        else:
+            self.decompressor = decompressor  # pragma: no cover
 
     def decode(self, data: bytes) -> bytes:
         assert zstandard is not None
@@ -185,19 +196,16 @@ class ZStandardDecoder(ContentDecoder):
             output.write(self.decompressor.decompress(data))
             while self.decompressor.eof and self.decompressor.unused_data:
                 unused_data = self.decompressor.unused_data
-                self.decompressor = zstandard.ZstdDecompressor().decompressobj()
+                self._new_decompressor()
                 output.write(self.decompressor.decompress(unused_data))
         except zstandard.ZstdError as exc:
             raise DecodingError(str(exc)) from exc
         return output.getvalue()
 
     def flush(self) -> bytes:
-        if not self.seen_data:
-            return b""
-        ret = self.decompressor.flush()  # note: this is a no-op
-        if not self.decompressor.eof:
+        if self.seen_data and not self.decompressor.eof:
             raise DecodingError("Zstandard data is incomplete")  # pragma: no cover
-        return bytes(ret)
+        return b""
 
 
 class MultiDecoder(ContentDecoder):
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ As well as these optional installs:
 * `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
 * `click` - Command line client support. *(Optional, with `httpx[cli]`)*
 * `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
-* `zstandard` - Decoding for "zstd" compressed responses. *(Optional, with `httpx[zstd]`)*
+* `zstandard` - Decoding for "zstd" compressed responses before Python 3.14. *(Optional, with `httpx[zstd]`)*
 
 A huge amount of credit is due to `requests` for the API layout that
 much of this work follows, as well as to `urllib3` for plenty of design
--- a/docs/index.md
+++ b/docs/index.md
@@ -119,7 +119,7 @@ As well as these optional installs:
 * `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
 * `click` - Command line client support. *(Optional, with `httpx[cli]`)*
 * `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
-* `zstandard` - Decoding for "zstd" compressed responses. *(Optional, with `httpx[zstd]`)*
+* `zstandard` - Decoding for "zstd" compressed responses before Python 3.14. *(Optional, with `httpx[zstd]`)*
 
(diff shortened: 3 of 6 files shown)


lovelydinosaur:
Ooh interesting, thanks.

Any idea on how widely zstd is currently supported? (Wrt both browsers and servers.)

Related to this... compression is one of the currently unimplemented features in the httpx 1.0 prerelease... https://www.encode.io/httpnext/

httpx/_decoders.py
```diff
@@ -174,9 +178,16 @@ def __init__(self) -> None:
...
-        self.decompressor = zstandard.ZstdDecompressor().decompressobj()
+        self._new_decompressor()
         self.seen_data = False
 
+    def _new_decompressor(self) -> None:
+        decompressor = zstandard.ZstdDecompressor()
+        if hasattr(decompressor, "decompressobj"):
+            self.decompressor = decompressor.decompressobj()  # prgama: no cover
+        else:
+            self.decompressor = decompressor  # pragma: no cover
```
tomchristie, line 189: Could you explain this part?

lilydjwg:
> Any idea on how widely zstd is currently supported? (Wrt both browsers and servers.)

At least crates.io, packagist and sourceforge support it. (My tests blow up because I cache http responses and there are zstd responses from other Python versions; the "zstandard" module doesn't support 3.14 yet.)

httpx/_decoders.py
mollymorphous: This is admittedly a little awkward. Python upstreamed the `pyzstd` package into `compress.zstd` because it's API was closer to existing standard library compression APIs. The `zstandard` package provides a [`ZstdDecompressObj`](https://python-zstandard.readthedocs.io/en/latest/decompressor.html#zstddecompressionobj) facade that implements the standard library style API.

An alternative would be to import the libraries under separate names so the method would look more like this:

```python
def _new_compressor(self) -> None:
    if compression_zstd is not None:
        self.decompressor = compression_zstd.ZstdDecompressor()
    else:
        self.decompressor = zstandard.ZstdDecompressor().decompressobj()
   
```

What do you think?

mollymorphous (commented):
Thanks for the review! Firefox and Chrome currently support zstd ([caniuse](https://caniuse.com/zstd)). Safari does not yet, but plans to: https://github.com/WebKit/standards-positions/issues/168

I wasn't able to find a hard number on server-side deployments, the answer seems to be not a lot, but CloudFlare recently added support to the CDN.

lovelydinosaur:
From a bit of time reviewing this I've not been able to track down good examples of URLs to use for comparison purposes here. 

Eg...

* CloudFlare blog pages don't appear to use compression for hosted images.
* GitHub pages sites don't appear to use compression for hosted images.
* WikiPedia supports gzip compression throughout.

Here's one example of a URL that does support zstd, though in this particular case it appears less efficient than gzip...

```
$ curl https://help.netflix.com/en/node/30081 -H "Accept-Encoding: br" --output netflix.br 
$ curl https://help.netflix.com/en/node/30081 -H "Accept-Encoding: gzip" --output netflix.gzip
$ curl https://help.netflix.com/en/node/30081 -H "Accept-Encoding: zstd" --output netflix.zstd
$ curl https://help.netflix.com/en/node/30081 -H "Accept-Encoding: identity" --output netflix
$ wc -c netflix*
   84920 netflix
   84920 netflix.br  # Not supported on this URL
   12602 netflix.gzip
   16761 netflix.zstd
```

The CloudFlare pitch for zstd https://blog.cloudflare.com/new-standards/ isn't *neccessarily* convincing... gzip is essentially just as fast, and looks to have slightly less efficient though notably more stable compression ratios.

I'm reviewing this for the purposes of `httpx` 1.0, and I'm expecting that *only supporting gzip* might be a reasonable default.

Does anyone have some useful real world examples that'd help verify if that is / isn't a good policy?

cclauss:
I suggest adding Python 3.14 to the test suite as in:
* #3645

tuffnatty:
I would suggest dropping `zstandard` and switch to stdlib-compatible `backports.zstd` on Python 3.9-3.13, as `httpx` has dropped Python 3.8 support already.

tuffnatty:
> The CloudFlare pitch for zstd https://blog.cloudflare.com/new-standards/ isn't _neccessarily_ convincing... gzip is essentially just as fast, and looks to have slightly less efficient though notably more stable compression ratios.

gzip is definitely not just as fast (without hardware offloading), it's just that they measure the whole response time, where the compression speed difference does not seem to matter much _on average_.
> 
> I'm reviewing this for the purposes of `httpx` 1.0, and I'm expecting that _only supporting gzip_ might be a reasonable default.

> Does anyone have some useful real world examples that'd help verify if that is / isn't a good policy?

I had used this to stream huge JSONL chunks over HTTP, and while it does not make traffic much less or a great overall speed improvement, the CPU load situation has improved a lot (relative to gzip).

cclauss:
@tuffnatty Would you be willing to create an alternative pull request that uses the backport and adds automated tests on Py3.14 like
* #3645

lovelydinosaur:
Okay, so my review of this was that *supporting gzip only* would be a sensible policy.
That's what we'll go for in 1.0. Let's not spend any more time rejigging zstd here.

lilydjwg:
> Okay, so my review of this was that _supporting gzip only_ would be a sensible policy.
> That's what we'll go for in 1.0. Let's not spend any more time rejigging zstd here.

Would there be a mechanism to add support for other compression methods via third-party code then? I'm a bit worried about interoperability with not-so-good servers and proxies.

lovelydinosaur:
That's a good question... ...I can't answer that fully at the moment.

There might not be any API *explicitly* for that purpose. Here's how interop. with the streams class would be...

```python
# A custom stream on top of the streams API...
class ZstdStream(httpx.Stream):
    def __init__(self, stream: httpx.Stream):
        self._wrapped = stream

    # Implement `.read()` and `.close()`

# Usage...
stream = ZStdStream(response.stream)
body = stream.read()
```

We probably don't want specific dials to "support for other compression methods via third-party code", since the less config the better. However we do want the tooling itself to be flexible enough to support customisation.

There's a current example in the docs demo'ing a custom client which is vastly more simple than with httpx 0.28. Perhaps building out that example to also demo custom response classes would be a good way to point users towards adaptability without increasing API surface area.

tuffnatty:
> @tuffnatty Would you be willing to create an alternative pull request that uses the backport 

@cclauss Thanks but the issue has been resolved in another way.

ddelange:
> Okay, so my review of this was that _supporting gzip only_ would be a sensible policy. That's what we'll go for in 1.0. Let's not spend any more time rejigging zstd here.

fwiw, the major alternatives to httpx support zstd (and for sure brotli):
- [aiohttp](https://github.com/aio-libs/aiohttp/commit/4872fce3426119e63e1a892c39b474786dafddac)
- [urllib3](https://github.com/urllib3/urllib3/commit/c4b5917e911a90c8bf279448df8952a682294135)
  - requests

supporting only gzip/deflate might be insufficient in 2025. especially brotli accounts for 33% of compressed http responses already in 2021 ([Figure 22.4](https://almanac.httparchive.org/en/2021/compression#trends-in-http-compression)) and 44% of compressed javascript versus 41% gzip in 2024 ([Figure 1.21](https://almanac.httparchive.org/en/2024/javascript#compression-methods))

cclauss:
> the issue has been resolved in another way.

How was the issue resolved?

tuffnatty:
> > the issue has been resolved in another way.
> 
> How was the issue resolved?

https://github.com/encode/httpx/pull/3613#issuecomment-3302376638

ddelange:
disregard my last comment, I see that they're all [supported](https://github.com/encode/httpx/blob/0.28.1/httpx/_decoders.py) :+1:

PR #3625, opened 2025-08-12
Fix: Raise helpful TypeError for invalid list 'data'

mahdirahimi1999:
# Summary

This PR fixes a misleading error that occurred when passing an invalid list (e.g., a list of dictionaries) to the `data` parameter.

Previously, `AsyncClient` would raise a confusing `RuntimeError` instead of a `TypeError` that accurately described the problem. This change introduces a validation check in `BaseClient` to catch this common mistake early and raises a helpful `TypeError`. The new error message guides the user to use the `json` parameter for JSON arrays, ensuring consistent behavior between both sync and async clients.

# Checklist

- [ ] I've updated the documentation accordingly. (N/A: This change improves an error message and does not require a documentation update.)

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -363,6 +363,16 @@ class BaseClient:
 
         [0]: /advanced/clients/#request-instances
         """
+        # Validate data parameter for better error messages
+        if data is not None and isinstance(data, list):
+            # Check if this looks like invalid JSON array (list of dicts/strings)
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Thanks tho no. Note that the upcoming `httpx` 1.0 pre-release comprehensively deals with getting to a much cleaner & more minimally typed API.

PR #3628, opened 2025-08-15
Migrate API docs to mkdocstrings

### LINKED ISSUE #3145: Cross reference documentation with mkdocstrings' `objects.inv`
Discussed in https://github.com/encode/httpx/discussions/3091

Well I say discussed but in reality it got no response so i'm trying to follow up. Does httpx provide `objects.inv` for documentation cross-reference? `mkdocstrings` generates this automatically unless explicitly disabled and it allows for third party packages relying on httpx to better document their projects and easily refer to httpx's documentation.

https://mkdocstrings.github.io/usage/#cross-references-to-other-projects-inventories

anuraaga:
# Summary

Migrates API docs from `autodoc` to `mkdocstrings`. I was hoping to be able to use an httpx inventory file in some documentation and noticed there seems to be interest in it from the attached issue so tried sending a PR.

I have made tweaks such as to the CSS to make it as close to the current site as possible. Though it could also make sense to have the full TOC.

The biggest difference in rendering is the docs don't have a inlet and are rendered directly. There are some other differences like font size and not showing the word `class` (mkdocstrings may allow it in the future as they have an insiders-only feature for it right now).

To really render using mkdocstrings, i.e. rendering out parameters in a structured way, the docstrings themselves would need to be updated to one of the supported styles like Google, numpy, or sphinx. This PR doesn't change that though to be smaller and still be quite close to what's already there, while also providing the inventory file.

To test, I run 

```
uv run --with mkdocs --with mkdocs-material --with mkautodoc --with 'mkdocstrings[python]' mkdocs build
open site/index.html
```

Fixes #3145 

image

# Checklist

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -9,43 +9,68 @@
     long-lived connections.
 
 ::: httpx.request
-    :docstring:
 
 ::: httpx.get
-    :docstring:
 
 ::: httpx.options
-    :docstring:
 
 ::: httpx.head
-    :docstring:
 
 ::: httpx.post
-    :docstring:
 
 ::: httpx.put
-    :docstring:
 
 ::: httpx.patch
-    :docstring:
 
 ::: httpx.delete
-    :docstring:
 
 ::: httpx.stream
-    :docstring:
 
 ## `Client`
 
 ::: httpx.Client
-    :docstring:
-    :members: headers cookies params auth request get head options post put patch delete stream build_request send close
+    options:
+      inherited_members: true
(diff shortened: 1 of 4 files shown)


anuraaga:
@lovelydinosaur Any interest in this? If not will go ahead and close this. Thanks.

pollier:
Hello, it would be great to be able to use objects.inv, generated by mkdocstring to have auto cross reference in project

PR #3629, opened 2025-08-15
Add `httpx-secure` to third party packages

Zaczero:
# Summary

Recently I worked on securing HTTPX against SSRF attacks (as none of the existing solutions worked for me) and decided to polish it up and publish it as a package for others to use.

# Checklist


lovelydinosaur:
Okay, thanks Kamil.

PR #3640, opened 2025-08-26
Add chunk_size to iter_lines to be compat with requests

HiromuHota:
# Summary
This PR adds chunk_size to iter_lines to be compat with requests.

# Checklist


HiromuHota:
Closing as adding chunk_size itself doesn't make it compat with requests.Response, which has decode_unicode and delimiter.
https://requests.readthedocs.io/en/latest/api/#requests.Response.iter_lines

PR #3649, opened 2025-09-02
Update compatibility.md with documentation of exceptions differences

GlenTiki:
# Summary

 This change just documents the high level differences in exceptions exposed from `httpx` vs `requests`.

# Checklist


lovelydinosaur:
Yes okay - thank you!

PR #3655, opened 2025-09-04
Ignore UTF-8 surrogate characters when encoding request content

nguyenhoan1988:
# Summary

- Add tests asserting surrogate code points are dropped from request bodies and JSON.
- Change all internal UTF-8 encodings to use errors="ignore" to avoid "surrogates not allowed" Unicode errors.

# Checklist

### DIFF
--- a/docs/third_party_packages.md
+++ b/docs/third_party_packages.md
@@ -24,6 +24,12 @@ Provides authentication classes to be used with HTTPX's [authentication paramete
 
 This package adds caching functionality to HTTPX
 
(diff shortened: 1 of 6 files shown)


cclauss:
Please rebase to pick up the fixes merged in:
* #3651
* #3654

lovelydinosaur:
> I understand that this PR may be closed in case there was no previous discussion.

Thank you, no. I wouldn't consider `errors='ignore'` an improvement here.

PR #3664, opened 2025-09-12
Add Python 3.14 to test matrix

lovelydinosaur:
Tests here were hanging on a previous run, let's see if that's resolved.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,7 @@ twine==6.1.0
 coverage[toml]==7.10.6
 cryptography==45.0.7
 mypy==1.17.1
-pytest==8.4.3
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
*Well that's tedious*.

We're not doing any gnarly Python magic here, so Python 3.13 -> 3.14 breaking almost certainly isn't something that the `httpx` package is causing. If anyone's got a good guess onto what part of dependency tooling is getting snarled up here???

karpetrosyan:
Hey Kim! I opened a PR (#3722 ) with the fixes. I would love to make a release with these updates, since httpx has not had a release in over a year! I would appreciate it if you could review it!
Best, Kar

PR #3698, opened 2025-10-28
support chunked upload in async file-like interfaces

### LINKED ISSUE #1620: Support async file types in `files = {}` and `content = ...`
We ought to support the following cases.

Raw upload content from an async file interface:

```python
import httpx
import trio

async def main():
        async with await trio.open_file(...) as f:
            client.post("https://www.example.com", content=f)

trio.run(main)
```

Multipart file upload from an async file interface:

```python
import httpx
import trio

async def main():
        async with await trio.open_file(...) as f:
            client.post("https://www.example.com", file={"upload": f})

trio.run(main)
```

We probably want to ensure that we're supporting both `trio`, `anyio` (Which have the same interfaces), and perhaps also `aiofiles. So eg, also supporting the following...

```python
# Supporting the same as above but using `asyncio`, with `anyio` for the file operations.
import anyio
import asyncio
import httpx

async def main():
        async with await anyio.open_file(...) as f:
            client.post("https://www.example.com", content=f)

asyncio.run(main())
```

The `content=...` case is a little simpler than the `data=...` case, since it really just need an async variant of `peek_filelike_length`, and a minor update to the `._content.encode_content()` function.

Also fiddly is what the type annotations ought to look like.

tsimoshka:
# Summary

This PR resolves #1620 and implements chunked upload for `anyio.open_file`, `trio.open_file` and `aiofiles.open_file` when used as `content=` parameter for `post` and `put` request. And implements support for multipart file upload for the same libraries. Most of the code for multipart file upload copied from my old stale PR https://github.com/encode/httpx/pull/3339.

changes:
- `_compat.py` file added to define `TypeIs` and `aclosing` for the range of supported python versions
- `_types.AsyncReadableFile` protocol was added along with `is_async_readable_file` type predicate function to detect and perform type narrowing for trio/asyncio/aiofiles async files
- `_types.FileContent` was extended to include the. `_types.AsyncReadableFile` protocol
- `_content.AsyncIteratorByteStream` updated to use `async read` methods for async files instead of looping over lines
- `_content.encode_content` updated to attach content length header for `_types.AsyncReadableFile` type
- `_multipart.FileField` updated to include async versions of `render` and `render_data`
- `_multipart.MultipartStream` updated with async version of `iter_data` and `__aiter__` method updated to use it
- `test_chunked_async_file_multipart` and `test_chunked_async_file_content` tests are added to test the new functionality
- `aiofiles` and `types-aiofiles` added in requirements to test aiofiles (can be removed with test if aiofiles support is not desirable)

# Checklist

- [ ] I've updated the documentation accordingly. (can update if PR is ok otherwise)


maretodoric:
It's 2026, python is booming with async functionality, feature requested since early 2021, multiple pull requests open - yet we still don't have this feature.
What gives?

PR #3710, opened 2025-11-14
Keep original host on IDNA decode errors

mimi89999:
# Summary

Emoji domains (e.g., `https://xn--i-7iq.ws/`, `https://xn--ls8h.la/`) are not allowed by IDNA2008 but are supported by some TLDs, have valid TLS certificates, and work in browsers, curl, and Python's `requests` library. 

Previously, httpx would raise an `IDNAError` when attempting to decode these Punycode domains, preventing legitimate requests.

This PR modifies the `host` property to gracefully handle IDNA decode errors by catching `idna.IDNAError` and keeping the original `xn--` encoded form. This allows httpx to work with these real-world emoji domains.

# Checklist

### DIFF
--- a/httpx/_urls.py
+++ b/httpx/_urls.py
@@ -170,6 +170,7 @@ class URL:
         """
         The URL host as a string.
         Always normalized to lowercase, with IDNA hosts decoded into unicode.
+        Invalid IDNA encodings are kept in their original xn-- form.
 
         Examples:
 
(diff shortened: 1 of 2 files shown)


mimi89999:
Alternatively, `httpx/_models.py` could be changed to check `self.url.raw_host` instead of checking `self.url.host`. I'm not sure which approach would be preferred. I prefer the approach that I proposed because it matches browser behavior of displaying the raw Punycode on decode errors rather than failing.

Issue #3712, opened 2025-11-24
MockTransport doesn't set the elapsed property


ionelmc:
This is not a nice solution:
```
def handler(request):
    resp = httpx.Response(200, json={"text": "Hello, world!"})
    resp.elapsed = timedelta(.......
    return resp
```

If you have dispatch code in that handler that can easily get boilerplatey.

I think httpx should do either of these:
* MockTransport take the duration as an optional argument
* httpx provide a decorator, so you can wrap your handler in it
* automatically set a duration of 0 for manually created Response objects - why error out if the data is already there?

karpetrosyan:
This issue already has two PRs with proposed fixes, but for now I suggest a very simple way to fix it without extra boilerplate code. We can close the PRs until we make a good design decision around this.

You can write a very simple wrapper around that mock transport and use it instead of the default one. Then you can handle the elapsed time for mocked responses however you want, like this.

```python
from __future__ import annotations

from datetime import timedelta

import httpx


class CustomMock(httpx.BaseTransport):
    def __init__(self, next_transport: httpx.BaseTransport | None = None) -> None:
        self._next_transport = httpx.MockTransport() if next_transport is None else next_transport

    def handle_request(self, request: httpx.Request) -> httpx.Response:
        response = self._next_transport.handle_request(request)
        response.elapsed = timedelta()
        return response

# use custom mock instead of the default mock
```

PR #3715, opened 2025-11-28
Feat/mock transport duration

BrunoMmS:
# Summary

This PR adds native support for simulating request latency in MockTransport by introducing an optional delay parameter.
When provided, the transport sets response.elapsed according to the delay value, unless the handler explicitly defines its own elapsed time.

This resolves the inconvenience described in issue #3712, where users needed to manually patch response.elapsed inside handlers to test timeout- or latency-sensitive logic.

# Checklist

### DIFF
--- a/httpx/_transports/mock.py
+++ b/httpx/_transports/mock.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import typing
+from datetime import timedelta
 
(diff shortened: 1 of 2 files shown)


karpetrosyan:
I'll close this for now until we have a good design decision around it

PR #3718, opened 2025-12-04
create_ssl_context: Don't skip ctx.load_cert_chain calls

scollinson:
…s a str

# Summary

`create_ssl_context` returns early when `verify` is a `str`, bypassing calls to `ctx.load_cert_chain` to load the client certificates.

# Checklist


zanieb:
Can you share why it matters / what the downstream affect is? Did you do investigation as to why it returns early today?

scollinson:
I have submitted this PR because I ran into this error when trying to use `httpx` for mTLS. Using wireshark I saw it presented empty certificates instead of the certificates I passed in, which led to this fix.

PR #3693, opened 2025-10-17
Python 3.14

mirthebeijers:
# Summary
Hi, this is my first contribution!!! Please be a little gentle with me ;-)

This is based on https://github.com/encode/httpx/pull/3664 but I add fixes for 

 - in the test suite deprecations are marked as fatal, and uvicorn is used to set up the tests, and uvicorn generates a deprecation warning because of iscoroutinefunction()
 - uvicorn is about to create a new release with a fix: https://github.com/Kludex/uvicorn/pull/2723 👀Python 3.14 has stricter requirements for event loops in background threads. The original code in httpx called server.run() which internally uses asyncio.run(), but this doesn't work properly in background threads in Python 3.14.
 - asyncio.get_event_loop() deprecated: The TestServer.serve() method was using asyncio.get_event_loop() which is now more restrictive in Python 3.14.

I'm not fully sure if this also passes on 3.9, it might need a little more work, but on 3.14 I think this fixes the problems ;-)

PR #3720, opened 2025-12-09
fix: resolve test hang with websockets 14+

SamMorrowDrums:
## Summary

Fixes #3708

The test suite hangs indefinitely when upgrading `websockets` from 13.x to 14.x or 15.x. This is because [websockets 14.0 switched to a new asyncio implementation](https://websockets.readthedocs.io/en/stable/project/changelog.html#id12) by default, which has different behavior around event loop handling.

## Root Cause

The issue stems from how the `TestServer` manages event loops when running in a background thread:

1. **In `TestServer.serve()`**: The deprecated `asyncio.get_event_loop()` + `loop.create_task()` pattern doesn't work correctly with websockets 14+'s new asyncio handling
2. **In `serve_in_thread()`**: The `server.run()` method relies on `asyncio.run()` which doesn't play well with the threaded setup when websockets is involved

## Changes

1. **`TestServer.serve()`**: Replace deprecated `asyncio.get_event_loop()` pattern with `asyncio.create_task()` which is the modern way to create tasks when already inside an async context.

2. **`serve_in_thread()`**: Instead of relying on `server.run()` which uses `asyncio.run()`, explicitly create a new event loop in the background thread with `asyncio.new_event_loop()` and `asyncio.set_event_loop()`.

3. **`server` fixture**: Remove `loop="asyncio"` from `Config()` since we now manage the event loop manually, avoiding conflicts with websockets' asyncio implementation.

## Testing

Tested locally with websockets 15.0.1 - all 43 integration tests pass without hanging.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -235,10 +235,9 @@ class TestServer(Server):
     async def serve(self, sockets=None):
         self.restart_requested = asyncio.Event()
 
-        loop = asyncio.get_event_loop()
         tasks = {
-            loop.create_task(super().serve(sockets=sockets)),
-            loop.create_task(self.watch_restarts()),
+            asyncio.create_task(super().serve(sockets=sockets)),
+            asyncio.create_task(self.watch_restarts()),
         }
         await asyncio.wait(tasks)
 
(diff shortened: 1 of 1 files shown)


karpetrosyan:
Duplicate of #3693, but thank you very much for the contribution. I will mention you in the fix as a co author!

mtelka:
@SamMorrowDrums I applied this change on top of `httpx 0.28.1` and testing still hangs for me with Python 3.9.25 and `websockets 15.0.1`.  Any suggestion?  Thank you.

mtelka:
When I uninstall `websockets 15.0.1` then all tests pass.

mtelka:
Please note that I use `uvicorn 0.39.0`.

PR #3725, opened 2025-12-13
Use zstandard implementation from stdlib (PEP-784)

Rogdham:
# Summary

## Context

Thanks to [PEP-784](https://peps.python.org/pep-0784/), Zstandard is included in Python starting from version 3.14 with the [`compression.zstd`](https://docs.python.org/3.14/library/compression.zstd.html) module.

So for Python 3.14+, we don't need an external lib. For older version of Python, I'm using [`backports.zstd`](https://github.com/Rogdham/backports.zstd) which exposes the same API as stdlib.

This also means that users of Python 3.14+ will benefit from Zstandard support even if they install `httpx` instead of `httpx[zstd]`.

A similar change has been adopted in various other libraries already, such as `urllib3` from version 2.6.0.

_Full disclosure: I'm the author and maintainer of `backports.zstd`, and the maintainer of `pyzstd` (which code was used as a base for the integration into Python). I also helped with PEP-784 and its integration into CPython._

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ As well as these optional installs:
 * `rich` - Rich terminal support. *(Optional, with `httpx[cli]`)*
 * `click` - Command line client support. *(Optional, with `httpx[cli]`)*
 * `brotli` or `brotlicffi` - Decoding for "brotli" compressed responses. *(Optional, with `httpx[brotli]`)*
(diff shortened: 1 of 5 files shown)


Rogdham:
Hello @mbeijen, as the author of Zstandard support in `httpx` your review would be appreciated! :hugs:

mbeijen:
Thanks for your work, this looks great (esp. together with #3726 !!!)

mbeijen (approved):

PR #3697, opened 2025-10-27
bugfix for zstd decompressobj reuse

mbeijen:
Fixes https://github.com/encode/httpx/discussions/3538

# Summary

This fixes an issue where zstd decoding fails. I've included a test case.

# Checklist

- [ x ] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [ x ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ x ] I've updated the documentation accordingly.

PR #3726, opened 2025-12-13
Zstandard: fix frame boundaries

Rogdham:
# Summary

An alternative fix to #3697. Added test taken from that PR, with credit to @mbeijen in the commit metadata.

Fixes #3538.
Closes #3697.

The first commit of this PR is from #3725. I am marking it as draft until #3725 is merged.

# Checklist


Rogdham:
The `test_zstd_truncated`, because it truncated the start of the frame, was not testing something different compared to `test_zstd_decoding_error`, I fixed the test by truncating the end only.

Also moved one statement out of the loop for speedup. I added an extra test as well.

PR #3727, opened 2025-12-14
feat: support "auth-int" quality-of-protection when using DigestAuth

dsillman2000:
# Summary

Adding support for "auth-int" (message integrity authentication) quality-of-protection implementation, as specified in [RFC-7616.](https://datatracker.ietf.org/doc/html/rfc7616). Additionally, updates unit tests to reflect support for the "auth-int" QoP.

## Note on "example" scenario from RFC-7616

I had to change the expected response hashes from the official ones stated in the [RFC-7616 Section 3.9.1](https://datatracker.ietf.org/doc/html/rfc7616#section-3.9.1) example since they don't provide an example where the client chooses to use "auth-int" QoP.

That being said, I was able to check the correctness of my implementation against `httpbin` using the `/digest-auth/auth-int/Mufasa/CircleofLife/[MD5|SHA-256]` endpoints to verify that a proper 200 status code is returned, meaning the httpbin server implementation of "auth-int" authentication agrees with the implementation added to this feature branch. Httpbin appears to support "http" requests to this endpoint, but doesn't properly support "https" requests (unexpected EoF encountered).

<details><summary>Integration test with HTTPBin</summary>
<p>

```py
from httpx import Client, DigestAuth, Request

if __name__ == "__main__":
    auth = DigestAuth(username="Mufasa", password="CircleofLife")
    # Note that I'm using locally-hosted HTTPBin over Docker since the official website is down today, Dec 14 2025.
    request = Request(
        "GET",
        # "http://httpbin.org/digest-auth/auth-int/Mufasa/CircleofLife/MD5",
        # "http://httpbin.org/digest-auth/auth-int/Mufasa/CircleofLife/SHA-256",
        # "http://0.0.0.0:80/digest-auth/auth-int/Mufasa/CircleofLife/MD5",
        "http://0.0.0.0:80/digest-auth/auth-int/Mufasa/CircleofLife/SHA-256",
    )

    with Client(auth=auth) as client:
        print(response.headers.raw)
        print(response.content)
        print(response.status_code) # We get a HTTP code 200 "OK"

```

</p>
</details> 

# Checklist

## GH Actions test failures

GH Actions seem to be failing non-deterministically for the `test_proxies.py` module (no changes from this PR), particularly due to timeouts when talking to `example.com`. I encourage a project maintainer to try re-running the checks for this PR during their review to demonstrate the non-deterministic behavior (take note of which python versions fail each time, if any).

Is there a default timeout being used for the tests which we can make more permissive?

### DIFF
--- a/docs/advanced/authentication.md
+++ b/docs/advanced/authentication.md
@@ -40,6 +40,18 @@ HTTP digest authentication is a challenge-response authentication scheme. Unlike
 [<Response [401 UNAUTHORIZED]>]
 ```
 
+HTTPX also supports digest authentication using `auth-int` quality-of-protection, which provides message integrity protection. When `qop="auth-int"` is used, the authentication response includes a hash of the request body, protecting against tampering with the message content during transmission. This provides additional security over `qop="auth"` by ensuring that both the authentication credentials and the message body integrity are verified.
+
+```pycon
+>>> auth = httpx.DigestAuth(username="olivia", password="secret")
+>>> client = httpx.Client(auth=auth)
+>>> response = client.get("https://httpbin.org/digest-auth/auth-int/olivia/secret")
+>>> response
+<Response [200 OK]>
+>>> response.history
+[<Response [401 UNAUTHORIZED]>]
+```
+
 ## NetRC authentication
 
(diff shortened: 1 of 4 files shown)


dsillman2000:
@karpetrosyan Please let me know if this PR looks good to you. Thanks in advance for your consideration!

PR #3749, opened 2026-01-26
Implement support for specifying the httpcore Network Backend when creating default transports

iciclespider:
# Summary

This enables specifying the httpcore Network Backend when creating the default transports as discussed here: https://github.com/encode/httpx/discussions/3180

# Checklist

### DIFF
--- a/docs/advanced/transports.md
+++ b/docs/advanced/transports.md
@@ -35,6 +35,18 @@ connecting via a Unix Domain Socket that is only available via this low-level AP
 {"ID": "...", "Containers": 4, "Images": 74, ...}
 ```
 
+Another advanced configuration is supplying a custom httpcore [Network Backend](https://www.encode.io/httpcore/network-backends/).
+
+```pycon
+>>> import httpx
+>>> import myk8s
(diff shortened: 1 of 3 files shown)


docs/advanced/transports.md
```diff
@@ -35,6 +35,18 @@ connecting via a Unix Domain Socket that is only available via this low-level AP
 {"ID": "...", "Containers": 4, "Images": 74, ...}
 ```
 
+Another advanced configuration is supplying a custom httpcore [Network Backend](https://www.encode.io/httpcore/network-backends/).
+
+```pycon
+>>> import httpx
+>>> import myk8s
```
lovelydinosaur: https://pypi.org/project/myk8s/

??

iciclespider: It was a theoretical package that returns a theoretical NetwokBackend. ~~Any suggestion for an alternative would be welcome.~~

iciclespider: I updated the network backend examples to use httpcore.MockBackend.

PR #3757, opened 2026-02-08
Fix MockTransport not setting elapsed property

### LINKED ISSUE #3712: MockTransport doesn't set the elapsed property

veeceey:
## Summary
Fixes MockTransport to automatically set `response.elapsed` to `timedelta(0)` for both sync and async requests, eliminating boilerplate and matching real transport behavior.

## Problem
Previously, accessing `response.elapsed` on MockTransport responses would raise `RuntimeError`:
```python
RuntimeError: '.elapsed' may only be accessed after the response has been read or closed.
```

Users had to add boilerplate in every handler:
```python
def handler(request):
    resp = httpx.Response(200, json={"text": "Hello, world!"})
    resp.elapsed = timedelta(0)  # Boilerplate needed
    return resp
```

## Solution
MockTransport now automatically sets `response.elapsed = timedelta(0)` after calling the handler, in both `handle_request()` and `handle_async_request()`.

## Test Plan
**Before fix:**
```python
def handler(request):
    return httpx.Response(200, json={"text": "Hello, world!"})

response = client.get("https://example.com")
print(response.elapsed)  # RuntimeError
```

**After fix:**
```python
def handler(request):
    return httpx.Response(200, json={"text": "Hello, world!"})

response = client.get("https://example.com")
print(response.elapsed)  # 0:00:00 ✓
```

**Tests:**
- Added `test_mock_transport_elapsed()` for sync clients
- Added `test_async_mock_transport_elapsed()` for async clients
- Both tests verify `elapsed == timedelta(0)` and type correctness

## Related
Fixes #3712

### DIFF
--- a/httpx/_transports/mock.py
+++ b/httpx/_transports/mock.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import typing
 
 from .._models import Request, Response
@@ -24,6 +25,7 @@ class MockTransport(AsyncBaseTransport, BaseTransport):
         response = self.handler(request)
         if not isinstance(response, Response):  # pragma: no cover
             raise TypeError("Cannot use an async handler in a sync Client")
+        response.elapsed = datetime.timedelta()
         return response
 
     async def handle_async_request(
@@ -40,4 +42,5 @@ class MockTransport(AsyncBaseTransport, BaseTransport):
         if not isinstance(response, Response):
             response = await response
 
+        response.elapsed = datetime.timedelta()
         return response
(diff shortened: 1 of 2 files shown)


veeceey:
All CI passing on Python 3.9-3.13, ready for merge

veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #3760, opened 2026-02-08
Fix: Prevent query parameter corruption in base_url

veeceey:
## Summary

Fixes #3614

When a `base_url` contains query parameters, the trailing slash enforcement was corrupting the query parameter values by appending the slash to the query string instead of just the path.

## Problem

```python
client = httpx.Client(base_url="https://api.com/get?data=1")
print(client.base_url.query)  
# Before: b'data=1/' ❌ (corrupted!)
# After:  b'data=1'  ✅ (correct)
```

The `_enforce_trailing_slash()` method was appending `/` to the entire `raw_path`, which includes both the path and query string. For a URL like `/get?data=1`, this became `/get?data=1/`, corrupting the query parameter value.

## Solution

Modified `_enforce_trailing_slash()` in `_client.py` to:
1. Split `raw_path` into path and query components (using `?` as delimiter)
2. Add trailing slash only to the path portion
3. Recombine path and query

## Examples

**base_url with query parameters:**
```python
client = httpx.Client(base_url="https://api.com/get?data=1")
# Before: https://api.com/get?data=1/  (corrupted)
# After:  https://api.com/get/?data=1  (correct)
```

**base_url with trailing slash and query:**
```python
client = httpx.Client(base_url="https://api.com/get/?key=value")
# Result: https://api.com/get/?key=value  (no double slash)
```

**base_url without query (unchanged behavior):**
```python
client = httpx.Client(base_url="https://api.com/get")
# Result: https://api.com/get/  (trailing slash added as before)
```

## Testing
- ✅ Manual testing with various base_url configurations
- ✅ Added regression tests in `tests/client/test_client.py`
- ✅ Verified backward compatibility with existing behavior

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -232,9 +232,19 @@ class BaseClient:
         return self._trust_env
 
     def _enforce_trailing_slash(self, url: URL) -> URL:
-        if url.raw_path.endswith(b"/"):
-            return url
-        return url.copy_with(raw_path=url.raw_path + b"/")
+        # Split raw_path into path and query to avoid corrupting query parameters
+        raw_path = url.raw_path
+        if b"?" in raw_path:
+            # URL has query parameters - only check/add slash to path portion
+            path, query = raw_path.split(b"?", 1)
+            if path.endswith(b"/"):
+                return url
+            return url.copy_with(raw_path=path + b"/?" + query)
+        else:
+            # No query parameters - check/add slash to entire raw_path
+            if raw_path.endswith(b"/"):
+                return url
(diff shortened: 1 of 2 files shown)


veeceey:
All CI passing on Python 3.9-3.13, ready for merge

veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

lovelydinosaur:
Heya... thanks for this.

Development of the next version is shifting across to a rewrite.
I'll communicate about this more fully in due course.

PR #3761, opened 2026-02-08
Merge params with existing URL query parameters instead of replacing

### LINKED ISSUE #3621: When the URL contains request parameters and the `params` parameter is set, the request parameters in the URL will disappear unexpectedly.
The starting point for issues should usually be a discussion...

https://github.com/encode/httpx/discussions

Possible bugs may be raised as a "Potential Issue" discussion, feature requests may be raised as an "Ideas" discussion. We can then determine if the discussion needs to be escalated into an "Issue" or not.

This will help us ensure that the "Issues" list properly reflects ongoing or needed work on the project.

---

veeceey:
## Summary

- Fixes #3621: When a URL already contains query parameters (e.g. `https://example.com/path?page=1&s=list`) and the `params` argument is also provided, the existing query parameters were silently dropped and replaced by only the new `params`. This was inconsistent with the `requests` library and surprised users.
- Changed `Request.__init__` to use `URL(url).copy_merge_params(params)` instead of `URL(url, params=params)`, so existing URL query parameters are preserved and merged with the new ones.
- Updated `test_request_params` to reflect the new merge behavior and added an additional test case for overlapping keys.

### Before (broken)
```python
request = httpx.Request("GET", "https://example.com?page=1&s=list", params={"pid": 0})
print(request.url)
# https://example.com?pid=0   <-- page and s are lost!
```

### After (fixed)
```python
request = httpx.Request("GET", "https://example.com?page=1&s=list", params={"pid": 0})
print(request.url)
# https://example.com?page=1&s=list&pid=0   <-- all params preserved
```

## Test plan

- [x] Existing `test_request_params` updated and passes
- [x] New test case for overlapping parameter keys added
- [x] All 173 model and client tests pass
- [x] Verified the exact reproduction case from the issue works correctly

### DIFF
--- a/httpx/_models.py
+++ b/httpx/_models.py
@@ -396,7 +396,7 @@ class Request:
         extensions: RequestExtensions | None = None,
     ) -> None:
         self.method = method.upper()
-        self.url = URL(url) if params is None else URL(url, params=params)
+        self.url = URL(url) if params is None else URL(url).copy_merge_params(params)
         self.headers = Headers(headers)
         self.extensions = {} if extensions is None else dict(extensions)
 
--- a/tests/models/test_requests.py
+++ b/tests/models/test_requests.py
@@ -232,10 +232,18 @@ def test_request_params():
     request = httpx.Request("GET", "http://example.com", params={})
     assert str(request.url) == "http://example.com"
 
+    # Params are merged with existing URL query parameters, not replaced.
     request = httpx.Request(
         "GET", "http://example.com?c=3", params={"a": "1", "b": "2"}
     )
-    assert str(request.url) == "http://example.com?a=1&b=2"
+    assert str(request.url) == "http://example.com?c=3&a=1&b=2"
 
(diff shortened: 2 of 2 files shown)


naman-netomi:
Hi team,
We’re extensively using httpx in our production setup, and some recent critical changes require us to upgrade from v0.27.2 → v0.28.1. However, this specific issue (where existing query parameters in the URL are lost when passing params) is breaking our flow.

Given the criticality of this change, deploying a custom patch isn’t an option for us at the moment.
Could you please share an estimated timeline for when this fix will be released?
It would be great if this could be prioritized, as it directly impacts our production rollout.

Thanks in advance!

veeceey:
Hi @naman-netomi, thanks for sharing your use case — I appreciate hearing that this fix would unblock your production upgrade. Unfortunately, release timelines are up to the httpx maintainers, so I can't speak to when this will land in a release. Hopefully the maintainers can weigh in when they get a chance.

PR #3769, opened 2026-02-20
Ensure mounted transports are closed even if main transport raises

bysiber:
When closing a client with proxy mounts, if the main transport's `close()`/`__exit__()` raises an exception, the mounted proxy transports would never be cleaned up. This wraps transport cleanup in `try/finally` to ensure all transports are properly closed.

Affects `Client.close()`, `Client.__exit__()`, `AsyncClient.aclose()`, and `AsyncClient.__aexit__()`.

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -1267,10 +1267,12 @@ class Client(BaseClient):
         if self._state != ClientState.CLOSED:
             self._state = ClientState.CLOSED
 
-            self._transport.close()
-            for transport in self._mounts.values():
-                if transport is not None:
-                    transport.close()
+            try:
+                self._transport.close()
+            finally:
+                for transport in self._mounts.values():
+                    if transport is not None:
+                        transport.close()
 
     def __enter__(self: T) -> T:
         if self._state != ClientState.UNOPENED:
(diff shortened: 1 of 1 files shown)


httpx/_client.py
```diff
@@ -1267,10 +1267,12 @@ def close(self) -> None:
...
-            self._transport.close()
-            for transport in self._mounts.values():
-                if transport is not None:
-                    transport.close()
+            try:
+                self._transport.close()
+            finally:
+                for transport in self._mounts.values():
+                    if transport is not None:
+                        transport.close()
```
noqt, line 1275: Cleanup still stops at the first mounted transport that raises, so later mounts remain open. On `ff7190e`, I used the public `Client(transport=..., mounts=...)` API with a main transport whose `close()` raises, a first mounted transport whose `close()` also raises, and a second mounted transport that records cleanup. `client.close()` produced `calls=['main', 'first']`; the second mount was never closed, and the first mount's exception replaced the original main-transport exception. The async loops have the same short-circuit behavior.

To meet this PR's all-transports cleanup goal, each mounted cleanup needs to be attempted independently while preserving or aggregating the raised exceptions.

Disclosure: I ran the changed client module through Bandit → SARIF → Lumi Trace at NOQT (which ranked `_client.py` first), then verified this with the focused runtime repro above.

PR #3771, opened 2026-02-20
Fix `unquote` IndexError on empty string input

bysiber:
`unquote('')` in `_utils.py` raises `IndexError` because it accesses `value[0]` without checking string length first.

This can occur when parsing digest auth `WWW-Authenticate` headers containing parameters with empty unquoted values (e.g. `realm=` instead of `realm=""`).

### DIFF
--- a/httpx/_utils.py
+++ b/httpx/_utils.py
@@ -89,7 +89,9 @@ def to_bytes_or_str(value: str, match_type_of: typing.AnyStr) -> typing.AnyStr:
 
 
 def unquote(value: str) -> str:
-    return value[1:-1] if value[0] == value[-1] == '"' else value
+    if len(value) >= 2 and value[0] == value[-1] == '"':
+        return value[1:-1]
+    return value
 
 
 def peek_filelike_length(stream: typing.Any) -> int | None:


httpx/_utils.py
```diff
@@ -89,7 +89,9 @@ def to_bytes_or_str(value: str, match_type_of: typing.AnyStr) -> typing.AnyStr:
 
 
 def unquote(value: str) -> str:
-    return value[1:-1] if value[0] == value[-1] == '"' else value
+    if len(value) >= 2 and value[0] == value[-1] == '"':
```
akx, line 92:
```suggestion
    if value and value[0] == value[-1] == '"':
```
to avoid a global lookup, a function call and a comparison...?

CAOShurong (approved):
Verified independently (no run, no claim). The PR was force-pushed after I first fetched it; I re-fetched and confirmed the current head before reviewing — head `55ead904eff7376ccbc7f23e93be19117c642441`.

Root cause: `unquote()` was `value[1:-1] if value[0] == value[-1] == '"' else value`. The truth condition `value[0] == value[-1] == '"'` accesses `value[0]` first, so for an empty string `unquote('')` raised `IndexError: string index out of range` before the quote-strip slice could ever run. This fires when parsing a `WWW-Authenticate` digest header carrying an empty unquoted value (e.g. `realm=`).

Behavior check at the current head:
- `unquote('')` returns `''` (no crash).
- `unquote('"a"')` returns `'a'` (quote-stripping preserved).
- `unquote('raw')` returns `'raw'` (unchanged).

On base `ae1b9f6` the same `unquote('')` raises `IndexError: string index out of range` — RED→GREEN confirmed. The fix guards with `len(value) >= 2` before the quote check, which is the minimal correct change.

One non-blocking suggestion: the PR has no regression test for the empty-string case. A tiny assertion (`assert unquote('') == ''`) in `tests/test_utils.py` would lock the behavior in. The change itself is correct and low-risk, so approving as-is; happy to see a follow-up test added.

(AI assistance used to collect verification evidence; the fix is the author's.)

Mukller:
LGTM. Simple fix: the len(value) >= 2 guard prevents the IndexError on empty string input. The original alue[0] == value[-1] == double-quote would raise on len(value) == 0.

Approved.

PR #3773, opened 2026-02-22
Adapt test_response_decode_text_using_autodetect for chardet 6.0

musicinmybrain:
# Summary

See https://github.com/encode/httpx/discussions/3772. Version 6.0.0 of `chardet` detects this byte string as `WINDOWS-1252`, while previous versions detected `ISO-8859-1`. Since both are plausible (the encoded byte string decodes to the same Unicode string under either encoding), this PR adjusts the test to accept either claim.

# Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. **No new test required, as this fixes a test regression.**
- [x] I've updated the documentation accordingly. **No documentation changes required.**

### DIFF
--- a/tests/models/test_responses.py
+++ b/tests/models/test_responses.py
@@ -1011,7 +1011,10 @@ def test_response_decode_text_using_autodetect():
 
     assert response.status_code == 200
     assert response.reason_phrase == "OK"
-    assert response.encoding == "ISO-8859-1"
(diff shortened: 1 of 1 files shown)


lovelydinosaur (approved):
There's more we could say about this, though... yep. Thanks

PR #3783, opened 2026-02-26
Add option to keep the same method for 301/302 redirects

kajinamit:
# Summary

RFC hasn't been clear enough about the expected behavior for 301 and 302 response. While it's (unfortunately) common for browsers to switch the original http method to GET when following redirects, some server applications expects the "legacy" behavior which keeps the same method over redirects.

Add new option to clients to select the behavior.

# Checklist

### DIFF
--- a/httpx/_client.py
+++ b/httpx/_client.py
@@ -196,6 +196,7 @@ class BaseClient:
         timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
         follow_redirects: bool = False,
         max_redirects: int = DEFAULT_MAX_REDIRECTS,
+        keep_method_for_redirects: bool = False,
         event_hooks: None | (typing.Mapping[str, list[EventHook]]) = None,
         base_url: URL | str = "",
         trust_env: bool = True,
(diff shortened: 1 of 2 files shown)


imbdb (approved):
Thank you for this PR. Looks good to me.
Helps in getting around these non standard conventions.

httpx/_client.py
```diff
@@ -502,6 +504,9 @@ def _redirect_method(self, request: Request, response: Response) -> str:
         if response.status_code == codes.SEE_OTHER and method != "HEAD":
             method = "GET"
 
+        if self.keep_method_for_redirects:
```
imbdb, line 507: If you are targeting 301/302, Please make them explicit in the condition and also add a comment for this behavior change for future humans/agents to understand the context.

imbdb (changes requested):

httpx/_client.py
kajinamit: I've added a few logics/comments to explain the intention and hope that addresses your concern.

motaharevadiaty-design: می‌خوام برنامه دانلود کنم

lovelydinosaur:
Thanks for the interest. I'd rather not have configuration options to support esoteric configurations.
You'd need to ensure redirects are off for this use case and handle that directly.

